PR #48, opened 2021-04-26
Refactor network

michalseweryn:


finality-aleph/src/environment.rs
```diff
@@ -131,18 +131,18 @@ impl<B: Block, H: Hash> UnitStore<B, H> {
 pub(crate) struct Environment<B: Block, H: Hash, C, BE, SC> {
     client: Arc<C>,
     select_chain: SC,
-    tx_consensus: Option<mpsc::UnboundedSender<NotificationIn<H>>>,
-    rx_consensus: Option<mpsc::UnboundedReceiver<NotificationOut<H>>>,
+    tx_consensus: mpsc::UnboundedSender<NotificationIn<H>>,
+    rx_consensus: mpsc::UnboundedReceiver<NotificationOut<H>>,
     tx_network: mpsc::UnboundedSender<NetworkCommand<B, H>>,
     rx_network: mpsc::UnboundedReceiver<NetworkEvent<B, H>>,
-    rx_order: Option<tokio::sync::mpsc::UnboundedReceiver<Vec<H>>>,
```
michalseweryn, line 143: There is no better moment to initialize fields than in the constructor

finality-aleph/src/environment.rs
```diff
@@ -131,18 +131,18 @@ impl<B: Block, H: Hash> UnitStore<B, H> {
...
+    rx_order: Fuse<tokio::sync::mpsc::UnboundedReceiver<Vec<H>>>,
     auth_cryptostore: AuthorityKeystore,
     store: UnitStore<B, H>,
     requests: BinaryHeap<ScheduledTask<H>>,
-    request_ticker: time::Interval,
+    request_ticker: Fuse<time::Interval>,
```
michalseweryn: Changing the implementation of the Future trait for Environment to an `async` method `run_epoch(self)` requires usage of fused streams

finality-aleph/src/environment.rs
```diff
@@ -131,18 +131,18 @@ impl<B: Block, H: Hash> UnitStore<B, H> {
...
     hashing: Box<dyn Fn(&[u8]) -> H + Send>,
     epoch_id: EpochId,
-    _phantom: std::marker::PhantomData<(B, BE)>,
+    _phantom: PhantomData<BE>,
```
michalseweryn, line 149: less is more :)

finality-aleph/src/environment.rs
```diff
@@ -156,56 +156,35 @@ where
     pub(crate) fn new(
```
michalseweryn, line 160: The large number of arguments makes sad not only clippy but also myself. Maybe we should get rid of the constructor completely?

finality-aleph/src/environment.rs
```diff
@@ -156,56 +156,35 @@ where
...
+            rx_consensus,
             tx_network,
             rx_network,
-            rx_order: None,
+            rx_order: rx_order.fuse(),
             auth_cryptostore,
             store: UnitStore::new(),
             requests: BinaryHeap::new(),
-            request_ticker: time::interval(TICK_INTERVAL),
+            request_ticker: time::interval(TICK_INTERVAL).fuse(),
```
michalseweryn: I have just noticed that since `Environment` is no longer a future, we do not need to store `Interval` as a field -- we can create it in the `run_epoch` method.  I will fix it

finality-aleph/src/environment.rs
```diff
@@ -247,55 +226,67 @@ where
     // Pulls requests from the priority queue (sorted by scheduled time) and sends them to random peers
     // as long as they are scheduled at time <= curr_time
     pub(crate) fn trigger_tasks(&mut self) {
-        loop {
+        while let Some(request) = self.requests.peek() {
```
michalseweryn, line 232: This equivalent approach to this loop seems more readable to me

finality-aleph/src/environment.rs
```diff
@@ -247,55 +226,67 @@ where
...
-                        let command = NetworkCommand::SendToRandPeer(message);
-                        self.place_network_command(command);
-                        debug!(target: "env", "Fetch request for {:?} sent.", coord);
-                        self.requests.push(ScheduledTask::new(
-                            Task::CoordRequest(coord),
-                            curr_time + FETCH_INTERVAL,
-                        ));
-                    } else {
-                        debug!(target: "env", "Request dropped as the unit is in store already {:?}", coord);
-                    }
```
michalseweryn, line 279: The method `trigger_tasks` was quite heave, so I delegated handling the two types of tasks to separate methods

finality-aleph/src/environment.rs
```diff
@@ -247,55 +226,67 @@ where
...
     }
 
+    fn schedule_coord_request(&mut self, coord: UnitCoord, curr_time: time::Instant) {
+        debug!(target: "env", "Starting request for {:?}", coord);
+        // If we already have a unit with such a coord in our store then there is no need to request it.
+        // It will be sent to consensus soon (or have already been sent).
+        if self.store.contains_coord(&coord) {
+            debug!(target: "env", "Request dropped as the unit is in store already {:?}", coord);
+            return;
+        }
```
michalseweryn, line 258: `else` is bad ;)

finality-aleph/src/environment.rs
```diff
@@ -410,83 +401,41 @@ where
...
-    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        debug!(target: "env", "Polling environment.");
-        if self.rx_consensus.is_some() {
-            loop {
-                match self.rx_consensus.as_mut().unwrap().poll_next_unpin(cx) {
-                    Poll::Ready(Some(notification)) => {
-                        self.on_consensus_notification(notification);
-                    }
-                    Poll::Ready(None) => {
+    pub(crate) async fn run_epoch(mut self) {
```
michalseweryn, line 423: using an async method improves readability

finality-aleph/src/network.rs
```diff
@@ -13,6 +13,9 @@ use crate::{
     AuthorityId, EpochId,
 };
 
+#[cfg(test)]
+mod tests;
+
```
michalseweryn: This is accidental, should not be here

finality-aleph/src/lib.rs
```diff
@@ -194,19 +195,5 @@ where
...
-        auth_keystore,
-        authorities,
-    } = config;
-    let consensus = party::ConsensusParty::new(client, network, select_chain, auth_keystore);
-    async move {
-        consensus
-            .run(authorities, consensus_config, spawn_handle.into())
-            .await;
-    }
+    run_consensus_party(AlephParams{ config })
```
michalseweryn: Earlier, the `run_aleph_consensus` permuted the fields of `AlephConfig`, put some of them to the `ConsensusParty` and passed remaining ones in the `run` method. Instead we can simply pass the config to an async function.  The config is wrapped in an `AlephParams` type, mimicking the style of grandpa.

finality-aleph/src/party.rs
```diff
@@ -1,101 +1,88 @@
...
-    client: Arc<C>,
-    select_chain: SC,
-    auth_keystore: AuthorityKeystore,
-    //NOTE: not sure if this phantom is necessary here
-    _phantom: std::marker::PhantomData<BE>,
+pub struct AlephParams<N, C, SC> {
+    pub config: crate::AlephConfig<N, C, SC>,
 }
 
-impl<B, N: network::Network<B>, C, BE, SC> ConsensusParty<B, N, C, BE, SC>
```
michalseweryn, line 30: As mention above, the interface of `ConsensusParty` made it isomorphic to a simple async function, so I replaced it with an async function.

finality-aleph/src/party.rs
```diff
@@ -1,101 +1,88 @@
...
-            None,
-            authorities,
-            auth_keystore,
-            hashing,
-            epoch_id,
-        );
+    let (notification_in_tx, notification_in_rx) = mpsc::unbounded();
+    let (notification_out_tx, notification_out_rx) = mpsc::unbounded();
+    // Making `rush::Consensus` accept only tokio sender sinks seems a questionable choice...
+    let (order_tx, order_rx) = tokio::sync::mpsc::unbounded_channel();
```
michalseweryn, line 57: A lot of txes and rxes are created here, but anyway, it seems to be the most reasonable approach to link different components with channels. Alternatively we could return sinks (txes) and streams (rxes) in the constructors of the components, but this would require forwarding the streams to sinks, which generates many futures to poll. Yet another solutions would be to choose a convention that e.g. the constructors create channels, keep the sinks return streams, keep the sinks, and the streams are initialized after the construction, which is not a clean solution neither.

finality-aleph/src/party.rs
```diff
@@ -1,101 +1,88 @@
...
-            rx_network,
-            None,
-            authorities,
-            auth_keystore,
-            hashing,
-            epoch_id,
-        );
+    let (notification_in_tx, notification_in_rx) = mpsc::unbounded();
+    let (notification_out_tx, notification_out_rx) = mpsc::unbounded();
+    // Making `rush::Consensus` accept only tokio sender sinks seems a questionable choice...
```
michalseweryn, line 56: Is there a reason, why we use both futures and tokio channels?

finality-aleph/src/environment.rs
DamianStraszak: Agreed, this was an artifact from times we thought environment will be a shared struct.

finality-aleph/src/environment.rs
DamianStraszak: I would keep new(), as there is a lot of fields to initialize and it would be a little cumbersome to do that from outside. At some point we can bundle together some of the arguments, but Environment is largely WIP at this point, so not sure it makes sense to do it now. I think we should turn off that warning from clippy :)

finality-aleph/src/environment.rs
DamianStraszak: Right.

finality-aleph/src/environment.rs
```diff
@@ -430,80 +419,37 @@ where
...
+        let mut request_ticker = time::interval(TICK_INTERVAL).fuse();
+        loop {
+            futures::select! {
```
DamianStraszak, line 426: You can take a look at tokio::select! -- it does not require fused futures. Not sure it matters though.

finality-aleph/src/party.rs
```diff
@@ -1,101 +1,87 @@
...
-    pub(crate) async fn run(
-        &self,
-        authorities: Vec<AuthorityId>,
-        conf: ConsensusConfig<NodeId>,
-        spawn_handle: SpawnHandle,
-    ) {
-        // TODO now it runs just a single instance of consensus but later it will
-        // orchestrate managing multiple instances for differents epoch
+    let (net_command_tx, net_command_rx) = mpsc::unbounded();
+    let network = network.clone();
```
DamianStraszak, line 42: This clone does not seem that useful :)?

finality-aleph/src/environment.rs
michalseweryn: Initializing the fields from the outside does not seem much more cumbersome than passing them to the constructor, and the constructor does nothing ;). But I am fine with leaving it as is. I will add an attribute to silence clippy

finality-aleph/src/environment.rs
michalseweryn: Good to know! I will keep it that way, one fuse for a local variable seems acceptable

finality-aleph/src/party.rs
michalseweryn: fixed!

finality-aleph/src/party.rs
DamianStraszak: I think we intended to use tokio (and started to do so in rush), but then substrate uses futures' and it was much easier to write code with futures' in the substrate part. Now there is inconsistency. I think we should decide for one, and go for it -- at this point I imagine it would be easier to convert everything to futures channels.

finality-aleph/src/party.rs
DamianStraszak: Yeah, I see what you are saying. I would keep it as it is -- I think this will be the only place with mess like that.

finality-aleph/src/party.rs
mike1729: You're right, this way is simpler :)

PR #49, opened 2021-04-26
Add session information into block and read it in block import.

kostekIV:


pallet/src/lib.rs
```diff
@@ -1,15 +1,39 @@
...
+
+    #[derive(Encode, Decode)]
+    pub struct SessionChange<T>
+    where
+        T: Config,
+    {
+        /// The block number the session was created.
+        pub created_at: T::BlockNumber,
+        pub session_id: u64,
+        pub changed: bool,
```
mike1729, line 34: Here the name `changed` is out of the context and the name is not descriptive enough. Add a comment or change the name.

pallet/src/lib.rs
```diff
@@ -49,20 +91,75 @@ pub mod pallet {
...
+                assert!(
+                    <Authorities<T>>::get().is_empty(),
+                    "Authorities are already initialized!"
+                );
+                <Authorities<T>>::put(authorities);
+            }
+
+            <SessionInfo<T>>::put(SessionChange {
+                session_id: 0,
+                changed: true,
```
mike1729, line 109: If you set `changed` to `true` here, doesn't it result in setting the authorities again in `on_finalize`?

pallet/src/tests.rs
```diff
@@ -0,0 +1,10 @@
+#![cfg(test)]
+
+use crate::mock::{new_test_ext, run_session};
+
+#[test]
+fn test_run_session_for_n_rounds() {
+    new_test_ext(&[(1u64, 1u64), (2u64, 2u64)]).execute_with(|| {
+        run_session(10);
```
mike1729, line 8: How about some assert?

pallet/src/lib.rs
kostekIV: I have changed it to `authorities_changed`.

pallet/src/lib.rs
kostekIV: Yes but only once for the first block and it will only update the store state to the same state it had. Other options I saw was some ugly boolean flags with if-else which I thought are not needed for such a one-time case.

pallet/src/tests.rs
kostekIV: Added some 'real' tests here.

PR #63, opened 2021-05-24
Add committee change handling to block import and pallet

maciejnems:
* Forward all justifications from BlockImport to JustificationNotification sender
* Remove WillChange storage from pallet
* Add AlephSessionApi
* Implement AlephSessionApi trait for runtime

### DIFF
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -29,8 +29,8 @@ pub use frame_support::{
     construct_runtime, parameter_types,
     sp_runtime::curve::PiecewiseLinear,
     traits::{
-        Currency, Imbalance, KeyOwnerProofSystem, LockIdentifier, OnUnbalanced, Randomness,
-        U128CurrencyToVote,
+        Currency, EstimateNextNewSession, Imbalance, KeyOwnerProofSystem, LockIdentifier,
+        OnUnbalanced, Randomness, U128CurrencyToVote, ValidatorSet,
     },
     weights::{
         constants::{BlockExecutionWeight, ExtrinsicBaseWeight, RocksDbWeight, WEIGHT_PER_SECOND},
@@ -38,7 +38,7 @@ pub use frame_support::{
     },
     StorageValue,
 };
-use primitives::AuthorityId as AlephId;
+use primitives::{ApiError as AlephApiError, AuthorityId as AlephId, Session as AuthoritySession};
 
 pub use pallet_balances::Call as BalancesCall;
 pub use pallet_timestamp::Call as TimestampCall;
@@ -518,9 +518,13 @@ impl_runtime_apis! {
         }
     }
 
-    impl primitives::AlephApi<Block> for Runtime {
-        fn authorities() -> Vec<AlephId> {
-            Aleph::authorities()
+    impl primitives::AlephSessionApi<Block, AlephId, BlockNumber> for Runtime {
+        fn current_session() -> AuthoritySession<AlephId, BlockNumber> {
+            Aleph::current_session()
+        }
+
+        fn next_session() -> Result<AuthoritySession<AlephId, BlockNumber>, AlephApiError> {
+            Aleph::next_session()
         }
     }
 }
--- a/finality-aleph/src/import.rs
+++ b/finality-aleph/src/import.rs
@@ -1,6 +1,7 @@
 use crate::{data_io::finalize_block, justification::AlephJustification, AuthorityId};
 use aleph_primitives::{AuthoritiesLog, ALEPH_ENGINE_ID};
 use codec::Encode;
+use futures::channel::mpsc;
 use sc_client_api::backend::Backend;
 use sp_api::TransactionFor;
 use sp_consensus::{
@@ -22,7 +23,8 @@ where
 {
     inner: Arc<I>,
     authorities: Vec<AuthorityId>,
-    _phantom: PhantomData<(Be, Block)>,
+    justification_rx: mpsc::UnboundedSender<JustificationNotification<Block>>,
+    _phantom: PhantomData<Be>,
 }
 
 impl<Block, Be, I> AlephBlockImport<Block, Be, I>
@@ -31,10 +33,15 @@ where
     Be: Backend<Block>,
     I: crate::ClientForAleph<Block, Be>,
 {
-    pub fn new(inner: Arc<I>, authorities: Vec<AuthorityId>) -> AlephBlockImport<Block, Be, I> {
+    pub fn new(
+        inner: Arc<I>,
+        authorities: Vec<AuthorityId>,
+        justification_rx: mpsc::UnboundedSender<JustificationNotification<Block>>,
+    ) -> AlephBlockImport<Block, Be, I> {
         AlephBlockImport {
             inner,
             authorities,
+            justification_rx,
             _phantom: PhantomData,
         }
     }
@@ -73,6 +80,7 @@ where
         AlephBlockImport {
             inner: self.inner.clone(),
             authorities: self.authorities.clone(),
+            justification_rx: self.justification_rx.clone(),
             _phantom: PhantomData,
         }
     }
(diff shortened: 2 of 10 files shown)


bin/runtime/src/lib.rs
```diff
@@ -518,9 +519,13 @@ impl_runtime_apis! {
         }
     }
 
-    impl primitives::AlephApi<Block> for Runtime {
-        fn authorities() -> Vec<AlephId> {
-            Aleph::authorities()
+    impl primitives::AlephApi<Block, AlephId, BlockNumber> for Runtime {
+        fn current_session() -> Option<AuthoritySession<AlephId,BlockNumber>> {
```
kostekIV, line 523: There are some styling issues, did you run `local-pipeline`?

bin/runtime/src/lib.rs
```diff
@@ -518,9 +519,13 @@ impl_runtime_apis! {
...
+            Aleph::current_session()
+        }
+
+        fn next_session() -> Option<AuthoritySession<AlephId,BlockNumber>> {
```
kostekIV, line 527: If possible I would prefer this to be a vector of next sessions even if always only 1 will be know in advance

pallet/src/lib.rs
```diff
@@ -42,38 +29,29 @@ pub mod pallet {
...
-                    Self::deposit_log(AuthoritiesLog::WillChange {
-                        session_id: session_info.session_id,
-                        // TODO: this is a stub for now.
-                        when: block_number,
-                        next_authorities: session_info.next_authorities,
-                    });
-                }
-            }
-        }
+        fn on_finalize(_block_number: T::BlockNumber) {}
```
kostekIV, line 40: not needed anymore

pallet/src/lib.rs
```diff
@@ -95,44 +73,43 @@ pub mod pallet {
...
-                    created_at: current_block,
-                    next_authorities: authorities,
+                    authorities,
+                    start_h: current_block,
+                    stop_h: current_block + T::Period::get(),
+                });
+                <NextSession<T>>::put(Session {
+                    session_id: old_session.session_id + 2,
+                    authorities: queued_authorities,
+                    start_h: current_block + T::Period::get(),
```
kostekIV, line 108: Shouldn't start_h = stop_h + 1 of previous session?

primitives/src/lib.rs
```diff
@@ -19,20 +19,24 @@ mod app {
...
     }
 }
 
 #[derive(Decode, Encode, PartialEq, Eq, Clone)]
-pub enum AuthoritiesLog<Id, Number>
+pub struct Session<Id, BlockNumber>
 where
     Id: Encode + Decode,
-    Number: Encode + Decode,
+    BlockNumber: Encode + Decode + AtLeast32BitUnsigned,
```
kostekIV, line 36: Is it necessary for BlockNumber to be ` AtLeast32BitUnsigned`?

pallet/src/lib.rs
maciejnems: It should, thanks for noticing :)

finality-aleph/src/import.rs
```diff
@@ -122,10 +103,14 @@ where
             justifications.and_then(|just| just.into_justification(ALEPH_ENGINE_ID))
         {
             let res = self.import_justification(hash, number, (ALEPH_ENGINE_ID, justification));
-            res.unwrap_or_else(|_err| {
-                imported_aux.bad_justification = true;
+            if let Err(ConsensusError::ClientImport(message)) = res {
+                if message == "Aleph can import only Aleph justifications." {
```
kostekIV, line 107: It looks strange that we are marking bad_justification only if the error message is matched.

finality-aleph/src/import.rs
```diff
@@ -158,22 +143,26 @@ where
...
-            );
-            Ok(())
-        } else {
-            Err(ConsensusError::ClientImport("Bad justification".into()))
+                justification: justification.1,
+            }) {
+            Err(_) => Err(ConsensusError::ClientImport(
+                "Could not send justification to ConsensusParty".into(),
+            )),
+            Ok(()) => Ok(()),
```
kostekIV, line 156: would map_err work here instead of match block?

finality-aleph/src/import.rs
maciejnems: I agree, but I don't have any other idea how we could distinguish this error and it should be distinguished. `ConsensusError::ClientImport` takes only string arguments.

bin/runtime/src/lib.rs
DamianStraszak: I would even say, the most convenient would be something like
`get_session(SessionId) -> Option<AuthoritySession...>`
but @kostekIV is the one who knows the best :)

kostekIV: +1 for querying based on SessionID, looks like the cleanest API to use and write. Although what should be the behaviour if we ask for session `k` in a block in which we are already in session `k + 1 + something`?  Should it return None? If not we should probably store some map from sessionID to AuthoritySession in the pallet.

bin/runtime/src/lib.rs
```diff
@@ -519,13 +518,13 @@ impl_runtime_apis! {
...
-        }
-
-        fn next_session() -> Option<AuthoritySession<AlephId,BlockNumber>> {
-            Aleph::next_session()
+    impl primitives::AlephSessionApi<Block, AlephId, BlockNumber> for Runtime {
+        fn current_session() -> AuthoritySession<AlephId,BlockNumber> {
+            AuthoritySession {
+                session_id: Session::session_index(),
+                authorities: Aleph::authorities(),
+                stop_h: Session::estimate_next_new_session(System::block_number()).0.unwrap(),
```
kostekIV, line 526: Does it returns start_h of the next session or start_h -1?

bin/runtime/src/lib.rs
maciejnems: Should be fixed now

primitives/src/lib.rs
maciejnems: It is not. I removed it

bin/runtime/src/lib.rs
maciejnems: We to Session pallet `Period` and `Offset` as parameters. This will always return `Offset + k*Period` here. If current block is `n`, then  `Offset + k*Period > n`. So for example if `n = Offset + k*Period` it will return `Offset + (k+1)*Period`

bin/runtime/src/lib.rs
```diff
@@ -518,9 +518,32 @@ impl_runtime_apis! {
...
+                    .unwrap(),
+            }
+        }
+
+        fn next_session() -> AuthoritySession<AlephId, BlockNumber> {
+            AuthoritySession {
+                session_id: Session::session_index() + 1,
+                authorities: Session::queued_keys()
+                    .iter()
+                    .map(|(_, key)| key.get(AlephId::ID).unwrap_or_default())
```
mike1729, line 537: Such `unwrap_or_default` could hide some mistake and be hard to find. Usually, such calls are implemented in pallets so that an error as missing key here should be logged. Generally, `uwrap`s are good for debug or in places where you are 100% sure that it is really not `None`.

finality-aleph/src/import.rs
maciejnems: Ok, I added a new function called `send_justification`

bin/runtime/src/lib.rs
maciejnems: I moved implementation to the palled and added error handling for this case. Other unwraps are ok and I added comments for them.

pallet/src/lib.rs
```diff
@@ -103,35 +74,43 @@ pub mod pallet {
...
+                .iter()
+                .map(|(_, key)| key.get(T::AuthorityId::ID).ok_or(AlephApiError::DecodeKey))
+                .collect::<Result<Vec<T::AuthorityId>, AlephApiError>>()
+                .and_then(|authorities| {
+                    Ok(AuthoritySession {
+                        session_id: Session::<T>::session_index() + 1,
+                        authorities,
+                        // System::<T>::block_number() is a valid block number, so unwrap should always be ok here
+                        stop_h: Session::<T>::estimate_next_new_session(
+                            Session::<T>::estimate_next_new_session(System::<T>::block_number())
```
mike1729, line 106: Could you define a tmp var for this with some name indicating what's going on? Like `next_session_start`.

mike1729 (approved):

PR #67, opened 2021-06-09
Bump AlephBFT version

timorl:
Mostly changes related to the new network trait.

### DIFF
--- a/bin/node/src/service.rs
+++ b/bin/node/src/service.rs
@@ -3,7 +3,7 @@
 use aleph_runtime::{self, opaque::Block, RuntimeApi};
 use codec::Decode;
 use finality_aleph::{
-    default_rush_config, run_aleph_consensus, AlephBlockImport, AlephConfig, AuthorityId,
+    default_aleph_config, run_aleph_consensus, AlephBlockImport, AlephConfig, AuthorityId,
     AuthorityKeystore, ConsensusConfig, JustificationNotification,
 };
 use futures::channel::mpsc;
@@ -151,7 +151,7 @@ fn consensus_config(auth: AuthorityId, authorities: &[AuthorityId]) -> Consensus
     let node_id = authorities.iter().position(|a| a == &auth).unwrap().into();
     let n_members = authorities.len().into();
 
-    default_rush_config(n_members, node_id, 0)
+    default_aleph_config(n_members, node_id, 0)
 }
 
 /// Builds a new service for a full client.
--- a/bin/node/src/service.rs
+++ b/bin/node/src/service.rs
@@ -3,8 +3,8 @@
 use aleph_runtime::{self, opaque::Block, RuntimeApi};
 use codec::Decode;
 use finality_aleph::{
-    run_aleph_consensus, AlephBlockImport, AlephConfig, AuthorityId, AuthorityKeystore,
-    ConsensusConfig, JustificationNotification,
+    default_rush_config, run_aleph_consensus, AlephBlockImport, AlephConfig, AuthorityId,
+    AuthorityKeystore, ConsensusConfig, JustificationNotification,
 };
 use futures::channel::mpsc;
 use sc_client_api::{CallExecutor, ExecutionStrategy, ExecutorProvider};
@@ -151,12 +151,7 @@ fn consensus_config(auth: AuthorityId, authorities: &[AuthorityId]) -> Consensus
     let node_id = authorities.iter().position(|a| a == &auth).unwrap().into();
     let n_members = authorities.len().into();
 
-    ConsensusConfig {
-        node_id,
-        session_id: 0,
-        n_members,
-        create_lag: std::time::Duration::from_millis(500),
-    }
+    default_rush_config(n_members, node_id, 0)
 }
 
 /// Builds a new service for a full client.
--- a/finality-aleph/Cargo.toml
+++ b/finality-aleph/Cargo.toml
@@ -21,7 +21,7 @@ sp-keystore = { git = "https://github.com/paritytech/substrate.git", rev = "1fa8
 sc-network = { git = "https://github.com/paritytech/substrate.git", rev = "1fa8cf7"}
 sc-network-gossip = { git = "https://github.com/paritytech/substrate.git", rev = "1fa8cf7"}
 sc-telemetry = { git = "https://github.com/paritytech/substrate.git", rev = "1fa8cf7"}
-rush = { git = "https://github.com/Cardinal-Cryptography/rush.git", rev = "caed0da" }
+rush = { git = "https://github.com/Cardinal-Cryptography/AlephBFT.git", rev = "381d388" }
 serde = "1.0"
 sc-service = { git = "https://github.com/paritytech/substrate.git", rev = "1fa8cf7"}
 sp-application-crypto = { git = "https://github.com/paritytech/substrate.git", rev = "1fa8cf7"}
--- a/finality-aleph/src/lib.rs
+++ b/finality-aleph/src/lib.rs
@@ -3,9 +3,8 @@ use sp_keystore::{SyncCryptoStore, SyncCryptoStorePtr};
 use codec::{Decode, Encode};
 
 use futures::Future;
-pub use rush::{
-    default_config as default_aleph_config, nodes::NodeIndex, Config as ConsensusConfig,
-};
+use rush::NodeIndex;
+pub use rush::{default_config as default_aleph_config, Config as ConsensusConfig};
 use sc_client_api::{backend::Backend, Finalizer, LockImportRun, TransactionFor};
 use sc_service::SpawnTaskHandle;
 use sp_api::{NumberFor, ProvideRuntimeApi};
--- a/finality-aleph/src/lib.rs
+++ b/finality-aleph/src/lib.rs
@@ -4,7 +4,7 @@ use codec::{Decode, Encode};
 
 use futures::Future;
 pub use rush::{
-    default_config as default_rush_config, nodes::NodeIndex, Config as ConsensusConfig,
+    default_config as default_aleph_config, nodes::NodeIndex, Config as ConsensusConfig,
 };
 use sc_client_api::{backend::Backend, Finalizer, LockImportRun, TransactionFor};
 use sc_service::SpawnTaskHandle;
--- a/finality-aleph/src/lib.rs
+++ b/finality-aleph/src/lib.rs
@@ -3,13 +3,18 @@ use sp_keystore::{SyncCryptoStore, SyncCryptoStorePtr};
 use codec::{Decode, Encode};
 
 use futures::Future;
-pub use rush::{nodes::NodeIndex, Config as ConsensusConfig};
+pub use rush::{
+    default_config as default_rush_config, nodes::NodeIndex, Config as ConsensusConfig,
+};
 use sc_client_api::{backend::Backend, Finalizer, LockImportRun, TransactionFor};
 use sc_service::SpawnTaskHandle;
 use sp_api::{NumberFor, ProvideRuntimeApi};
 use sp_blockchain::{HeaderBackend, HeaderMetadata};
 use sp_consensus::{BlockImport, SelectChain};
-use sp_runtime::{traits::Block, RuntimeAppPublic};
+use sp_runtime::{
+    traits::{BlakeTwo256, Block},
+    RuntimeAppPublic,
+};
 use std::{convert::TryInto, fmt::Debug, sync::Arc};
 pub mod config;
 mod data_io;
@@ -137,12 +142,15 @@ where
 {
 }
 
-#[derive(Clone, Default, Debug, Decode, Encode)]
+type Hasher = hash::Wrapper<BlakeTwo256>;
+
+#[derive(Clone, Debug, Decode, Encode)]
 struct Signature {
     id: NodeIndex,
     sgn: AuthoritySignature,
 }
 
+#[derive(Clone)]
 struct KeyBox {
     id: NodeIndex,
     auth_keystore: AuthorityKeystore,
--- a/finality-aleph/src/network.rs
+++ b/finality-aleph/src/network.rs
@@ -15,8 +15,8 @@ use crate::{Error, Hasher, KeyBox, SessionId, Signature};
 #[cfg(test)]
 mod tests;
 
-#[derive(PartialEq, Eq, Clone, Debug, Hash)]
-struct PeerId(ScPeerId);
+#[derive(PartialEq, Eq, Copy, Clone, Debug, Hash)]
+pub struct PeerId(ScPeerId);
 
 impl From<PeerId> for ScPeerId {
     fn from(wrapper: PeerId) -> Self {
@@ -31,8 +31,6 @@ impl From<ScPeerId> for PeerId {
 }
 
 impl Encode for PeerId {
-    // TODO: this probably has a constant number of bytes when encoded,
-    // so `size_hint` would be useful.
     fn using_encoded<R, F: FnOnce(&[u8]) -> R>(&self, f: F) -> R {
         self.0.to_bytes().using_encoded(f)
     }
@@ -41,10 +39,9 @@ impl Encode for PeerId {
 impl Decode for PeerId {
     fn decode<I: codec::Input>(value: &mut I) -> Result<Self, codec::Error> {
         let bytes = Vec::<u8>::decode(value)?;
-        match ScPeerId::from_bytes(&bytes) {
-            Err(_) => Err("PeerId not encoded with to_bytes".into()),
-            Ok(result) => Ok(result.into()),
-        }
+        ScPeerId::from_bytes(&bytes)
+            .map_err(|_| "PeerId not encoded with to_bytes".into())
+            .map(|pid| pid.into())
     }
 }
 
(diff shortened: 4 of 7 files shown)


finality-aleph/src/network.rs
```diff
@@ -87,51 +124,95 @@ impl<B: BlockT, H: ExHashT> Network<B> for Arc<NetworkService<B, H>> {
...
+        self.all_peers.remove(peer);
     }
 
-    pub(crate) fn iter(&self) -> impl Iterator<Item = (&PeerId, &PeerInfo)> {
-        self.peers.iter()
+    pub(crate) fn iter(&self, session_id: SessionId) -> impl Iterator<Item = &PeerId> {
+        self.to_peer
+            .iter()
+            .filter(move |((s_id, _), _)| *s_id == session_id)
+            .map(|(_, p_id)| p_id)
```
michalseweryn, line 211: `iter` is a bit unfortunate name. How about `authenticated_peers_for`?
Maybe we would like to make  `to_peer` a `HashMap<SessionId, HashMap<NodeIndex, PeerId>>` to avoid iterating the whole hashmap?

michalseweryn (changes requested):
Gonna continue tomorrow

bin/node/src/service.rs
```diff
@@ -151,12 +151,7 @@ fn consensus_config(auth: AuthorityId, authorities: &[AuthorityId]) -> Consensus
     let node_id = authorities.iter().position(|a| a == &auth).unwrap().into();
     let n_members = authorities.len().into();
 
-    ConsensusConfig {
-        node_id,
-        session_id: 0,
-        n_members,
-        create_lag: std::time::Duration::from_millis(500),
-    }
+    default_rush_config(n_members, node_id, 0)
```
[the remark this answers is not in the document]
timorl: Removed this instance and some instances in network itself – I don't think using `rush` was necessary in either case, so leaving less work for anyone who will be fixing it later.

finality-aleph/src/network.rs
```diff
@@ -1,18 +1,53 @@
...
+
+impl From<ScPeerId> for PeerId {
+    fn from(id: ScPeerId) -> Self {
+        PeerId(id)
+    }
+}
+
+impl Encode for PeerId {
+    // TODO: this probably has a constant number of bytes when encoded,
+    // so `size_hint` would be useful.
```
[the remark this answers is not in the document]
timorl: Because it's undocumented and even if true most likely subject to change without notice. I guess I should remove the TODO then...

finality-aleph/src/network.rs
```diff
@@ -141,191 +222,253 @@ enum SessionStatus {
...
         }
     }
 }
 
-impl<B: BlockT + 'static, N: Network<B> + Clone> ConsensusNetwork<B, N> {
+impl<D: Clone + Encode + Decode, B: BlockT + 'static, N: Network<B> + Clone>
+    ConsensusNetwork<D, B, N>
+{
     /// Create a new instance.
     pub(crate) fn new(network: N, protocol: impl Into<Cow<'static, str>>) -> Self {
```
michalseweryn, line 340: `protocol` should be just a `Cow` :cow:

finality-aleph/src/network/tests.rs
```diff
@@ -235,178 +252,182 @@ async fn test_network_event_sync_disconnected() {
...
-                assert!(peer_id == data.bob.peer_id || peer_id == data.charlie.peer_id);
-                assert_eq!(protocol, PROTOCOL_NAME);
-                match <(SessionId, Vec<u8>)>::decode(&mut message.as_slice()) {
-                    Ok((session_id, message)) => {
-                        assert_eq!(session_id.0, 0);
-                        assert_eq!(message, fake_message);
-                    }
-                    _ => panic!("Expected a properly encoded message"),
+    // Wait for acknowledgement that Alice noted Bob's presence.
+    let _ = data
```
michalseweryn, line 348: Isn't `let _ = ` pointless here and below?

michalseweryn (changes requested):

finality-aleph/src/network.rs
```diff
@@ -24,48 +59,50 @@ pub trait Network<B: BlockT>: Clone + Send + Sync + 'static {
...
-    fn add_set_reserved(&self, who: PeerId, protocol: Cow<'static, str>);
+    fn add_set_reserved(&self, who: ScPeerId, protocol: Cow<'static, str>);
 
     /// TODO: figure out what does this actually do...
-    fn remove_set_reserved(&self, who: PeerId, protocol: Cow<'static, str>);
+    fn remove_set_reserved(&self, who: ScPeerId, protocol: Cow<'static, str>);
+
+    // The PeerId of this node.
+    fn peer_id(&self) -> ScPeerId;
 }
```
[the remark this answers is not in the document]
timorl: No idea actually, I wrote this part over a week ago and apparently changing these this way seemed simpler to me then. Changed back, although there are quite a couple more `into`s now everywhere.

I think it might have been due to `PeerId` having to be `pub`?

finality-aleph/src/network.rs
```diff
@@ -87,51 +124,95 @@ impl<B: BlockT, H: ExHashT> Network<B> for Arc<NetworkService<B, H>> {
...
     }
+
     pub(crate) fn insert(&mut self, peer: PeerId) {
-        self.peers.insert(peer, PeerInfo::new());
+        self.all_peers.insert(peer, PeerInfo::new());
     }
 
-    pub(crate) fn remove(&mut self, peer: &PeerId) {
-        self.peers.remove(peer);
+    pub(crate) fn authenticated(&self, peer: &PeerId, session_id: &SessionId) -> bool {
```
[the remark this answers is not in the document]
timorl: I prefer just `authenticated` in such cases, but this is not a strong preference so changed.

finality-aleph/src/network.rs
```diff
@@ -87,51 +124,95 @@ impl<B: BlockT, H: ExHashT> Network<B> for Arc<NetworkService<B, H>> {
...
-}
 
-use rand::{seq::SliceRandom, thread_rng};
+    fn peer_id(&self) -> ScPeerId {
+        *self.local_peer_id()
+    }
+}
 
 #[derive(Debug)]
-pub(crate) struct PeerInfo {}
```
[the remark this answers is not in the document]
timorl: We could, but I really like the separation in logic.

finality-aleph/src/network.rs
```diff
@@ -87,51 +124,95 @@ impl<B: BlockT, H: ExHashT> Network<B> for Arc<NetworkService<B, H>> {
...
+        peer: &PeerId,
+        session_id: SessionId,
+        node_id: NodeIndex,
+    ) {
+        if self.all_peers.get(peer).is_none() {
+            self.insert(peer.clone());
+        }
+        self.all_peers
+            .get_mut(peer)
+            .unwrap()
```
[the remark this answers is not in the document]
timorl: Yup that is definitely better, changed.

finality-aleph/src/network.rs
```diff
@@ -87,51 +124,95 @@ impl<B: BlockT, H: ExHashT> Network<B> for Arc<NetworkService<B, H>> {
...
+        self.to_peer.insert((session_id, node_id), peer.clone());
     }
 
-    //TODO: optimize this (it does not need to be perfectly random, if this helps)
-    pub(crate) fn sample_random(&self) -> Option<PeerId> {
-        let peers: Vec<&PeerId> = self.peers.keys().collect();
-        let mut rng = thread_rng();
-        peers.choose(&mut rng).cloned().cloned()
+    pub(crate) fn remove(&mut self, peer: &PeerId) {
+        if let Some(peer_info) = self.all_peers.get(peer) {
```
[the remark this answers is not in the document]
timorl: Ha, yup, done.

finality-aleph/src/network.rs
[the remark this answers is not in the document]
michalseweryn: Well, afterall it is just a hint, so even if the obvious meaning is that it is the size of encoding in bytes turns out wrong, it shouldn't break anything. I tend to implement it.

finality-aleph/src/network.rs
```diff
@@ -1,18 +1,53 @@
...
+    }
+}
+
+impl Decode for PeerId {
+    fn decode<I: codec::Input>(value: &mut I) -> Result<Self, codec::Error> {
+        let bytes = Vec::<u8>::decode(value)?;
+        match ScPeerId::from_bytes(&bytes) {
+            Err(_) => Err("PeerId not encoded with to_bytes".into()),
+            Ok(result) => Ok(result.into()),
+        }
```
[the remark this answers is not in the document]
michalseweryn: I mean't something like:
```
let result = ScPeerId::from_bytes(&bytes).map_err(_ => "PeerId not encoded with to_bytes".into())?;
Ok(result.into())
```

finality-aleph/src/network.rs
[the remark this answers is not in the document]
michalseweryn: It seems that this introduces `into`s only in the implementation of the trait, as everywhere else we consequently use `PeerId`

finality-aleph/src/network.rs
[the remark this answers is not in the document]
michalseweryn: I do have a strong preference :P It is a standard to name methods returning `bool` as `is_*` (`is_empty`, `is_some`, `is_err` to name few). `authenticated` alone is more ambiguous as it does not suggest that is a predicate returning a bool.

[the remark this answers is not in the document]
michalseweryn: Well, now I see that it is different, as in this case the method takes arguments other than `self`... so I am not so convinced about `is_authenticated` anymore :sweat_smile:

finality-aleph/src/network.rs
timorl: Yup, that's definitely better. Unfortunately some of the code became potentially less readable due to the half-baked nature of functional features (well, maybe 7/8ths-baked, but still).

finality-aleph/src/network.rs
timorl: Heh, definitely, not my fault though I did not touch this specific line. ;p Changed.

finality-aleph/src/network/tests.rs
timorl: Yup, must have been some leftovers, thanks!

finality-aleph/src/network.rs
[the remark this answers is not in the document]
timorl: Yup, there was some fun with the compiler, but now it seems mostly fine.

finality-aleph/src/network.rs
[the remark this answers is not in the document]
timorl: Well, I would have to figure out how many bytes in a peer id and how these bytes translate into final size... I'm definitely too lazy to do all that, and even if I did this would be disgustingly unstable with changes in either `PeerId` or `Encode` of byte slices.

finality-aleph/src/network.rs
```diff
@@ -141,148 +228,195 @@ enum SessionStatus {
...
+            commands_for_session,
+            commands_from_user,
             peers: Arc::new(Mutex::new(Peers::new())),
-            phantom: PhantomData,
+            _phantom: PhantomData,
         }
     }
 
-    pub(crate) fn session_manager(&self) -> SessionManager {
+    pub(crate) fn session_manager(&self) -> SessionManager<D> {
```
michalseweryn, line 359: Why not store the SessionManager as a field of `ConsensusNetwork`?

michalseweryn (changes requested):
I know that those were not introduced by your changes, but can you fix the clippy errors?

finality-aleph/src/network.rs
```diff
@@ -398,3 +543,32 @@ impl<B: BlockT + 'static, N: Network<B> + Clone> ConsensusNetwork<B, N> {
         }
     }
 }
+
+pub(crate) type RushNetworkData<B> = rush::NetworkData<Hasher, <B as BlockT>::Hash, Signature>;
+
+pub(crate) struct RushNetwork<B: BlockT> {
+    inner: GenericNetwork<RushNetworkData<B>>,
+}
```
michalseweryn, line 551: Why are `RushNetwork` and `GenericNetwork` separate beings?

finality-aleph/src/network.rs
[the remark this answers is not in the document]
michalseweryn: Fair enough, can you make it module private?

finality-aleph/src/network.rs
```diff
@@ -87,51 +121,104 @@ impl<B: BlockT, H: ExHashT> Network<B> for Arc<NetworkService<B, H>> {
...
 }
 
-#[derive(Debug)]
 struct Peers {
-    pub(crate) peers: HashMap<PeerId, PeerInfo>,
+    all_peers: HashMap<PeerId, PeerInfo>,
+    to_peer: HashMap<SessionId, HashMap<NodeIndex, PeerId>>,
 }
 
 impl Peers {
```
michalseweryn, line 160: Methods of a private struct should be private

finality-aleph/Cargo.toml
```diff
@@ -21,7 +21,7 @@ sp-keystore = { git = "https://github.com/paritytech/substrate.git", rev = "1fa8
 sc-network = { git = "https://github.com/paritytech/substrate.git", rev = "1fa8cf7"}
 sc-network-gossip = { git = "https://github.com/paritytech/substrate.git", rev = "1fa8cf7"}
 sc-telemetry = { git = "https://github.com/paritytech/substrate.git", rev = "1fa8cf7"}
-rush = { git = "https://github.com/Cardinal-Cryptography/rush.git", rev = "caed0da" }
+rush = { git = "https://github.com/Cardinal-Cryptography/AlephBFT.git", rev = "381d388" }
```
DamianStraszak, line 24: Is the plan to rename rush -> alephbft in a separate PR? Otherwise you can do it here already.

finality-aleph/src/network.rs
timorl: Mostly for testing purposes – the logic is completely independent of the data being transferred over this network, but if I were to use AlephBFT's structs directly I would have to expose ways of constructing them outside of AlephBFT.

timorl:
> I know that those were not introduced by your changes, but can you fix the clippy errors?

I fixed all the clippy errors already, the failing pipeline is for different reasons which I cannot change. I believe you can bother Michał about it, because he has someone he can bother further.

finality-aleph/src/lib.rs
```diff
@@ -3,13 +3,18 @@ use sp_keystore::{SyncCryptoStore, SyncCryptoStorePtr};
 use codec::{Decode, Encode};
 
 use futures::Future;
-pub use rush::{nodes::NodeIndex, Config as ConsensusConfig};
+pub use rush::{
```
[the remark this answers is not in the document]
timorl: These are more like externalities, things that are supposed to be used outside of AlephBFT. And re-exporting all the ones we really need to use in one place seems like a prudent approach. However, I'll make sure that actually all of these need to be exported – I am relatively sure only about `default_config`.

finality-aleph/src/network.rs
```diff
@@ -141,148 +228,195 @@ enum SessionStatus {
...
+        let enc_auth_data = auth_data.encode();
+        let AuthData {
+            session_id,
             peer_id,
-            self.protocol.clone(),
-            (session_id, message).encode(),
-        );
+            node_id,
+        } = auth_data;
+        if let Some(session_data) = self.sessions.lock().get(&session_id) {
```
DamianStraszak, line 393: There might be (?) this annoying issue that some node starts a session 1-2 secs later than others and many authentications for the new session get lost forever.  I would say -- leave a TODO for this, and one can work on this in a separate task, as the PR is quite large already.

finality-aleph/src/party.rs
```diff
@@ -96,38 +95,30 @@ where
...
+        authorities: session.authorities.clone(),
+        id: node_id,
+    };
     let session_id = session.session_id as u64;
     let session_network =
-        session_manager.start_session(SessionId(session_id), session.authorities.clone());
+        RushNetwork::<B>::new(session_manager.start_session(SessionId(session_id), keybox.clone()));
 
+    let consensus_config =
+        default_aleph_config(session.authorities.len().into(), node_id, session_id);
```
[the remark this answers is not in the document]
timorl: Oh goodness, no. There is some proper misdesign here, the config contains `session_id`, which should change each session. In the future also the node indes might change between sessions, as well as potentially the node count. Anyway, there pretty clearly is a part of the config that we want to keep the same between sessions and part that we want to change. Right now this is not expressed in any way in the code and requires a bigger refactor to make it even remotely sane – potentially needing even changes in AlephBFT (instead of `default_config` separate the delay config out and have a function for creating it – `default_delay_config`; perhaps also additional changes that I haven't though of yet).

I created a task for that but I definitely don't want to do this in this PR.

finality-aleph/Cargo.toml
timorl: Doesn't work, at this commit there is no aleph-bft crate in the AlephBFT repository.

finality-aleph/src/network.rs
timorl: Huh, right, done.

finality-aleph/src/network.rs
timorl: `ConsensusNetwork` is passed to a separate thread to do its work there, so `SessionManager` has to be a separate object. It might work as a `Arc<Mutex<...>>` thing, but I think at this point it would be more complicated than having a separate object.

finality-aleph/src/lib.rs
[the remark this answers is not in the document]
timorl: Made `NodeIndex` a purely local import, the others are still necessary although the refactor mentioned below should provide opportunities for simplifying this.

finality-aleph/src/network.rs
timorl: Authentications are sent not only when a new session starts, but also on every `peer_connected` event, so I think this should actually work. Although I'm not convinced I get the semantics of the `peer_connected` events quite right, so I'm not sure.

DamianStraszak: The thing is though that the peer is already connected when a new session starts.

timorl: Ah, I see what you mean, the session is not started so we don't iterate over it. Yeah, that is a serious problem – either keeping a forward set of sessions, or resending authentications in the way we do with some other messages would solve it, but both require significant work. I'll add a task.

finality-aleph/src/network.rs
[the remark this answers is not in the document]
timorl: Whoops, missed this comment, force pushed this change too.

finality-aleph/src/network.rs
michalseweryn: I mean the session manager is constructed from three fields. why not couple these 3 fields into one field of type SessionManager?

timorl: One of them is actually taken out of `network`, so this would introduce slight data duplication, another is the channel, which not being hidden within a substruct is somewhat more readable in the code that uses it (I started writing this before noticing the thread issue). The last one, `sessions`, would actually be better within a substruct, but well, it's only one.

michalseweryn:
l

michalseweryn (approved):

DamianStraszak (approved):

PR #73, opened 2021-06-22
Add faucet account and http endpoint ports

timorl:


run_nodes.sh
```diff
@@ -35,6 +35,7 @@ for i in ${!authorities[@]}; do
     --chain dev \
     --base-path /tmp/$auth \
     --name $auth \
+    --rpc-port $(expr 9933 + $i) \
```
DamianStraszak, line 38: is it possible to move these port ranges 9933 --- 9944 further apart? I would imagine it's possible that someday one would run it with more than 10 nodes and then there could be some collisions (not sure how dangerous this is though :) )

bin/node/src/chain_spec.rs
```diff
@@ -72,19 +73,19 @@ pub fn development_config() -> Result<ChainSpec, String> {
...
         "Bob//stash",
         "Charlie",
         "Dave",
         "Eve",
-        "Ferdie",
     ]
     .iter()
     .map(get_account_id_from_seed::<sr25519::Public>)
     .collect();
+    rich_accounts.push(hex!["eaefd9d9b42915bda608154f17bb03e407cbf244318a0499912c2fb1cd879b74"].into());
```
DamianStraszak, line 88: Would be nice to leave a comment or so, saying that "This is the faucet address", to not get confused later.

run_nodes.sh
timorl: Port 9933 is the default port for the HTTP RPC endpoint, so I would prefer to leave it as is. Especially since this is only intended to run nodes on a single machine, so I wouldn't expect anyone running more than 10.

In good news I'm pretty sure collisions don't break much  – without this argument all nodes tried starting their HTTP RPC endpoint on port 9933 and one of them always succeeded, the rest failed silently. So even if someone runs more than 10 nodes at some point, they only have to make sure that they attempt HTTP RPC connections to one of the first ports and the WebSocket RPC connections to one of the last.

bin/node/src/chain_spec.rs
timorl: Good idea, done.

DamianStraszak (approved):

bin/node/src/chain_spec.rs
```diff
@@ -72,19 +73,20 @@ pub fn development_config() -> Result<ChainSpec, String> {
...
-    let rich_accounts: Vec<_> = [
+    let mut rich_accounts: Vec<_> = [
         "Alice",
         "Alice//stash",
         "Bob",
         "Bob//stash",
         "Charlie",
         "Dave",
         "Eve",
-        "Ferdie",
```
maciejnems, line 83: Poor Ferdie, no longer rich :(

maciejnems (approved):

PR #74, opened 2021-06-24
Gossip

mike1729:
Implements a first proposition of gossip.

### DIFF
--- a/finality-aleph/src/network.rs
+++ b/finality-aleph/src/network.rs
@@ -15,7 +15,7 @@ use std::{
     sync::Arc,
 };
 
-use log::debug;
+use log::{debug, trace};
 
 use crate::{Error, Hasher, KeyBox, SessionId, Signature};
 
@@ -25,7 +25,7 @@ mod tests;
 // TODO below constants should be calculated based on the size of validators set for given session
 const GOSSIP_FORWARD: usize = 5;
 const SEND_FORWARD: usize = 2;
-const CACHE_SIZE: usize = 1_000_000;
+const CACHE_SIZE: usize = 10000;
 
 #[derive(PartialEq, Eq, Copy, Clone, Debug, Hash)]
 pub struct PeerId(ScPeerId);
@@ -463,7 +463,7 @@ where
         }
     }
 
-    fn forward(
+    fn forward_to_peers(
         &self,
         session_id: SessionId,
         index: MessageIndex,
@@ -491,6 +491,14 @@ where
         }
     }
 
+    fn send_to_user(&self, session_id: SessionId, data: D, session_data: &mut SessionData<D>) {
+        if let Err(e) = session_data.data_for_user.unbounded_send(data) {
+            //TODO: need to write some logic on when an session should be terminated and make sure
+            // that there are no issues with synchronization when terminating.
+            session_data.status = SessionStatus::Terminated;
+            debug!(target: "afa", "Error {:?} when passing a message event to session {:?}.", e, session_id);
+        }
+    }
     fn authenticate_to(&self, session_data: &SessionData<D>, peer_id: PeerId) {
         self.commands_for_session
             .unbounded_send(SessionCommand::Meta(
@@ -512,14 +520,14 @@ where
                 if peer_id == auth_data.peer_id {
                     self.on_incoming_authentication(auth_data, signature);
                 } else {
-                    debug!(target: "afa", "Peer {:?} attempting to authenticate as peer {:?}.", peer_id, auth_data.peer_id);
+                    trace!(target: "afa", "Peer {:?} attempting to authenticate as peer {:?}.", peer_id, auth_data.peer_id);
                 }
             }
             AuthenticationRequest(session_id) => {
                 if let Some(session_data) = self.sessions.lock().get(&session_id) {
                     self.authenticate_to(session_data, peer_id);
                 } else {
-                    debug!(target: "afa", "Received authentication request for unknown session: {:?}.", session_id);
+                    trace!(target: "afa", "Received authentication request for unknown session: {:?}.", session_id);
                 }
             }
         }
@@ -542,23 +550,23 @@ where
         if let Some(session_data) = sessions.get_mut(&session_id) {
             if session_data.status == SessionStatus::InProgress {
                 if session_data.messages.contains(&message) {
-                    debug!(target: "afa", "Received data with old index in session {:?}.", session_id);
+                    trace!(target: "afa", "Received data with old index in session {:?}.", session_id);
                     return;
                 } else {
                     session_data.messages.put(message, ());
                 }
-                if let Recipient::Target(node_id) = recipient {
-                    if node_id == session_data.auth_data.node_id {
-                        if let Err(e) = session_data.data_for_user.unbounded_send(data) {
-                            //TODO: need to write some logic on when an session should be terminated and make sure
-                            // that there are no issues with synchronization when terminating.
-                            session_data.status = SessionStatus::Terminated;
-                            debug!(target: "afa", "Error {:?} when passing a message event to session {:?}.", e, session_id);
-                        }
-                        return;
+                match recipient {
+                    Recipient::Target(node_id) if node_id == session_data.auth_data.node_id => {
+                        self.send_to_user(session_id, data, session_data);
+                    }
+                    Recipient::All => {
+                        self.send_to_user(session_id, data.clone(), session_data);
+                        self.forward_to_peers(session_id, index, data, recipient);
+                    }
+                    _ => {
+                        self.forward_to_peers(session_id, index, data, recipient);
                     }
                 }
-                self.forward(session_id, index, data, recipient);
             }
         }
     }
@@ -597,7 +605,7 @@ where
                 if self.peers.lock().is_authenticated(&peer_id, &session_id) {
                     self.on_incoming_data(session_id, index, data, recipient);
                 } else {
-                    debug!(target: "afa", "Received unauthenticated message from {:?} for session {:?}, requesting authentication.", peer_id, session_id);
+                    trace!(target: "afa", "Received unauthenticated message from {:?} for session {:?}, requesting authentication.", peer_id, session_id);
                     self.commands_for_session
                         .unbounded_send(SessionCommand::Meta(
                             MetaMessage::AuthenticationRequest(session_id),
--- a/finality-aleph/src/network.rs
+++ b/finality-aleph/src/network.rs
@@ -2,12 +2,17 @@ use codec::{Decode, Encode};
 use futures::{channel::mpsc, stream::Stream, StreamExt};
 use lru::LruCache;
 use parking_lot::Mutex;
-use rand::prelude::SliceRandom;
 use rush::{nodes::NodeIndex, Index, KeyBox as _};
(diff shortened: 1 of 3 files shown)


finality-aleph/src/network.rs
```diff
@@ -196,22 +209,29 @@ impl Peers {
...
     fn get(&self, session_id: SessionId, node_id: NodeIndex) -> Option<&PeerId> {
         self.to_peer.get(&session_id)?.get(&node_id)
     }
+
+    fn get_rand(&self, session_id: SessionId, n_peers: usize) -> impl Iterator<Item = &PeerId> {
+        self.session_peers
+            .get(&session_id)
+            .expect("the session already started")
+            .as_slice()
+            .choose_multiple(&mut rand::thread_rng(), n_peers)
```
timorl, line 233: I'm a bit concerned about the `thread_rng` here, this is usually quite a slow operation and sending messages is a very common occurrence. Perhaps instead of making this random do a simple round-robin on the set of peers to send to? Or at least keep the thread RNG somewhere so that further calls are only pseudorandom?

finality-aleph/src/network.rs
```diff
@@ -233,24 +259,27 @@ enum MetaMessage {
     AuthenticationRequest(SessionId),
 }
 
+type MessageIndex = u64;
+
 #[derive(Clone, Encode, Decode)]
 enum InternalMessage<D: Clone + Encode + Decode> {
     Meta(MetaMessage),
-    Data(SessionId, D),
+    Data(SessionId, MessageIndex, D, Recipient<NodeIndex>),
```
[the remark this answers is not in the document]
timorl: And simplify some function interfaces etc.

finality-aleph/src/network.rs
```diff
@@ -470,27 +573,20 @@ impl<D: Clone + Encode + Decode, B: BlockT + 'static, N: Network<B> + Clone>
...
-                    Some(node_id) => {
-                        if let Some(peer_id) = self.peers.lock().get(session_id, node_id) {
-                            self.send_message(peer_id, message);
-                        }
-                    }
+                let msg_bytes = data.encode();
+                let index = self.check_or_insert_cache(session_id, msg_bytes);
+                let message = InternalMessage::Data(session_id, index, data, recipient);
+                for peer_id in self.peers.lock().peers_authenticated_for(session_id) {
+                    self.send_message(peer_id, message.clone());
```
timorl, line 589: Am I reading this correctly that you always send the message to everyone, even if it is addressed to one of the nodes you actually know?

timorl (changes requested):
Very reasonable simple implementation, only a couple comments.

finality-aleph/src/network.rs
DamianStraszak, line 233: Will it work properly if the current number of peers is `< n_peers`? In which case it should choose all.

timorl: Interestingly the `RandIterator` version of the `choose_multiple` method states directly that this is the case (returns all if  `n_peers` is too big), but this is not specified in this variant...

finality-aleph/src/network.rs
```diff
@@ -155,13 +162,15 @@ impl PeerInfo {
 struct Peers {
     all_peers: HashMap<PeerId, PeerInfo>,
     to_peer: HashMap<SessionId, HashMap<NodeIndex, PeerId>>,
+    session_peers: HashMap<SessionId, Vec<PeerId>>,
```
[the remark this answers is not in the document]
mike1729: I'm not sure if `values` returns the same iterator every time it's called but here it doesn't make a change. Changed.

finality-aleph/src/network.rs
```diff
@@ -233,24 +259,27 @@ enum MetaMessage {
...
     Meta(MetaMessage),
-    Data(SessionId, D),
+    Data(SessionId, MessageIndex, D, Recipient<NodeIndex>),
 }
 
 struct SessionData<D: Clone + Encode + Decode> {
     pub(crate) data_for_user: mpsc::UnboundedSender<D>,
     pub(crate) status: SessionStatus,
     pub(crate) keychain: KeyBox,
+    messages: LruCache<Vec<u8>, MessageIndex>,
```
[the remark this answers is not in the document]
mike1729: Changed.

finality-aleph/src/network.rs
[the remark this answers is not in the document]
mike1729: Added `DataMesage`.

finality-aleph/src/network.rs
```diff
@@ -365,19 +398,66 @@ impl<D: Clone + Encode + Decode, B: BlockT + 'static, N: Network<B> + Clone>
         }
     }
 
+    fn check_or_insert_cache(&self, session_id: SessionId, msg_bytes: Vec<u8>) -> MessageIndex {
+        let mut sessions = self.sessions.lock();
+        let messages = &mut sessions
+            .get_mut(&session_id)
+            .expect("session has started.")
```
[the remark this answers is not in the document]
mike1729: We run this only after `new_session` has been called.

finality-aleph/src/network.rs
mike1729: Added a pseudorandom method but I'm not entirely satisfied with it. I'll welcome any improvements.

finality-aleph/src/network.rs
mike1729: Fixed.

finality-aleph/src/network.rs
mike1729: Worked in previous version and works in the new one.

finality-aleph/src/network.rs
```diff
@@ -404,16 +525,40 @@ impl<D: Clone + Encode + Decode, B: BlockT + 'static, N: Network<B> + Clone>
...
         let mut sessions = self.sessions.lock();
         if let Some(session_data) = sessions.get_mut(&session_id) {
             if session_data.status == SessionStatus::InProgress {
-                if let Err(e) = session_data.data_for_user.unbounded_send(data) {
-                    //TODO: need to write some logic on when an session should be terminated and make sure
-                    // that there are no issues with synchronization when terminating.
-                    session_data.status = SessionStatus::Terminated;
-                    debug!(target: "afa", "Error {:?} when passing a message event to session {:?}.", e, session_id);
+                if session_data.messages.contains(&message) {
+                    debug!(target: "afa", "Received data with old index in session {:?}.", session_id);
```
DamianStraszak, line 545: Received duplicate message. Can also make this TRACE already :)

finality-aleph/src/network.rs
```diff
@@ -404,16 +525,40 @@ impl<D: Clone + Encode + Decode, B: BlockT + 'static, N: Network<B> + Clone>
...
+                    return;
+                } else {
+                    session_data.messages.put(message, ());
                 }
+                if let Recipient::Target(node_id) = recipient {
+                    if node_id == session_data.auth_data.node_id {
+                        if let Err(e) = session_data.data_for_user.unbounded_send(data) {
```
DamianStraszak, line 552: I might be wrong, but for Recipient::All we never place the message in the data_for_user channel?

finality-aleph/src/network.rs
```diff
@@ -233,24 +271,51 @@ enum MetaMessage {
...
     Meta(MetaMessage),
-    Data(SessionId, D),
+    Data(DataMessage<D>),
 }
 
 struct SessionData<D: Clone + Encode + Decode> {
     pub(crate) data_for_user: mpsc::UnboundedSender<D>,
     pub(crate) status: SessionStatus,
     pub(crate) keychain: KeyBox,
+    messages: LruCache<DataMessage<D>, ()>,
```
DamianStraszak, line 309: I would either make this hold hashes of DataMessage's, or make the CACHE_SIZE something like 10^4 rather than 10^6. Actually, even for hashes I would reduce the size. The smaller the cache is, the more cache friendly the cache :)

finality-aleph/src/network.rs
```diff
@@ -212,6 +227,23 @@ impl Peers {
...
         self.to_peer.get(&session_id)?.get(&node_id)
     }
+
+    fn get_rand(
+        &self,
+        session_id: SessionId,
+        amount: usize,
+        offset: usize,
+    ) -> impl Iterator<Item = &PeerId> {
+        let skip = if let Some(n_peers) = self.to_peer.get(&session_id).map(|hm| hm.len()) {
```
timorl, line 237: I might be confused, but is this really needed if you use `cycle` below? Or are you worried about the performance of `skip`?

finality-aleph/src/network.rs
```diff
@@ -233,24 +271,51 @@ enum MetaMessage {
...
+    session_id: SessionId,
+    index: MessageIndex,
+    data: D,
+    recipient: Recipient<NodeIndex>,
+}
+
+impl<D: Encode> Hash for DataMessage<D> {
+    fn hash<H: StdHasher>(&self, state: &mut H) {
+        self.index.hash(state);
+        self.data.encode().hash(state);
```
timorl, line 287: Unlikely to surface, but I think this is technically a bug – if we have two nodes that want to send the same data, they  might assign it the same index and then one of the packets might be discarded. Since we cannot depend on the source of the data anywhere this can only be a problem if both packets are being sent to different destinations. Adding `recipient` to the hash would solve the problem.

It's quite possible this wouldn't be a problem for AlephBFT in practice, bu it's a strange limitation to have to remember.

timorl: Wait, no, this is not a bug, because we already assume any packet might not reach it's destination for random reasons. Leaving above comment, but just informational.

finality-aleph/src/network.rs
```diff
@@ -233,24 +271,51 @@ enum MetaMessage {
...
+
+#[derive(Clone, Encode, Decode)]
+struct DataMessage<D> {
+    session_id: SessionId,
+    index: MessageIndex,
+    data: D,
+    recipient: Recipient<NodeIndex>,
+}
+
+impl<D: Encode> Hash for DataMessage<D> {
```
timorl, line 284: Maybe derive these (`Hash` and `PartialEq`) using the `derivative` crate, like in AlephBTF?

timorl (approved):
Only minor comments that shouldn't block merging as far as I'm concerned.

finality-aleph/src/network.rs
mike1729: I worry about the performance of `cycle`. It clones the iterator, so I'm afraid that it wouldn't jump straight to index `skip` but actually make all the skipped clones.

finality-aleph/src/network.rs
mike1729: Sure thing!

finality-aleph/src/network.rs
mike1729: You might but you are not!

finality-aleph/src/network.rs
mike1729: Reducing the size.

finality-aleph/src/network.rs
mike1729: Huh, I missed that, adding!

mike1729: Unfortunately, it doesn't work:( as D doesn't implement std::hash::Hash.

timorl: Aw. :< Oh well.

DamianStraszak (approved):

PR #75, opened 2021-06-29
Justification handler

kostekIV:


finality-aleph/src/justification.rs
```diff
@@ -1,70 +1,121 @@
...
+        Ok(aleph_justification)
     }
 }
 
-// For now it is glorified proxy channel not doing anything useful
-pub struct JustificationHandler<Block: BlockT> {
+pub struct JustificationHandler<Block: BlockT, N: network::Network<Block> + 'static> {
     finalization_proposals_tx: mpsc::UnboundedSender<JustificationNotification<Block>>,
     justification_rx: mpsc::UnboundedReceiver<JustificationNotification<Block>>,
+    sessions: Arc<HashMap<u32, SessionInstance<Block>>>,
```
michalseweryn, line 49: I believe we should use `Arc<Mutex<_>>` for sessions here and everywhere else. Otherwise there is no way to modify the hashmap in `member.rs`.

PR #76, opened 2021-06-30
Fix rotating party.

kostekIV:


finality-aleph/src/party.rs
```diff
@@ -258,20 +262,27 @@ where
...
         };
 
-        let current_stop_h = session.stop_h;
         let client = self.client.clone();
         let finalization_proposals_rx = &mut self.finalization_proposals_rx;
         let proposals_task = {
             let client = self.client.clone();
             async move {
-                while client.info().finalized_number < current_stop_h {
+                loop {
```
kostekIV, line 285: I am not sure about this but it looks fine to just loop here

finality-aleph/src/party.rs
```diff
@@ -258,20 +262,27 @@ where
             .runtime_api()
             .current_session(&BlockId::Number(prev_block_number))
         {
-            Ok(session) => session,
+            Ok(session) => {
+                self.sessions.insert(
+                    session_id,
+                    SessionInstance {
+                        session: session.clone(),
+                    },
```
michalseweryn, line 270: I was wondering if we could replace the value type of the hashmap to `Session<_, _>` and get rid of the `SessionInstance` type completely?

finality-aleph/src/party.rs
```diff
@@ -305,7 +315,6 @@ where
                 .collect();
         // wait for the first task returning true
         tasks.next().await;
-        //tasks.filter(|b| std::future::ready(*b)).next().await;
```
michalseweryn, line 308: I think we actually want `tasks.filter(|b| std::future::ready(*b)).next()` instead of `tasks.next()`. If we do not participate in the consensus, the `session_task` will complete immediately returning false.

michalseweryn (changes requested):

finality-aleph/src/party.rs
michalseweryn: We need `while` here. Otherwise, if we do not participate in the consensus (i.e. we are not an authority), we wouldn't end the task, which would prevent us from moving on to the next session

michalseweryn (approved):

DamianStraszak (approved):

PR #81, opened 2021-07-11
Justifications + debugging for testnet

DamianStraszak:


bin/runtime/src/lib.rs
```diff
@@ -260,7 +260,7 @@ impl_opaque_keys! {
 }
 
 parameter_types! {
-    pub const Period: u32 = 50;
+    pub const Period: u32 = 5;
```
maciejnems, line 263: Do we want period in final version to be so small?

finality-aleph/Cargo.toml
```diff
@@ -21,7 +21,10 @@ prometheus-endpoint = { package = "substrate-prometheus-endpoint", git = "https:
 sp-keystore = { git = "https://github.com/paritytech/substrate.git", rev = "1fa8cf7"}
 sc-network = { git = "https://github.com/paritytech/substrate.git", rev = "1fa8cf7"}
 sc-telemetry = { git = "https://github.com/paritytech/substrate.git", rev = "1fa8cf7"}
-aleph-bft = { git = "https://github.com/Cardinal-Cryptography/AlephBFT.git", rev = "3e2a0d0" }
+
+aleph-bft = { git = "https://github.com/DamianStraszak/rush.git", branch = "testnet-fixes" }
+# aleph-bft = {path = "../../rush"}
```
maciejnems: I guess it is here for simplicity to test locally, but maybe we shouldn't keep it on main?

finality-aleph/src/finalization.rs
```diff
@@ -195,7 +192,7 @@ impl<
     pub(crate) fn new(network: RmcNetwork<B>, keychain: &'a MK) -> Self {
         let (messages_for_rmc, messages_from_network) = mpsc::unbounded();
         let (messages_for_network, messages_from_rmc) = mpsc::unbounded();
-        let scheduler = DoublingDelayScheduler::new(Duration::from_millis(10));
+        let scheduler = DoublingDelayScheduler::new(Duration::from_millis(500));
```
maciejnems, line 195: Maybe this could be some parameter of finalization or BlockSignatureAggregator?

finality-aleph/src/finalization.rs
```diff
@@ -254,20 +250,24 @@ impl<
             loop {
                 tokio::select! {
                     multisigned_hash = self.rmc.next_multisigned_hash() => {
-                            let unchecked = multisigned_hash.into_unchecked();
+
+                            //let unchecked = multisigned_hash.into_unchecked().clone();
+                            //debug!(target: "afa", "New multisigned_hash {:?}.", unchecked);
```
maciejnems, line 255: Shouldn't we log this information here?

michalseweryn (approved):
Looks good to me, just some minor improvment ideas :)

finality-aleph/src/finalization.rs
michalseweryn: Also, can we not have a commented out code? ;)

bin/runtime/src/lib.rs
DamianStraszak: Honestly, I don't think it matters -- this is typically run only for testing purposes, where you want to have short sessions, to catch bugs with session change. For actual testnet, this is changed anyway, and we will soon have this field as part of a config file.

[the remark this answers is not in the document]
DamianStraszak: Yeah, to me this constant at this place is only set for testing purposes, so `5` works great :) I think this and other constants will soon go to a config file which will resolve this issue.

finality-aleph/src/finalization.rs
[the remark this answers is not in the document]
DamianStraszak: Hard to say as this is only a fallback mechanism. I think of this in the following way: if you join time `T` after a relevant event happened (let's say constructing a multisig) then after another at most `T` time you receive the message -- which I would say is not that bad.

finality-aleph/src/finalization.rs
```diff
@@ -216,25 +213,24 @@ impl<
...
 
     pub(crate) async fn start_aggregation(&mut self, hash: B::Hash) {
+        debug!(target: "afa", "Started aggregation for block hash {:?}", hash);
         if !self.started_hashes.insert(hash) {
             return;
         }
         self.hash_queue.push_back(hash);
-        self.rmc.start_rmc(SignableHash { hash }).await
+        self.rmc.start_rmc(SignableHash { hash }).await;
+        debug!(target: "afa", "Started rmc for block hash {:?}", hash);
```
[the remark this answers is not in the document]
DamianStraszak: Actually I have added the second one, because I saw `.await` and wanted to catch whether something bad happens there, but you are right, enough to have one debug here.

finality-aleph/src/finalization.rs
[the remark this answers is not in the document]
michalseweryn: You receive the message assuming that I will still be running the rmc instance

michalseweryn (changes requested):
Nice work! There are a few things to fix and some place for small improvements. Also, since you are refactoring the consensus party, can you move network spawning to the `run_consensus_party`, so that we would spawn it in the same place as justification handler and only pass the session manager to the consensus party?

finality-aleph/src/party.rs
```diff
@@ -98,41 +139,109 @@ where
...
+            }
+        }
+    }
 }
 
 /// If we are on the authority list for the given session, runs an
 /// AlephBFT task and returns `true` upon completion. Otherwise, immediately returns `false`.
-async fn maybe_run_session_as_authority<B, C, BE, SC>(
-    authority: AuthorityId,
+#[allow(clippy::too_many_arguments)]
```
timorl, line 225: :<

I don't see how to fix this easily, but this almost surely needs at least a refactor in the future. With the zero bug policy I'm not convinced this can be ensured, so could you try fixing this somehow?

finality-aleph/src/party.rs
```diff
@@ -216,36 +351,39 @@ where
     SC: SelectChain<B> + 'static,
     NumberFor<B>: From<u32>,
 {
+    #[allow(clippy::too_many_arguments)]
```
timorl, line 354: Ouch, same comment as before.

finality-aleph/src/network.rs
```diff
@@ -87,8 +75,11 @@ pub trait Network<B: BlockT>: Clone + Send + Sync + 'static {
     /// TODO: figure out what does this actually do...
     fn remove_set_reserved(&self, who: PeerId, protocol: Cow<'static, str>);
 
-    // The PeerId of this node.
+    /// The PeerId of this node.
     fn peer_id(&self) -> PeerId;
+
+    /// Request the justification for the given block
+    fn request_justification(&self, hash: &B::Hash, number: NumberFor<B>);
```
timorl, line 82: This seems like a strange place to require this function. Cannot we send it as one of the enum variants in `DataNetwork`? Or does this work in a strange way that is very entangled with substrate?

timorl (changes requested):
Mostly nits, but also some important questions (especially the network one).

finality-aleph/src/finalization.rs
DamianStraszak: Fixed.

finality-aleph/src/justification.rs
```diff
@@ -1,70 +1,122 @@
...
     finalization_proposals_tx: mpsc::UnboundedSender<JustificationNotification<Block>>,
     justification_rx: mpsc::UnboundedReceiver<JustificationNotification<Block>>,
+    sessions: Arc<Mutex<SessionMap<Block>>>,
+    auth_keystore: AuthorityKeystore,
+    session_period: u32,
+    network: N,
 }
 
-impl<Block: BlockT> JustificationHandler<Block> {
+impl<Block: BlockT, N: network::Network<Block> + 'static> JustificationHandler<Block, N>
```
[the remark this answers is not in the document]
DamianStraszak: This is now outdated but was definitely correct. Now JustificationHandler has an importatn role: verifying justifications from 2 different sources + finalizing blocks + requesting justifications.

finality-aleph/src/party.rs
michalseweryn: The typical solution is to intruduce a `RunSessionParams` struct. This is technically equivalent, but is usually clearer because fields of a struct have names unlike function parameters

finality-aleph/src/network.rs
michalseweryn: The `Network` trait is something that should be discussed in more detail. There are unused methods and methods which we don't know what are supposed to do?

finality-aleph/src/justification.rs
```diff
@@ -1,70 +1,220 @@
...
+                        self.handle_justification_notification(notification, keybox,  last_finalized_number, stop_h.into());
+                    } else {
+                        error!(target: "afa", "Got a None from justification aggregator stream.");
+                    }
+                }
+                _ = &mut ticker_fut => {
+                    continue;
+                }
+            }
+            self.handle_requests(stop_h.into());
```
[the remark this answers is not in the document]
DamianStraszak: Yeah, I was thinking about that and the short answer is: I'm pretty sure it is enough for this to work "reasonably" well.

By requesting `last_finalized_number + 1` (and having some delays between consecutive requests) we will be catching up very slowly so that might not be the best idea. The current vesion is minimal and should be enough to prevent disasters. If the session time is say ~10mins, then I would say it is reasonable.

finality-aleph/src/justification.rs
```diff
@@ -1,70 +1,220 @@
...
-        Self {
-            signature: auth_crypto_store.sign(&hash.encode()[..]),
-        }
+    pub(crate) fn new<Block: BlockT>(signature: SignatureSet<Signature>) -> Self {
+        Self { signature }
     }
 
-    pub(crate) fn _decode_and_verify<Block: BlockT>(
-        justification: &[u8],
+    pub(crate) fn verify<
```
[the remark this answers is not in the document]
DamianStraszak: Ok, corrected.

finality-aleph/src/justification.rs
```diff
@@ -1,70 +1,220 @@
...
+                    debug!(target: "afa", "Not finalizing block {:?}. Last finalized {:?}, stop_h {:?}", num, last_finalized, stop_h);
+                }
+            }
+            Err(err) => {
+                error!(target: "afa", "Error when verifying justification {:?} for block {:?} {:?}", err, num, block_hash);
+            }
+        }
+    }
+
+    pub(crate) fn handle_requests(&mut self, num: NumberFor<B>) {
```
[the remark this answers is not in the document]
DamianStraszak: Yes, done.

finality-aleph/src/justification.rs
```diff
@@ -1,70 +1,220 @@
...
+                maybe_notification= self.authority_justification_rx.next() => {
+                    if let Some(notification) = maybe_notification {
+                        debug!(target: "afa", "Got justification (local aggregator)");
+                        self.handle_justification_notification(notification, keybox,  last_finalized_number, stop_h.into());
+                    } else {
+                        error!(target: "afa", "Got a None from justification aggregator stream.");
+                    }
+                }
+                _ = &mut ticker_fut => {
+                    continue;
```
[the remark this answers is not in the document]
DamianStraszak: Huh, you are right of course! Changed to "nothing" instead of `continue`

finality-aleph/src/network.rs
DamianStraszak: This is part of some very specific API that substrate's NetworkService provides. There are probably alternatives on how to do that, but I don't see how to remove it from this trait.

finality-aleph/src/party.rs
```diff
@@ -165,44 +200,53 @@ where
...
+                    debug!(target: "afa", "the stream of multisigned hashes has ended");
                     break;
                 }
             },
-            else => {
-                debug!(target: "afa", "finished party {:?} with finalized block at {:?}", session_id.0, client.info().finalized_number);
-                break;
-            }
         }
     }
```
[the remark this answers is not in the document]
DamianStraszak: This changed a little bit in the new version. A new file `aggregator.rs` contains now some of the logic and finalization has little stuff left. Not sure if this has been addressed?

finality-aleph/src/network/tests.rs
```diff
@@ -393,11 +389,10 @@ async fn requests_authentication_from_unauthenticated() {
         .next()
         .await
         .expect("got auth request");
-    println!("ta-daa");
     assert_eq!(peer_id, bob_peer_id);
     assert_eq!(protocol, PROTOCOL_NAME);
     let message =
-        InternalMessage::<MockData>::decode_all(message.as_slice()).expect("a correct message");
+        InternalMessage::<MockData>::decode(&mut message.as_slice()).expect("a correct message");
```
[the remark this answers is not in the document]
DamianStraszak: Corrected.

finality-aleph/src/justification.rs
```diff
@@ -1,70 +1,220 @@
...
+            if keybox.is_none() {
+                error!(target: "afa", "Cannot fetch the keybox for session {:?}. Waiting 500ms and will try again ...", current_session);
+                Delay::new(Duration::from_millis(500)).await;
+                continue;
+            }
+            let keybox = keybox.expect("We loop until this is some.");
+            let mut ticker_fut = Delay::new(Duration::from_millis(1000));
+            tokio::select! {
+                maybe_notification = self.import_justification_rx.next() => {
+                    if let Some(notification) = maybe_notification {
```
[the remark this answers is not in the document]
DamianStraszak: Right, added log and `return`

finality-aleph/src/justification.rs
```diff
@@ -1,70 +1,220 @@
...
+                        debug!(target: "afa", "Got justification (block import)");
+                        self.handle_justification_notification(notification, keybox, last_finalized_number, stop_h.into());
+                    }
+                }
+                maybe_notification= self.authority_justification_rx.next() => {
+                    if let Some(notification) = maybe_notification {
+                        debug!(target: "afa", "Got justification (local aggregator)");
+                        self.handle_justification_notification(notification, keybox,  last_finalized_number, stop_h.into());
+                    } else {
+                        error!(target: "afa", "Got a None from justification aggregator stream.");
```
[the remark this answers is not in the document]
DamianStraszak: Yes, corrected.

finality-aleph/src/party.rs
```diff
@@ -98,41 +139,109 @@ where
...
+            _ = &mut exit_rx => {
+                debug!(target: "afa", "Aggregator received exit signal. Terminating.");
+                return;
+            }
+        }
+    }
 }
 
 /// If we are on the authority list for the given session, runs an
 /// AlephBFT task and returns `true` upon completion. Otherwise, immediately returns `false`.
```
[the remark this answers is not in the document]
DamianStraszak: Removed

finality-aleph/src/party.rs
```diff
@@ -142,68 +251,94 @@ where
...
+            member.run_session(aleph_network, exit_member_rx).await;
+            debug!(target: "afa", "Member task stopped for {:?}", session_id.0);
         }
     };
-    spawn_handle.0.spawn("aleph/consensus_session", aleph_task);
 
-    debug!(target: "afa", "Consensus party #{} has started.", session_id.0);
+    let forwarder_task = async move {
+        debug!(target: "afa", "Running the forwarder task for {:?}", session_id.0);
+        forwarder.await;
```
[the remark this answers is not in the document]
DamianStraszak: Ok, I changed it. Up to one `pin_mut!` what you wrote was correct :)

finality-aleph/src/finalization.rs
```diff
@@ -210,31 +152,29 @@ impl<
...
-    }
-
-    pub(crate) async fn finish(&mut self) {
-        self.finished = true;
+        self.rmc.start_rmc(SignableHash { hash }).await;
     }
 
-    pub(crate) fn is_finished(&self) -> bool {
-        self.finished
+    pub(crate) async fn notify_last_hash(&mut self) {
```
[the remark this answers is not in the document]
DamianStraszak: Nice, you passed the perception test for reviewers!

finality-aleph/src/finalization.rs
```diff
@@ -246,28 +186,32 @@ impl<
...
             loop {
                 tokio::select! {
                     multisigned_hash = self.rmc.next_multisigned_hash() => {
-                            let unchecked = multisigned_hash.into_unchecked();
+                        let hash = multisigned_hash.as_signable().hash;
+                        let unchecked = multisigned_hash.into_unchecked().signature();
+                            debug!(target: "afa", "New multisigned_hash {:?}.", unchecked);
                             self.signatures
-                                .insert(unchecked.signable.hash, unchecked.signature);
+                                .insert(hash, unchecked );
```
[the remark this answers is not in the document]
DamianStraszak: Yeah, fixed. What's inside macros is not affected by `fmt`.

finality-aleph/src/finalization.rs
```diff
@@ -366,47 +309,4 @@ mod tests {
             }
         }
     }
-
-    #[tokio::test]
-    async fn chain_extension_scenario() {
```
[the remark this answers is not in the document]
DamianStraszak: Yeah, but it was testing something I removed actually :)

finality-aleph/src/party.rs
mike1729: Here the solution is to remove `authority` param as it may be read from `auth_keystore`.

finality-aleph/src/finalization.rs
DamianStraszak: This is a very low-level parameter, which we most likely don't want to play with. So I would leave it as it is...

finality-aleph/src/justification.rs
```diff
@@ -1,70 +1,220 @@
...
+{
+    authority_justification_rx: mpsc::UnboundedReceiver<JustificationNotification<B>>,
+    import_justification_rx: mpsc::UnboundedReceiver<JustificationNotification<B>>,
+    sessions: Arc<Mutex<SessionMap<B>>>,
+    auth_keystore: AuthorityKeystore,
+    session_period: u32,
+    network: N,
+    client: Arc<C>,
+    last_request: Instant,
+    last_finalization: Instant,
```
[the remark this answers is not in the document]
DamianStraszak: I changed these to `last_request_time` and `last_finalization_time` and I think it's better to leave them as they are. 1) these names are descriptive thus easy to understand when reading, 2) I have changed the condition for requesting justifications, so that the `last_event_time` would not be enough and I expect there will be likely more experimentation with this condition.

finality-aleph/src/justification.rs
```diff
@@ -1,70 +1,220 @@
...
+            let current_session = self.current_session_id(last_finalized_number);
+            let stop_h = (current_session + 1) * self.session_period - 1;
+            let keybox = self.session_keybox(current_session);
+            if keybox.is_none() {
+                error!(target: "afa", "Cannot fetch the keybox for session {:?}. Waiting 500ms and will try again ...", current_session);
+                Delay::new(Duration::from_millis(500)).await;
+                continue;
+            }
+            let keybox = keybox.expect("We loop until this is some.");
+            let mut ticker_fut = Delay::new(Duration::from_millis(1000));
```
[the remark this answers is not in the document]
DamianStraszak: Hmm, we can I guess, but then the version with select requires you to play with Either::Left/Right which I think makes the code obscure. If there is a way to use timeout without this syntax than I would be happy to change this :P

finality-aleph/src/justification.rs
```diff
@@ -1,70 +1,220 @@
...
-            self.finalization_proposals_tx
-                .unbounded_send(notification)
-                .expect("Notification should succeed");
+        loop {
+            let last_finalized_number = self.client.info().finalized_number;
+            let current_session = self.current_session_id(last_finalized_number);
+            let stop_h = (current_session + 1) * self.session_period - 1;
+            let keybox = self.session_keybox(current_session);
+            if keybox.is_none() {
+                error!(target: "afa", "Cannot fetch the keybox for session {:?}. Waiting 500ms and will try again ...", current_session);
```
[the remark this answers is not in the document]
DamianStraszak: Actually, this keybox not being there once or twice is not even something unexpected, so made it into `debug` now.

finality-aleph/src/party.rs
DamianStraszak: Yeah, I'm aware of this fix, but I don't see how this makes the code better (we can get rid of `#[allow(clippy::too_many_arguments)]` but let's not count that :P) -- there would be then a few new "Params"-type structs defined for the sole purpose of silencing clippy.  Reading the code and (to a lesser extent) writing is then harder, not simpler.

finality-aleph/src/party.rs
DamianStraszak: Yeah, I think @mike1729 is correct, but then I would have to place this low level line `let auth = SyncCryptoStore::ed25519_public_keys(&*keystore, finality_aleph::KEY_TYPE)[0];` somewhere here -- which well I can do, but if it's only playing around this arbitrary limit of 7 parameters then I have my doubts :P.

finality-aleph/src/import.rs
```diff
@@ -163,21 +169,16 @@ where
         number: NumberFor<Block>,
         justification: Justification,
     ) -> Result<(), Self::Error> {
+        log::debug!(target: "afa", "import_justification called on {:?}", justification);
         match self.send_justification(hash, number, justification) {
```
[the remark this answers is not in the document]
DamianStraszak: Right, corrected.

finality-aleph/src/party.rs
mike1729: No, just `let authority = auth_keystore.authority()` :)

finality-aleph/src/network.rs
michalseweryn: Well, the only interface needed by `ConsensusNetwork` are the methods
* `async fn next_event(&self) -> Event<D>` where `Event<D>` is an enum with variants `PeerConnected(PeerId)`, `PeerDisconnected(PeerId)` and `IncomingMessage(InternalMessage<D>)`,
* `fn send_message(&self, peer_id: PeerId, msg: InternalMessage<D>)`,
* and maybe `fn request_justification(&self, hash: &B::Hash, number: NumberFor<B>)`, but this can be just another variant `InternalMessage` which is handled specially in the implementation.

This require writing a (slightly) non-trivial implementation of the trait, and maybe even having an adapter, but this seems more reasonable than having a trait which precisely reflects a substrate struct

finality-aleph/src/justification.rs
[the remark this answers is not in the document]
michalseweryn: Well, you can use `futures::stream::select(self.import_justification_rx, self.authority_justification_rx).next()`. The select on streams expect that the streams have the same type and returns a stream of elements of that type. This however loses the information about the source of a notifification (import/aggregator). We have the following options:

* accept that this bit of information is lost (it is used only in logs),
* add a field in the notification which tells the source of the notification,
* instead of calling select on receivers, call it on `self.import_justification_rx.inspect(...)` so that we log an appropriate message before handling the justification

finality-aleph/src/party.rs
DamianStraszak: Ok, sorry for that, you are right! Fixed and removed clippy silencer.

finality-aleph/src/justification.rs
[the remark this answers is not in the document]
DamianStraszak: Ok, this is addressed, I think :)

finality-aleph/src/network.rs
timorl: I'm gonna assume this will be dealt with as part of the network refactor task and I'm resolving this.

finality-aleph/src/party.rs
DamianStraszak: There is now a task for this in the backlog.

finality-aleph/src/import.rs
```diff
@@ -116,17 +126,17 @@ where
...
+        let post_hash = block.post_hash();
         if let Some(m) = &self.metrics {
-            m.report_block(hash, Instant::now(), "importing");
+            m.report_block(post_hash, Instant::now(), "importing");
         };
 
-        log::debug!(target: "afa", "Importing block #{:?}", number);
+        let justifications = block.justifications.take();
+
+        log::debug!(target: "afa", "Importing block {:?} {:?} {:?}", number, block.header.hash(), block.post_hash());
```
timorl, line 136: We almost everywhere `use log::{what, ever}` and then call the macros without the namespace, could you do this here too, for consistency?

timorl: There are a couple places this happens, just `grep -Rn log::`.

finality-aleph/src/justification.rs
```diff
@@ -1,70 +1,233 @@
...
+use sp_api::{BlockId, BlockT, NumberFor};
+use sp_runtime::traits::Header;
+use std::{
+    collections::HashMap,
+    marker::PhantomData,
+    sync::Arc,
+    time::{Duration, Instant},
+};
+use tokio::time::timeout;
+//use tokio::stream::StreamExt;
```
timorl, line 22: Commented code.

finality-aleph/src/party.rs
```diff
@@ -70,194 +101,276 @@ fn get_node_index(authorities: &[AuthorityId], my_id: &AuthorityId) -> Option<No
...
+                    // The unwrap might actually fail if data availability is not implemented correctly.
+                    let notification = JustificationNotification {
+                        justification: AlephJustification::new::<B>(multisignature),
+                        hash,
+                        number
+                    };
+                    if let Err(e) = justification_tx.unbounded_send(notification)  {
+                        error!(target: "afa", "Issue with sending justification from Aggregator to JustificationHandler {:?}.", e);
+                    }
+                    //finalize_block_as_authority(client.clone(), hash, multisignature);
```
timorl, line 217: Commented code.

finality-aleph/src/party.rs
```diff
@@ -70,194 +101,276 @@ fn get_node_index(authorities: &[AuthorityId], my_id: &AuthorityId) -> Option<No
...
+                    }
+                } else {
+                    debug!(target: "afa", "Batches ended in aggregator. Terminating.");
+                    return;
+                }
+            }
+            multisigned_hash = aggregator.next_multisigned_hash() => {
+                if let Some((hash, multisignature)) = multisigned_hash {
+                    let number = client.number(hash).unwrap().unwrap();
+                    // The unwrap might actually fail if data availability is not implemented correctly.
```
timorl, line 208: And it isn't right now, right? Not a blocker, just making sure I understand.

Although technically against the Zero Bug Policy, but I'm more and more convinced that one is just pure fiction...

timorl (changes requested):
Mostly nits. I really hope we won't need such enormous PRs doing many things at once in the future. :/

finality-aleph/src/party.rs
DamianStraszak: Right now it is not at all implemented, Maciek working on it.

finality-aleph/src/import.rs
DamianStraszak: Thanks, seems I caught a couple of them but not all. Now it is fixed everywhere.

finality-aleph/src/justification.rs
DamianStraszak: Removed.

finality-aleph/src/party.rs
DamianStraszak: Removed.

finality-aleph/src/justification.rs
```diff
@@ -128,8 +117,23 @@ where
         ) {
             if num > last_finalized && num <= stop_h {
                 debug!(target: "afa", "Finalizing block {:?} {:?}", num, block_hash);
-                self.last_finalization_time = Instant::now();
-                self.finalize(num, block_hash, notification.justification);
+
+                debug!(target: "afa", "Finalizing block {:?}", num);
```
timorl, line 121: Repeated log.

timorl (approved):
One nit, in general nice.

michalseweryn (approved):

finality-aleph/src/justification.rs
DamianStraszak: Thanks, fixed.

PR #82, opened 2021-07-15
fixes for the CI pipeline

fbielejec:


.github/workflows/ci.yml
```diff
@@ -1,109 +1,52 @@
...
-        with:
-          command: check
-      - name: test
-        if: ${{ matrix.toolchain == 'stable' }}
-        uses: actions-rs/cargo@v1
+      - run: rustup target add wasm32-unknown-unknown --toolchain nightly-2021-07-05
+      - uses: actions-rs/cargo@v1
         with:
           command: test
-  lint:
```
michalseweryn, line 80: I don't understand everything in the configuration, but does it turn the lint check off?

.github/workflows/ci.yml
```diff
@@ -1,109 +1,52 @@
...
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: ${{ matrix.toolchain }}
+          toolchain: nightly-2021-07-05
```
timorl, line 47: I'm a bit uneasy about having this as a fixed version. Could we replace this with the equivalent of "use latest", or is that either technically unfeasible or just unwise for stability reasons?

.github/workflows/ci.yml
```diff
@@ -1,109 +1,52 @@
...
-    runs-on: ${{ matrix.os }}
-    needs: lint
+
+  test:
+    name: Run unit test suite
+    runs-on: ubuntu-latest
     steps:
-      - name: checkout the source code
-        uses: actions/checkout@v2
+      - uses: actions/checkout@v2
```
timorl, line 34: There is quite a bit duplication between this and the previous section. I'm not particularly familiar with CI configs – can this be somehow unified?

timorl (changes requested):

.github/workflows/ci.yml
fbielejec: there is a bug atm: https://github.community/t/failed-sending-data-to-the-peer-connection-died-tried-5-times-before-giving-up/189130 
But maybe having a fixed nightly is not a bad idea anyway (builds are reproducible)?

.github/workflows/ci.yml
fbielejec: correct, but if we want we can bring back lints and fmt?

.github/workflows/ci.yml
fbielejec: It's mostly to have these jobs running in parallel - but good point, I''ll see if there's any mechanism to unify these steps in actions, I'm less familiar with them than with CircleCI, but it should exists

.github/workflows/ci.yml
timorl: Yup, reproducibility is a very good argument, especially if nightly is sufficiently unstable for such bugs. I would say we still want to have some procedures around bumping the version once in a while, could you write this down somewhere so we don't forget? I'm trying to avoid a situation in which it turns out we are using a year old version of "nightly" that is actually behind the stable compiler. :P

timorl (approved):
Running sequentially is probably fine. Generally <3

.github/workflows/ci.yml
michalseweryn: I do want them :P

fbielejec: you want it, you got it :)

.github/workflows/ci.yml
fbielejec: any ideas where/how, is wiki a good place?

timorl: Yeah, I think somewhere there, but I'm not sure where exactly. For now at top level of "procedures" I guess – we are unlikely to forget about it then.

PR #85, opened 2021-07-19
Make constants `Period` and `MillisecsPerBlock` specified by environment variables

michalseweryn:
The parameters `SessionPeriod` (formerly `Period`) and `MillisecsPerBlock` no longer have to be known at compile time. Instead, you specify them if environment variables `SESSION_PERIOD` and `MILLISECS_PER_BLOCK`, respectively. If a variable is not set, a default value is used (5 for `SessionPeriod` and 4000 for `MilliSecsPerBlock`).

This PR also fixes the bug with incorrect generation of local keys. The bug was cause by the fact that when keys were supposed to be generated (and saved in the `/tmp/authorities_keys` file), an instance of `ChainSpec` is created. However, constructing a `ChainSpec` involves reading keys from `/tmp/authorities_keys`, and panics if that file does not exist. The bug is fixed by returning an empty list of keys instead of panicking in the described case. This simple fix is not the cleanest solution, but is simple and avoids repetition of code

### DIFF
--- a/bin/node/src/chain_spec.rs
+++ b/bin/node/src/chain_spec.rs
@@ -39,9 +39,12 @@ where
 }
 
 fn read_keys(n_members: usize) -> (Vec<AuraId>, Vec<AlephId>) {
-    let auth_keys = std::fs::read_to_string(KEY_PATH).expect("keys were not generated");
     let auth_keys: HashMap<u32, Vec<[u8; 32]>> =
-        serde_json::from_str(&auth_keys).expect("should contain list of keys");
+        if let Ok(auth_keys) = std::fs::read_to_string(KEY_PATH) {
+            serde_json::from_str(&auth_keys).expect("should contain list of keys")
+        } else {
+            return Default::default()
+        };
 
     let aura_keys: Vec<_> = auth_keys
         .get(&key_types::AURA.into())
--- a/bin/node/src/chain_spec.rs
+++ b/bin/node/src/chain_spec.rs
@@ -15,9 +15,6 @@ use sp_core::{ed25519, sr25519, Pair, Public};
 use sp_runtime::traits::{IdentifyAccount, Verify};
 use std::{env::VarError, fmt::Display, str::FromStr};
 
-const SESSION_PERIOD_ENV_VAR: &str = "SESSION_PERIOD";
-const MILLISECS_PER_BLOCK_ENV_VAR: &str = "MILLISECS_PER_BLOCK";
-
 const FAUCET_HASH: [u8; 32] =
     hex!("eaefd9d9b42915bda608154f17bb03e407cbf244318a0499912c2fb1cd879b74");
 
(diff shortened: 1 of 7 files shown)


bin/node/src/chain_spec.rs
```diff
@@ -38,30 +47,85 @@ where
...
+}
+
+// TODO: rename to AlephEnvVarParams
+#[derive(Clone, Copy)]
+struct EnvironmentVariables {
+    session_period: u32,
+    millisecs_per_block: u64,
+}
+
+impl EnvironmentVariables {
```
fbielejec, line 102: name suggestion: could we cal this struct Config

bin/node/src/chain_spec.rs
```diff
@@ -38,30 +47,85 @@ where
...
+            aura_key,
+            aleph_key,
+            account_id,
+        })
+        .collect()
+}
+
+// TODO: rename to AlephEnvVarParams
+#[derive(Clone, Copy)]
+struct EnvironmentVariables {
```
fbielejec, line 97: naming suggestion: could we call this Config?

bin/node/src/chain_spec.rs
```diff
@@ -11,6 +13,13 @@ use sp_application_crypto::key_types;
 use sp_consensus_aura::sr25519::AuthorityId as AuraId;
 use sp_core::{ed25519, sr25519, Pair, Public};
 use sp_runtime::traits::{IdentifyAccount, Verify};
+use std::{env::VarError, fmt::Display, str::FromStr};
+
+const SESSION_PERIOD_ENV_VAR: &str = "SESSION_PERIOD";
+const MILLISECS_PER_BLOCK_ENV_VAR: &str = "MILLISECS_PER_BLOCK";
+
+const FAUCET_HASH: [u8; 32] =
+    hex!("eaefd9d9b42915bda608154f17bb03e407cbf244318a0499912c2fb1cd879b74");
```
fbielejec, line 22: is this always hardcoded or could be a configurable variable in itself?

bin/node/src/chain_spec.rs
```diff
@@ -38,30 +47,85 @@ where
...
+#[derive(Clone, Copy)]
+struct EnvironmentVariables {
+    session_period: u32,
+    millisecs_per_block: u64,
+}
+
+impl EnvironmentVariables {
+    fn fetch() -> Result<Self, String> {
+        Ok(Self {
+            session_period: Self::fetch_var_or(SESSION_PERIOD_ENV_VAR, DEFAULT_SESSION_PERIOD)?,
```
fbielejec, line 105: YMMV but if the defaults are only read here, there might be no point in storing them as const, just hardcode them here directly?

fbielejec, line 105: YMMV but if the name sof the vars are read only here, there might be no point in storing them as const, just hardcode them here directly?

DamianStraszak (approved):
Code looks good to me. Tested, and it works :)

fbielejec (commented):
see comments

fbielejec:
I tired running the keygen command:
```
./aleph-node dev-keys --base-path data --chain testnet1 --key-types aura alp0
```

and I get:

```
thread 'main' panicked at 'should contain list of keys: Error("EOF while parsing a value", line: 1, column: 0)', bin/node/src/chain_spec.rs:60:46 :thinking: 
```

bin/node/src/chain_spec.rs
michalseweryn: I'll leave it hardcoded for now. There is no benefit in having it configurable for now, and making it specifiable when starting nodes is nontrivial: the hex macro which we use right now works only with string literals, not any values.

bin/node/src/chain_spec.rs
michalseweryn: Done

bin/node/src/chain_spec.rs
michalseweryn: Fixed

fbielejec (approved):
Looks good!

PR #89, opened 2021-07-24
Substrate upgrade

mike1729:


finality-aleph/src/data_io.rs
```diff
@@ -1,28 +1,62 @@
...
+    }
+}
+
+impl<B: Block> aleph_bft::DataIO<B::Hash> for DataIO<B> {
     type Error = Error;
 
     fn get_data(&self) -> B::Hash {
-        let header = self.select_chain.best_chain().expect("No best chain");
+        if self.refresh_best_chain_tx.unbounded_send(()).is_err() {
+            debug!(target: "afa", "Channel for new best chain requests is closed.");
```
fbielejec, line 50: ymmv, but this used to result in a panic, now is a silent debug statement, maybe a `warning!` is in order?

finality-aleph/src/party.rs
```diff
@@ -328,6 +351,7 @@ async fn run_session_as_authority<B, C, BE, SC>(
     // This is a temporary solution -- need to fix this in AlephBFT.
     let _ = exit_aggregator_tx.send(());
     let _ = exit_forwarder_tx.send(());
+    let _ = exit_refresher_tx.send(());
```
michalseweryn, line 354: Can we replace all these `let _ = *tx.send(());` with `if let Err(e) = *tx.send(()) {debug!(target: "afa", "*tx closed before terminating it manually: {}", e)}`? The current solution silences warnings which probably shouldn't be silenced?

finality-aleph/src/data_io.rs
michalseweryn: The diff does not reflect the corresponding part of the code. There is still a panic, but now it is in the line 33, in the refresh task.

finality-aleph/src/data_io.rs
```diff
@@ -1,28 +1,62 @@
...
 use sp_runtime::traits::{Block, Header};
+use std::sync::Arc;
 
 #[derive(Clone)]
-pub(crate) struct DataIO<B: Block, SC: SelectChain<B>> {
-    pub(crate) select_chain: SC,
+pub(crate) struct DataIO<B: Block> {
+    pub(crate) best_chain: Arc<Mutex<B::Hash>>,
     pub(crate) ordered_batch_tx: mpsc::UnboundedSender<OrderedBatch<B::Hash>>,
+    pub(crate) refresh_best_chain_tx: mpsc::UnboundedSender<()>,
```
michalseweryn, line 17: If the sole purpose of the channel is to periodically inform the refresh task to request a new best chain, maybe it would be easier to simply have a ticker in the refresh task? Not sure about that though, but the channel seems a bit unintuitive, especially that when we send the request before we read the best chain from the shared variable...

michalseweryn (changes requested):

finality-aleph/src/data_io.rs
mike1729: I thought about ticker for some time, it has advantage of being simpler, but I thought that the solution with channel has this advantage that the pace of the calls to `best_chain` follows the pace of `get_data`, which depends on AlephBFT and may change. 
I think that refreshing before reading is fine, since we want to get the freshest version.
However, I don't find neither of those two satisfactory, so I you think that ticker is better, I can change it.

finality-aleph/src/party.rs
mike1729: Sure, I followed the old pattern instead of taking the opportunity to fix it.

upgrade_substrate.sh
```diff
@@ -0,0 +1,20 @@
...
+paths=(bin/node/Cargo.toml bin/runtime/Cargo.toml finality-aleph/Cargo.toml
+    primitives/Cargo.toml pallet/Cargo.toml)
+
+for path in ${paths[@]}; do
+    echo upgrade $path
+    sed -e 's/\(substrate.git.*rev = "\).*"/\1'$1'"/' < $path > x
+    mv x $path
+done
+
+# cargo update
```
timorl, line 20: Commented code.

timorl: Either do this, or `echo` some instructions about what to do next.

finality-aleph/src/data_io.rs
timorl: I don't think the refresh is likely to be in effect when we get the hash, especially since `get_data` is not `async`, so won't give control away. I think with the current semantics of `get_data` (get the newest hash we have at this point) a timer is just better. Unless there is a way to receive push notifications about new blocks being available, but even if that is possible that would be a future improvement rather than a part of this task (unless it's trivial, but I doubt it).

timorl (changes requested):
Nice, the only reason this is not an approve is the commented code – the rest of the changes are optional imo.

upgrade_substrate.sh
mike1729: I forgot to uncomment it!

upgrade_substrate.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
mike1729: It simple, but convenient. +1 for renaming it. It's silly since I don't know curl magic, if you can provide me with some info than I'm more than happy to improve the script.

finality-aleph/src/data_io.rs
mike1729: Changed to ticker.

upgrade_substrate.sh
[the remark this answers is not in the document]
timorl: Turns out it's not even `curl` magic, just good old `git` wizardry:
```
git ls-remote https://github.com/paritytech/substrate.git | grep refs/heads/master | cut -c 1-7
```

finality-aleph/src/data_io.rs
michalseweryn: How about using the long-forgotten bounded channel? We could use `futures::channel::mpsc::channel` which returns a sender which blocks until the previously sent hash is received. And the Receiver has a synchronous `try_next` method which checks whether an item is available. This way the type of `best_chain` can be changed from `Arc<Mutex<B::Hash>>` to `RefCell<B::Hash>`, and `refresh_best_chain` only needs `Sender<B::Hash>` instead of `mpsc::UnboundedReceiver<()>` and `Arc<Mutex<B::Hash>>`.

timorl: I would say making this mechanism better should be a separate task, since I suspect there might be a much better approach with push notifications.

Besides, I think what @michalseweryn proposes might have the problem of giving us the next block, rather than the newest? But I'm not sure I understood this correctly, so I may be way off.

finality-aleph/src/party.rs
```diff
@@ -319,15 +332,25 @@ async fn run_session_as_authority<B, C, BE, SC>(
     spawn_handle
         .0
         .spawn("aleph/consensus_session_aggregator", aggregator_task);
+    spawn_handle
+        .0
+        .spawn("aleph/consensus_session_refresher", refresher_task);
 
     let _ = exit_rx.await;
     info!(target: "afa", "Shutting down authority session {}", session_id.0);
     let _ = exit_member_tx.send(());
```
michalseweryn, line 341: Can we please also log error when these futures return `Err`?

mike1729: Sure, just missed that.

finality-aleph/src/party.rs
mike1729: fixed.

upgrade_substrate.sh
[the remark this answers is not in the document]
mike1729: Changed.

finality-aleph/src/party.rs
mike1729: Fixed.

finality-aleph/src/data_io.rs
mike1729: I'll add a task to Jira, to come up with optimal solution and let stick to the ticker for now. Is it ok with you?

timorl (approved):
<3

finality-aleph/src/data_io.rs
michalseweryn: sure

michalseweryn (approved):

PR #91, opened 2021-07-27
Graceful exit

fixxxedpoint:
`party::run_session_as_authority` awaits its components to finish gracefully.

### DIFF
--- a/finality-aleph/src/lib.rs
+++ b/finality-aleph/src/lib.rs
@@ -16,7 +16,7 @@ use sp_runtime::{
     traits::{BlakeTwo256, Block},
     RuntimeAppPublic,
 };
-use std::{convert::TryInto, fmt::Debug, sync::Arc};
+use std::{convert::TryInto, fmt::Debug, pin::Pin, sync::Arc};
 mod aggregator;
 pub mod config;
 mod data_io;
@@ -189,21 +189,36 @@ impl From<SpawnTaskHandle> for SpawnHandle {
     }
 }
 
-impl aleph_bft::SpawnHandle for SpawnHandle {
+impl SpawnHandle {
     fn spawn(&self, name: &'static str, task: impl Future<Output = ()> + Send + 'static) {
         self.0.spawn(name, task)
     }
+
     fn spawn_essential(
         &self,
         name: &'static str,
         task: impl Future<Output = ()> + Send + 'static,
-    ) -> TaskHandle {
+    ) -> Pin<Box<impl Future<Output = Result<(), ()>> + Send>> {
         let (tx, rx) = oneshot::channel();
         self.0.spawn(name, async move {
             task.await;
             let _ = tx.send(());
         });
         Box::pin(rx.map_err(|_| ()))
+    }
+}
+
+impl aleph_bft::SpawnHandle for SpawnHandle {
+    fn spawn(&self, name: &'static str, task: impl Future<Output = ()> + Send + 'static) {
+        <SpawnHandle>::spawn(self, name, task)
+    }
+
+    fn spawn_essential(
+        &self,
+        name: &'static str,
+        task: impl Future<Output = ()> + Send + 'static,
+    ) -> TaskHandle {
+        <SpawnHandle>::spawn_essential(self, name, task)
     }
 }
 
@@ -245,3 +260,5 @@ where
 {
     run_consensus_party(AlephParams { config })
 }
+
+pub fn into_join_handle() {}
--- a/finality-aleph/src/lib.rs
+++ b/finality-aleph/src/lib.rs
@@ -201,11 +201,7 @@ impl SpawnHandle {
     ) -> impl Future<Output = Result<(), ()>> + Send {
         let (tx, rx) = oneshot::channel();
         self.spawn(name, async move {
-            {
-                // force task to be dropped before we send `()` on tx
-                let task = task;
-                task.await;
-            }
+            task.await;
             let _ = tx.send(());
         });
         rx.map_err(|_| ())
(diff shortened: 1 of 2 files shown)


finality-aleph/src/lib.rs
```diff
@@ -189,15 +189,16 @@ impl From<SpawnTaskHandle> for SpawnHandle {
...
     fn spawn(&self, name: &'static str, task: impl Future<Output = ()> + Send + 'static) {
         self.0.spawn(name, task)
     }
+
     fn spawn_essential(
         &self,
         name: &'static str,
         task: impl Future<Output = ()> + Send + 'static,
-    ) -> TaskHandle {
+    ) -> Pin<Box<impl Future<Output = Result<(), ()>> + Send>> {
```
mike1729, line 201: Why this instead of `TaskHandle`?

finality-aleph/src/lib.rs
```diff
@@ -189,15 +189,16 @@ impl From<SpawnTaskHandle> for SpawnHandle {
     }
 }
 
-impl aleph_bft::SpawnHandle for SpawnHandle {
```
mike1729, line 192: Why you changed this since in the implementation of `aleph_bft::SpawnHandle` trait you just call them.

finality-aleph/src/lib.rs
```diff
@@ -245,3 +260,5 @@ where
 {
     run_consensus_party(AlephParams { config })
 }
+
+pub fn into_join_handle() {}
```
mike1729, line 264: What is the purpose of this?

mike1729 (changes requested):

finality-aleph/src/lib.rs
fixxxedpoint: I didn't want to make the internal api of aleph-node to depend on aleph-bft. But actually all of this is a bit confusing. That spawn_essential in general is a bit confusing too - in sp_core the intention of analogous method is to take down the whole node in case of an error. In our case it is just a wrapper allowing to recognize that a task finished and it could be an external to method.

finality-aleph/src/lib.rs
fixxxedpoint: sorry for that

finality-aleph/src/party.rs
```diff
@@ -310,24 +310,20 @@ async fn run_session_as_authority<B, C, BE, SC>(
...
     info!(target: "afa", "Shutting down authority session {}", session_id.0);
     let _ = exit_member_tx.send(());
-    debug!(target: "afa", "Waiting 5000ms for Member to shut down without panic");
-    Delay::new(Duration::from_millis(5000)).await;
-    // This is a temporary solution -- need to fix this in AlephBFT.
+    let _ = member_handle.await;
     let _ = exit_aggregator_tx.send(());
+    let _ = aggregator_handle.await;
     let _ = exit_forwarder_tx.send(());
+    let _ = forwarder_handle.await;
```
DamianStraszak, line 326: This indeed should work, but it is a little unsatisfying that the ordering matters here :/  Maybe add some comment saying that this order is correct and why...

finality-aleph/src/lib.rs
fixxxedpoint: removed

finality-aleph/src/party.rs
fixxxedpoint: I've changed the order in which they are spawned too. Worker tasks are spawned before forwarder now and forwarder is "dropped" last to avoid those "send on closed channel" panics. I've added some comment regarding that.

finality-aleph/src/lib.rs
fixxxedpoint: That block was unnecessary - .await moves its Future.

timorl: I would say even heavily depending on AlephBFT here and in `party.rs` is fine, and then there would be no need for accessing internal fields, just use the trait methods. The advantage of less boilerplate seems worth the minimal cost to me.

timorl (approved):
Nothing the other reviewers didn't say, other than than LGTM.

finality-aleph/src/lib.rs
fixxxedpoint: I can't fully agree but in this case it doesn't matter that much. Generally, it's easier to maintain a stable internal api that just needs to be adapted to an external trait, than directly using some external trait. Any change in that external trait requires changes everywhere it was used internally and you have no "control" over it since you do not own it. When you just adapt to it, only the adapting part needs to be changed.

timorl: That is a very good point, I just feel in our case the integration with AlephBFT is tight enough that any changes to the API will require quite a lot of work anyway. And we will most likely keep similar style and approaches in both repositories, so we will likely want to change the interface to mimic the one in AlephBFT anyway.

Eh, actually I don't have very strong feelings one way or another in this case, so I would say it's up to @mike1729. You have my approve anyway.

DamianStraszak (changes requested):
There are some ERRORs in the logs thrown by AlephBFT (seemingly after exiting) -- would be good to take a look if this can be fixed.

finality-aleph/src/lib.rs
mike1729: We are nonetheless depend on alephBFT so I don't see any gain here.
As to the confusion, the problem is that we changed AlephBFT to meet audit requirements. That should be refactored now.

finality-aleph/src/lib.rs
fixxxedpoint: Now it uses alephbft's trait instead.

fixxxedpoint:
> There are some ERRORs in the logs thrown by AlephBFT (seemingly after exiting) -- would be good to take a look if this can be fixed.

There are two separate bugs related with this. One of them was fixed in aleph-node by this PR. The other can be only fixed inside of alephbft.

DamianStraszak (approved):

mike1729 (approved):

PR #92, opened 2021-07-28
Serve our paperclip overlords

timorl:


finality-aleph/src/party.rs
```diff
@@ -229,144 +231,109 @@ async fn run_aggregator<B, C, BE>(
...
-    debug!(target: "afa", "Authority task {:?}", session.session_id);
-
-    let (ordered_batch_tx, ordered_batch_rx) = mpsc::unbounded();
+    fn run_session_as_authority(
+        &self,
+        node_id: NodeIndex,
+        data_network: DataNetwork<NetworkData<B>>,
+        session: Session<AuthorityId, NumberFor<B>>,
+        exit_rx: futures::channel::oneshot::Receiver<()>,
+    ) -> impl crate::Future<Output = ()> {
```
michalseweryn, line 249: Why `crate::`?

michalseweryn (changes requested):
Looks good, just two small improvments

finality-aleph/src/party.rs
timorl: To use the same `Future` crate-wide. But you are right that this looks strange, so added an explicit import instead.

finality-aleph/src/party.rs
```diff
@@ -78,17 +77,20 @@ where
...
         session_manager,
         client,
-        select_chain,
-        spawn_handle,
         auth_keystore,
-        authority_justification_tx,
+        select_chain,
         metrics,
-        sessions.clone(),
+        authority,
```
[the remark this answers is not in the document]
timorl: Yup, much better now.

michalseweryn (approved):

finality-aleph/src/party.rs
michalseweryn: I still think it would more natural to import `futures::Future` directly, but you can leave it as is

timorl: I'm not sure why this is exported in the top level of the crate, but since it is using the version from there sounds better to me. I guess we could set up clearer guidelines what we want to do in such situations (whether we can reexport external types, and what to use if we do).

Yithis (approved):

PR #96, opened 2021-08-02
A0-194: Implement Data Availability in Aleph Node

maciejnems:


finality-aleph/src/data_io.rs
```diff
@@ -1,13 +1,189 @@
...
+    fn add_message(&mut self, message: AlephNetworkData<B>) {
+        let requirements: Vec<_> = message
+            .included_data()
+            .into_iter()
+            .filter(|b| {
+                if self.available_blocks.contains(b) {
+                    return false;
+                }
+                if self.client.header(BlockId::Hash(*b)).is_ok() {
+                    self.available_blocks.put(*b, ());
```
DamianStraszak, line 139: it's not clear to me why to put directly instead of via `self.add_block`.

DamianStraszak (changes requested):
Up to some minor issues it seems everything is alright.

In regular circumstances the complex part of this code is never executed so it's hard to say whether this actually works :/ I would suggest running a local test like this: (no need to add anything to repo, just run a local sanity check) try to artificially delay block import (i.e. can add a sleep somewhere in import.rs) and run the code (maybe with some additional logging) to check that critical pieces of code like `add_pending_message` etc. are executed and work correctly.

finality-aleph/src/data_io.rs
maciejnems: There was none and I think `self.add_block` is even better here

maciejnems:
> Up to some minor issues it seems everything is alright.
> 
> In regular circumstances the complex part of this code is never executed so it's hard to say whether this actually works :/ I would suggest running a local test like this: (no need to add anything to repo, just run a local sanity check) try to artificially delay block import (i.e. can add a sleep somewhere in import.rs) and run the code (maybe with some additional logging) to check that critical pieces of code like `add_pending_message` etc. are executed and work correctly.

I tested it locally a little bit differently. I removed the filter in `add_message`, so every message is always added to pending messages. This way after the timeout messages are checked and sent to network. I also tested it without the stream from block import. From local testing this works as intended (the messages are first saved as pending, after timeout they are sent and removed from data store). Still I think we should have some way to test it during pipeline. This is difficult, because there is no way of creating `alephbft::NetworkData` here.

finality-aleph/src/data_io.rs
```diff
@@ -1,13 +1,190 @@
...
+use futures::channel::mpsc::{UnboundedReceiver, UnboundedSender};
+use futures_timer::Delay;
+use log::{debug, trace};
+use sc_client_api::{backend::Backend, BlockchainEvents};
+use sp_runtime::generic::BlockId;
+use tokio::stream::StreamExt;
+
+type MessageId = u64;
+const AVAILABLE_BLOCKS_CACHE_SIZE: usize = 1000;
+const PENDING_MESSAGE_LIFETIME: MessageId = 100_000;
```
fbielejec, line 25: the naming here is a bit confusing, is it an id or a duration?

finality-aleph/src/data_io.rs
```diff
@@ -1,13 +1,190 @@
...
+            self.dependent_messages
+                .entry(*block_hash)
+                .or_insert_with(HashSet::new)
+                .insert(message_id);
+        }
+        self.message_requirements
+            .insert(message_id, requirements.len());
+        self.pending_messages.insert(message_id, message);
+
+        if message_id >= PENDING_MESSAGE_LIFETIME {
```
fbielejec, line 126: same here, it's super confusing to be comparing duration and id (something unique), maybe just needs better naming?

finality-aleph/src/data_io.rs
```diff
@@ -1,13 +1,190 @@
...
+            pending_messages: HashMap::new(),
+            available_blocks: LruCache::new(AVAILABLE_BLOCKS_CACHE_SIZE),
+            ready_messages_tx,
+            messages_rx,
+            _phantom: PhantomData,
+        }
+    }
+
+    pub(crate) async fn run(&mut self, mut exit: oneshot::Receiver<()>) {
+        let mut timeout = Delay::new(TIMEOUT_DURATION);
```
DamianStraszak, line 70: Normally I'm quite lenient when it comes to names, but I don't like timeout -- this has a negative association that something went wrong. Here this is something closer to a "periodic maintenance"

finality-aleph/src/data_io.rs
maciejnems: Yes, it isn't really a lifetime. I changed it to `MESSAGE_ID_BOUNDARY` as this is a boundary for oldest message id that should be kept in pending messages.

finality-aleph/src/data_io.rs
maciejnems: I changed it to `PERIODIC_MAINTENANCE_INTERVAL`

finality-aleph/src/data_io.rs
```diff
@@ -1,13 +1,190 @@
...
+            dependent_messages: HashMap::new(),
+            pending_messages: HashMap::new(),
+            available_blocks: LruCache::new(AVAILABLE_BLOCKS_CACHE_SIZE),
+            ready_messages_tx,
+            messages_rx,
+            _phantom: PhantomData,
+        }
+    }
+
+    pub(crate) async fn run(&mut self, mut exit: oneshot::Receiver<()>) {
```
DamianStraszak, line 69: Please write a comment documenting how this component works internally. In particular it should mention why is "periodic maintenance" necessary.

maciejnems: Done

finality-aleph/src/data_io.rs
```diff
@@ -1,22 +1,214 @@
...
+        } else {
+            self.add_pending_message(message, requirements);
+        }
+    }
+
+    fn push_messages(&mut self, block_hash: B::Hash) {
+        for message_id in self
+            .dependent_messages
+            .entry(block_hash)
+            .or_insert_with(HashSet::new)
```
mike1729, line 176: [Hashmap::remove](https://doc.rust-lang.org/beta/std/collections/struct.HashMap.html#method.remove) returns `Option<V>` so can check that instead of creating unnecessary empty HashSet that you removes in the next block.

finality-aleph/src/data_io.rs
```diff
@@ -1,22 +1,214 @@
...
+                if self.available_blocks.contains(block_hash) {
+                    return false;
+                }
+                if let Ok(Some(_)) = self.client.header(BlockId::Hash(*block_hash)) {
+                    self.add_block(*block_hash);
+                    return false;
+                }
+                true
+            })
+            .collect();
```
mike1729, line 160: You don't store anywhere requirements as a Vector, so you could leave it as Iterator.

finality-aleph/src/testing/data_io.rs
```diff
@@ -0,0 +1,156 @@
...
+};
+
+#[derive(Debug)]
+struct TestNetworkData {
+    data: Vec<Hash>,
+}
+
+impl AlephNetworkMessage<Block> for TestNetworkData {
+    fn included_blocks(&self) -> Vec<Hash> {
+        self.data.to_vec()
```
mike1729, line 26: `clone` is more natural.

finality-aleph/src/testing/data_io.rs
```diff
@@ -0,0 +1,156 @@
...
+    }
+    blocks
+}
+
+#[tokio::test]
+async fn sends_messages_with_imported_blocks() {
+    let (task_handle, mut client, store_tx, mut store_rx, exit_data_store_tx) =
+        prepare_data_store();
+
+    let consensus_network_handle = tokio::spawn(task_handle);
```
mike1729, line 82: `consensus_network_handle` doesn't seem accurate. `data_store_handle`?

finality-aleph/src/testing/data_io.rs
```diff
@@ -0,0 +1,156 @@
...
+    store_tx
+        .unbounded_send(TestNetworkData {
+            data: blocks.clone(),
+        })
+        .unwrap();
+
+    Delay::new(Duration::from_millis(10)).await;
+    exit_data_store_tx.send(()).unwrap();
+    consensus_network_handle.await.unwrap();
+    let message = store_rx.next().await.expect("We own the tx");
```
mike1729, line 95: Can you shuffle the 4 lines above so that delay is not needed? 
1. `store_rx.next().await`
2. `exit_data_store_tx.send(())`
3. `consensus_network_handle.await`

finality-aleph/src/testing/data_io.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    assert_eq!(message.included_blocks(), blocks);
+}
+
+#[tokio::test]
+async fn sends_messages_after_import() {
+    println!("ehhhhh");
```
mike1729, line 102: I assume the test doesn't pass without it?

finality-aleph/src/testing/data_io.rs
```diff
@@ -0,0 +1,156 @@
...
+    let (task_handle, mut client, store_tx, mut store_rx, exit_data_store_tx) =
+        prepare_data_store();
+
+    let consensus_network_handle = tokio::spawn(task_handle);
```
mike1729, line 106: As above.

finality-aleph/src/testing/data_io.rs
```diff
@@ -0,0 +1,156 @@
...
+        .unwrap();
+    client
+        .import(BlockOrigin::Own, block.clone())
+        .await
+        .unwrap();
+
+    Delay::new(Duration::from_millis(10)).await;
+    exit_data_store_tx.send(()).unwrap();
+    consensus_network_handle.await.unwrap();
+    let message = store_rx.next().await.expect("We own the tx");
```
mike1729, line 127: As above.

finality-aleph/src/testing/data_io.rs
```diff
@@ -0,0 +1,156 @@
...
+        .build()
+        .unwrap()
+        .block;
+    store_tx
+        .unbounded_send(TestNetworkData {
+            data: vec![block.header.hash()],
+        })
+        .unwrap();
+
+    Delay::new(Duration::from_millis(10)).await;
```
mike1729, line 150: This doesn't seem accurate since there is a possibility that after 11ms you would get the message. The problem is bigger unfortunately and I'm not sure you can test `not sending`. You could remove this sleep, shuffle the rest as above, and just call `UnboundedReceiver::try_next` and assert that it's `Some(None)`.

mike1729 (changes requested):

finality-aleph/src/data_io.rs
```diff
@@ -1,22 +1,214 @@
...
+                    trace!(target: "afa", "Received message at Data Store {:?}", message);
+                    self.add_message(message);
+                }
+                Some(block) = &mut import_stream.next() => {
+                    trace!(target: "afa", "Block import notification at Data Store for block {:?}", block);
+                    self.add_block(block.hash);
+                }
+                _ = &mut maintenance_timeout => {
+                    trace!(target: "afa", "Data Store maintenance timeout");
+                    let keys : Vec<_> = self.dependent_messages.keys().cloned().collect();
```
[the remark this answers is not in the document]
maciejnems: Because inside `for` I would like to call a method which is mutating `self`, so we need to clone all keys

finality-aleph/src/testing/data_io.rs
maciejnems: I wanted this test to not run infinitely in case the DataStore wouldn't send the message. The delay is needed here so that DataStore would receive message before exit signal. If we send the exit signal after waiting for `store_rx.next()` it is possible that if DataStore has a bug this test will run forever.

finality-aleph/src/data_io.rs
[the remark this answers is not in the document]
mike1729: Right, but still you don't need to collect.

finality-aleph/src/testing/data_io.rs
maciejnems: Unfortunately when I wast testing this approach the `one_shot` would always be handled before `unbounded` (in run method of DataStore) so this test would be useless, because it wouldn't even recognize that DataStore always sends messages further

finality-aleph/src/testing/data_io.rs
mike1729: If there was a bug, you would see it after 60s, i.e. `cargo test` would output a warning. I wouldn't add sleeps for such a case.

finality-aleph/src/testing/data_io.rs
mike1729: I see, so how about that: add two messages dependent on different blocks, add just one block, and after receiving message on store_rx check if we got only one and the second is still on data store.

finality-aleph/src/data_io.rs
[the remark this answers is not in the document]
maciejnems: Unfortunately I do, because using just `cloned()` is still an immutable borrow on `self`

finality-aleph/src/testing/data_io.rs
maciejnems: Yes, there is a warning, but it doesn't stop the test, so I'm not sure how long it would run for example in our pipeline on github. The delay is here just so that `unbounded_channel` is handled before `one_shot`. I could also add here a limit, so that after some time it panics (for example 30 seconds).

mike1729: I'm not sure what are the limits on GH but hanging test may happen in other situations as well and we don't add artificial panics to dodge them.

maciejnems: Ok, then I will remove the delay

finality-aleph/src/data_io.rs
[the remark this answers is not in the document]
mike1729: Uh...yeah, this is obvious, sorry for the stir.

finality-aleph/src/testing/data_io.rs
maciejnems: I think that is a good idea. I'm just not sure about checking if the second message is still in DataStore (I don't think we should check internal state here). Sending the not imported message first and the imported message second should be enough for this test. And then calling `.next()` once and `try_next()` after that. The second call should return an error.

mike1729: Fair enough.

finality-aleph/src/data_io.rs
maciejnems: After that I am passing requirements to `self.add_pending_messages(message, requirements)` and also checking if it is empty and checking its size, so collect is more convenient in use here. Otherwise I would have to use `.peekable()` or something like that here.

mike1729: Vector allocates so sometimes convenience is not justified. In this case, it seems ok though.

mike1729 (approved):

DamianStraszak (approved):

PR #99, opened 2021-08-04
A0-232 : pass the settings from genesis block to the justification handler

fbielejec:


finality-aleph/src/justification.rs
```diff
@@ -179,14 +182,13 @@ where
 
         let mut notification_stream = futures::stream::select(import_stream, authority_stream);
 
+        let (session_period, _, _) = self.chain_cadence;
```
DamianStraszak, line 185: Maybe it's better for chain_cadence to be a struct with 3 named fields instead of a 3-tuple? :)

finality-aleph/src/party.rs
```diff
@@ -54,21 +57,34 @@ where
...
             },
     } = aleph_params;
 
     let sessions = Arc::new(Mutex::new(HashMap::new()));
 
+    // NOTE: justifications are requested every so often
+    let justifications_cadence = min(
+        (session_period.0 as u64 * 1000u64) * millisecs_per_block.0,
+        millisecs_per_block.0,
+    ) / 1000;
```
DamianStraszak, line 72: This is not the same as the formula mentioned in the task description, we definitely don't want this to be as frequent as block production.

Also, be careful with integer division, we don't want the delay to be 0 if the result turns out to be subsecond.

DamianStraszak (changes requested):

finality-aleph/src/party.rs
maciejnems: Also isn't `millisecs_per_block.0` always smaller than `(session_period.0 as u64 * 1000u64) * millisecs_per_block.0`?
`(session_period.0 as u64 * 1000u64)` is an integer (and I don't think it makes sense when set to 0).

maciejnems: Maybe instead of division we could use `JustificationsCadence::from_millis(...)`?

finality-aleph/src/justification.rs
maciejnems: I'm also wondering why is `millisecs_per_block` in `ChainCadence`. I haven't seen it used anywhere.

finality-aleph/src/justification.rs
```diff
@@ -45,6 +45,9 @@ impl AlephJustification {
     }
 }
 
+pub type JustificationsCadence = Duration;
+pub type ChainCadence = (SessionPeriod, MillisecsPerBlock, JustificationsCadence);
```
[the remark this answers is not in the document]
michalseweryn: Furthermore, it seems that the `JustificationsCadence` can be calculated from the other two fields, so maybe this should be a method?

michalseweryn (changes requested):

finality-aleph/src/party.rs
fbielejec: good point, although technically we never assert that session is longer than time till block production :P

bin/node/rustfmt.toml
```diff
@@ -0,0 +1 @@
+edition = "2018"
```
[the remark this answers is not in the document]
fbielejec: rust-analyzer goes berserk, without it - although I don't know why GH is marking this as change, when we already have these on main :thinking:

finality-aleph/src/party.rs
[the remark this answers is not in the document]
fbielejec: I'm adding an assert that checks at runtime whether session setting is longer than block production time

finality-aleph/src/justification.rs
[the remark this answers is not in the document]
fbielejec: This tuple/struct came about only because of clippy warning you when a function has more than 7 arguments :shrug:

YMMV but wouldn't this be an overkill here?  There must be some (granted, negligible) cost of indirection when using a method.
We want to be able to pass these constants further (in a struct) so the said method would have to mutate the interior state of the struct, therefore we'd also have to initiate it's fields (so probably introduce `new` method), when all we want is a _de-facto_ map with some named fields to silence linter warnings?

finality-aleph/src/justification.rs
fbielejec: removed

bin/node/src/service.rs
```diff
@@ -204,6 +204,11 @@ pub fn new_full(mut config: Configuration) -> Result<TaskManager, ServiceError>
             .unwrap(),
     );
 
+    assert!(
+        session_period.0 as u64 * 1000 > millisecs_per_block.0,
```
DamianStraszak, line 208: session_period is measured in "blocks" not in seconds 
"session_time" would be session_period * MillisecsPerBlock

maciejnems (approved):

finality-aleph/src/party.rs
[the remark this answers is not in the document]
maciejnems: Changed. I'm not sure about assert, as Session period set to zero doesn't really make sense

maciejnems: The formula is changed now

michalseweryn (approved):

PR #104, opened 2021-08-25
A0-268: Validators rotating based on authority list

Yithis:


pallet/src/lib.rs
```diff
@@ -32,16 +43,37 @@ pub mod pallet {
...
     pub struct Pallet<T>(sp_std::marker::PhantomData<T>);
 
     #[pallet::hooks]
     impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
 
+    //ensure_root
     #[pallet::call]
-    impl<T: Config> Pallet<T> {}
+    impl<T: Config> Pallet<T> {
+        #[pallet::weight(10_000)]
```
mike1729, line 67: In our case it probably doesn't matter, but normally the weight should be related to the number of operations on storage.

pallet/src/lib.rs
```diff
@@ -32,16 +43,37 @@ pub mod pallet {
...
 
+    pub struct AlephSessionManager<T>(sp_std::marker::PhantomData<T>);
+
     #[pallet::pallet]
     pub struct Pallet<T>(sp_std::marker::PhantomData<T>);
 
     #[pallet::hooks]
     impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
 
+    //ensure_root
```
mike1729, line 64: The first line in the below call should be: `frame_system::ensure_root(origin)?`.

mike1729 (changes requested):
Have you run your code and check that it works on a live chain? It seems that some support for setting `ValidatorsList` in genesis is missing, but I'm not sure.

pallet/src/lib.rs
```diff
@@ -32,16 +43,37 @@ pub mod pallet {
...
     #[pallet::call]
-    impl<T: Config> Pallet<T> {}
+    impl<T: Config> Pallet<T> {
+        #[pallet::weight(10_000)]
+        pub fn change_validators(
```
fbielejec, line 68: shouldn't we also have another call to remove a validator?

bin/runtime/src/lib.rs
```diff
@@ -312,7 +313,7 @@ impl pallet_session::Config for Runtime {
     type ShouldEndSession = pallet_session::PeriodicSessions<SessionPeriod, Offset>;
     type NextSessionRotation = pallet_session::PeriodicSessions<SessionPeriod, Offset>;
     // The () SessionManager makes always a new session with the same set of validators as before.
-    type SessionManager = ();
+    type SessionManager = pallet_aleph::AlephSessionManager<Self>;
```
fbielejec, line 316: I might be wrong here, but since the validator_list starts as None, doesn't it mean that at the genesis (block) there will be no validators? Maybe you should add some initial validators too, similar to how authorities are read from a file at genesis?

pallet/Cargo.toml
```diff
@@ -14,6 +14,7 @@ sp-std = { default-features = false, git = "https://github.com/paritytech/substr
 serde = "1.0"
 primitives = { path = "../primitives", default-features = false}
 pallet-session = { default-features = false, git = "https://github.com/paritytech/substrate.git", rev = "8bbad73f5" }
+sp-staking = { default-features = false, git = "https://github.com/paritytech/substrate.git", rev = "8bbad73f5" }
```
fbielejec, line 17: needed?

bin/runtime/src/lib.rs
mike1729: Probably it should be added in chain_spec but I'm not sure. It depends if SessionManager is called before first session or not. I don't know.

pallet/Cargo.toml
mike1729: Needed for `sp_staking::SessionIndex`.

bin/runtime/src/lib.rs
Yithis: It should be, but I `None`, since currently it is `None`. @mike1729 if you will give me a list of validators I should put there I can add it into chain_spec. Just I don't know what values should there be.

fbielejec: maybe you could just set it to the same accounts as authorities to start with?

mike1729: This should work but it would be good to know why, eg when the Session Manager is called for the first time. First session is special and there are several places where it's handled in a special way.

pallet/src/lib.rs
fbielejec: My comment was from operational perspective: in most situtations we will rather be adding or removing a single validator, and this way we do not need to keep track of a list of validators ourselves, making sure it's up to date (which can be error prone). YMMV, it's a non-blocker from my perspective, can be merged as-is, we can always change it later.

pallet/src/lib.rs
```diff
@@ -20,10 +20,21 @@ pub mod pallet {
         sp_std,
     };
     use frame_system::pallet_prelude::*;
-    use pallet_session::Pallet as Session;
+    use pallet_session::{Pallet as Session, SessionManager};
     use primitives::{
         ApiError as AlephApiError, DEFAULT_MILLISECS_PER_BLOCK, DEFAULT_SESSION_PERIOD,
     };
+    use sp_staking::SessionIndex;
```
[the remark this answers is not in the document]
Yithis: Done

fbielejec (approved):

pallet/Cargo.toml
mike1729, line 17: Shouldn't this be removed now as well?

pallet/src/lib.rs
```diff
@@ -20,28 +20,59 @@ pub mod pallet {
...
     #[pallet::pallet]
     pub struct Pallet<T>(sp_std::marker::PhantomData<T>);
 
     #[pallet::hooks]
     impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
 
     #[pallet::call]
-    impl<T: Config> Pallet<T> {}
+    impl<T: Config> Pallet<T> {
+        #[pallet::weight(0)]
```
mike1729, line 65: Weight should be: `(T::BlockWeights::get().max_block, DispatchClass::Operational)`

mike1729 (changes requested):

DamianStraszak (approved):

pallet/src/lib.rs
Yithis: Done

pallet/Cargo.toml
Yithis: Done

DamianStraszak (changes requested):

pallet/src/lib.rs
```diff
@@ -116,6 +154,23 @@ pub mod pallet {
         }
     }
 
+    impl<T: Config> SessionManager<T::AccountId> for AlephSessionManager<T> {
+        fn new_session(session: u32) -> Option<Vec<T::AccountId>> {
+            if let Some((validators_list, validators_list_session)) = Pallet::<T>::validators_list()
```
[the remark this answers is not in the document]
DamianStraszak: While fetching "the whole `validators_list`" might sounds innocuous, in reality what is stored in the database and needs to be fetched (from disk) is `20-100` times larger then the original data.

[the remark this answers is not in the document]
mike1729: I believe that overhead is related to the number of calls not the size of single element, but I'm not 100% sure.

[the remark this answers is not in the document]
DamianStraszak: The rule of thumb in designing storage is to bundle several items into structs (or pair like here) only if you access and modify the fields "together". According to my understanding this is not quite the case here -- you would use the session number many times and the validators list only once.

Now I think that the "Vec" type data would be stored under a single key, so the additional cost here would only be fetching the validators list from disk (so the `20-100` times overhead is probably not quite correct -- it would be correct when talking about a single scalar storage value). But this is still not cheap, as it is serialized, hashed and whatnot.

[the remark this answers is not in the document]
mike1729: Hashed is only the key to the db, but the vector is indeed deserialized from bytes, so you are right that in this case it's better to store separately.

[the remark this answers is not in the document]
Yithis: Done

pallet/src/lib.rs
```diff
@@ -156,11 +168,15 @@ pub mod pallet {
...
         fn new_session(session: u32) -> Option<Vec<T::AccountId>> {
-            if let Some((validators_list, validators_list_session)) = Pallet::<T>::validators_list()
+            if let Some(session_for_validators_change) =
+                Pallet::<T>::session_for_validators_change()
             {
-                if validators_list_session <= session {
-                    ValidatorsList::<T>::put(None::<(Vec<T::AccountId>, u32)>);
-                    return Some(validators_list);
+                if session_for_validators_change <= session {
+                    if let Some(validators) = Pallet::<T>::validators() {
```
DamianStraszak, line 175: There should be either a pair of `Some`s or a pair of `None`'s (or at least that's what I would think), so it's better to use `expect` in place of `if let`.

Yithis: Done

DamianStraszak (approved):

bin/runtime/src/lib.rs
```diff
@@ -328,7 +329,7 @@ impl pallet_session::Config for Runtime {
     type ShouldEndSession = pallet_session::PeriodicSessions<SessionPeriod, Offset>;
     type NextSessionRotation = pallet_session::PeriodicSessions<SessionPeriod, Offset>;
     // The () SessionManager makes always a new session with the same set of validators as before.
```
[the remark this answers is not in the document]

PR #107, opened 2021-08-30
Draft: A0-302: Store validator keys in a json file

michalseweryn:


bin/node/src/command.rs
```diff
@@ -49,8 +49,10 @@ impl SubstrateCli for Cli {
         let ExtraParams {
             session_period,
             millisecs_per_block,
-        } = self.extra;
-        let chain_params = chain_spec::ChainParams::from_cli(session_period, millisecs_per_block);
+            keys_path,
+        } = self.extra.clone();
+        let keys_path = keys_path.unwrap_or_else(|| "authority-keys.json".into());
```
michalseweryn, line 54: This is needed because the binary does not parse the `--keys-path` flag for some reason. To be investigated

PR #110, opened 2021-09-01
A0-340: wait in CI until consensus begins

fbielejec:


docker/docker_entrypoint.sh
```diff
@@ -11,6 +11,7 @@ BASE_PATH=${BASE_PATH:?'Base path should be specified'}
 RPC_PORT=${RPC_PORT:-9933}
 WS_PORT=${WS_PORT:-9943}
 PORT=${PORT:-30333}
+PUBLIC_ADDR=${PUBLIC_ADDR:-/ip4/127.0.0.1/tcp/$PORT}
```
bartoszjedrzejewski: Replace that, by adding 

> if [[ -n "${PUBLIC_ADDR:-}" ]]; then
>   ARGS+=(--public-addr "${PUBLIC_ADDR}")
> fi

at the end, after FLAG_L_ALEPH_BFT.

docker/docker_entrypoint.sh
```diff
@@ -28,6 +29,7 @@ ARGS=(
   --rpc-cors all --rpc-methods Safe # TODO: should we allow to specify them?
   --no-prometheus --no-telemetry # Currently not using. plan to start as soon as capacity is available
   --no-mdns
+  --public-addr "${PUBLIC_ADDR}"
```
bartoszjedrzejewski: We don't want to hardcode this. See comment above.

docker/docker_entrypoint.sh
```diff
@@ -50,4 +52,6 @@ if [[ -n "${FLAG_L_ALEPH_BFT:-}" ]]; then
   ARGS+=(-lAlephBFT=debug)
 fi
 
+echo "My public address to advertize is: $PUBLIC_ADDR"
```
bartoszjedrzejewski: There's really no need for this in final form.

bartoszjedrzejewski (changes requested):

bartoszjedrzejewski (approved):

skrolikowski-10c (approved):

PR #114, opened 2021-09-02
Added early session-skip to finality-aleph.

DamianStraszak:
Now, when a node is catching up, it will not need to initialize consensus-related processes, just skip through the session right away.

### DIFF
--- a/finality-aleph/src/party.rs
+++ b/finality-aleph/src/party.rs
@@ -437,11 +437,17 @@ where
         let last_block = last_block_of_session::<B>(session_id, self.session_period);
 
         // Early skip attempt -- this will trigger during catching up (initial sync).
-        let last_finalized_number = self.client.info().finalized_number;
-        if last_finalized_number >= last_block {
-            debug!(target: "afa", "Skipping session {:?} early because block {:?} already finalized", session_id, last_finalized_number);
-            return;
+        if self.client.info().best_number >= last_block {
+            // We need to give the JustificationHandler some time to pick up the keybox for the new session,
+            // validate justifications and finalize blocks.
+            Delay::new(Duration::from_millis(2000)).await;
+            let last_finalized_number = self.client.info().finalized_number;
+            if last_finalized_number >= last_block {
+                debug!(target: "afa", "Skipping session {:?} early because block {:?} already finalized", session_id, last_finalized_number);
+                return;
+            }
         }
+
 
         let maybe_node_id =
             get_node_index(&authorities, &self.auth_keystore.authority_id().clone());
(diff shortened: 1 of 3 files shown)


timorl (approved):

timorl (approved):
Cool, it got better.

finality-aleph/src/party.rs
```diff
@@ -431,25 +431,32 @@ where
...
+            .insert(session_id, authorities.clone());
+        let last_block = last_block_of_session::<B>(session_id, self.session_period);
+
+        // Early skip attempt -- this will trigger during catching up (initial sync).
+        if self.client.info().best_number >= last_block {
+            // We need to give the JustificationHandler some time to pick up the keybox for the new session,
+            // validate justifications and finalize blocks.
+            Delay::new(Duration::from_millis(2000)).await;
+            let last_finalized_number = self.client.info().finalized_number;
+            if last_finalized_number >= last_block {
```
timorl, line 445: Actually, could you do this check both before and after the delay? Otherwise we might be waiting for completely no reason.

finality-aleph/src/party.rs
```diff
@@ -482,10 +494,22 @@ where
...
+        // In this method we make sure that the amount of data we keep in RAM in finality-aleph
+        // does not grow with the size of the blockchain.
+        debug!(target: "afa", "Pruning session data below {:?}.", prune_below);
+        self.session_authorities
+            .lock()
+            .retain(|&s, _| s >= prune_below);
+    }
+
     async fn run(mut self) {
         for curr_id in 0.. {
```
timorl, line 507: Another thing that came to me – this is being run at every lunch, including one that would start from block 1000000 or whatever, so the loop below would have to go through all sessions, even if we started with the 1000000 blocks already loaded in the DB, right? Maybe we could before this point at least look in the DB we are using for the latest finalized block and use that to set the beginning of this loop? Or is my understanding somehow wrong?

timorl (changes requested):
Thought about it a bit more and wanted to comment after all. <_<"

finality-aleph/src/party.rs
DamianStraszak: Right, I was surprised you didn't comment on that :) Yeah, this is bad, I changed it now into a loop that checks 10 times, every 200ms. It's a little ugly, but works much better.

finality-aleph/src/party.rs
DamianStraszak: You are absolutely correct. I have added a few lines that check what's the current session (based on last finalized block) and run the iteration starting from it.

timorl (approved):

mike1729 (approved):

PR #115, opened 2021-09-02
Feature/chainspec and keygen

fbielejec:
This PR introduces new command(s)

* bootstrap-chain (alias: build-spec) command [4/4]
- [x] takes chain identifier and base-path parameters
- [x] writes aura + aleph keys to the keystore of all validators
- [x] creates and prints to stdout a corresponding chainspec (with validators account ids and public keys written to pallet_session keys at genesis)
- [x] `dev` is a "special" chain identifier as it creates a development chain and a corresponding chainspec

Example:

``` bash
./aleph-node bootstrap-chain --base-path chain/data --chain a0tnet1 > chain/data/chainspec.json
```

* run chain (default) command [1/1]
- [x] takes a path to JSON chainspec

**Follow-up work**

This is an MVP required to geneate a chainspec for mainnet version of aleph-node, therefore the following items are moved for a follow-up PRs

* for the bootstrap chain command
- [ ] OPTIONAL: create libp2p network secret keys and print peerIds to the stdout

* OPTIONAL bootstrap-node command [0/2]
- [ ] takes base-path chain and node-name arguments, writes aura + aleph keys to the keystore for one validator
- [ ] OPTIONAL: creates libp2p network secret key and prints peerId to the stdout

### DIFF
--- a/bin/node/src/chain_spec.rs
+++ b/bin/node/src/chain_spec.rs
@@ -1,5 +1,3 @@
-use std::collections::HashMap;
-
 use aleph_primitives::{
     AuthorityId as AlephId, DEFAULT_MILLISECS_PER_BLOCK, DEFAULT_SESSION_PERIOD,
 };
@@ -9,23 +7,18 @@ use aleph_runtime::{
 };
 use hex_literal::hex;
 use sc_service::ChainType;
-use sp_application_crypto::key_types;
 use sp_consensus_aura::sr25519::AuthorityId as AuraId;
-use sp_core::{ed25519, sr25519, Pair, Public};
+use sp_core::{sr25519, Pair, Public};
 use sp_runtime::traits::{IdentifyAccount, Verify};
-use std::{env::VarError, fmt::Display, str::FromStr};
 
 const FAUCET_HASH: [u8; 32] =
     hex!("eaefd9d9b42915bda608154f17bb03e407cbf244318a0499912c2fb1cd879b74");
 
-pub(crate) const LOCAL_AUTHORITIES: [&str; 8] = [
+pub const LOCAL_AUTHORITIES: [&str; 8] = [
     "Damian", "Tomasz", "Zbyszko", "Hansu", "Adam", "Matt", "Antoni", "Michal",
 ];
 
-pub(crate) const KEY_PATH: &str = "/tmp/authorities_keys";
-
-pub(crate) const TESTNET_ID: &str = "a0tnet1";
-pub(crate) const DEVNET_ID: &str = "dev";
+pub const DEVNET_ID: &str = "dev";
 
 /// Specialized `ChainSpec`. This is a specialization of the general Substrate ChainSpec type.
 pub type ChainSpec = sc_service::GenericChainSpec<GenesisConfig>;
@@ -49,12 +42,12 @@ where
 
 #[derive(Clone)]
 pub struct AuthorityKeys {
-    account_id: AccountId,
-    aura_key: AuraId,
-    aleph_key: AlephId,
+    pub account_id: AccountId,
+    pub aura_key: AuraId,
+    pub aleph_key: AlephId,
 }
 
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, Debug)]
 pub struct ChainParams {
     session_period: u32,
     millisecs_per_block: u64,
@@ -63,113 +56,17 @@ pub struct ChainParams {
 impl ChainParams {
     pub fn from_cli(session_period: Option<u32>, millisecs_per_block: Option<u64>) -> Self {
         ChainParams {
-            session_period: Self::param(
-                "session period",
-                session_period,
-                "SESSION_PERIOD",
-                DEFAULT_SESSION_PERIOD,
-            ),
-            millisecs_per_block: Self::param(
-                "millisecs per block",
-                millisecs_per_block,
-                "MILLISECS_PER_BLOCK",
-                DEFAULT_MILLISECS_PER_BLOCK,
-            ),
-        }
-    }
-
-    fn param<T: FromStr + Display>(
-        debug_name: &str,
-        cli_value: Option<T>,
-        var: &str,
-        default: T,
-    ) -> T
-    where
-        <T as FromStr>::Err: ToString,
-    {
-        cli_value
-            .or_else(|| Self::parse_env_var(var))
-            .unwrap_or_else(|| {
-                log::debug!(
-                    "{} parameter not specified, using default value: {}",
-                    debug_name,
-                    default
-                );
-                default
-            })
-    }
-
-    fn parse_env_var<T: FromStr>(var: &str) -> Option<T>
-    where
-        <T as FromStr>::Err: ToString,
-    {
-        match std::env::var(var) {
-            Ok(value) => match value.parse() {
-                Ok(value) => Some(value),
-                Err(err) => {
-                    panic!(
-                        "error parsing environment variable {}: {}",
-                        var,
-                        err.to_string()
-                    );
-                }
-            },
-            Err(VarError::NotPresent) => None,
-            Err(err @ VarError::NotUnicode(_)) => {
-                panic!("environment variable {} is not unicode: {}", var, err);
-            }
+            session_period: session_period.unwrap_or(DEFAULT_SESSION_PERIOD),
+            millisecs_per_block: millisecs_per_block.unwrap_or(DEFAULT_MILLISECS_PER_BLOCK),
         }
     }
 }
 
-fn read_keys(n_members: usize) -> Vec<AuthorityKeys> {
-    let auth_keys: HashMap<u32, Vec<[u8; 32]>> =
-        if let Ok(auth_keys) = std::fs::read_to_string(KEY_PATH) {
-            serde_json::from_str(&auth_keys).expect("should contain list of keys")
-        } else {
-            return Default::default();
-        };
-
-    let aura_keys = auth_keys
-        .get(&key_types::AURA.into())
-        .unwrap()
-        .iter()
-        .copied()
-        .map(|bytes| AuraId::from(sr25519::Public::from_raw(bytes)));
-
-    let aleph_keys = auth_keys
-        .get(&aleph_primitives::KEY_TYPE.into())
-        .unwrap()
-        .iter()
-        .copied()
-        .map(|bytes| AlephId::from(ed25519::Public::from_raw(bytes)));
-
-    let account_ids = LOCAL_AUTHORITIES
-        .iter()
-        .map(get_account_id_from_seed::<sr25519::Public>);
-
-    aura_keys
-        .zip(aleph_keys)
-        .zip(account_ids)
-        .take(n_members)
-        .map(|((aura_key, aleph_key), account_id)| AuthorityKeys {
-            account_id,
-            aura_key,
-            aleph_key,
-        })
-        .collect()
-}
-
-pub fn development_config(chain_params: ChainParams) -> Result<ChainSpec, String> {
+pub fn development_config(
+    chain_params: ChainParams,
+    authorities: Vec<AuthorityKeys>,
+) -> Result<ChainSpec, String> {
     let wasm_binary = WASM_BINARY.ok_or_else(|| "Development wasm not available".to_string())?;
-
-    let n_members = std::fs::read_to_string("/tmp/n_members")
-        .expect("Committee size is not specified")
-        .trim()
-        .parse::<usize>()
-        .expect("Wrong committee size");
-
-    let authorities = read_keys(n_members);
 
     let rich_accounts: Vec<_> = [
         "Alice",
@@ -195,7 +92,7 @@ pub fn development_config(chain_params: ChainParams) -> Result<ChainSpec, String
         DEVNET_ID,
         ChainType::Development,
         move || {
-            testnet_genesis(
+            genesis(
                 wasm_binary,
                 // Initial PoA authorities
                 authorities.clone(),
@@ -226,16 +123,12 @@ pub fn development_config(chain_params: ChainParams) -> Result<ChainSpec, String
     ))
 }
 
-pub fn testnet1_config(chain_params: ChainParams) -> Result<ChainSpec, String> {
+pub fn config(
+    chain_params: ChainParams,
+    authorities: Vec<AuthorityKeys>,
+    chain_id: &str,
+) -> Result<ChainSpec, String> {
     let wasm_binary = WASM_BINARY.ok_or_else(|| "Development wasm not available".to_string())?;
-
-    let n_members = std::fs::read_to_string("/tmp/n_members")
-        .expect("Committee size is not specified")
-        .trim()
-        .parse::<usize>()
-        .expect("Wrong committee size");
-
-    let authorities = read_keys(n_members);
 
     let sudo_account: AccountId = hex![
         // 5F4SvwaUEQubiqkPF8YnRfcN77cLsT2DfG4vFeQmSXNjR7hD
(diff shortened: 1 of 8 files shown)


bin/node/src/chain_spec.rs
```diff
@@ -226,38 +123,36 @@ pub fn development_config(chain_params: ChainParams) -> Result<ChainSpec, String
...
 
-    // Give money to the faucet account.
+    // Give money to the faucet account
     let faucet: AccountId = FAUCET_HASH.into();
     let rich_accounts = vec![faucet, sudo_account.clone()];
+
     Ok(ChainSpec::from_genesis(
         // Name
-        "Aleph Zero",
+        "AlephZero",
```
timorl, line 145: Antek was insisting there should be a space there.

bin/node/src/cli.rs
```diff
@@ -19,116 +23,55 @@ pub struct Cli {
...
-    /// Build a chain specification.
-    BuildSpec(sc_cli::BuildSpecCmd),
+
+    // NOTE: similarly we could have a BootstrapNode command that takes a node-name parameter
+    // and writes aura, aleph and (optionally) libp2p private keys to the base-path of a single node
+    // and prints accountId and peerId to the stdout
+    /// Populate authorities keystore and generate JSON chainspec (printed to stdout)    
+    BootstrapChain(BootstrapChainCmd),
+
+    /// Same as BootstrapChain
```
timorl, line 73: If you are just keeping it for backwards compatibility, then... don't? I'm not completely sure, but if we can avoid supporting something then we should and in this case I don't think much depends on the exact name of the command.

bin/node/src/commands.rs
```diff
@@ -0,0 +1,178 @@
...
+/// The `bootstrap-chain` command is used generate keys for the genesis authorities
+/// keys are written to the keystore of the authorities
+/// and the chain specification is printed to stdout in the JSON format
+#[derive(Debug, StructOpt)]
+pub struct BootstrapChainCmd {
+    /// Force raw genesis storage output.
+    #[structopt(long = "raw")]
+    pub raw: bool,
+
+    #[allow(missing_docs)]
```
timorl, line 26: This sounds bad?

bin/node/src/chain_spec.rs
```diff
@@ -272,7 +167,7 @@ pub fn testnet1_config(chain_params: ChainParams) -> Result<ChainSpec, String> {
         Some(
             [(
                 "tokenSymbol".to_string(),
-                serde_json::Value::String("TZERO".into()),
+                serde_json::Value::String("DZERO".into()),
```
timorl, line 170: `DZERO` is supposed to be the devnet token symbol, did you change it here by mistake?

timorl (changes requested):

bin/node/src/chain_spec.rs
fbielejec: there is one common `config` used by all "nets". The idea being that everything that we do not want hard-coded is passed as command line argument. Could add it to cli arguments or hard-code if we have a decision on what the token should be (AZERO?).

bin/node/src/cli.rs
fbielejec: Ok

bin/node/src/commands.rs
fbielejec: copied from Substrate :) Removing

bin/node/src/chain_spec.rs
maciejnems: I guess we want it to be a command line argument as well? We want to provide different name for Testnet and for Mainnet.

bin/node/src/chain_spec.rs
timorl: It was `TZERO` for the testnet and `AZERO` for the mainnet, although you should probably confirm that last bit with Antek.

bin/node/src/commands.rs
```diff
@@ -0,0 +1,168 @@
...
+        &self.shared_params
+    }
+}
+
+impl BootstrapChainCmd {
+    fn upsert_aura_key(
+        &self,
+        keystore: &LocalKeystore,
+        key_type: KeyTypeId,
+    ) -> Result<AuraId, Error> {
```
michalseweryn, line 46: There are several things with this method.
0. I don't like the `upsert` name as we never update an existing value. Can we rename it to something like `get_aura_key`?
1. We don't use `self` at all, so can make it a "static" method or even a top-level function,
2. We don't really make use of the exact type of `keystore`, so I think it would be cleaner to use the type  `impl SyncCryptoStore`,
3. We only call it with `key_type = key_types::AURA`, and it wouldn't make sense to call it with any other name given the value, so I think we could get rid of it.
4.  the return type is `Result`, but we never return `Err(_)` variant, so we can simply return `AuraId`.

These comments apply to `upsert_aleph_key` too.

bin/node/src/commands.rs
```diff
@@ -0,0 +1,168 @@
...
+        let key = SyncCryptoStore::sr25519_public_keys(&*keystore, key_type)
+            .into_iter()
+            .next()
+            .map_or_else(
+                || {
+                    SyncCryptoStore::sr25519_generate_new(&*keystore, key_type, None)
+                        .map_err(|_| Error::KeyStoreOperation)
```
michalseweryn, line 53: We wrap the value in `Err` only to unwrap it below. Why not call `expect` on the result of calling `sr25519_generate_new`? Then `map_or_else` would be replaced with `unwrap_or_else`.

bin/node/src/commands.rs
```diff
@@ -0,0 +1,168 @@
...
+
+impl BootstrapChainCmd {
+    fn upsert_aura_key(
+        &self,
+        keystore: &LocalKeystore,
+        key_type: KeyTypeId,
+    ) -> Result<AuraId, Error> {
+        let key = SyncCryptoStore::sr25519_public_keys(&*keystore, key_type)
+            .into_iter()
+            .next()
```
michalseweryn, line 49: Pedantic: Since we want to retrieve any element of the vector, `.into_iter().next()` can be simplified to `.pop()`.

bin/node/src/commands.rs
```diff
@@ -0,0 +1,168 @@
...
+            Err(why) => panic!("{}", why),
+        }
+    }
+
+    pub fn run(&self, cli: &Cli) -> Result<(), Error> {
+        let chain_id = self.shared_params.chain_id(self.shared_params.is_dev());
+
+        let mut genesis_authorities: Vec<AuthorityKeys> = Vec::new();
+
+        LOCAL_AUTHORITIES.iter().for_each(|authority| {
```
michalseweryn, line 97: calling `.map(|authority| {...}).collect()` instead of `for_each(...)` would be cleaner and would allow us to get rid of the ugly `mut` in `genesis_authorities`. Also, `for` loop is more idiomatic than `for_each`.

bin/node/src/commands.rs
```diff
@@ -0,0 +1,168 @@
...
+            session_period,
+            millisecs_per_block,
+        } = cli.extra;
+
+        let chain_params = chain_spec::ChainParams::from_cli(session_period, millisecs_per_block);
+        let chain_spec = match chain_id.as_str() {
+            chain_spec::DEVNET_ID => {
+                chain_spec::development_config(chain_params, genesis_authorities)
+            }
+            _ => chain_spec::config(chain_params, genesis_authorities, &chain_id),
```
michalseweryn: I don't like that testnet is called by passing any id other than `DEVNET_ID`

bin/node/src/commands.rs
```diff
@@ -0,0 +1,166 @@
...
+            });
+        });
+
+        info!("Building chain spec");
+
+        let chain_spec = match self.chain_params.chain_id().as_str() {
+            chain_spec::DEVNET_ID => {
+                chain_spec::development_config(self.chain_params.clone(), genesis_authorities)
+            }
+            _ => chain_spec::config(self.chain_params.clone(), genesis_authorities, &chain_id),
```
michalseweryn, line 126: I don't like that you run a testnet by passing as chain_id anything other than `DEVNET_ID` (and the value of the id is ignored). Can we add a `TESTNET_ID` constant , and panic when the chain_id is anything else than `DEVNET_ID` or `TESTNET_ID`?

bin/node/src/commands.rs
```diff
@@ -0,0 +1,166 @@
...
+        };
+
+        match chain_spec {
+            Ok(spec) => {
+                let json = sc_service::chain_ops::build_spec(&spec, self.raw)?;
+                if std::io::stdout().write_all(json.as_bytes()).is_err() {
+                    let _ = std::io::stderr().write_all(b"Error writing to stdout\n");
+                }
+            }
+            Err(why) => panic!("{}", why),
```
michalseweryn, line 136: We might as well `.unwrap()` the `chain_spec`

bin/node/src/chain_spec.rs
fbielejec: ok, I will just make all of these parameters part of the CLI struct (ChainParams)

bin/node/src/chain_spec.rs
fbielejec: ok, this and the token symbol will now come form cli (with some sensible defaults)

michalseweryn (changes requested):

bin/node/src/commands.rs
```diff
@@ -0,0 +1,168 @@
+use crate::chain_spec::{self, get_account_id_from_seed, AuthorityKeys, LOCAL_AUTHORITIES};
```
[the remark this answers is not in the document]
fbielejec: we will have other aleph specific commands there - e.g. BootstrapNode. Substrate goes even one step further, with every command in a separate file but I reckon this is an overkill for us - not to mention they might share some common functionalities (split as e.g private functions)

bin/node/src/cli.rs
```diff
@@ -19,116 +23,55 @@ pub struct Cli {
...
+    BuildSpec(BootstrapChainCmd),
```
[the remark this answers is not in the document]
fbielejec: you might have commented on outdated code, this was already removed.

bin/node/src/cli.rs
```diff
@@ -14,121 +9,54 @@ pub struct Cli {
...
 }
 
 #[derive(Debug, StructOpt)]
 pub enum Subcommand {
     /// Key management cli utilities
     Key(sc_cli::KeySubcommand),
-    /// Build a chain specification.
-    BuildSpec(sc_cli::BuildSpecCmd),
+
+    // NOTE: similarly we could have a BootstrapNode command that takes a node-name parameter
```
mike1729, line 55: Instead of a note in the code, it would be better to have a task in jira.

bin/node/src/commands.rs
```diff
@@ -0,0 +1,157 @@
...
+use sp_consensus_aura::sr25519::AuthorityId as AuraId;
+use sp_core::crypto::KeyTypeId;
+use sp_core::{ed25519, sr25519};
+use sp_keystore::SyncCryptoStore;
+use std::io::Write;
+use std::sync::Arc;
+use structopt::StructOpt;
+
+/// The `bootstrap-chain` command is used generate keys for the genesis authorities
+/// keys are written to the keystore of the authorities
```
mike1729, line 19: private keys

bin/node/src/commands.rs
fbielejec: Done

bin/node/src/commands.rs
```diff
@@ -0,0 +1,157 @@
...
+                    SyncCryptoStore::sr25519_generate_new(&*keystore, key_type, None)
+                        .map_err(|_| Error::KeyStoreOperation)
+                },
+                Ok,
+            );
+
+        match key {
+            Ok(k) => {
+                let bytes = k.as_array_ref();
+                Ok(AuraId::from(sr25519::Public::from_raw(*bytes)))
```
mike1729, line 54: Isn't `Ok(k.into())` enough?

bin/node/src/commands.rs
fbielejec: Sorry, but I don't follow the above two comments - we don't have a testnet config as it stands, the devnet is the only "special" kind of config created when the chain-id passed equals "dev" - this is in line with the vanilla BuildSpec command of Substrate (they have an is_dev() method on the command struct).

mike1729 (changes requested):
After Misza's comments, I have only cosmetic comments.

bin/node/src/commands.rs
fbielejec: Done

bin/node/src/commands.rs
fbielejec: a lot cleaner indeed :)

bin/node/src/commands.rs
fbielejec: good spot, refactored

bin/node/src/commands.rs
fbielejec: good spot, a lot more functional, thanks :)

bin/node/src/cli.rs
fbielejec: I wasn't sure we want to do it in the end, but I wanted it mentioned somewhere (rather than nowhere :) ) Anyway: https://cardinal-cryptography.atlassian.net/browse/A0-368

bin/node/src/commands.rs
fbielejec: Ad 0) Intended behaviour of this code is to insert a new key into the keystore or retrieve an existing one - get_or_set?
Ad 1) Done
Ad 2) I don't follow this comment - open-keystore returns an Arc<LocalKeystore> which is a type, not a trait?
Ad 3) Done
Ad 4) Done

bin/node/src/commands.rs
fbielejec: Done, thanks!

michalseweryn (changes requested):
Some more comments

bin/node/src/commands.rs
```diff
@@ -0,0 +1,118 @@
...
+        let genesis_authorities = LOCAL_AUTHORITIES
+            .iter()
+            .map(|authority| {
+                let authority_keystore = self
+                    .open_keystore(authority, &chain_id)
+                    .unwrap_or_else(|_| panic!("Cannot open keystore for {}", authority));
+
+                let aura_key = get_or_set_aura_key(&authority_keystore);
+                let aleph_key = get_or_set_aleph_key(&authority_keystore);
+                let account_id = get_account_id_from_seed::<sr25519::Public>(authority);
```
DamianStraszak, line 68: Don't know whether this should be fixed in this PR, but the account_id's should definitely not be generated from seed in case of testnet and mainnet (for dev runs we don't care much, seed is fine then). 
If you decide to not touch it here, then a high-priority jira task should be created for this -- to be done before the binary is finalized, I guess.

fbielejec: https://cardinal-cryptography.atlassian.net/browse/A0-370

michalseweryn (approved):

bin/node/src/chain_spec.rs
```diff
@@ -49,128 +45,78 @@ where
...
-            millisecs_per_block: Self::param(
-                "millisecs per block",
-                millisecs_per_block,
-                "MILLISECS_PER_BLOCK",
-                DEFAULT_MILLISECS_PER_BLOCK,
-            ),
+    pub fn chain_id(&self) -> &str {
+        match &self.chain_name {
+            Some(id) => id,
+            None => "a0tnet1",
```
timorl, line 82: All the other defaults are for the devnet and this is for the testnet (right?). If this is on purpose please add a comment as to why.

timorl (approved):
One comment about a slight inconsistency, other than that nice work.

bin/node/src/chain_spec.rs
fbielejec: Fixed

PR #116, opened 2021-09-06
Downgrade to 3.0.0

mike1729:
Using @fixxxedpoint  magic now the build works. 

TODO:
  - clippy
  - tests


bin/node/src/service.rs
```diff
@@ -207,64 +178,44 @@ pub fn new_full(mut config: Configuration) -> Result<TaskManager, ServiceError>
...
-        rpc_extensions_builder,
-        on_demand: None,
-        remote_blockchain: None,
-        backend,
-        system_rpc_tx,
-        config,
-        telemetry: telemetry.as_mut(),
-    })?;
-
+    let (_rpc_handlers, _telemetry_connection_notifier) =
```
timorl, line 181: I'm a bit confused – telemetry has been removed as a crate, but there is something here. Do we have telemetry? Have we lost it? Did we ever have or use it?

bin/runtime/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aleph-runtime"
-version = "0.3.0"
+version = "0.1.0"
```
timorl, line 3: Going back with a version sounds very wrong when it's not a literal rollback. Could we bump all the versions instead?

This is a problem in a couple of places, but I'll only write it here, for discussion purposes.

bin/runtime/src/lib.rs
```diff
@@ -194,10 +193,9 @@ impl frame_system::Config for Runtime {
     type SystemWeightInfo = ();
     /// This is used as an identifier of the chain. 42 is the generic substrate prefix.
     type SS58Prefix = SS58Prefix;
-    type OnSetCode = ();
 }
 
-impl pallet_randomness_collective_flip::Config for Runtime {}
+// impl pallet_randomness_collective_flip::Config for Runtime {}
```
timorl, line 198: Commented code.

bin/runtime/src/lib.rs
```diff
@@ -449,21 +445,24 @@ impl_runtime_apis! {
         ) -> sp_inherents::CheckInherentsResult {
             data.check_extrinsics(&block)
         }
+
+        fn random_seed() -> <Block as BlockT>::Hash {
+            RandomnessCollectiveFlip::random_seed()
```
timorl, line 450: I'm pretty sure we were supposed to remove the flip pallet.

finality-aleph/Cargo.toml
```diff
@@ -17,30 +17,28 @@ rand = "0.8"
...
+sp-api = "3.0.0"
+sp-blockchain = "3.0.0"
+sp-consensus= "0.9.0"
+sc-client-api = "3.0.0"
+sp-io= "3.0.0"
 derive_more = "0.99.7"
 env_logger = "0.8.3"
 async-trait = "0.1.42"
-
+
```
timorl, line 42: Whitespace on empty line.

finality-aleph/src/import.rs
```diff
@@ -132,24 +120,20 @@ where
...
 
-        if let Some(justification) =
-            justifications.and_then(|just| just.into_justification(ALEPH_ENGINE_ID))
-        {
-            debug!(target: "afa", "Got justification along imported block #{:?}", number);
-
+        if let Some(justification) = justification {
             if self
-                .send_justification(post_hash, number, (ALEPH_ENGINE_ID, justification))
+                .send_justification(post_hash, number, justification)
```
timorl, line 136: Kinda confusing – did we send `ALEPH_ENGINE_ID` together with any justifications and this was completely unnecessary because we actually only ever get this kind of justification anyway, and other ones will not be validated?

timorl: Or are other types of justifications something that is only introduced in 4.0.0?

timorl: NVM, found the comment about using GRANDPA justification ID. Question though – I saw `FRNK` in the wallet UI as the justification, what will I see now? Does it matter?

finality-aleph/src/data_io.rs
```diff
@@ -233,46 +231,23 @@ where
...
-    }
-}
-
-impl<B: BlockT> aleph_bft::DataIO<AlephDataFor<B>> for DataIO<B> {
+impl<B: BlockT, SC: SelectChain<B>> aleph_bft::DataIO<AlephDataFor<B>> for DataIO<B, SC> {
     type Error = Error;
 
     fn get_data(&self) -> AlephDataFor<B> {
-        let best = *self.best_chain.lock();
+        let new_best_header = self.select_chain.best_chain().expect("No best chain");
```
timorl, line 244: Just to make sure I understand correctly – `best_chain` is not async in 3.0.0, so we don't need all the mechanism we had for refreshing it before?

timorl (changes requested):
Small nitpicks and a lot of questions that are mostly out of curiosity/to learn more about what we are doing.

bin/node/src/service.rs
mike1729: Right now, we have only support for prometheus. If we wanted telemetry, some work would have to be done.

bin/runtime/Cargo.toml
mike1729: Definitely, an rollback error.

bin/runtime/src/lib.rs
mike1729: Removed.

bin/runtime/src/lib.rs
mike1729: It turns out that the trait `sp_block_builder::BlockBuilder<Block>` has a method `fn random_seed() -> <Block as BlockT>::Hash` where the flip pallet is used. It is removed in newer versions.

finality-aleph/Cargo.toml
mike1729: Fixed.

finality-aleph/src/import.rs
mike1729: I believe it doesn't matter. Now probably it will be `Vec<u8>`.

finality-aleph/src/data_io.rs
mike1729: Precisely.

bin/runtime/src/lib.rs
timorl: Ah, interesting. How sure are you nothing uses this method? If pretty sure then I'd prefer the `default` implementation, without the additional pallet.

timorl (approved):
Only one unresolved comment above, other than that LGTM.

PR #117, opened 2021-09-07
A0-304: create common docker file

Jamakase:
## What

Improves dockerfile usage experience.

## How

- Expose ports used in command
- Use Entrypoint default params in `docker-compose`
- Specify ubuntu version of docker for extension
- Add .env file for environment variables

From now docker-compose should be run as `docker-compose --env-file ./docker/.env -f docker/docker-compose.yml up`

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -17,7 +17,7 @@ services:
       - NAME=Damian
       - BASE_PATH=/tmp/Damian
       - NODE_KEY_PATH=/tmp/Damian/libp2p_secret
-      - CHAIN_NAME=/tmp/chainspec.json
+      - CHAIN=/tmp/chainspec.json
       - BOOT_NODES='/ip4/127.0.0.1/tcp/30333/p2p/12D3KooWAbWDCsZiA6EXrXh33E7hBYUFK53b5CSajoPsYL5n2Zdm'
     volumes:
       - ./data/:/tmp/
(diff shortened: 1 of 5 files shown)


bartoszjedrzejewski (approved):
Looks good to me. Make sure @fbielejec takes a look at that also and make sure that the docker-compose works fine for him.

fbielejec (changes requested):
Looks good but the CI fails to start the chain: https://github.com/Cardinal-Cryptography/aleph-node/pull/117/checks?check_run_id=3532928886#step:8:15

```
"host" network_mode is incompatible with port_bindings
```

We can keep instrumenting containers in the host mode and just remove the bindings, or if we intend to use port bindings add a docker network and assign IP addresses to different node containers within. 

I don't have a strong preference but I found it useful in a handful of scenarios to start 3-4 nodes using compose and than another validator as a host process for debugging purposes, so I slightly lean towards just using host networking and no port bindings. YMMV

docker/docker-compose.yml
```diff
@@ -3,98 +3,69 @@ services:
...
-                      --name Damian \
-                      --base-path /tmp/Damian \
-                      --node-key-file /tmp/Damian/libp2p_secret \
-                      --rpc-port 9933 --ws-port 9943 --port 30333 \
-                      --rpc-cors all --rpc-methods Safe \
-                      --execution Native \
-                      --no-mdns --no-prometheus --no-telemetry \
-                      --chain /tmp/chainspec.json \
-                      --bootnodes '/ip4/127.0.0.1/tcp/30333/p2p/12D3KooWAbWDCsZiA6EXrXh33E7hBYUFK53b5CSajoPsYL5n2Zdm'
+    ports:
```
fbielejec, line 10: this is incompatible with host networking, might be the easiest to just remove the binding (see general comment).

Jamakase: Indeed. Haven't noticed that ports were misconfigured here ( partially host network used, partially ports ).

To be honest, I would prefer port binding, but as it is easy to support both of the modes - let's use one that suits dev needs more.

fbielejec (approved):

PR #118, opened 2021-09-07
A0-369: Adapted keybox to changing session keys.

DamianStraszak:


finality-aleph/src/justification.rs
```diff
@@ -232,11 +233,12 @@ where
             Some(authorities) => authorities.to_vec(),
             None => return None,
         };
-
-        Some(KeyBox {
+        // Below we use a fake index 0 of a node -- we never use this keybox for signing so that's OK, but should consider
```
Yithis, line 236: I would love to see here a "TODO" mark

Yithis (approved):

maciejnems (approved):

finality-aleph/src/party.rs
```diff
@@ -113,18 +120,23 @@ where
...
+fn get_node_index(authorities: &[AuthorityId], keystore: &SyncCryptoStorePtr) -> Option<NodeIndex> {
+    let our_consensus_keys: HashSet<_> = SyncCryptoStore::keys(&**keystore, KEY_TYPE)
+        .unwrap()
+        .into_iter()
+        .collect();
+    trace!(target: "afa", "Found {:?} consensus keys in our local keystore {:?}", our_consensus_keys.len(), our_consensus_keys);
     authorities
         .iter()
-        .position(|a| a == my_id)
+        .position(|pkey| our_consensus_keys.contains(&pkey.into()))
```
michalseweryn, line 131: I wanted to suggest using [`SyncCryptoStorePtr::has_keys`](https://docs.rs/sp-keystore/0.9.0/sp_keystore/trait.SyncCryptoStore.html#tymethod.has_keys) method, but its interface is quite ugly, so nevermind...

michalseweryn (changes requested):
Just a single naming issue and two optional improvement suggestions

finality-aleph/src/party.rs
DamianStraszak: Yeah, another point is also that this way I can make a call to the keystore only once and just create a set, while using `has_keys` I would need to make a call for every authority, I guess.

finality-aleph/src/justification.rs
```diff
@@ -232,11 +233,12 @@ where
...
+        // refactoring this in the future so that verification and signing a separate (not combined within a KeyBox).
```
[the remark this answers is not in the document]
DamianStraszak: Ok, I started doing that, but in the middle of the process I realized this requires more changes than I was initially aware of, so now I think it's not worth it...

finality-aleph/src/justification.rs
DamianStraszak: Done.

finality-aleph/src/justification.rs
```diff
@@ -67,7 +68,7 @@ where
     BE: Backend<B> + 'static,
 {
     session_authorities: Arc<Mutex<SessionMap>>,
-    auth_keystore: AuthorityKeystore,
+    auth_keystore: SyncCryptoStorePtr,
```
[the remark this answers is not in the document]
DamianStraszak: Done.

finality-aleph/src/justification.rs
```diff
@@ -232,11 +233,12 @@ where
             Some(authorities) => authorities.to_vec(),
             None => return None,
         };
-
-        Some(KeyBox {
+        // Below we use a fake index 0 of a node -- we never use this keybox for signing so that's OK, but should consider
+        // refactoring this in the future so that verification and signing a separate (not combined within a KeyBox).
+        Some(KeyBox::new(
```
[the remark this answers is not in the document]
DamianStraszak: I made a small improvement so that the keybox is created twice instead of 3 times overall. The general issue with session_authorities I'm aware of and I think that this deserves a deeper refactor, just not sure if we can afford it now :)

michalseweryn (approved):

PR #120, opened 2021-09-09
read AccountIds from cli args

fbielejec:


bin/node/src/chain_spec.rs
```diff
@@ -109,6 +112,33 @@ impl ChainParams {
...
     }
+
+    pub fn account_ids(&self) -> Vec<AccountId> {
+        match &self.account_ids {
+            Some(ids) => ids
+                .iter()
+                .map(|id| {
+                    let id = id.as_str();
+                    let key: sr25519::Public =
+                        hex::decode(id).unwrap().as_slice().try_into().unwrap();
```
mike1729, line 123: Can you call `map(|id| hex_literal::hex!(id).into())`?

mike1729 (changes requested):

bin/node/src/chain_spec.rs
DamianStraszak, line 123: Since this expects public keys on input and not account_ids (technically they might be the same thing, I'm not sure, but the types are different) it would be better to rename this cli argument to `authority_pkeys` or some such. I think you can also write a description of what this argument is and what format is expected, by placing a comment above the appropriate structopt.

DamianStraszak (changes requested):
LGTM, a small comment only.

bin/node/src/chain_spec.rs
michalseweryn: Doesn't the `hex` macro only work with string literals?

fbielejec:
```
error: proc macro panicked
   --> bin/node/src/chain_spec.rs:124:21
    |
124 |                     hex![id].into()
    |                     ^^^^^^^^
    |
    = help: message: expected string literals
```
:cry:

DamianStraszak (approved):

bin/node/src/chain_spec.rs
```diff
@@ -72,15 +72,15 @@ pub struct ChainParams {
...
+    authorities_pkeys: Option<Vec<String>>,
 
     #[structopt(long)]
     n_members: Option<u32>,
 }
 
 impl ChainParams {
     pub fn chain_id(&self) -> &str {
-        match &self.chain_name {
+        match &self.chain_id {
```
michalseweryn, line 83: Can we instead do something like adding `default_value = "a0dnet1"` in `#[structopt(...)]` before `chain_id`?

michalseweryn (approved):
One suggestion, otherwise lgtm

bin/node/src/chain_spec.rs
fbielejec: done

bin/node/src/chain_spec.rs
michalseweryn: Actually, you can add a task in jira instead

mike1729 (approved):

bin/node/src/chain_spec.rs
fbielejec: done

PR #127, opened 2021-09-15
A0-406: Fix script to work with bootstrapping and add non-validator nodes

michalseweryn:
Fix the script to work with chain/node bootstrapping
Added an extra argument, which specifies the number of non validator nodes to run. These nodes will not be included in the list of validators in the chainspec, but will be helpful in testing committee changes.

### DIFF
--- a/run_nodes.sh
+++ b/run_nodes.sh
@@ -2,21 +2,23 @@
 
 function usage(){
   echo "Usage:
-      ./run_nodes.sh [-v N_VALIDATORS] [-n N_NON_VALIDATORS] [-s] [ALEPH_NODE_ARG]...
+      ./run_nodes.sh [-v N_VALIDATORS] [-n N_NON_VALIDATORS] [-b false] [-p BASE_PATH] [ALEPH_NODE_ARG]...
   where 2 <= N_VALIDATORS <= N_VALIDATORS + N_NON_VALIDATORS <= 10
-  (by default, N_VALIDATORS=4 and N_NON_VALIDATORS=0)"
+  (by default, N_VALIDATORS=4, N_NON_VALIDATORS=0 and BASE_PATH=/tmp)"
 }
 
 N_VALIDATORS=4
 N_NON_VALIDATORS=0
 BUILD_ALEPH_NODE='true'
+BASE_PATH='/tmp'
 
-while getopts "v:n:s" flag
+while getopts "v:n:b:p" flag
 do
   case "${flag}" in
     v) N_VALIDATORS=${OPTARG};;
     n) N_NON_VALIDATORS=${OPTARG};;
-    s) BUILD_ALEPH_NODE='false';;
+    b) BUILD_ALEPH_NODE=${OPTARG};;
+    p) BASE_PATH=${OPTARG};;
     *)
       usage
       exit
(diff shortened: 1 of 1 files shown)


fbielejec:
You could make the chain a tad bit ergonomic and user friendly to use by providing argument options with some sensible defaults, and by adding flags e.g. (untested!):

```bash

while getopts v:n:k:h: flag
do
  case "${flag}" in
    v) N_VALIDATORS=${OPTARG};;
    n) N_NON_VALIDATORS=${OPTARG};;
    k) KILL=${OPTARG};;    
    h) HELP=${OPTARG};;
  esac
done

# defaults

if [ -z ${N_VALIDATORS+x} ]; then
   N_VALIDATORS=4
fi

if [ -z ${N_NON_VALIDATORS+x} ]; then
  N_NON_VALIDATORS=0
fi

if [[ $HELP = true ]]; then
  echo "Usage:
    ./run_nodes.sh -v <n_validators> -n <n_non_validators> [Additional Arguments to ./target/debug/aleph-node]
where 2 <= n_validators <= n_validators + n_non_validators <= 10"  
  exit 0;
fi

if [[ $KILL = true ]]; then
  killall -9 aleph-node
  exit 0;
fi

```
then the script becomes `./run_nodes -v 4 -n 0` or `./run_nodes -k true`  etc

michalseweryn:
@fbielejec 

I followed your suggestion, but I resigned from the `-k` flag, as all it does is calling `killall -9 aleph-node`

run_nodes.sh
```diff
@@ -1,45 +1,78 @@
...
 killall -9 aleph-node
 
 set -e
 
 clear
 
-n_members="$1"
-echo "$n_members" > /tmp/n_members
-shift
+cargo build --release -p aleph-node
```
fbielejec, line 33: this could also be a flag, that defaults to true, else LGTM

michalseweryn: done

run_nodes.sh
```diff
@@ -1,45 +1,82 @@
...
+validator_ids_string="${validator_ids[*]}"
+# comma separated ids
+validator_ids_string="${validator_ids_string//${IFS:0:1}/,}"
 
-# cargo build --release -p aleph-node
 
-authorities=(Damian Tomasz Zbyszko Hansu Adam Matt Antoni Michal)
-authorities=("${authorities[@]::$n_members}")
+echo "Bootstrapping chain for nodes 0..$((N_VALIDATORS - 1))"
+./target/release/aleph-node bootstrap-chain --base-path docker/data --chain-id dev --account-ids "$validator_ids_string" > docker/data/chainspec.json
```
fixxxedpoint, line 58: Suggestion about `chainspec.json` and `base-path` parameter: it would be nice to not modify folders that are under git governance by default. We already tackled some bug because one of our scripts was writing to that folder. Copy it somewhere else or something...

michalseweryn: done

run_nodes.sh
```diff
@@ -1,45 +1,84 @@
...
-    ./run_nodes.sh SIZE [Additional Arguments to ./target/debug/aleph-node]
+N_VALIDATORS=4
+N_NON_VALIDATORS=0
+BUILD_ALEPH_NODE='true'
+BASE_PATH='/tmp'
 
-where 2 <= SIZE <= 8"
-    exit
-fi
+while getopts "v:n:b:p" flag
```
fixxxedpoint, line 15: missing colon after `p`

```suggestion
while getopts "v:n:b:p:" flag
```

fbielejec (approved):

Yithis (approved):

fixxxedpoint (approved):

PR #131, opened 2021-09-17
Add support for EXTERNAL_PORT and ENV_FILE to Docker Entrypoint.

bartoszjedrzejewski:


docker/docker_entrypoint.sh
```diff
@@ -11,6 +18,7 @@ BASE_PATH=${BASE_PATH:?'Base path should be specified'}
 RPC_PORT=${RPC_PORT:-9933}
 WS_PORT=${WS_PORT:-9943}
 PORT=${PORT:-30333}
+EXTERNAL_PORT=${EXTERNAL_PORT:-${PORT}}
```
Jamakase, line 21: I see that you are assigning `EXTERNAL_PORT` but where is it used?

Jamakase (approved):

docker/docker_entrypoint.sh
```diff
@@ -1,6 +1,13 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+# load env variables from a file if ENV_FILE is set
+if [[ -n "${ENV_FILE:-}" ]] && [[ -f "${ENV_FILE}" ]]; then
```
Jamakase, line 5: What is the expected behaviour if both env file and env variables are set? Should we prefer file one over env? Or fail?

bartoszjedrzejewski: Expected behaviour is that env file overrides previously set env variables. For now this is what is needed for: https://github.com/Cardinal-Cryptography/aleph-infra/blob/main/initContainer/entrypoint.sh there's no overlap currently.

docker/docker_entrypoint.sh
bartoszjedrzejewski: https://github.com/Cardinal-Cryptography/aleph-infra/blob/main/initContainer/entrypoint.sh this is where it's used. External port can be different than port variable while setting up a public-addr for testnet for example.

skrolikowski-10c (approved):

PR #135, opened 2021-09-17
Network logs

DamianStraszak:
Added useful logging in network.

### DIFF
--- a/finality-aleph/src/network.rs
+++ b/finality-aleph/src/network.rs
@@ -8,7 +8,8 @@ use std::{
     borrow::Cow, collections::HashMap, hash::Hash, iter, marker::PhantomData, pin::Pin, sync::Arc,
 };
 
-use log::{debug, error, trace, warn};
+use log::{debug, error, info, trace, warn};
+use std::time::Duration;
 
 use crate::{
     aggregator::SignableHash,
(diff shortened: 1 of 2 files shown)


mike1729 (approved):

fbielejec (approved):

finality-aleph/src/network.rs
```diff
@@ -599,6 +604,15 @@ where
                         break;
                     }
                 }
+                _ = status_ticker.next() => {
+                    debug!(target: "afa", "Total peers in aleph network {:?}", self.peers.all_peers.len());
+                    for (session_id, session_data) in self.sessions.lock().iter() {
+                        let authenticated: Vec<usize> = self.peers.to_peer.get(session_id).into_iter().map(|hm| hm.keys()).flatten().map(|x| x.0).collect();
+                        let n_members:usize = session_data.keychain.node_count().into();
+                        info!(target: "afa", "Network nodes in session {:?}: {:?}/{:?}", session_id, authenticated.len(), n_members);
```
fbielejec, line 612: debug ?

DamianStraszak: I marked this `info` intentionally, because when we run testnet nodes with standard flags, then the debug logs are not shown... Also, this is the only way to learn about connectivity in the Aleph network, so would be good to see it in testnet logs.

PR #137, opened 2021-09-20
Feature/bootstrap with p2p keys

fbielejec:
This PR introduces:

- p2p key generation happens now in one step, bundled together with the bootstrap-node/bootstrap-chain commands (important for DevOps engineers)
- Some changes to the docker image to facilitate running consensus locally within a private docker/k8 networks using IP (and not DNS) addressing for the bootstrap nodes (passing `discover-local`  sets an inner flag `allow_non_globals_in_dht` to `true`, together with `--allow-private-ipv4` this results in all IP addresses participating in the peer discovery).


bin/node/src/chain_spec.rs
```diff
@@ -33,6 +34,38 @@ pub fn get_from_seed<TPublic: Public>(seed: &str) -> <TPublic::Pair as Pair>::Pu
...
+    }
+}
+
+impl<'de> Deserialize<'de> for SerializablePeerId {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let s = String::deserialize(deserializer)?;
+        let inner = PeerId::from_str(&s).expect("Could not deserialize as a PeerId");
```
michalseweryn, line 64: We shouldn't panic in  a function returning `Result`. Replace `.expect(...)` with something like 
`.map_err(|_| D::Error::custom(format!("iinvalid PeerId: {}", s)))?` instead?

bin/node/src/chain_spec.rs
```diff
@@ -93,6 +132,13 @@ impl ChainParams {
         self.base_path.clone().into()
     }
 
+    pub fn node_key_file(&self) -> &str {
+        match &self.node_key_file {
+            Some(filename) => filename,
+            None => "p2p_secret",
+        }
+    }
+
```
michalseweryn, line 141: Some fields have `default_value` specified in the attribute macro, and some in the getter. Can you add a ticket to unify this (preferrably specify the default value in the attribute macro)?

bin/node/src/commands.rs
```diff
@@ -35,17 +39,49 @@ fn aleph_key(keystore: &impl SyncCryptoStore) -> AlephId {
...
+        }
+        false => {
+            let keypair = libp2p_ed25519::Keypair::generate();
+            let secret = keypair.secret();
+            let secret_hex = hex::encode(secret.as_ref());
+            fs::write(file, secret_hex).expect("Could not write p2p secret");
+            SerializablePeerId::new(PublicKey::Ed25519(keypair.public()).into_peer_id())
+        }
+    }
+}
```
michalseweryn, line 68: There is a syntax for that, and it is called if-and-else ;)

bin/node/src/chain_spec.rs
```diff
@@ -6,14 +6,15 @@ use aleph_runtime::{
     Signature, SudoConfig, SystemConfig, VestingConfig, WASM_BINARY,
 };
 use hex_literal::hex;
+use libp2p::PeerId;
 use sc_service::config::BasePath;
 use sc_service::ChainType;
-use serde::{Deserialize, Serialize};
+use serde::{ser::Serializer, Deserialize, Deserializer, Serialize};
```
michalseweryn, line 12: `ser::` is redundant

bin/node/src/chain_spec.rs
```diff
@@ -60,6 +94,11 @@ pub struct ChainParams {
     #[structopt(long, short = "d", value_name = "PATH", parse(from_os_str))]
     pub base_path: PathBuf,
 
+    /// Specify filename to write node private p2p keys to
+    /// Resulting keys will be stored at: base_path/account_id/node_key_file for each node
+    #[structopt(long)]
+    pub node_key_file: Option<String>,
```
michalseweryn, line 100: Is there a reason why a user would prefer to specify a `node_key_file` different from `p2p_secret`? Maybe we should always use the defailt value?

michalseweryn (changes requested):
Reviewed the changes in the code, unfortunately I am unable to follow what is going on in the `,.sh` and `.yml` files

bin/node/src/chain_spec.rs
fbielejec: https://cardinal-cryptography.atlassian.net/browse/A0-428

bin/node/src/chain_spec.rs
fbielejec: devops guys might need - better safe than sorry IMHO

bin/node/src/chain_spec.rs
fbielejec: :+1:

fbielejec:
> Reviewed the changes in the code, unfortunately I am unable to follow what is going on in the `,.sh` and `.yml` files

I added top level comments in the PR that explain the changes

michalseweryn (approved):

bartoszjedrzejewski (approved):

PR #140, opened 2021-09-22
Feature/token spec

fbielejec:
tested, chainspec can be swapped on a live chain

### DIFF
--- a/bin/node/src/chain_spec.rs
+++ b/bin/node/src/chain_spec.rs
@@ -1,5 +1,6 @@
 use aleph_primitives::{
-    AuthorityId as AlephId, DEFAULT_MILLISECS_PER_BLOCK, DEFAULT_SESSION_PERIOD,
+    AuthorityId as AlephId, ADDRESSES_ENCODING, DEFAULT_MILLISECS_PER_BLOCK,
+    DEFAULT_SESSION_PERIOD, TOKEN_DECIMALS,
 };
 use aleph_runtime::{
     AccountId, AlephConfig, AuraConfig, BalancesConfig, GenesisConfig, SessionConfig, SessionKeys,
@@ -187,8 +188,14 @@ impl ChainParams {
 fn system_properties(token_symbol: String) -> serde_json::map::Map<String, Value> {
     [
         ("tokenSymbol".to_string(), Value::String(token_symbol)),
-        ("tokenDecimals".to_string(), Value::Number(Number::from(12))),
-        ("ss58Format".to_string(), Value::Number(Number::from(42))),
+        (
+            "tokenDecimals".to_string(),
+            Value::Number(Number::from(TOKEN_DECIMALS)),
+        ),
+        (
+            "ss58Format".to_string(),
+            Value::Number(Number::from(ADDRESSES_ENCODING)),
+        ),
     ]
     .iter()
     .cloned()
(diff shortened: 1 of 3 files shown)


bin/node/src/chain_spec.rs
```diff
@@ -183,6 +184,17 @@ impl ChainParams {
     }
 }
 
+fn system_properties(token_symbol: String) -> serde_json::map::Map<String, Value> {
+    [
+        ("tokenSymbol".to_string(), Value::String(token_symbol)),
+        ("tokenDecimals".to_string(), Value::Number(Number::from(12))),
+        ("ss58Format".to_string(), Value::Number(Number::from(42))),
```
maciejnems, line 191: Just a small nit, but maybe we could have token decimals and ss58 fromat defined in aleph_primitives?

Also I wonder if in future we would like to have more tokens apart from `AZERO` on mainnet or `TZERO` on testnet, because that would require changing this part. Probably not and this is an unnecessary concern

docker/docker-compose.yml
```diff
@@ -5,12 +5,16 @@ services:
   damian:
     image: aleph-node:latest
     container_name: damian
+    ports:
+      - 30333:30333
+      - 9933:9933
+      - 9943:9943
```
maciejnems, line 11: I'm not familiar with docker. Does it change anything?

timorl, line 11: What happened here?

timorl (approved):
Looks good, just a single question.

maciejnems (approved):

docker/docker-compose.yml
fbielejec: binds these ports from the container to the host network port.

timorl: Why did you do that now? So that you can reach the locally running or devnet node? Is this just for the devnet, or something more?

primitives/src/lib.rs
```diff
@@ -25,6 +25,8 @@ pub type AuthorityId = app::Public;
 
 pub const DEFAULT_SESSION_PERIOD: u32 = 5;
 pub const DEFAULT_MILLISECS_PER_BLOCK: u64 = 4000;
+pub const TOKEN_DECIMALS: u32 = 5;
```
timorl, line 28: s/5/12

bin/node/src/chain_spec.rs
fbielejec: token symbol can be passed form command line when generating chainspec

docker/docker-compose.yml
fbielejec: exposed to the host to query with curl

bin/node/src/chain_spec.rs
maciejnems: Yes, it cam be passed from command line now, but chainspec supports multiple tokens at once. That's what i meant. But to be fair we don't need it right now

primitives/src/lib.rs
fbielejec: ouch :)

docker/docker-compose.yml
fbielejec: this compose stack is run only in the CI and locally, devnet is just using the same image, but is based on k8 swarm. CI actually uses another docker container (with curl) to connect to the docker container network and run RPC queries (see bash script for testing finality), because it's easier and faster to pull than installing curl on the CI server. Locally everyone has curl installed so it's nice to expose these ports.

You can even potentially run 4 nodes in a compose  stack and connect a 5-th that is `debug` compiled, record what is going on with a tracer/debugger etc.

bin/node/src/chain_spec.rs
fbielejec: yup, the client part of the config does not end up in the genesis block and can be updated later at will

PR #141, opened 2021-09-22
Adjusted aleph-bft data and justification policy, to allow catching-up.

DamianStraszak:


finality-aleph/src/data_io.rs
```diff
@@ -252,7 +272,28 @@ pub(crate) async fn refresh_best_chain<B: BlockT, SC: SelectChain<B>>(
...
-                *best_chain.lock() = AlephData::new(new_best_header.hash(), *new_best_header.number());
+                if *new_best_header.number() <= max_block_num {
+                    *data.lock() = AlephData::new(new_best_header.hash(), *new_best_header.number());
+                } else {
+                    // we check if prev_best_header is an ancestor of new_best_header:
+                    if data.lock().number < max_block_num {
+                        let reduced_header = reduce_to_num(new_best_header.clone(), max_block_num);
+                        *data.lock() = AlephData::new(reduced_header.hash(), *reduced_header.number());
+                    } else {
+                        let is_ancestor = if let Some(prev_header) = prev_best_header {
```
timorl, line 283: This whole thing seems unnecessary – replacing the best block by the latest within-session ancestor of the new best chain is what we want to do anyway, right? Unless I'm reading the code wrong this is what we end up doing anyway, only sometimes doing some unnecessary additional ancestor tracking.

finality-aleph/src/data_io.rs
```diff
@@ -239,11 +239,31 @@ pub(crate) struct DataIO<B: BlockT> {
...
     mut exit: oneshot::Receiver<()>,
-) {
+) where
+    B: BlockT,
+    C: crate::ClientForAleph<B, BE> + Send + Sync + 'static,
+    BE: Backend<B> + 'static,
+    SC: SelectChain<B> + 'static,
+{
+    // A function to reduce block to the level given by num, by traversing down via parents.
+    let reduce_to_num = |header: B::Header, num: NumberFor<B>| {
```
timorl, line 255: I'm pretty sure you can just make this a normal function taking the client as an argument, right?

finality-aleph/src/data_io.rs
timorl: Actually, the whole if structure seems irrelevant – `reduce_to_num` is an approximate noop when `header.num() <= num`, so you can just always run it and replace with the result. This would make even the function unnecessary, since the code would be in only one place.

I am quite worried that I might be missing some big thing that makes all I wrote extremely stupid. xd

finality-aleph/src/data_io.rs
timorl: Although see below about the function being needed at all.

timorl (changes requested):
I'm really unsure whether some of my reasoning below is correct. <_<

finality-aleph/src/data_io.rs
DamianStraszak: The point why this is so complicated is that it optimizes for the case when `best_block` is much higher than `max_block_num`. The simplest thing would be to always go down via ancestors from `new_best_header` to `max_block_num`: however, this might be quite expensive (it's not clear to me whether this even sits in RAM, or maybe we need to read stuff from disk, if the gap is say 20k blocks)... So the optimization is to essentially remember the last call to `best_block` as `prev_best_header` and check is `new_best_header` extends the branch of `prev_best_header` -- if so, then we don't need to do the expensive backtracking again, because we did it in the last try...

If there is a simple, yet correct way to do than I would be happy to hear :P

timorl: Ah, yes, now I see. >.< Thanks for the explanation.

Since this is a bit confusing maybe a comment with a slightly longer explanation would be useful. Although I am not very sure about that, I kinda feel stupid for not understanding it before, so it might not actually be necessary.

finality-aleph/src/data_io.rs
DamianStraszak: I could make it into a normal function, but it annoys me that I need to make it generic over 3 arguments (B, C, BE + their bounds). Somehow I thought that this way it's a little bit more elegant, shall I change it?

timorl: I have a strong preference for code-level functions, which is way stronger than the dislike of multiple template arguments for me. I think this keeps the code much more readable, but it's also just a preference. I would suggest changing it, but it's not a requirement for approval from me.

finality-aleph/src/party.rs
```diff
@@ -209,13 +209,12 @@ async fn run_aggregator<B, C, BE>(
...
-                            if *header.number() <= last_block_in_session {
-                                aggregator.start_aggregation(header.hash()).await;
-                                last_finalized = header.hash();
+                        if let Some(data) = should_finalize(last_finalized, new_block_data, client.as_ref(), last_block_in_session) {
+                            if data.number <= last_block_in_session {
+                                aggregator.start_aggregation(data.hash).await;
+                                last_finalized = data.hash;
                             }
-                            if *header.number() >= last_block_in_session {
+                            if data.number >= last_block_in_session {
```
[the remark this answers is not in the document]
DamianStraszak: Yes, correct, I fixed this and what was a little above it.

finality-aleph/src/finalization.rs
```diff
@@ -41,40 +41,44 @@ where
 /// the sequence of headers of the blocks on the path from `last_finalized` to `new_data`
 /// excluding the header corresponding to `last_finalized`, or an empty sequence if
 /// `new_data` is not a descendant of `last_finalized`.
-pub(crate) fn chain_extension_step<BE, B, C>(
+pub(crate) fn should_finalize<BE, B, C>(
```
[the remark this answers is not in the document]
DamianStraszak: Updated

finality-aleph/src/finalization.rs
```diff
@@ -83,21 +87,20 @@ where
             Ok(Some(header)) => header,
             _ => {
                 error!(target: "afa", "No header for hash {}", hash);
-                return VecDeque::new();
+                return None;
```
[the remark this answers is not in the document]
DamianStraszak: Yes, this is the purpose of the loop.
Updated commend.

finality-aleph/src/data_io.rs
DamianStraszak: Yeah, it's probably not easy to see this at first -- added a comment.

finality-aleph/src/data_io.rs
DamianStraszak: Ok, makes sense to me. Changed to a regular function.

finality-aleph/src/data_io.rs
```diff
@@ -277,17 +288,17 @@ pub(crate) async fn refresh_best_chain<B, BE, SC, C>(
                 } else {
                     // we check if prev_best_header is an ancestor of new_best_header:
                     if data.lock().number < max_block_num {
-                        let reduced_header = reduce_to_num(new_best_header.clone(), max_block_num);
+                        let reduced_header = reduce_header_to_num::<B, BE, C>(client.clone(), new_best_header.clone(), max_block_num);
```
timorl, line 291: Are the template args necessary here? I would have thought `client` contains all the information the function needs.

timorl (approved):
Just one note, other than that LGTM.

finality-aleph/src/data_io.rs
DamianStraszak: Cool, it works without the explicit args, thanks :)

PR #144, opened 2021-09-25
Adds support for multisignatures

mike1729:


bin/node/src/service.rs
```diff
@@ -151,10 +151,10 @@ pub fn new_full(mut config: Configuration) -> Result<TaskManager, ServiceError>
...
     } = new_partial(&config)?;
 
-    config
-        .network
-        .extra_sets
-        .push(finality_aleph::peers_set_config());
+    let mut peers_set_config = finality_aleph::peers_set_config();
+    peers_set_config.set_config.reserved_nodes =
+        config.network.default_peers_set.reserved_nodes.clone();
+    config.network.extra_sets.push(peers_set_config);
```
michalseweryn, line 157: That `mut` here is kinda awkward. Can we pass a `Vec<MultiaddrWithPeerId>` as a parameter to `peers_set_config`?

bin/runtime/Cargo.toml
```diff
@@ -53,32 +54,33 @@ substrate-wasm-builder = { version = "4.0.0"}
...
+    "pallet-aleph/std",
+    "pallet-aura/std",
+    "pallet-balances/std",
+    "pallet-randomness-collective-flip/std",
+    "pallet-session/std",
+    "pallet-sudo/std",
+    "pallet-timestamp/std",
+    "pallet-transaction-payment/std",
+    "pallet-vesting/std",
+	"pallet-multisig/std",
```
michalseweryn, line 69: this line is misaligned ;)

bin/runtime/src/lib.rs
```diff
@@ -356,6 +356,29 @@ impl pallet_vesting::Config for Runtime {
...
+pub const fn deposit(items: u32, bytes: u32) -> Balance {
+    items as Balance * 15 * CENTS + (bytes as Balance) * 6 * CENTS
+}
+parameter_types! {
+    // One storage item; key size is 32; value is size 4+4+16+32 bytes = 56 bytes.
+    pub const DepositBase: Balance = deposit(1, 88);
+    // Additional storage item size of 32 bytes.
+    pub const DepositFactor: Balance = deposit(0, 32);
+    pub const MaxSignatories: u16 = 100;
+}
```
michalseweryn, line 370: I must admit I do not follow what is going here. Are we sure that the parameters are correct, or are we hoping that polkadot's parameters will work for us?

michalseweryn (changes requested):

bin/runtime/Cargo.toml
mike1729: A tab snicked in :)

bin/runtime/src/lib.rs
mike1729: Those parameters govern the weight and the price of the calls from the multisig pallet, so I believe that pulling those values from polkadot is ok.

bin/node/src/service.rs
mike1729: Done.

fbielejec: `finality_aleph::peers_set_config();` seems to be doing `SetConfig::default()` modulo the number of in/out of peers:
https://github.com/paritytech/substrate/blob/nil/client/network/src/config.rs#L556

finality-aleph/src/lib.rs
```diff
@@ -48,8 +48,8 @@ pub fn peers_set_config() -> sc_network::config::NonDefaultSetConfig {
         set_config: sc_network::config::SetConfig {
             // This seems to be a way to configure the AlephBFT network to have a prespecified set of nodes or at least
             // set suitable limits on the number of nodes we should connect to.
-            in_peers: 0,
-            out_peers: 0,
+            in_peers: 50,
+            out_peers: 50,
```
fbielejec, line 55: I honestly never saw these numbers as having any effect whatsoever when trying it. 
`NetworkConfiguration` struct seems to have yet another `SetConfig`:
https://github.com/paritytech/substrate/blob/nil/client/network/src/config.rs#L419
that carries it's own setting for in and out peers (25 and 75 by default). 

Looking through the substrate code there seems to be a few places where they iterate over the `extra_sets.reserved_nodes` but not in/out peers... 
I have a suspicion someone just re-used the struct to carry the vector of reserved_nodes...

Maybe just call `SetConfig::default()` as extra_sets in the upstream and remove this code entirely?

michalseweryn (approved):

finality-aleph/src/lib.rs
mike1729: Done.

bin/node/src/service.rs
mike1729: We need to set our name, despite that we do nothing as apparently it hasn't had any effect :(

fbielejec: we could still comply with the note though, something along the lines of:
```
pub fn peers_set_config() -> sc_network::config::NonDefaultSetConfig {
    let mut non_default_set = sc_network::config::NonDefaultSetConfig::new(
        network::ALEPH_PROTOCOL_NAME.into(),
        // max_notification_size should be larger than the maximum possible honest message size (in bytes).
        // Max size of alert is UNIT_SIZE * MAX_UNITS_IN_ALERT ~ 100 * 5000 = 50000 bytes
        // Max size of parents response UNIT_SIZE * N_MEMBERS ~ 100 * N_MEMBERS
        // When adding other (large) message types we need to make sure this limit is fine.
        1024 * 1024,
    );

    non_default_set.set_config = sc_network::config::SetConfig::default();
    non_default_set.set_config
}
```

fbielejec (approved):

bin/node/src/service.rs
mike1729: done.

PR #145, opened 2021-09-27
Fixed old hanging sessions in aleph network.

DamianStraszak:


mike1729 (approved):

michalseweryn (approved):

finality-aleph/src/network.rs
```diff
@@ -614,10 +654,6 @@ where
                     }
                 }
             }
-
-            self.sessions
```
fbielejec, line 618: if we don't use these statuses for filtering, maybe they are not needed at all?

PR #152, opened 2021-10-04
Add ability to reach ws and rpc ports from external sources.

bartoszjedrzejewski:


mike1729 (approved):

timorl (approved):

timorl (changes requested):

docker/docker_entrypoint.sh
```diff
@@ -35,7 +35,7 @@ ARGS=(
   --chain "${CHAIN}"
   --node-key-file "${NODE_KEY_PATH}"
   --rpc-port "${RPC_PORT}" --ws-port "${WS_PORT}" --port "${PORT}"
-  --rpc-cors all --rpc-methods Safe # TODO: should we allow to specify them?
+  --rpc-cors all
   --no-prometheus --no-telemetry # Currently not using. plan to start as soon as capacity is available
```
fbielejec, line 39: could we add `--ws-max-connections <int>` while we are at it? It could default to 100 (which is the default anyway as we have learnt yesterday)

fbielejec (commented):
Pls add one more flag as per comment, else LGTM

docker/docker_entrypoint.sh
timorl: Oh, right, preferable add it only to the bootnode and set it to "a lot", e.g. 10000 or something.

PR #155, opened 2021-10-08
 [Don't Merge] Clean-up of the Aleph pallet and preparation for authority discovery.

DamianStraszak:
Cleaned up the Aleph pallet: removed unnecessary genesis arguments and improved internal logic regarding validator changes.

Added new runtime api calls for the purpose of our authority discovery.

### DIFF
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -15,10 +15,7 @@ use sp_consensus_aura::sr25519::AuthorityId as AuraId;
 use sp_core::{crypto::KeyTypeId, OpaqueMetadata};
 use sp_runtime::{
     create_runtime_str, generic, impl_opaque_keys,
-    traits::{
-        AccountIdLookup, BlakeTwo256, Block as BlockT, ConvertInto, IdentifyAccount, OpaqueKeys,
-        Verify,
-    },
+    traits::{AccountIdLookup, BlakeTwo256, Block as BlockT, ConvertInto, OpaqueKeys},
     transaction_validity::{TransactionSource, TransactionValidity},
     ApplyExtrinsicResult, MultiSignature,
 };
@@ -42,7 +39,10 @@ pub use frame_support::{
     },
     StorageValue,
 };
-use primitives::{ApiError as AlephApiError, AuthorityId as AlephId};
+use primitives::{
+    AccountId as AlephAccountId, ApiError as AlephApiError, AuthDiscoveryError,
+    AuthorityId as AlephId,
+};
 
 pub use pallet_balances::Call as BalancesCall;
 pub use pallet_timestamp::Call as TimestampCall;
@@ -58,9 +58,7 @@ pub type BlockNumber = u32;
 /// Alias to 512-bit hash when used in the context of a transaction signature on the chain.
 pub type Signature = MultiSignature;
 
-/// Some way of identifying an account on the chain. We intentionally make it equivalent
-/// to the public key of our transaction signing scheme.
-pub type AccountId = <<Signature as Verify>::Signer as IdentifyAccount>::AccountId;
+pub type AccountId = AlephAccountId;
 
 /// The type for looking up accounts. We don't expect more than 4 billion of them, but you
 /// never know...
@@ -106,10 +104,10 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("aleph-node"),
     impl_name: create_runtime_str!("aleph-node"),
     authoring_version: 1,
-    spec_version: 1,
+    spec_version: 2,
     impl_version: 1,
     apis: RUNTIME_API_VERSIONS,
-    transaction_version: 1,
+    transaction_version: 2,
 };
 
 /// This determines the average expected block time that we are targetting.
@@ -526,7 +524,8 @@ impl_runtime_apis! {
     }
 
     impl primitives::AlephSessionApi<Block> for Runtime {
-        fn next_session_authorities() -> Result<Vec<AlephId>, AlephApiError> {
+
+        fn next_session_authorities() -> Result<Vec<AlephId>, AlephApiError>{
             Aleph::next_session_authorities()
         }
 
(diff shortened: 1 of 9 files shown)


pallet/src/lib.rs
```diff
@@ -158,11 +155,43 @@ pub mod pallet {
...
 
-        pub fn next_session_authorities() -> Result<Vec<T::AuthorityId>, AlephApiError> {
+        pub fn next_session_authorities() -> Vec<T::AuthorityId> {
             Session::<T>::queued_keys()
                 .iter()
-                .map(|(_, key)| key.get(T::AuthorityId::ID).ok_or(AlephApiError::DecodeKey))
-                .collect::<Result<Vec<T::AuthorityId>, AlephApiError>>()
+                .map(|(_, key)| {
+                    key.get(T::AuthorityId::ID)
+                        .expect("Keys in storage decode correctly.")
```
kostekIV, line 163: Do we want to potentially panic in the `pallet` code?

pallet/src/lib.rs
```diff
@@ -158,11 +155,43 @@ pub mod pallet {
...
+                })
+                .collect()
+        }
+
+        pub fn future_session_validator_ids(
+            session_id: u32,
```
kostekIV, line 169: So this function either returns the next validators or the current validator set (ignoring the error for now). The naming is a little bit confusing I would assume that it always returns the next validator set, not sure what is an appropriate name here. `validator_ids_for_session`?

pallet/src/lib.rs
DamianStraszak: 1) This cannot quite panic, the only way for this to panic is to use a wrong `AuthoridyId::ID`, which you can catch during the first compilation. Also, why would panicking in the runtime be worse than panicking in the client code? In the previous version the same panic was just one level up...
2) This makes the output type of the calls
`next_session_validator_keys`
`current_session_validator_keys`
identical.

kostekIV: Sure I agree. I am only asking this because the substrate explicitly says the runtime should handle errors and not panic: `https://substrate.dev/docs/en/knowledgebase/runtime/errors`. I know there are more places like this in code so I wanted to know what is our stand on this topic.

pallet/src/lib.rs
```diff
@@ -79,12 +77,15 @@ pub mod pallet {
             session_for_validators_change: u32,
         ) -> DispatchResult {
             ensure_root(origin)?;
+            let current_session = Session::<T>::current_index();
+            let trigger_at_session = if current_session + 2 > session_for_validators_change {
+                current_session + 2
```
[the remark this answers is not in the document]
kostekIV: Thanks, now I get it

bin/runtime/src/lib.rs
```diff
@@ -56,7 +53,7 @@ pub type Signature = MultiSignature;
 
 /// Some way of identifying an account on the chain. We intentionally make it equivalent
 /// to the public key of our transaction signing scheme.
-pub type AccountId = <<Signature as Verify>::Signer as IdentifyAccount>::AccountId;
+pub type AccountId = AlephAccountId;
```
timorl, line 56: Perhaps the comment above this line should be updated/moved to `primitives`?

pallet/src/lib.rs
```diff
@@ -16,14 +16,12 @@ pub mod pallet {
     use super::*;
     use frame_support::{
         pallet_prelude::*,
-        sp_runtime::{traits::OpaqueKeys, RuntimeAppPublic},
+        sp_runtime::{traits::Convert, traits::OpaqueKeys, RuntimeAppPublic},
```
timorl, line 19: Huh, clippy doesn't complain about `traits` twice?

pallet/src/lib.rs
```diff
@@ -79,12 +77,20 @@ pub mod pallet {
             session_for_validators_change: u32,
         ) -> DispatchResult {
             ensure_root(origin)?;
+            let current_session = Session::<T>::current_index();
+            // Because of how pallet `Session` works, it is not possible to make the change happen at
+            // `session_for_validators_change` if it is < `current_session` + 2. In order to store
+            // accurate information in our storage and give the user accurate feedback (in the form
```
timorl, line 83: Something that has been bugging me a bit – by default I would return an error in such a case, since when you cannot do exactly what the user asks it's often better to avoid guessing. Is there a good reason in this case to assume that the user would want this change to happen at the earliest opportunity?

timorl: Followup comment – would the obvious solution of adding a `force_change` boolean argument be possible? Is complicating the API worth it?

pallet/src/lib.rs
timorl: The substrate doc seems pretty adamant about this, so I would support changing this to an `Err`.

pallet/src/lib.rs
timorl: I would say that it's pretty clear due to the argument, so imo the name is fine.

pallet/src/lib.rs
```diff
@@ -158,11 +160,43 @@ pub mod pallet {
...
+
+        pub fn future_session_validator_ids(
+            session_id: u32,
+        ) -> Result<Vec<T::ValidatorId>, ApiError> {
+            let current_session = Session::<T>::current_index();
+            if session_id < current_session {
+                return Err(ApiError::TooLowSessionId);
+            }
+            // If the session_id is higher than the one for which the validator rotation is scheduled, we output
+            // the new validator set, otherwise we output the current set (we read it off the session pallet).
```
timorl, line 181: It would be nice to document the behaviour, especially in context of potentially getting inaccurate information – it should be made clear to users that caching a response from this function about a session more than 2(?) away from the current session is a bad idea, since an authority change might happen.

Bonus question about the far future – will we at some point keep a more complex schedule of authorities? E.g. information about authorities for the next 50 sessions, with explicit lack of knowledge of later sessions? I'm asking because it would be nice to be able to return a `TooHighSessionId` error in such cases, rather than potentially providing inaccurate information (it's probably not a good idea now though).

timorl (changes requested):
Mostly nits, and some questions with probably short answers.

pallet/src/lib.rs
timorl: Exactly, and return an error when cancelling.

pallet/src/lib.rs
DamianStraszak: Replaced.

PR #156, opened 2021-10-11
Enable setting unit creation delay with a flag

pmikolajczyk41:
A new flag `--base-unit-creation-delay` enables manipulating delay for unit creation. The mechanics is analogous to `--millisecs-per-block`.

### DIFF
--- a/bin/node/src/chain_spec.rs
+++ b/bin/node/src/chain_spec.rs
@@ -1,6 +1,6 @@
 use aleph_primitives::{
-    AuthorityId as AlephId, ADDRESSES_ENCODING, DEFAULT_MILLISECS_PER_BLOCK,
-    DEFAULT_SESSION_PERIOD, TOKEN_DECIMALS,
+    AuthorityId as AlephId, ADDRESSES_ENCODING, DEFAULT_BASE_UNIT_CREATION_DELAY,
+    DEFAULT_MILLISECS_PER_BLOCK, DEFAULT_SESSION_PERIOD, TOKEN_DECIMALS,
 };
 use aleph_runtime::{
     AccountId, AlephConfig, AuraConfig, BalancesConfig, GenesisConfig, SessionConfig, SessionKeys,
@@ -110,6 +110,9 @@ pub struct ChainParams {
     pub millisecs_per_block: Option<u64>,
 
     #[structopt(long)]
+    pub base_unit_creation_delay: Option<u64>,
+
+    #[structopt(long)]
     pub chain_name: Option<String>,
 
     #[structopt(long)]
@@ -139,6 +142,11 @@ impl ChainParams {
     pub fn millisecs_per_block(&self) -> u64 {
         self.millisecs_per_block
             .unwrap_or(DEFAULT_MILLISECS_PER_BLOCK)
+    }
+
+    pub fn base_unit_creation_delay(&self) -> u64 {
+        self.base_unit_creation_delay
+            .unwrap_or(DEFAULT_BASE_UNIT_CREATION_DELAY)
     }
 
     pub fn session_period(&self) -> u32 {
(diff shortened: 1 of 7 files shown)


bin/node/src/chain_spec.rs
```diff
@@ -109,6 +109,9 @@ pub struct ChainParams {
     #[structopt(long)]
     pub millisecs_per_block: Option<u64>,
 
+    #[structopt(long)]
+    pub base_unit_creation_delay: Option<u64>,
```
michalseweryn, line 113: Effectively, we only use it after convertig it to `f64`, is there anything preventing us from changing its type globally to `f64`?

finality-aleph/src/party.rs
```diff
@@ -543,14 +551,15 @@ pub(crate) fn create_aleph_config(
     n_members: usize,
     node_id: NodeIndex,
     session_id: SessionId,
+    base_unit_creation_delay: u64,
```
michalseweryn, line 554: Can you change the type to `BaseUnitCreationDelay`? the float should be unwrapped as late as possible, so that its meaning is reflected by the type for as long as possible. Ideally, I would see this in the `exponential_slowdown`, but this would require changing public interface, so maybe not?

finality-aleph/src/lib.rs
```diff
@@ -59,6 +59,9 @@ pub struct SessionPeriod(pub u32);
 #[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, Ord, PartialOrd, Encode, Decode)]
 pub struct MillisecsPerBlock(pub u64);
 
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, Ord, PartialOrd, Encode, Decode)]
+pub struct BaseUnitCreationDelay(pub u64);
```
michalseweryn, line 63: Please add a comment documenting what units we are using (milliseconds)

bin/node/src/chain_spec.rs
```diff
@@ -141,6 +144,11 @@ impl ChainParams {
             .unwrap_or(DEFAULT_MILLISECS_PER_BLOCK)
     }
 
+    pub fn base_unit_creation_delay(&self) -> u64 {
```
michalseweryn, line 147: Can we return BaseUnitCreationDelay here? And to make the code consistent, can you change the return types of `ChainParams::millisecs_per_block()` and `ChainParams::session_period()` to `MillisecsPerBlock` and `SessionPeriod`, respectively? By the way, shouldn't these types should probably defined in `aleph-primitives`?

michalseweryn (changes requested):
Several small changes requested

bin/node/src/chain_spec.rs
pmikolajczyk41: You are right, however, when using `f64` the compiler screams:
```
the trait `WrapperTypeDecode` is not implemented for `f64`
```
for the piece of code below (from `primitives/src/lib.rs`):
```rust
sp_api::decl_runtime_apis! {
    pub trait AlephSessionApi
    {
        fn next_session_authorities() -> Result<Vec<AuthorityId>, ApiError>;
        fn authorities() -> Vec<AuthorityId>;
        fn session_period() -> u32;
        fn base_unit_creation_delay() -> f64;
    }
}
```

michalseweryn: Ugh, this compiler error is so confusing. It means that actually `Decode` is not implemented for `f64`, and it checks if `WrapperTypeDecode` is implemented for `f64` because we have

```
impl<T, X> Decode for X where
    T: Decode + Into<X>,
    X: WrapperTypeDecode<Wrapped = T>
```

Apparently, acccording to authors of `codec`, we don't need floating point types. And maybe they are right? why not store the delay as an unsigned integer denoting milliseconds?

finality-aleph/src/lib.rs
pmikolajczyk41: Comment moved

PR #158, opened 2021-10-15
Monitoring

pmikolajczyk41:
Adds:

- docker-compose image with Prometheus and Grafana
- dashboard for main checkpoints and txps

@coauthored with @kostekIV

### DIFF
--- a/benchmark/docker-compose.yml
+++ b/benchmark/docker-compose.yml
@@ -2,8 +2,9 @@ version: '3'
 
 services:
   prometheus:
-    network_mode: "host" # only needed if run locally
     image: prom/prometheus:v2.30.3
+    extra_hosts:
+      - host.docker.internal:host-gateway
     ports:
       - "9090:9090"
     volumes:
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -2,22 +2,28 @@
 
 ## Installation and usage
 
-1. You will need `docker-compose` installed.
+1. You will need `docker` (version `20.10+`) and `docker-compose` installed.
 2. Specify where Prometheus should be fetching metrics from, i.e. add IPs of the machines running the protocol 
 (together with the port, usually `9615`) to the `targets` entry in `prometheus.yml`, e.g.:
 ```yml
 ...
   - targets: [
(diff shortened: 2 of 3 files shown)


benchmark/docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+version: '3'
+
+services:
+  prometheus:
+    network_mode: "host" # only needed if run locally
+    image: prom/prometheus:latest
```
fbielejec, line 6: could we fix versions just in case (to current recent)? Sometimes a breaking change can sneak in and a `latest` flag just dumbly follows it. At the same time `latest` pulls the `latest` image when you start the stack and then does not update it unless you remove the image, can easilly end up with two ppl silently having two different versions trying to figure out what happened

benchmark/docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+    ports:
+      - "9090:9090"
+    volumes:
+      - ./prometheus.yml:/etc/prometheus/prometheus.yml
+
+  grafana:
+    image: grafana/grafana:latest
```
fbielejec, line 13: same, I suggest fixing the version

benchmark/docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+version: '3'
+
+services:
+  prometheus:
+    network_mode: "host" # only needed if run locally
+    image: prom/prometheus:latest
+    ports:
```
fbielejec, line 7: host networking and port bindings are incompatible, have you tested if this stack starts?

benchmark/docker-compose.yml
pmikolajczyk41: done

benchmark/docker-compose.yml
pmikolajczyk41: it worked, but I replaced it with `extra_hosts` as suggested off-line

benchmark/docker-compose.yml
pmikolajczyk41: done

PR #161, opened 2021-10-18
Feature/tx benchmarks

fbielejec:
### Tx flooding tool written in Rust

Using:

```bash
cd flooder/
cargo build --release

RUST_LOG=warn ./target/release/flooder --phrase <mnemonic-or-path> --nodes 127.0.0.1:9943 127.0.0.1:9944 127.0.0.1:9945 127.0.0.1:9946 --transactions 10000 --throughput 2000
```

Following settings will result in 10_000 txs split over 10 independent tasks and sent to 4 nodes in round-robin fashion


flooder/src/main.rs
```diff
@@ -0,0 +1,365 @@
...
+                // retrieve signed txs
+                let signed_tx = &txs[thread as usize][batch as usize][tx as usize];
+
+                // send txs using a connection pool
+                futures_batch.push(send_tx(
+                    pool.get(tx as usize % pool.len()).unwrap().to_owned(),
+                    signed_tx.to_owned(),
+                    Arc::clone(histogram),
+                ));
+            }
```
kostekIV, line 142: I think one future/thread should get more txs to send at once. So the `for tx in 0..transactions_per_batch` I would put into the future code.

flooder/src/main.rs
```diff
@@ -0,0 +1,365 @@
...
+        }
+        // block on batches of futures
+        join_all(futures_batch).await;
+    }
+}
+
+async fn estimate_tx_fee(
```
kostekIV, line 149: Does this need to be an `async` function>

flooder/src/main.rs
```diff
@@ -0,0 +1,365 @@
...
+type Block = generic::Block<Header, OpaqueExtrinsic>;
+
+#[tokio::main(flavor = "multi_thread")]
+async fn main() -> Result<(), anyhow::Error> {
+    env_logger::init();
+    let config: Config = Config::parse();
+    info!("Starting benchmark with config {:#?}", &config);
+
+    let (account, _seed) =
+        sr25519::Pair::from_phrase(&config::read_phrase(config.phrase), None).unwrap();
```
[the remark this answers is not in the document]
fbielejec: Doesn't this conflate _phrase_ (mnemonic) with _seed_ though? For the well-known-accounts  the mnemonic is known anyways IMO? 
Potentially, why not add another command line argument (--seed) and support both?

flooder/src/config.rs
```diff
@@ -0,0 +1,35 @@
+use clap::Parser;
+use std::{fs, path::PathBuf};
+
+#[derive(Debug, Parser)]
+#[clap(version = "1.0")]
+pub struct Config {
+    /// URL address(es) of the nodes to send transactions to
+    #[clap(long, default_value = "127.0.0.1:9943")]
```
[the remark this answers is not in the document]
fbielejec: in CI this is the bootstrap nodes WS endpoint (I tend to use dockerized chain for local development). 9945 is good middle ground :)

flooder/src/main.rs
fbielejec: good catch!

flooder/src/main.rs
fbielejec: refactored, see if you like it more this way. Found a great lib (rayon, https://github.com/rayon-rs/rayon) that is an abstraction over async/await and provides sugar for parallel iterators

flooder/src/main.rs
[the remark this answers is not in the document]
kostekIV: Cool, thanks

flooder/src/main.rs
```diff
@@ -0,0 +1,308 @@
...
+
+fn flood(
+    pool: Vec<Api<sr25519::Pair, WsRpcClient>>,
+    txs: Vec<TransferTransaction>,
+    // total_threads: u64,
+    // total_batches: u64,
+    transactions_per_batch: usize,
+    histogram: &Arc<Mutex<HdrHistogram<u64>>>,
+) {
+    txs.par_chunks(transactions_per_batch).for_each(|batch| {
```
kostekIV, line 103: looks really nice!

flooder/src/main.rs
```diff
@@ -0,0 +1,308 @@
...
+    );
+
+    Ok(())
+}
+
+fn flood(
+    pool: Vec<Api<sr25519::Pair, WsRpcClient>>,
+    txs: Vec<TransferTransaction>,
+    // total_threads: u64,
+    // total_batches: u64,
```
kostekIV, line 99: We can remove it now, right?

flooder/src/config.rs
```diff
@@ -0,0 +1,39 @@
...
+    #[clap(long, default_value = "10000")]
+    pub transactions: u64,
+
+    /// what throughput to use (transactions/s)
+    #[clap(long, default_value = "1000")]
+    pub throughput: u64,
+
+    /// how many threads to create
+    #[clap(long, default_value = "10")]
+    pub threads: u64,
```
kostekIV, line 21: Correct me if I am wrong, but with rayon-rs the threads are no longer needed?

kostekIV: Or alternatively, we can set threads number for rayon, but I think I would prefer the default nr to be chosen.

flooder/src/config.rs
```diff
@@ -0,0 +1,39 @@
+use clap::Parser;
+use std::{fs, path::PathBuf};
+
+#[derive(Debug, Parser)]
+#[clap(version = "1.0")]
+pub struct Config {
+    /// URL address(es) of the nodes to send transactions to
+    #[clap(long, default_value = "127.0.0.1:9945g")]
```
kostekIV, line 8: typo leftover `g`?

flooder/src/main.rs
```diff
@@ -0,0 +1,308 @@
...
+            Some(seed) => sr25519::Pair::from_string(&seed, None).unwrap(),
+            None => panic!("Needs --phrase or --seed"),
+        },
+    };
+
+    let pool = create_connection_pool(config.nodes);
+    let connection = pool.get(0).unwrap();
+
+    let total_users = config.transactions;
+    let transactions_per_batch = config.throughput / config.threads;
```
kostekIV, line 47: Now it should be something like 
```suggestion
    let transactions_per_batch = config.throughput / current_num_threads();
```

flooder/src/config.rs
fbielejec: emacs bites back.. fixed

flooder/src/config.rs
fbielejec: done

kostekIV (approved):

Yithis (approved):
Nice!

flooder/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+edition = "2018"
+
+# client
+[dependencies.substrate-api-client]
+git = "https://github.com/scs/substrate-api-client"
+rev = "f9d3a773bb5c76a0c0689b012fc08599beb90ab7"
+
+[dependencies]
+# Substrate dependencies
+sp-core = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", branch = "master", features = ["full_crypto"] }
```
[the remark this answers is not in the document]
fbielejec: incompatible with client lib, they follow master :shrug:

PR #181, opened 2021-11-02
AZ-194: Set default block size limit

pmikolajczyk41:


bin/node/src/service.rs
```diff
@@ -217,13 +217,15 @@ pub fn new_full(mut config: Configuration) -> Result<TaskManager, ServiceError>
...
-        let proposer_factory = sc_basic_authorship::ProposerFactory::new(
+        let mut proposer_factory = sc_basic_authorship::ProposerFactory::new(
             task_manager.spawn_handle(),
             client.clone(),
             transaction_pool,
             prometheus_registry.as_ref(),
             None,
         );
+        proposer_factory
+            .set_default_block_size_limit(*BlockLength::get().max.get(Default::default()) as usize);
```
mike1729, line 228: 1. Could you replace `Default::default` with something more informative? Or add a comment explaining it purpose?
2. In polkadot, the Runtime limit is `5 * 1024 * 1024` while the `DEFAULT_BLOCK_SIZE_LIMIT` is `4 * 1024 * 1024 + 512`, so 90%.

mike1729 (changes requested):

bin/node/src/service.rs
pmikolajczyk41: Finally, I extracted the limit into a constant and thus got rid of the problem `Default::default` vs an awkward dependency.

As discussed offline, the `DEFAULT_BLOCK_SIZE_LIMIT` does not matter (as long as it is not greater than its counterpart in runtime), so I'm just using the same value (5MB).

mike1729 (approved):

DamianStraszak (approved):

PR #188, opened 2021-11-08
AZ-256: Add treasury pallet

pmikolajczyk41:
- `pallet_treasury` added; the parameters and setting were discussed offline
- the governance over treasury is passed to the same authority as `sudo`
- tests covering channeling fee and approving/rejecting proposals; the only thing missing is verifying whether tips are channeled as well; however, for now there is no easy way to do it - a new Jira ticket should be created unless discussed here otherwise

_explanation for tips problem_:
In out e2e tests we use `substrate-client-api` package. Unfortunately, it does not support specifying a tip for an extrinsic. There are basically three solutions, neither of which seems appropriate/possible for now:
1. extend `substrate-client-api` -- although we have already been contributing recently to their repository, we cannot integrate the newest commit, which operates on metadata v14 (giving up older v13 which we are using); hence, as long as we do not upgrade our code, we cannot enjoy new versions of  `substrate-client-api`
2. hard fork (probably temporarily) `substrate-client-api` and extend it with our needs -- this seems to be an overkill and introduces possible problems with maintanence
3. copy some functionality from `substrate-client-api` and edit it within our code -- very ugly
4. operate on hex data and do magic with appending tip -- awful


bin/runtime/src/lib.rs
```diff
@@ -277,8 +280,24 @@ impl MultiplierUpdate for ConstantFeeMultiplierUpdate {
     }
 }
 
+type NegativeImbalance = <Balances as Currency<AccountId>>::NegativeImbalance;
+
+pub struct DealWithFees;
```
timorl, line 285: Maybe call this `EverythingToTheTreasury` or something, then you can avoid the comments below.

bin/runtime/src/lib.rs
```diff
@@ -416,6 +435,51 @@ impl pallet_multisig::Config for Runtime {
...
+// We allow at most 20 approvals in the queue at once.
+pub const TREASURY_MAX_APPROVALS: u32 = 20;
+
+parameter_types! {
+    pub const Burn: Permill = Permill::from_percent(TREASURY_BURN);
+    pub const ProposalBond: Permill = Permill::from_percent(TREASURY_PROPOSAL_BOND);
+	pub const ProposalBondMinimum: Balance = TREASURY_MINIMUM_BOND;
+    pub const MaxApprovals: u32 = TREASURY_MAX_APPROVALS;
+    pub const SpendPeriod: BlockNumber = TREASURY_SPEND_PERIOD;
+    pub const TreasuryPalletId: PalletId = PalletId(*b"py/trsry");
```
timorl, line 456: What does this ID do?

e2e-tests/src/main.rs
```diff
@@ -125,15 +118,81 @@ fn test_token_transfer(config: Config) -> anyhow::Result<()> {
     Ok(())
 }
 
+// todo: check channeling tips
```
timorl, line 121: This would be better as a task in Jira.

e2e-tests/src/main.rs
```diff
@@ -125,15 +118,81 @@ fn test_token_transfer(config: Config) -> anyhow::Result<()> {
...
+    let proposals_counter = get_proposals_counter(&connection);
+    assert!(proposals_counter >= 2, "Proposal was not created");
+
+    let sudo = get_sudo(config);
+    let connection = connection.set_signer(sudo);
+
+    send_treasury_approval(proposals_counter - 2, &connection);
+    wait_for_approval(&connection, proposals_counter - 2)?;
+
+    treasury_reject(proposals_counter - 1, &connection)?;
```
timorl, line 186: This is a single function, but the above is two, confusing. Please make this consistent.

e2e-tests/src/utils.rs
```diff
@@ -70,20 +89,127 @@ pub fn get_free_balance(account: &AccountId32, connection: &Connection) -> Balan
...
+
+pub fn get_total_issuance(connection: &Connection) -> u128 {
+    connection
+        .get_storage_value("Balances", "TotalIssuance", None)
+        .unwrap()
+        .unwrap()
+}
+
+//////////////////////
+// Treasury related //
```
timorl, line 121: Yeah, this is why `util` modules are a bad idea.

Generally the organization of code around here is a bit of a mess, including a cluttered `main.rs`. It would be great if you could split the code somewhat more readably. Possible approach – split them by component tested and just import the test functions into main (making it tiny and keeping the per-component code in one place). I'm not sure how much work that would be though, if you think it's too much for this PR at least do this for the treasury tests.

You might still end up with a `util` module (e.g. containing the function above this comment?), but hopefully much smaller and if everything goes well with a better name (e.g. `finance` for the above?). Kinda guessing here, 'cause I'm not intimately familiar with this code.

If you have a better idea as to how to keep this organized that would likely also be fine.

e2e-tests/src/utils.rs
```diff
@@ -30,15 +38,26 @@ pub fn accounts(seeds: Option<Vec<String>>) -> Vec<KeyPair> {
     seeds.into_iter().map(keypair_from_string).collect()
 }
 
+pub fn get_first_account(accounts: &[KeyPair]) -> KeyPair {
+    accounts.get(0).expect("No accounts passed").to_owned()
+}
+
 pub fn get_first_two_accounts(accounts: &[KeyPair]) -> (KeyPair, KeyPair) {
```
timorl, line 45: These two functions here are kinda bizarre – you can just use `[0]` and `[1]` everywhere and this will panic when out of bounds (which is fine, the test will fail, and it's terribly unexpected if I understand correctly – it would mean the test has been set up incorrectly, right?).

bin/runtime/src/lib.rs
timorl: Oh, it's the ID of the associated account? Why is it called what it's called then, is there a standard here?

e2e-tests/src/waiting.rs
```diff
@@ -68,3 +66,42 @@ pub fn wait_for_finalized_block(connection: Connection, block_number: u32) -> an
 
     Err(anyhow::anyhow!("Giving up"))
 }
+
+/// blocks the main thread waiting for an approval for proposal with id `proposal_id`
+pub fn wait_for_approval(connection: &Connection, proposal_id: u32) -> anyhow::Result<()> {
```
timorl, line 71: One of these things is not like the others... is there no "approval" event?

timorl (changes requested):
Some comments, mostly about code organization.

I'm slightly more worried that this changes the runtime and chainspec, and I don't feel confident that's a good idea right now, especially that we are working towards the testnet update which is turning out to be annoyingly complicated. I would feel safer if @DamianStraszak confirmed this is ok before merging (just the runtime+chainspec changes, not the whole PR).

bin/runtime/src/lib.rs
pmikolajczyk41: Indeed, the `PalletId` is used to derive the corresponding account, an example in the pallet_treasury:
```rust
/// The treasury's pallet id, used for deriving its sovereign account ID.
#[pallet::constant]
type PalletId: Get<PalletId>;
```

In substrate these are some used names:
 - `PalletId(*b"py/trsry")`
 - `PalletId(*b"py/socie")`
 - `PalletId(*b"py/lotto")`

I followed their way. Also, afaik, it has to be of length 8

e2e-tests/src/main.rs
pmikolajczyk41: surely, but I wanted to wait for the review, in case somebody come up with a better solution to be applied here

e2e-tests/src/main.rs
pmikolajczyk41: I tried :P but unfortunately, these operations must be done in different ways :(
1. **approval** - sending an extrinsic is straightforward, but pallet_treasury does _not_ emit any corresponding event (apart from system.extrinsicSucceeded, but such event is in every block due to the timestamp extrinsic and there is no 'good' way of matching such result with the origin call); therefore, the only reliable way of verifying whether the approval was successful is just to look at the storage and check the `approvals` vec (and have a hope that the spending period has not just occurred :P)
2. **rejection** - again, sending an extrinsic is straightforward, and in this case there is a corresponding event, namely `Treasury::Rejected` so we can listen for this event; _however_, it often happens, that this event is dispatched before we start listening, so we need to setup listening before sending the extrinsic

e2e-tests/src/waiting.rs
pmikolajczyk41: answered in another conversation

bin/runtime/src/lib.rs
timorl: > it has to be of length 8

...what? lol

In this case it cannot be improved much, but it would still be nice to know what the "py" refers to – it would be a bit silly if it was somehow supposed to be associated with e.g. Polkadot. Although the 'p' probably stands for "pallet"...?

e2e-tests/src/main.rs
timorl: Ah, these are definitely annoyances.

To make this less confusing I would suggest creating a `treasury_approve` function, probably just consisting of the two lines above, at least then they look the same in the test and the reader doesn't have to spend time wondering why you only check the approval. Internally the functions will still differ, but the code will be more readable.

bin/runtime/src/lib.rs
pmikolajczyk41: regarding length: `pub struct PalletId(pub [u8; 8]);` (https://github.com/paritytech/substrate/blob/4581dd99ed59f65ede46a34c81133177d5824a68/frame/support/src/lib.rs#L111)

pmikolajczyk41: image

pmikolajczyk41: I dug it up, and now I wonder whether e.g. `a0/trsry` would be better :P

bin/runtime/src/lib.rs
```diff
@@ -106,7 +109,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("aleph-node"),
     impl_name: create_runtime_str!("aleph-node"),
     authoring_version: 1,
-    spec_version: 4,
+    spec_version: 5,
     impl_version: 1,
     apis: RUNTIME_API_VERSIONS,
     transaction_version: 2,
```
DamianStraszak, line 115: You also need to bump `transaction_version`

bin/runtime/src/lib.rs
timorl: Excellent, I was afraid that it was something like this and well done finding such a direct proof. :D

"a0" as a prefix is indeed much better.

DamianStraszak (commented):
When it comes to the runtime update associated with this PR:
1) I think it is fine to make changes in the chainspec, but it will not affect any already running chain, because according to my understanding we should always pass the old, original chainspec when running mainnet or testnet binaries. And even if we passed the new one then nothing would actually happen (??). But that is a topic worth researching...
2) I seems that the pallet does not have any storage items that are added in the genesis block (and I'm now pretty sure no pallet should ever do that -- and our does :P) so I think adding it without any migration should work with no issues. 
3) Even though 2), I think it would be great to merge this change after we make the first testnet update (hopefully Monday next week?), so that we don't update too many things at once, because, you know, we don't have much experience in updating yet.

bin/runtime/src/lib.rs
pmikolajczyk41: renamed

bin/runtime/src/lib.rs
pmikolajczyk41: renamed

e2e-tests/src/main.rs
pmikolajczyk41: Jira #329

e2e-tests/src/main.rs
pmikolajczyk41: done

e2e-tests/src/utils.rs
pmikolajczyk41: I reorganized the package, but if it was just a transition between mess and mess', I can make further changes

e2e-tests/src/utils.rs
pmikolajczyk41: done

bin/runtime/src/lib.rs
pmikolajczyk41: bumped

e2e-tests/src/utils.rs
```diff
@@ -1,89 +1,6 @@
...
-    info!("[+] Transfer transaction hash: {}", tx_hash);
-
-    tx
-}
+pub mod accounts;
+pub mod extrinsic;
+pub mod fee;
+pub mod transfer;
+pub mod types;
+pub mod waiting;
```
timorl, line 6: This setup seems backwards to me. I would put all the `util` modules on the top level, and create a `test` module with all the modules that are currently in the top level. Then the test module would export all the functions that currently start with `test_*`, but without this prefix, and all main would do would be `use crate::test` and then call `test::fee_calculation` etc. There could also be a `lib.rs` file which would export all the utility functions from the currently-util modules and the things in `test` would import straight from `crate`.

This is incredibly pedantic tho, not a blocker for merging in the slightest, only do this if you are convinced it actually sounds better than the current version, and are in the mood. The current version is already massively better organized than what was here before, so thanks!

timorl: Oh, and at least extrinsic.rs and types.rs would be merged into lib.rs in the above, maybe some others too (waiting.rs?).

timorl (approved):
Very nice, although keep Damian's comment in mind before merging.

One comment, but extremely nitpicky and feel free to ignore.

bin/runtime/src/lib.rs
```diff
@@ -416,6 +433,51 @@ impl pallet_multisig::Config for Runtime {
...
+pub const TREASURY_MINIMUM_BOND: Balance = 1000 * CENTS;
+// Every 4h we implement accepted proposals.
+pub const TREASURY_SPEND_PERIOD: BlockNumber = 4 * HOURS;
+// We allow at most 20 approvals in the queue at once.
+pub const TREASURY_MAX_APPROVALS: u32 = 20;
+
+parameter_types! {
+    pub const Burn: Permill = Permill::from_percent(TREASURY_BURN);
+    pub const ProposalBond: Permill = Permill::from_percent(TREASURY_PROPOSAL_BOND);
+	pub const ProposalBondMinimum: Balance = TREASURY_MINIMUM_BOND;
```
fbielejec, line 451: weird formatting?

bin/runtime/src/lib.rs
```diff
@@ -433,6 +495,7 @@ construct_runtime!(
         Aleph: pallet_aleph::{Pallet, Call, Config<T>, Storage, Event<T>},
         Session: pallet_session::{Pallet, Call, Storage, Event, Config<T>},
         Scheduler: pallet_scheduler::{Pallet, Call, Storage, Event<T>},
+		Treasury: pallet_treasury::{Pallet, Call, Storage, Config, Event<T>},
```
fbielejec, line 498: same, weird formatting (but seems to be passing fmt checks :thinking: )

e2e-tests/src/main.rs
```diff
@@ -1,39 +1,36 @@
...
-    run(test_finalization, "finalization", config.clone())?;
-    run(test_fee_calculation, "fee calculation", config.clone())?;
-    run(test_token_transfer, "token transfer", config.clone())?;
-    run(test_change_validators, "validators change", config)?;
+    run(test::finalization, "finalization", config.clone())?;
+    run(test::fee_calculation, "fee calculation", config.clone())?;
+    run(test::token_transfer, "token transfer", config.clone())?;
+    run(test::channeling_fee, "channeling fee", config.clone())?;
+    run(test::treasury_access, "treasury access", config.clone())?;
+    run(test::change_validators, "validators change", config)?;
```
fbielejec, line 22: ince these test run single threaded, ooone thing comes to mind - we could use that fact to implicitely test finalization after changing validators (if there was ever to be a bug there), by moving the test earlier in the suite? For example:

```
    run(test::finalization, "finalization", config.clone())?;
    run(test::fee_calculation, "fee calculation", config.clone())?;
    run(test::change_validators, "validators change", config)?;
    run(test::token_transfer, "token transfer", config.clone())?;
    run(test::channeling_fee, "channeling fee", config.clone())?;
    run(test::treasury_access, "treasury access", config.clone())?;
```

bin/runtime/src/lib.rs
DamianStraszak: It's because this is macro code. Fmt does not have power there :P

e2e-tests/src/test/treasury.rs
```diff
@@ -0,0 +1,199 @@
...
+        .module_with_constants_by_name("Treasury")
+        .unwrap()
+        .constant_by_name("PalletId")
+        .unwrap()
+        .get_value();
+    PalletId(pallet_id.try_into().unwrap()).into_account()
+}
+
+type ProposalTransaction =
+    UncheckedExtrinsicV4<([u8; 2], Compact<u128>, MultiAddress<AccountId, ()>)>;
```
fbielejec, line 108: extra newline here?

e2e-tests/src/test/treasury.rs
```diff
@@ -0,0 +1,199 @@
...
+}
+
+pub fn treasury_access(config: Config) -> anyhow::Result<()> {
+    let Config { node, seeds, .. } = config.clone();
+
+    let proposer = accounts_from_seeds(seeds)[0].to_owned();
+    let beneficiary = AccountId::from(proposer.public());
+    let connection = create_connection(node).set_signer(proposer);
+
+    propose_treasury_spend(0u128, &beneficiary, &connection);
```
fbielejec, line 75: put some real value pls

e2e-tests/src/test/treasury.rs
```diff
@@ -0,0 +1,199 @@
...
+        .metadata
+        .module_with_constants_by_name("Treasury")
+        .unwrap()
+        .constant_by_name("PalletId")
+        .unwrap()
+        .get_value();
+    PalletId(pallet_id.try_into().unwrap()).into_account()
+}
+
+type ProposalTransaction =
```
fbielejec, line 107: move to /utils/types.rs ?

e2e-tests/src/test/treasury.rs
```diff
@@ -0,0 +1,199 @@
...
+}
+
+fn get_proposals_counter(connection: &Connection) -> u32 {
+    connection
+        .get_storage_value("Treasury", "ProposalCount", None)
+        .unwrap()
+        .unwrap()
+}
+
+type GovernanceTransaction = UncheckedExtrinsicV4<([u8; 2], Compact<u32>)>;
```
fbielejec, line 131: to /utils/types.rs ?

fbielejec (commented):
minor comment, solid work!

e2e-tests/src/test/treasury.rs
pmikolajczyk41: I have left this type (similarly to `ProposalTransaction`) as it is not returned from any public function and is for internal use only

e2e-tests/src/test/treasury.rs
pmikolajczyk41: similarly to `GovernanceTransaction`

e2e-tests/src/test/treasury.rs
pmikolajczyk41: changed

e2e-tests/src/main.rs
pmikolajczyk41: reordered

bin/runtime/src/lib.rs
pmikolajczyk41: changed tabs to spaces (must have been introduced while copying from substrate :|)

bin/runtime/src/lib.rs
pmikolajczyk41: changed tabs to spaces (must have been introduced while copying from substrate :|)

e2e-tests/src/utils.rs
pmikolajczyk41: reorganized, thanks a lot!

fbielejec (approved):

timorl (approved):

PR #190, opened 2021-11-12
AZ-62: Add synchronization workflow with foundation repository

maciejnems:
I tested it here:
* workflow - https://github.com/Cardinal-Cryptography/aleph-node/runs/4187960088?check_suite_focus=true
* commit - https://github.com/Cardinal-Cryptography/aleph-node/commit/bba0b589d47c2786167ab00366065832accd1bd4
* foundation branch - https://github.com/aleph-zero-foundation/aleph-node/tree/AZ-62-add-sync-with-foundation-repo

### DIFF
--- a/.github/workflows/push-foundation-repo.yml
+++ b/.github/workflows/push-foundation-repo.yml
@@ -1,0 +1,17 @@
+name: Sync Cardinal-Cryptography repo with Aleph-Zero-Foundation repo
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    if: ${{ github.repository == 'Cardinal-Cryptography/aleph-node'}}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.SYNCAZF }}
+      - name: Push to Aleph-Zero-Foundation
+        run: git push https://x-access-token:${{ secrets.SYNCAZF }}@github.com/aleph-zero-foundation/aleph-node.git


.github/workflows/push-foundation-repo.yml
```diff
@@ -0,0 +1,17 @@
...
+    branches: [main]
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    if: ${{ github.repository == 'Cardinal-Cryptography/aleph-node'}}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
kostekIV, line 14: What is the reason for `fetch-depth` set to 0?

maciejnems: By default checkout@v2 fetches only one commit, so the push is not possible. We need to fetch the full history

kostekIV (approved):

.github/workflows/push-foundation-repo.yml
```diff
@@ -0,0 +1,17 @@
...
+          token: ${{ secrets.SYNCAZF }}
```
Yithis, line 15: There is the same token for both AlephBFT and aleph-node?

maciejnems: Yes. More specifically this token has write rights to aleph-foundation organization repos.

Yithis (approved):

.github/workflows/push-foundation-repo.yml
Yithis: I'm not sure if it is safe to have one token for all organization repos, but won't also fight for changing it

maciejnems: I don't think there is a problem here. We keep all our work in Cardinal Cryptography organization, so having write token for repos in Aleph Foundation organization won't mess with our work in any way

PR #193, opened 2021-11-12
Preparations to the runtime update. Reverting back some types.

DamianStraszak:
This PR makes some unpleasant changes in order to make the runtime happy (and to make us more confident that the coming and future runtime updates will go as expected). In particular it changes back to raw types from wrapped types for some of the runtime apis and removes the unit-creation-delay param from the runtime.


timorl (approved):
LGTM code-wise, I'm not good enough to be sure the update will work, but we'll see. :>

pmikolajczyk41 (approved):
seems ok; 

as a side effect, passing `unit_creation_delay` in benchmark script will not work and thus requires fixing -- I created a Jira for this (https://cardinal-cryptography.atlassian.net/browse/AZ-288)

fbielejec (approved):
I like these changes, less wrapper structs, more primitives = better

PR #195, opened 2021-11-15
AZ-151: calculate fee from weight

pmikolajczyk41:


e2e-tests/src/main.rs
```diff
@@ -44,25 +49,58 @@ fn test_finalization(config: Config) -> anyhow::Result<u32> {
...
+    let balance_after = balance_of(&from, &connection);
+    info!("[+] Account {} balance after tx: {}", to, balance_after);
+
+    let FeeInfo {
+        base_fee,
+        len_fee,
+        unadjusted_weight,
+        adjusted_weight,
+        tip,
+    } = get_tx_fee_info(&connection, &tx);
```
kostekIV, line 75: Maybe the FeeInfo should have just 3 fields:
* `fee`
* `unadjusted_weight`
* `adjusted_weight`

e2e-tests/src/main.rs
```diff
@@ -248,3 +276,68 @@ fn accounts(seeds: Option<Vec<String>>) -> Vec<sr25519::Pair> {
...
+
+fn get_tx_fee_info(connection: &Connection, tx: &TransferTransaction) -> FeeInfo {
+    let block = connection.get_block::<Block>(None).unwrap().unwrap();
+    let block_hash = block.header.hash();
+
+    let unadjusted_weight = connection
+        .get_payment_info(&tx.hex_encode(), Some(block_hash))
+        .unwrap()
+        .unwrap()
+        .weight as Balance;
```
kostekIV, line 306: looks suspicious  :raised_eyebrow:

e2e-tests/src/main.rs
```diff
@@ -248,3 +276,68 @@ fn accounts(seeds: Option<Vec<String>>) -> Vec<sr25519::Pair> {
...
+        adjusted_weight: inclusion_fee.adjusted_weight_fee,
+        tip: fee.tip,
+    }
+}
+
+fn balance_of(account: &AccountId32, connection: &Connection) -> Balance {
+    connection.get_account_data(account).unwrap().unwrap().free
+}
+
+fn transfer(target: &AccountId32, value: u128, connection: &Connection) -> TransferTransaction {
```
kostekIV, line 327: apply this function also to lines 108 to 119

e2e-tests/Cargo.toml
```diff
@@ -1,12 +1,12 @@
 [package]
 name = "aleph-e2e-client"
 version = "0.1.0"
-edition = "2018"
+edition = "2021"
```
fbielejec, line 4: there is also a bunch of rustfmt files that shoudl be bumped as well

e2e-tests/src/main.rs
pmikolajczyk41: thank you, done

e2e-tests/src/main.rs
pmikolajczyk41: well, using the exact type `Weight` requires adding new dependency, so I chose 'suspicious' cast; I guess that in tests it is acceptable somehow, especially since effectively we use weight as balance

e2e-tests/src/main.rs
pmikolajczyk41: good catch, done

e2e-tests/Cargo.toml
pmikolajczyk41: I bumped edition in two corresponding files

e2e-tests/src/main.rs
kostekIV: ok

kostekIV (approved):

e2e-tests/src/main.rs
```diff
@@ -248,3 +260,64 @@ fn accounts(seeds: Option<Vec<String>>) -> Vec<sr25519::Pair> {
...
+    let tx: UncheckedExtrinsicV4<_> = compose_extrinsic!(
+        connection,
+        "Balances",
+        "transfer",
+        GenericAddress::Id(target.clone()),
+        Compact(value)
+    );
+
+    let tx_hash = connection
+        .send_extrinsic(tx.hex_encode(), XtStatus::InBlock)
```
fbielejec, line 317: why not XtStatus::Finalized ?

pmikolajczyk41: good question; I left it as it has already been, but I will change it

pmikolajczyk41: changed

fbielejec (approved):

PR #197, opened 2021-11-16
runtime storage migration in aleph pallet

fbielejec:
Changes to storage in our pallet.
This PR introduces a light optimization to the storage.

- we continue storing validators change in two storage values (session and actual validations set), to avoid de-serializing a potentially large vector of validators at every session boundary,
- however we store those types raw i.e. not as Option wrapped types (instead we query them as Options).

This is a very minor optimization, one we could potentially live without, however the change is an opportunity to test writing and applying storage migrations.
A relevant document accompanying this PR can be found here: https://www.notion.so/cardinalcryptography/Storage-migrations-4cbec0f087d5465b86d7b7adde3d0409


pmikolajczyk41 (dismissed):
lgtm, but please, shorten the PR title

DamianStraszak (changes requested):
This is a change that would require a storage migration, because it alters the types in the runtime storage. I don't have the confidence or knowledge to play with these things at this stage, so would avoid doing such changes at any cost :/

Yithis (commented):
As code change it looks good to me, but I totally agree with Damian about changing runtime

fbielejec:
> This is a change that would require a storage migration, because it alters the types in the runtime storage. I don't have the confidence or knowledge to play with these things at this stage, so would avoid doing such changes at any cost :/

We have a task about runtime upgrades in the pipeline anyway - maybe a nice occasion to try writing (and testing) migrations?

pallet/src/lib.rs
```diff
@@ -13,12 +13,20 @@ mod mock;
...
 
-use frame_support::{sp_runtime::BoundToRuntimeAppPublic, traits::OneSessionHandler};
+use frame_support::{
+    sp_runtime::BoundToRuntimeAppPublic,
+    traits::{OneSessionHandler, StorageVersion},
+};
 pub use pallet::*;
 
+/// The current storage version.
+const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
```
pmikolajczyk41, line 28: Unfortunately "current" is a little bit misleading in this context (at least for me) - as far as I understand, this corresponds to the storage version represented by the code, not the one on-chain, and will not be changed by a possible migration. Am I right?

pallet/src/migrations/v0_to_v1.rs
```diff
@@ -0,0 +1,56 @@
...
+
+pub fn migrate<T: Config, P: GetStorageVersion + PalletInfoAccess>() -> Weight {
+    let on_chain_storage_version = <P as GetStorageVersion>::on_chain_storage_version();
+    let new_storage_version = crate::STORAGE_VERSION;
+
+    if on_chain_storage_version == StorageVersion::default() && new_storage_version == 1 {
+        log::info!(target: "pallet_aleph", "Running migration from STORAGE_VERSION 0 to 1");
+
+        let mut writes = 0;
+        let _ = OldSessionForValidatorsChange::translate(|old| match old {
```
pmikolajczyk41, line 25: hopefully, it will always return `Ok(_)`, but maybe it is worth verifying?

pallet/src/tests.rs
```diff
@@ -1,7 +1,57 @@
...
+            "Storage version before applying migration should be default"
+        );
+
+        let _weight = migrations::v0_to_v1::migrate::<Test, Aleph>();
+
+        let v1 = <pallet::Pallet<Test> as GetStorageVersion>::on_chain_storage_version();
+
+        assert_eq!(
+            v1,
+            StorageVersion::new(1),
```
pmikolajczyk41, line 38: surely it is not the case, but I would be more explicit with ensuring that `v0 != v1`, either by a direct comparison or by using `StorageVersion::new(0)` instead of `StorageVersion::new(1)`

pallet/src/tests.rs
```diff
@@ -1,7 +1,57 @@
 #![cfg(test)]
 
-use crate::mock::*;
+use crate::{migrations, mock::*, pallet};
 use frame_support::assert_ok;
+use frame_support::traits::{GetStorageVersion, StorageVersion};
+
+#[test]
+fn migration_from_v0_to_v1_works() {
```
pmikolajczyk41, line 8: should we also test migration when there is no validators change pending?

pallet/src/migrations/v0_to_v1.rs
```diff
@@ -0,0 +1,56 @@
...
+            Some(current) => {
+                writes += 1;
+                current
+            }
+            None => None,
+        });
+
+        let _ = OldValidators::<T>::translate(|old| match old {
```
pmikolajczyk41, line 33: Can you help me figuring out the mapping between alias `OldValidators` and the actual storage entry? I doubt that it is done by types or naive prefix deduction :P

pallet/src/tests.rs
```diff
@@ -1,7 +1,57 @@
 #![cfg(test)]
 
-use crate::mock::*;
+use crate::{migrations, mock::*, pallet};
 use frame_support::assert_ok;
+use frame_support::traits::{GetStorageVersion, StorageVersion};
+
+#[test]
```
pmikolajczyk41, line 7: should we also test that nothing is done when the storage version is already updated?

pmikolajczyk41 (commented):
In our ADR regarding updates, there's a statement
> We will need E2E tests that check inter-version compatibility.

I'm just wondering how this is relevant and reasonable to perform such check

bin/runtime/src/lib.rs
```diff
@@ -33,7 +33,7 @@ pub use frame_support::{
     sp_runtime::curve::PiecewiseLinear,
     traits::{
         Currency, EstimateNextNewSession, Imbalance, KeyOwnerProofSystem, LockIdentifier,
-        OnUnbalanced, Randomness, U128CurrencyToVote, ValidatorSet,
+        OnRuntimeUpgrade, OnUnbalanced, Randomness, U128CurrencyToVote, ValidatorSet,
```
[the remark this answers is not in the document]
fbielejec: leftover, good spot, removing

bin/runtime/src/lib.rs
```diff
@@ -106,7 +106,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("aleph-node"),
     impl_name: create_runtime_str!("aleph-node"),
     authoring_version: 1,
-    spec_version: 4,
+    spec_version: 5,
```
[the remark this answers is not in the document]
fbielejec: updated and bumped

pallet/src/lib.rs
fbielejec: this is Parity parlance: https://crates.parity.io/frame_support/pallet_prelude/trait.GetStorageVersion.html
I figured it's best to stick with it (for consistency and stuff)

pallet/src/migrations/v0_to_v1.rs
fbielejec: Good point, added +  logging

pallet/src/tests.rs
fbielejec: In this case we would probbaly be fine not running any migration at all (or en empty on that just updates the version) - the storage layer automatically removes the storage entry if the value is None, so next time sth is stored there it's stored with a correct schema.

pallet/src/migrations/v0_to_v1.rs
fbielejec: You got me thinking here. The macro signature is :
```
generate_storage_alias!(Prefix, StorageName => Value<u32>);
```

so it must be storing it under:

hex(128-bit hash of `Aleph` + 128-bit hash of `OldValidators`)

the `put_storage_value` does exactly that:
```
pub fn put_storage_value<T: Encode>(module: &[u8], item: &[u8], hash: &[u8], value: T) {
	let mut key = vec![0u8; 32 + hash.len()];
	key[0..16].copy_from_slice(&Twox128::hash(module));
	key[16..32].copy_from_slice(&Twox128::hash(item));
	key[32..].copy_from_slice(hash);
	frame_support::storage::unhashed::put(&key, &value);
}
```
but the the tests were passing even with it. I'll investigate

fbielejec: Ok, as per my comment - query(Option(value)) is the same as option-query(value). Translate is therefore an ... identity

pallet/src/tests.rs
fbielejec: done

pallet/src/migrations/v0_to_v1.rs
```diff
@@ -0,0 +1,60 @@
...
+        writes += 1;
+
+        T::DbWeight::get().reads(3) + T::DbWeight::get().writes(writes)
+    } else {
+        log::warn!(
+            target: "pallet_aleph",
+            "Do not know which migration to apply because on-chain storage version is {:?} and the version declared in the aleph pallet is {:?}",
+            on_chain_storage_version,
+            current_storage_version
+        );
```
DamianStraszak, line 56: Will this `else` execute on every future update? (Assuming we don't remove this migration and we don't make any other migrations)

fbielejec: if we don't remove the migration it will, yes

fbielejec:
> In our ADR regarding updates, there's a statement
> 
> > We will need E2E tests that check inter-version compatibility.
> 
> I'm just wondering how this is relevant and reasonable to perform such check

This would be incredibly cumbersome to achieve, we would have to:
- compile the node with old runtime code (using a past commit hash?)
- run the chain (native)
- compile new runtime code
- apply setCode extrinsic
- perform assertions

This would double the time to run the pipeline, because we still need to test the current code, for something that is already covered in a unit test.

pallet/src/migrations/v0_to_v1.rs
DamianStraszak: Ok, in case we don't remove it soon, I would have a small request to change this comment to "Not applying any migration, because on-chain ...." -- this is a less worrisome log message :)

DamianStraszak (approved):

pallet/src/tests.rs
pmikolajczyk41: I guess so, but for me this is still a scenario worth exposing in tests. Up to you, a non-blocker

pmikolajczyk41 (approved):

pallet/src/migrations/v0_to_v1.rs
fbielejec: Done

PR #202, opened 2021-11-22
Update readme

h4nsu:


README.md
```diff
@@ -1,9 +1,51 @@
...
+### Installation
+
+1. Set up the Rust development environment following the instructions [here][rust-installation]
+2. Install the Substrate framework as described [here][substrate-installation]
+3. Clone this repo to your machine
+4. Run `cargo build --release -p aleph-node`
+
+### Running
+
+To experiment with Aleph Node you can locally run a small blockchain network using the `run_nodes.sh` script from the root of this repository.  Please consult the script or the output of `run_nodes.sh -help` for additional parameters (like the number of nodes etc.). The script starts multiple instances of Aleph Node on your local machine, so please adjust the number of nodes carefully with performance of your system in mind. By default 4 nodes are started.
```
fbielejec, line 24: this is one option, but one can also run a small chain using the docker compose scripts: 

```bash
.github/scripts/run_consensus. 
```
Maybe worth it to document both?

h4nsu: Didn't know about that ^^' I'll try it out and add a mention here

Yithis:
Did you contact marketing about this PR?

fbielejec (commented):
Might be good to document the other way of running the chain, it basically boils down to doing locally what the CI server is doing build docker image of the node, generate chain spec and keys, inspect p2p id of the bootnode (it's non deterministic), run the chain  - ping me for help explaining it

README.md
```diff
@@ -1,9 +1,51 @@
...
 
-Node implementation for aleph blockchain built with Substrate framework.
+Aleph Zero is an open-source layer 1 blockchain platform focused on privacy, scalability and energy efficiency. It is based on a unique, peer-reviewed consensus algorithm - AlephBFT (as described in our [paper][aleph-bft-paper] and implemented [here][aleph-bft-link]).
+
+Aleph node is based on a Substrate node where the default finality gadget (GRANDPA) has been replaced with AlephBFT. Block authoring is realized with Substrate's Aura. The chain is run in periodic sesssions (900 blocks each) utilizing the Session pallet. The authorities in each session serve for both Aura and AlephBFT authorities, and on top of that are responsible for running the Aggregator protocol producing multisignatures of finalized blocks.
+
+
+### Installation
+
+1. Set up the Rust development environment following the instructions [here][rust-installation]
```
Yithis, line 17: Shouldn't we inform about exact version of nightly?

h4nsu: We probably should. Which one is that?

h4nsu:
> Did you contact marketing about this PR?

No, should I?

README.md
Yithis: https://www.notion.so/cardinalcryptography/How-to-setup-a-repository-7e240b93ab14471d98d59df9430375b0

Yithis:
If any - line 10. But since it will go public I thought that maybe someone should look at this for 1 minute. I don't want to force any pipeline here, so if you think it is to technical for them, just ignore it. I will accept without it.

h4nsu:
I asked marketing hotline, they made some minimal corrections. Good idea!

fbielejec (approved):

Yithis (approved):

README.md
h4nsu: After trying it for some time I run into some compatibility problems - the instructions would need to be somehow OS distro dependent and I think it would be a bit of an overkill to include so much details for an alternative running way.

PR #210, opened 2021-11-26
Az-107: Add per-session connection manager

timorl:
This one has proper tests. They even caught one major bug I left in when developing, so yeah, quality coding. <_<

Still dead code.

### DIFF
--- a/finality-aleph/src/new_network/connection_manager/addresses.rs
+++ b/finality-aleph/src/new_network/connection_manager/addresses.rs
@@ -1,0 +1,181 @@
+use crate::new_network::{connection_manager::Multiaddr, PeerId};
+use ip_network::IpNetwork;
+use sc_network::{multiaddr::Protocol, PeerId as ScPeerId};
+
+/// Checks whether the given Multiaddr is globally accessible.
+pub fn is_global(address: &Multiaddr) -> bool {
+    address.0.iter().all(|protocol| match protocol {
+        Protocol::Ip4(ip) => IpNetwork::from(ip).is_global(),
+        Protocol::Ip6(ip) => IpNetwork::from(ip).is_global(),
+        _ => true,
+    })
+}
+
+/// Checks whether the given Multiaddr contains a libp2p component.
+pub fn is_p2p(address: &Multiaddr) -> bool {
+    address
+        .0
+        .iter()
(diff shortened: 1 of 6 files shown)


finality-aleph/src/new_network/connection_manager/session.rs
```diff
@@ -0,0 +1,417 @@
...
+    session_id: SessionId,
+    addresses: Vec<Multiaddr>,
+) -> Result<(Authentication, PeerId), AddressError> {
+    let addresses: Vec<_> = addresses
+        .into_iter()
+        .filter(is_global)
+        .filter(is_p2p)
+        .collect();
+    if addresses.is_empty() {
+        return Err(AddressError::NoExternalP2pAddresses);
```
DamianStraszak, line 45: In a conversation with @timorl we realized that there might be issues with local networks if we try to do authentications for external addresses only. Leaving this comment for the record only. This is being worked on by @timorl now.

Yithis (approved):

finality-aleph/src/new_network/connection_manager/session.rs
timorl: Removed the filtering on global addresses, it will instead happen on the `add_reserved` level, since there it shouldn't matter for local tests.

finality-aleph/src/new_network/connection_manager/addresses.rs
```diff
@@ -0,0 +1,181 @@
...
+        .iter()
+        .fold(UniquePeerId::Unknown, |result, protocol| {
+            result.accumulate(peer_id(&protocol))
+        })
+        .into_option()
+}
+
+/// Returns the peer id contained in the set of multiaddresses if it's unique and present in every
+/// address, None otherwise.
+pub fn get_unique_peer_id(addresses: &[Multiaddr]) -> Option<PeerId> {
```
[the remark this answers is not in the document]
timorl: Yeah, the names are not ideal. The function above works on a single `Multiaddr`, which generally should either have exactly one or no peer IDs, so I did not add "unique" in the name there. This is not the greatest naming schema, but I have no better idea. :/

Yithis (changes requested):
A test of the newest changes?

Yithis (approved):

maciejnems (approved):

PR #214, opened 2021-11-29
AZ-365: Improve testability of JustificationHandler

pmikolajczyk41:


finality-aleph/src/justification.rs
```diff
@@ -64,111 +85,114 @@ where
     pub number: NumberFor<Block>,
 }
 
-pub(crate) struct JustificationHandler<B, RB, C, BE>
+pub(crate) struct JustificationHandler<B, RB, C, BE, D, SI, F>
 where
     B: BlockT,
     RB: network::RequestBlocks<B> + 'static,
     C: crate::ClientForAleph<B, BE> + Send + Sync + 'static,
```
maciejnems, line 92: I think it might be a good step, to completely remove ClientForAleph as a set of traits. For example this line could be changed to:
```
    C: HeaderBackend<B> + LockImportRun<B, BE> + Finalizer<B, BE> + Send + Sync + 'static,
```
as in JustificationHandler we only use `info`, `header` and we call `finalizer::finalize_block`, which uses `lock_import_and_run`, `apply_finality` and `info`. We don't need other traits here. This way mocking client would be easier and maybe we could mock `block_requester` too?

finality-aleph/src/finalization.rs
```diff
@@ -9,34 +9,39 @@ use sp_runtime::{
...
+            // NOTE: all other finalization logic should come here, inside the lock
+            client.apply_finality(import_op, BlockId::Hash(hash), justification, true)
+        });
+        let status = client.info();
+        debug!(target: "afa", "Attempted to finalize block with hash {:?}. Current best: #{:?}.", hash, status.finalized_number);
+        update_res
+    }
 }
 
+impl<BE, B, C> Finalizer<BE, B, C> for () {}
```
maciejnems, line 43: I feel strange about this implementation for unit here. Maybe it would be better to create struct `AlephFinalizer` which implements `Finalizer`?

finality-aleph/src/justification.rs
```diff
@@ -29,26 +27,49 @@ pub struct AlephJustification {
...
+    pub(crate) last_block_height: NumberFor<B>,
+    pub(crate) verifier: Option<AuthorityVerifier>,
+}
+
+/// Returns `SessionInfo` for the session regarding block with no. `number`.
+pub(crate) trait SessionInfoProvider<B: BlockT> {
+    fn for_block_num(&self, number: NumberFor<B>) -> SessionInfo<B>;
+}
+
+impl<F, B> SessionInfoProvider<B> for F
```
maciejnems, line 65: This setup for mocking is very smart and tricky, but I feel like it might be confusing. Right now I don't have any ideas how to mock those functions without such tricks. @DamianStraszak do you have any ideas?

finality-aleph/src/justification.rs
```diff
@@ -226,24 +235,27 @@ where
...
         }
     }
+}
 
-    fn session_verifier(&self, session_id: SessionId) -> Option<AuthorityVerifier> {
-        Some(AuthorityVerifier::new(
-            self.session_authorities.lock().get(&session_id)?.to_vec(),
-        ))
-    }
+fn wrap_channel<B: BlockT>(
```
DamianStraszak, line 247: I would love a more descriptive name, like `wrap_channel_with_logging` or some such :)

DamianStraszak (changes requested):
Some smallish comments. In general I think this is a step in the right direction.

finality-aleph/src/justification.rs
pmikolajczyk41: renamed

finality-aleph/src/justification.rs
```diff
@@ -64,111 +85,114 @@ where
...
+                if let Some(metrics) = &self.metrics {
+                    metrics.report_block(hash, self.last_finalization_time, Checkpoint::Finalized);
                 }
-            } else {
-                debug!(target: "afa", "Not finalizing block {:?}. Last finalized {:?}, stop_h {:?}", num, last_finalized, stop_h);
             }
-        } else {
-            error!(target: "afa", "Error when verifying justification for block {:?} {:?}", num, block_hash);
+            Err(e) => {
+                warn!(target: "afa", "Fail in finalization of {:?} {:?} -- {:?}", number, hash, e);
```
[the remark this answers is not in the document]
pmikolajczyk41: changed

finality-aleph/src/justification.rs
```diff
@@ -64,111 +85,114 @@ where
...
+            hash,
+        } = notification;
+
+        if number <= last_finalized || number > stop_h {
+            debug!(target: "afa", "Not finalizing block {:?}. Last finalized {:?}, stop_h {:?}", number, last_finalized, stop_h);
+            return;
+        };
+
+        if !(justification.verify::<B>(hash, &verifier)) {
+            error!(target: "afa", "Error when verifying justification for block {:?} {:?}", number, hash);
```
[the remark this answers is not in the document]
pmikolajczyk41: changed

finality-aleph/src/justification.rs
```diff
@@ -29,26 +27,49 @@ pub struct AlephJustification {
...
+/// Decides whether enough time has elapsed for making another request for justification.
+pub(crate) trait JustificationRequestDelay {
+    fn can_request_now(&self, last_request_time: Instant, last_finalization_time: Instant) -> bool;
+}
+
+impl<F: Fn(Instant, Instant) -> bool> JustificationRequestDelay for F {
+    fn can_request_now(&self, last_request_time: Instant, last_finalization_time: Instant) -> bool {
+        self(last_request_time, last_finalization_time)
+    }
+}
```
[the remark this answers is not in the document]
pmikolajczyk41: cool! refined

finality-aleph/src/finalization.rs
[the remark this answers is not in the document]
pmikolajczyk41: it required less changes :P I have replaced it with `AlephFinalizer`

pmikolajczyk41: changed to `AlephFinalizer`; surely better, thanks!

finality-aleph/src/justification.rs
pmikolajczyk41: good catch, thanks! I have minimized the required trait; I think that `RequestBlocks` is relatively easy to mock

finality-aleph/src/justification.rs
pmikolajczyk41: there is only `SessionInfoProvider` left; I can remove the blanket implementation for closures and just create a normal struct in `party.rs` if you think that it would increase legibility. I just miss something like functional interfaces from Java :P

DamianStraszak: I'm fine with either version, but there is indeed some trickiness in the current solution that beginner rusteceans might have hard time enjoying :)

DamianStraszak (approved):

maciejnems (approved):

maciejnems:
One nit. Could you also change `should_finalize` and `run_justification_handler`, to not use ClientForAleph?

pmikolajczyk41:
@maciejnems
good point, I must have overlooked it, thank you! changed

PR #215, opened 2021-11-29
Bump substrate-api-client version in flooder

pmikolajczyk41:


fbielejec (approved):
LGTM, like I mentioned in a DM, could use a workflow in GA that runs cargo clippy on these workspaces that are not part of the node binary, to avoid breaking changes happening in the future

fixxxedpoint (approved):

PR #217, opened 2021-11-29
Clean with regions

kostekIV:


benchmark/benchmark.py
```diff
@@ -35,8 +35,10 @@ def get_args() -> argparse.Namespace:
     parser_clean = subparsers.add_parser('clean')
     parser_clean.add_argument('--kill-monitoring', '-k', action='store_true',
                               help='whether to stop docker with Prometheus and Grafana')
+    parser_clean.add_argument('--regions', nargs='+', type=str,
+                              help='list of regions where the machines should be created')
```
pmikolajczyk41, line 39: unify this with corresponding flag for `run` subcommand

benchmark/benchmark.py
```diff
@@ -35,8 +35,10 @@ def get_args() -> argparse.Namespace:
...
     parser_clean.add_argument(
-        'tag', type=str, help='tag for the machines', default='b2')
+        '--tag', type=str, help='tag for the machines', default='b2')
```
pmikolajczyk41, line 41: leftover?

pmikolajczyk41 (changes requested):

benchmark/benchmark.py
pmikolajczyk41: because this change is unrelated to the PR (?)

and is there any reason for "b2"? seems like some working abbreviation for 'benchmark2'

kostekIV: Ok I see what you have meant. It is related though, because the regions argument accept a list of args so you cant have a positional arg after it. Regarding 'b2' I just didnt change what was already here.

PR #219, opened 2021-11-30
AZ-107: Another tiny step towards new network

timorl:


Yithis (commented):
Looks good to me with one small comment

finality-aleph/src/new_network/connection_manager/connections.rs
```diff
@@ -0,0 +1,85 @@
...
+    use crate::{new_network::PeerId, SessionId};
+    use sc_network::PeerId as ScPeerId;
+    use std::collections::HashSet;
+
+    fn random_peer_ids(num: usize) -> HashSet<PeerId> {
+        (0..num).map(|_| PeerId(ScPeerId::random())).collect()
+    }
+
+    #[test]
+    fn removes_peer_after_single_session() {
```
Yithis, line 65: I hate to be that pettifogging, but you wrote unit tests, that are assuming correctness of functions `add_peers` and `remove_session` without testing them separately

timorl: They cannot be tested separately using the public interface, without inspecting the implementation details of the struct. The only behaviour I can test is by calling them in various configurations and looking at the outputs of the latter. This is also the contract that this struct is supposed to fulfill, so I don't consider this a problem.

"Pettifogging" is a brilliant word (and approach ;3), I did not know that one.

finality-aleph/src/new_network/connection_manager/connections.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// Creates a new object, initially without any connections.
+    pub fn new() -> Self {
+        Connections {
+            associated_sessions: HashMap::new(),
+            peers_by_session: HashMap::new(),
+        }
+    }
+
+    /// Mark the specified peers as ones we should be connected to for the given session.
+    pub fn add_peers(&mut self, session_id: SessionId, peers: HashSet<PeerId>) {
```
fixxxedpoint, line 20:
```suggestion
         pub fn add_peers(&mut self, session_id: SessionId, peers: impl Iterator<Item = PeerId>) {
```

finality-aleph/src/new_network/connection_manager/connections.rs
```diff
@@ -0,0 +1,85 @@
...
+    fn does_not_remove_peer_if_still_in_session() {
+        let session_id = SessionId(43);
+        let other_session_id = SessionId(2137);
+        let peer_ids = random_peer_ids(1);
+        let mut connections = Connections::new();
+        connections.add_peers(session_id, peer_ids.clone());
+        connections.add_peers(other_session_id, peer_ids);
+        let to_remove = connections.remove_session(session_id);
+        assert!(to_remove.is_empty());
+    }
```
fixxxedpoint, line 84: Consider adding few more simple tests, e.g. similar to `does_not_remove_peer_if_still_in_session` but check if it actually removes peers after you call `.remove_session(other_session_id_`; with some other combinations (how they intersect) between two different sessions, etc.

fixxxedpoint (changes requested):
few suggestions and request for more tests

finality-aleph/src/new_network/connection_manager/connections.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// Mark the specified peers as ones we should be connected to for the given session.
+    pub fn add_peers(&mut self, session_id: SessionId, peers: HashSet<PeerId>) {
+        for peer in peers {
+            self.associated_sessions
+                .entry(peer)
+                .or_insert_with(HashSet::new)
+                .insert(session_id);
+            self.peers_by_session
+                .entry(session_id)
+                .or_insert_with(HashSet::new)
```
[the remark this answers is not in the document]
timorl: Done. I've been wondering whether this would work, but I was too lazy to check. <_<"

finality-aleph/src/new_network/connection_manager/connections.rs
timorl: Done.

timorl: Although I did `IntoIterator` rather than `Iterator`.

finality-aleph/src/new_network/connection_manager/connections.rs
timorl: Added only a single test, not to overdo this. ;p

If you have more concrete suggestions let me know.

finality-aleph/src/new_network/connection_manager/connections.rs
fixxxedpoint: That's even better.

fixxxedpoint (approved):
lgtm

PR #221, opened 2021-11-30
Pipeline for mainnet deployment on release added

skrolikowski-10c:
The deployment job is placed within github mainnet environment, that requires pipeline to be accept before launch. It's an additional fuse in order to make sure that the deployment is indeed what the user wants to do.

### DIFF
--- a/.github/workflows/deploy-mainnet.yml
+++ b/.github/workflows/deploy-mainnet.yml
@@ -1,0 +1,96 @@
+name: Deploy to Mainnet
+
+on:
+  release:
+    types: [released]
+
+jobs:
+  deploy-mainnet:
+    name: Deploy new aleph-node image to Mainnet EKS
+    runs-on: ubuntu-latest
+    environment: 'mainnet'
+    env:
+      AWS_REGION: us-east-1 # this region is used by all public ECR repos
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-mainnet.yml
```diff
@@ -0,0 +1,96 @@
...
+          GITHUB_TOKEN: ${{ secrets.CI_GH_TOKEN }}
+
+      - name: Send Slack Notification
+        uses: rtCamp/action-slack-notify@v2
+        continue-on-error: true
+        env:
+          SLACK_COLOR: ${{ job.status }}
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+          SLACK_USERNAME: GithubActions
+          SLACK_TITLE: deploy-testnet job has finished
```
bartoszjedrzejewski, line 95: deploy-testnet, should be mainnet

bartoszjedrzejewski:
There's no actual check if this image exists. If someone would do release right away, it would change the images in aleph-apps repo, even though those images don't exist in ECR. Please add an additional step, checking if that image actually exists.

bartoszjedrzejewski (changes requested):

bartoszjedrzejewski (approved):

PR #225, opened 2021-12-03
AZ-107: Validator discovery logic

timorl:


finality-aleph/src/new_network/connection_manager/session.rs
```diff
@@ -396,6 +410,7 @@ mod tests {
         let expected_missing: Vec<_> = (1..NUM_NODES).map(NodeIndex).collect();
         assert_eq!(missing_nodes, expected_missing);
         assert!(handler0.peer_id(&NodeIndex(1)).is_none());
+        assert!(handler0.authentication_for(&NodeIndex(1)).is_none());
     }
 
     #[tokio::test]
```
pmikolajczyk41, line 416: I guess it is worth checking whether `authentication_for` adapts well to the cached update

finality-aleph/src/new_network/connection_manager/session.rs
```diff
@@ -342,6 +355,7 @@ mod tests {
         let missing_nodes = handler0.missing_nodes();
         let expected_missing: Vec<_> = (1..NUM_NODES).map(NodeIndex).collect();
         assert_eq!(missing_nodes, expected_missing);
+        assert!(handler0.authentication_for(&NodeIndex(1)).is_none());
```
pmikolajczyk41, line 358: such check is missing from e.g. `ignores_wrong_session_authentication`; as I understand, there is an implicit implication: `x \in missing nodes  => authentication_for(x).is_none()`, but still we should check this anyway

finality-aleph/src/new_network/connection_manager/session.rs
```diff
@@ -88,6 +88,14 @@ impl Handler {
         self.own_authentication.clone()
     }
 
+    /// Returns the authentication for the node with the given index, if we have it.
+    pub fn authentication_for(&self, node_id: &NodeIndex) -> Option<Authentication> {
```
pmikolajczyk41, line 92: for completeness, I would add a testcase presenting what this method returns for own id

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+                    .unwrap(),
+            );
+        }
+        (
+            Discovery::new(time::Duration::from_millis(MS_COOLDOWN)),
+            handlers,
+        )
+    }
+
+    #[tokio::test]
```
pmikolajczyk41, line 306: follow-up to the today slack discussion: is there any testcase in this module, that should be rather run with multithreading? I do not see such, but just to be doubly sure

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+    collections::{HashMap, HashSet},
+    time,
+};
+
+/// Messages used for discovery and authentication.
+#[derive(Clone, Debug, Encode, Decode)]
+pub enum DiscoveryMessage {
+    // Contains the id of the broadcast, to avoid spam.
+    AuthenticationBroadcast(Authentication),
+    // Requests always contain own authentication, to avoid asymetric trust.
```
pmikolajczyk41, line 19: typo: asymmetric

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+use codec::{Decode, Encode};
+use std::{
+    collections::{HashMap, HashSet},
+    time,
+};
+
+/// Messages used for discovery and authentication.
+#[derive(Clone, Debug, Encode, Decode)]
+pub enum DiscoveryMessage {
+    // Contains the id of the broadcast, to avoid spam.
```
pmikolajczyk41, line 17: I have problems with understanding the term 'id of the broadcast'

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+    Authentications(Vec<Authentication>),
+}
+
+impl DiscoveryMessage {
+    pub fn session_id(&self) -> SessionId {
+        use DiscoveryMessage::*;
+        match self {
+            AuthenticationBroadcast((auth_data, _)) => auth_data.session(),
+            Request(_, (auth_data, _)) => auth_data.session(),
+            Authentications(auths) => match auths.get(0) {
```
pmikolajczyk41, line 30: as far as I understand, we assume that all the authentications come from the same session; I would add either a check or at least a comment near the variant

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+                Some((auth_data, _)) => auth_data.session(),
+                None => SessionId(0), // Broken message anyway, value doesn't matter.
+            },
+        }
+    }
+}
+
+/// Handles creating and responding to discovery messages.
+pub struct Discovery {
+    cooldown: time::Duration,
```
pmikolajczyk41, line 40: is there a reason for a qualified name? also for `Instant`

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+    last_broadcast: HashMap<NodeIndex, time::Instant>,
+    last_response: HashMap<NodeIndex, time::Instant>,
+    requested_authentications: HashMap<NodeIndex, HashSet<NodeIndex>>,
+    next_query: NodeIndex,
```
pmikolajczyk41, line 44: probably deserves a brief explanation

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+}
+
+fn authentication_broadcast(authentication: Authentication) -> (DiscoveryMessage, DataCommand) {
```
pmikolajczyk41, line 47: maybe it is a matter of taste, but I would introduce a new auxiliary struct `DMDC` (just a placeholder, you're far better with naming than I, so up to you :P) as an alias for `(DiscoveryMessage, DataCommand)`
it would have (at least) three constructor-methods corresponding to these functions

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+    ) -> Vec<Multiaddr> {
+        authentications
+            .into_iter()
+            .flat_map(|authentication| self.handle_authentication(authentication, handler))
+            .collect()
+    }
+
+    /// Analyzes the provided message and returns all the new multiaddresses we should
+    /// be connected to and any messages that we should send as a result of it.
+    pub fn handle_message(
```
pmikolajczyk41, line 217: again, matter of taste, but I highly prefer public API to be at the top of class definition

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+    }
+
+    fn handle_response(
+        &mut self,
+        authentications: Vec<Authentication>,
+        handler: &mut SessionHandler,
+    ) -> Vec<Multiaddr> {
+        authentications
+            .into_iter()
+            .flat_map(|authentication| self.handle_authentication(authentication, handler))
```
pmikolajczyk41, line 211: should we remove the duplicates?

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+            .collect()
+    }
+
+    /// Analyzes the provided message and returns all the new multiaddresses we should
+    /// be connected to and any messages that we should send as a result of it.
+    pub fn handle_message(
+        &mut self,
+        message: DiscoveryMessage,
+        handler: &mut SessionHandler,
+    ) -> (Vec<Multiaddr>, Option<(DiscoveryMessage, DataCommand)>) {
```
pmikolajczyk41, line 221: I woudl go for an alias for `(Vec<Multiaddr>, Option<(DiscoveryMessage, DataCommand)>)` - it would both better document the auxiliary functions and make the public API more descriptive

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+            None => true,
+        };
+        let message = match rebroadcast {
+            true => {
+                self.last_broadcast.insert(node_id, time::Instant::now());
+                Some(authentication_broadcast(authentication))
+            }
+            false => None,
+        };
+        (addresses, message)
```
pmikolajczyk41, line 153: sorry, this is still a new context for me and I have some problems with connecting things and this is some sanity check: the node who gets a broadcast message not necessarily will interact with asked peers, i.e. `addresses`, so there may not be a need to keep connections with them, right?

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+    ) -> (Vec<Multiaddr>, Option<(DiscoveryMessage, DataCommand)>) {
+        let addresses = self.handle_authentication(authentication.clone(), handler);
+        if addresses.is_empty() {
+            return (addresses, None);
+        }
+        let node_id = authentication.0.creator();
+        let rebroadcast = match self.last_broadcast.get(&node_id) {
+            Some(instant) => time::Instant::now() > *instant + self.cooldown,
+            None => true,
+        };
```
pmikolajczyk41, line 145: - do I understand correctly: if the last broadcast from `node_id` was long time ago, we are trying to help them?
 - what about a tiny pretty auxiliary method `should_rebroadcast()`?

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+        &mut self,
+        requester_id: NodeIndex,
+        node_ids: Vec<NodeIndex>,
+        handler: &mut SessionHandler,
+    ) -> Option<(DiscoveryMessage, DataCommand)> {
+        let requested_authentications = self
+            .requested_authentications
+            .entry(requester_id)
+            .or_default();
+        requested_authentications.extend(node_ids);
```
pmikolajczyk41, line 166: what about possible duplicates?

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+}
+
+impl Discovery {
+    /// Create a new discovery handler with the given response/broadcast cooldown.
+    pub fn new(cooldown: time::Duration) -> Self {
+        Discovery {
+            cooldown,
+            last_broadcast: HashMap::new(),
+            last_response: HashMap::new(),
+            requested_authentications: HashMap::new(),
```
pmikolajczyk41, line 82: can we say anything about (upper bound for) the size of these containers?

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+        node_ids: Vec<NodeIndex>,
+        handler: &mut SessionHandler,
+    ) -> Option<(DiscoveryMessage, DataCommand)> {
+        let requested_authentications = self
+            .requested_authentications
+            .entry(requester_id)
+            .or_default();
+        requested_authentications.extend(node_ids);
+        if let Some(instant) = self.last_response.get(&requester_id) {
+            if time::Instant::now() < *instant + self.cooldown {
```
pmikolajczyk41, line 168: just thinking: is a common cooldown for both broadcast and direct messages appropriate? probably yes :thinking:

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+        let (addresses, maybe_command) = discovery.handle_message(
+            DiscoveryMessage::Request(vec![requested_node_id], requester_authentication.clone()),
+            handler,
+        );
+        assert_eq!(addresses.len(), requester_authentication.0.address().len());
+        assert_eq!(
+            addresses[0].encode(),
+            requester_authentication.0.address()[0].encode()
+        );
+        match maybe_command {
```
pmikolajczyk41, line 462: there is also identical piece of code below, and a few very similar; I know that these are tests etc., but just consider whether there is an easy way of simplifying things here; if not, then there is no problem ofc

pmikolajczyk41:
I have left just a few nits, but generally great work, kudos! :1st_place_medal:

finality-aleph/src/new_network/connection_manager/mod.rs
```diff
@@ -43,6 +46,20 @@ pub struct AuthData {
...
+impl AuthData {
+    pub fn session(&self) -> SessionId {
+        self.session_id
+    }
+
+    pub fn creator(&self) -> NodeIndex {
+        self.node_id
+    }
+
+    pub fn address(&self) -> Vec<Multiaddr> {
```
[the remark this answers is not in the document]
timorl: Done, for some reason I've been scared this would conflict with the field name. <_<"

finality-aleph/src/new_network/connection_manager/session.rs
timorl: Added check.

finality-aleph/src/new_network/connection_manager/session.rs
timorl: Added.

finality-aleph/src/new_network/connection_manager/session.rs
timorl: Added.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: I am assuming all the calls to a specific `Discovery` are performed on one thread, so probably not. On the other hand many of the tests have multiple `Discovery`s, which theoretically could be run on separate threads, only this would be hard to arrange and not really add any value (as they are completely separate entities sharing nothing).

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Fixed.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Uhhh, that was from a former version of spam avoidance, outdated comment removed. ">_>

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Added comment to the variant.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Ehhh~, it kinda feels right to have these explicitly part of `time`, but I cannot actually figure out why I feel that way. Unqalified these.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Added short comment.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Done, called `DiscoveryCommand`. Unfortunately official constructors are not possible with type aliases.

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+        let node_count = handler.node_count();
+        let authentication = handler.authentication();
+        if missing_authorities.len() * 3 > 2 * node_count.0 {
+            // We know of fewer than 1/3 authorities, broadcast our authentication and hope others
+            // respond in kind.
+            vec![authentication_broadcast(authentication)]
+        } else {
+            // Attempt learning about more authorities from the ones you already know.
+            let mut result = Vec::new();
+            let mut target = self.next_query;
```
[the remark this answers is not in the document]
timorl: A loopback message wouldn't actually hurt us (and it shouldn't since if it did getting it from outside would be a possible attack), but this is now explicitly checked due to one unit test you proposed in another comment. Additionally filtering it here seems superfluous.

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+            while result.len() < 2 {
```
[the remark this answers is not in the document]
timorl: Extracted constant. I also noticed that for sufficiently small committee sizes (<7?) this might end up sending two queries to a single node, but I don't think that's much of a problem.

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+            }
+            self.next_query = target;
+            result
+        }
+    }
+
+    fn handle_authentication(
+        &mut self,
+        authentication: Authentication,
+        handler: &mut SessionHandler,
```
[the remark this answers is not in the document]
timorl: Ouch, I've been thinking a lot about this. If we want to do this this way it would probably have to be behind an `Arc<Mutex<..>>` or expose a separate interface for accessing the internal handler. Both are kinda ugly, but so is the current solution... I ended up choosing this one, and assuming I would keep pairs `(Discovery, SessionHandler)` in the object that will end up using this, but it should be easy to convince me that this is not the correct choice.

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+                    ));
+                }
+                target = NodeIndex((target.0 + 1) % node_count.0);
+            }
+            self.next_query = target;
+            result
+        }
+    }
+
+    fn handle_authentication(
```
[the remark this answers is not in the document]
timorl: Added comment.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Well, we clearly have different taste then, I prefer functions being defined mostly in order of calls, so you read a function before seeing it used.

Do you think this (say, your approach) should be enforced as part of code style in our repos?

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Shouldn't really matter, we might return the same address many times for other reasons, so I'm counting on the idempotence of `add_reserved` anyway.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: I think after the change you proposed above, making an alias for the type inside the internal parentheses, aliasing this would obfuscate more than help.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Yes, but this module does not contain logic that decides that – it always returns the addresses the node should be connected to if it wants to stay connected to the committee.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: If it was long ago then we can do it again, so we do. It's just a spam/congestion avoidance scheme.

Added auxiliary method.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: It's a `HashSet` :sunglasses:

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Yup, at most one item per node in session (we only add stuff that is in some way authenticated). The instants are probably small, the hashsets of nodeindices are again bound by the number of nodes in the session.

Whoops, the latter actually weren't, added a check.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Lets say it's not inappropriate. <_<"

They could be separate, but I couldn't think of any reason we would make them such.

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+            .map(|id| {
+                ScMultiaddr::empty()
+                    .with(ScProtocol::Ip4(Ipv4Addr::new(192, 168, 1, id)))
+                    .with(ScProtocol::Tcp(30333))
+                    .with(ScProtocol::P2p(ScPeerId::random().into()))
+            })
+            .collect()
+    }
+
+    async fn keyboxes() -> Vec<KeyBox> {
```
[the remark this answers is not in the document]
timorl: Yeah... I probably should do that, I've been postponing it a tad. Unfortunately this will be an annoying conflict with Jędrzej's PR. <_<"

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+                    .await
+                    .expect("The keys should sign successfully"),
+            ));
+        }
+        result
+    }
+
+    async fn build() -> (Discovery, Vec<SessionHandler>) {
+        let mut handlers = Vec::new();
+        for (keybox, address) in keyboxes().await.into_iter().zip(addresses()) {
```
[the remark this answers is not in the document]
timorl: Doesn't work with `async` ¯\\\_( ツ )_/¯

finality-aleph/src/new_network/connection_manager/discovery.rs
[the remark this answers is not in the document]
pmikolajczyk41: Then I do not see a problem to do this later, I mean in a separate PR.

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+        authentication: Authentication,
+        handler: &mut SessionHandler,
+    ) -> Vec<Multiaddr> {
+        if !handler.handle_authentication(authentication.clone()) {
+            return Vec::new();
+        }
+        authentication.0.address()
+    }
+
+    fn handle_broadcast(
```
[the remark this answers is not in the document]
timorl: Whoops, you are right, we should, I thought it wasn't necessary, but there is a very strange failure mode possible here. I will think about this a bit before fixing it, as it might require changing all the return types, which would be quite annoying...

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,690 @@
...
+        let mut messages = discovery.discover_authorities(handler);
+        assert_eq!(messages.len(), 1);
+        let message = messages.pop().unwrap();
+        assert_eq!(message.1, DataCommand::Broadcast);
+        match message.0 {
+            DiscoveryMessage::AuthenticationBroadcast(authentication) => {
+                assert_eq!(authentication.encode(), handler.authentication().encode())
+            }
+            _ => panic!("Expected an authentication broadcast, got {:?}", message.0),
+        }
```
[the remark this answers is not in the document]
timorl: Egh, I would prefer not to do that, equality of authentications is kinda icky?

finality-aleph/src/new_network/connection_manager/discovery.rs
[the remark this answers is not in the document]
pmikolajczyk41: Could you please explain this failure mode?

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: These have small annoying differences between them, I think implementing a general abstraction over them would be slightly more work than it's worth.

finality-aleph/src/new_network/connection_manager/discovery.rs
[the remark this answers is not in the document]
pmikolajczyk41: I meant exactly the check you do here, i.e. pattern-matching + `==` on authentication (this is what `assert_eq` does I believe)

finality-aleph/src/new_network/connection_manager/discovery.rs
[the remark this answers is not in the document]
pmikolajczyk41: This is for low number of network connection, not the committee size, right? I think it is worth keeping in mind, but rather out of scope of this code

finality-aleph/src/new_network/connection_manager/discovery.rs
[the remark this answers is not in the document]
pmikolajczyk41: I would go probably for the `Arc<Mutex<..>>` solution (btw, maybe it deserves a shortcut type), since I really don't like a bunch of methods with the same argument passed all around. However, unwrapping it everywhere would be awful and maybe some setter would be necessary (for changing `SessionHandler`?). 
I would leave the decision to the second reviewer, @Yithis :stuck_out_tongue_closed_eyes:

finality-aleph/src/new_network/connection_manager/discovery.rs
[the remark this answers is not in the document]
pmikolajczyk41: I see, but now I am also not so convinced why we should keep these connections to future committee. We do not necessarily need them, since we can currently be just a passive node participating in the network. 

But maybe such enriching network is somehow useful / potentially improving overall performance :thinking:

finality-aleph/src/new_network/connection_manager/discovery.rs
pmikolajczyk41: Yes, I think that a uniform approach should be taken. However, I do not insist.

finality-aleph/src/new_network/connection_manager/discovery.rs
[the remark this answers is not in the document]
pmikolajczyk41: Okay, you answered it here: https://github.com/Cardinal-Cryptography/aleph-node/pull/225#discussion_r763829773

Resolving

finality-aleph/src/new_network/connection_manager/discovery.rs
pmikolajczyk41: Fair point, but then, to avoid such confusion, I would modify just a little bit the comment documenting `handle_message`, to emphasize that this: _should be connected to **if it wants to stay connected to the committee**_

finality-aleph/src/new_network/connection_manager/discovery.rs
pmikolajczyk41: but over time, the number of entries can be as large as the number of all nodes that could reach us? should we be careful with that?

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,719 @@
...
+
+impl DiscoveryMessage {
+    pub fn session_id(&self) -> SessionId {
+        use DiscoveryMessage::*;
+        match self {
+            AuthenticationBroadcast((auth_data, _)) => auth_data.session(),
+            Request(_, (auth_data, _)) => auth_data.session(),
+            Authentications(auths) => match auths.get(0) {
+                Some((auth_data, _)) => auth_data.session(),
+                None => SessionId(0), // Broken message anyway, value doesn't matter.
```
Yithis, line 37: Don't we want to log it somehow in any case?

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,719 @@
...
+
+impl Discovery {
+    /// Create a new discovery handler with the given response/broadcast cooldown.
+    pub fn new(cooldown: Duration) -> Self {
+        Discovery {
+            cooldown,
+            last_broadcast: HashMap::new(),
+            last_response: HashMap::new(),
+            requested_authentications: HashMap::new(),
+            next_query: NodeIndex(0),
```
Yithis, line 88: Wouldn't it be better to somehow randomize this point?

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,719 @@
...
+            let mut target = self.next_query;
+            while result.len() < NODES_TO_QUERY {
+                if let Some(peer_id) = handler.peer_id(&target) {
+                    result.push(request(
+                        missing_authorities.clone(),
+                        authentication.clone(),
+                        peer_id,
+                    ));
+                }
+                target = NodeIndex((target.0 + 1) % node_count.0);
```
Yithis, line 121: Maybe also randomize?

finality-aleph/src/new_network/connection_manager/discovery.rs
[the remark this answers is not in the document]
Yithis: So sorry @pmikolajczyk41, but I'm with @timorl here :sweat_smile:

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,719 @@
...
+    }
+
+    fn handle_broadcast(
+        &mut self,
+        authentication: Authentication,
+        handler: &mut SessionHandler,
+    ) -> (Vec<Multiaddr>, Option<DiscoveryCommand>) {
+        let addresses = self.handle_authentication(authentication.clone(), handler);
+        if addresses.is_empty() {
+            return (addresses, None);
```
Yithis, line 155: in some sense I don't like the syntax, where you assign var, check if it isn't empty and return it, but don't have a good idea how to solve it.

What if addresses isn't empty, but only address there is our own? Or there is only address of message sender? I think here are some more cases

Yithis: Second comment applies also to other functions

finality-aleph/src/new_network/connection_manager/discovery.rs
Yithis: Again I'm here with @timorl, but I agree that it should be uniform across all repos.

finality-aleph/src/new_network/connection_manager/discovery.rs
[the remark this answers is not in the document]
pmikolajczyk41: sure, thus resolving the conversation

finality-aleph/src/new_network/connection_manager/discovery.rs
pmikolajczyk41: okay, I'm resolving the conversation, but probably the issue is worth keeping in mind

finality-aleph/src/new_network/connection_manager/discovery.rs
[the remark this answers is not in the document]
timorl: Committee size actually, since this code only runs if we are connected to at least 1/3 of the nodes, which is >=2 for committees >=7.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Huh? I'm kind of confused, the number of nodes is the number of committee members in this case, and that is bounded. Discovery exists per-session, so it will get deleted after a session passes.

finality-aleph/src/new_network/connection_manager/discovery.rs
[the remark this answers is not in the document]
timorl: Rather than doing `==` on authentications I compare their encodings. It's a workaround, but maybe it's stupid, I'll check how annoying it is to derive `PartialEq` for them.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Good point, although I'll probably do this outside of this method (unwritten code as of now <_<"). Here I only want to return a session id without worrying about this corner case.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Possibly, but I'm not convinced the added value is significant, while randomness is sorta annoying to deal with.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Same response as above, I believe this is good enough.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: I've been wondering whether it's better to return `Vec::new()` here, very unsure. Do you think this would be clearer?

If someone sends us an authentication with addresses that don't actually resolve to them then they just wasted their time. Our own libp2p address cannot be there, since then the handler would not have accepted it. If there is our own address with another peer id then using it shouldn't break anything in libp2p, other than not being able to connect to the node who sent us this, so again just wasted time.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Added.

finality-aleph/src/new_network/connection_manager/discovery.rs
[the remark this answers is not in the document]
timorl: Phew, done by changing the responses into `Vec`s. Only slightly annoying and probably actually better this way anyway. <_<"

finality-aleph/src/new_network/connection_manager/discovery.rs
Yithis: Why? It doesn't have to be perfect randomness, I just think it can help with distribution of network traffic

finality-aleph/src/new_network/connection_manager/discovery.rs
Yithis: Yes, I thought about it also and I think it would be better.

But such messages shouldn't also waste our time. But I understand if it is not a case here.

finality-aleph/src/new_network/connection_manager/discovery.rs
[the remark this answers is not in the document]
timorl: It wasn't, I guess I should just always derive that. <_<"

Anyway, done.

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,719 @@
...
+    #[tokio::test]
+    async fn rebroadcasts_after_cooldown() {
+        let (mut discovery, mut handlers) = build().await;
+        let authentication = handlers[1].authentication();
+        let handler = &mut handlers[0];
+        discovery.handle_message(
+            DiscoveryMessage::AuthenticationBroadcast(authentication.clone()),
+            handler,
+        );
+        sleep(Duration::from_millis(MS_COOLDOWN + 5));
```
Yithis, line 440: I don't really like the idea of using sleep in tests, but I'm not sure if we can overcome this with lost cost

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Changed.

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -0,0 +1,628 @@
...
+        for i in 1..num_nodes - 1 {
+            let authentication = handlers[i].authentication();
+            assert!(handlers[0].handle_authentication(authentication));
+        }
+        let handler = &mut handlers[0];
+        let messages = discovery.discover_authorities(handler);
+        assert_eq!(messages.len(), 2);
+        for message in messages {
+            assert!(
+                matches!(message, (DiscoveryMessage::Request(node_ids, authentication), DataCommand::SendTo(_, _)) if node_ids == vec![NodeIndex(6)] && authentication == handler.authentication())
```
pmikolajczyk41, line 355: uhm, please wrap such long lines :| clippy/fmt have no power within macros

finality-aleph/src/new_network/connection_manager/discovery.rs
pmikolajczyk41: +1 for taste, but the only reasonable solution for this I see would be to expose some (private) methods within `Discovery` on which we can listen asynchronously. But probably in this case this is an overkill

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Added some randomness, not quite happy with the outcome, but it should help with the failure mode you mentioned to me.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Ouch, I missed that, bleh, fixing.

timorl: Fixed, although it seems fmt has some power somewhere, only not consistently? Strange anyway.

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: I really don't have better ideas here, as this feature is entirely time-based. :<

pmikolajczyk41 (approved):

finality-aleph/src/new_network/connection_manager/discovery.rs
Yithis: Theoretically you can mock time suppliers, but it is probably an overkill

Yithis (approved):

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: I agree this would be overkill, resolving as per private communication.

PR #226, opened 2021-12-06
AZ-363: Split Keybox in the new network

Yithis:


finality-aleph/src/new_network/connection_manager/session.rs
```diff
@@ -30,10 +31,10 @@ pub enum AddressError {
 }
 
 async fn construct_authentication(
-    keychain: &KeyBox,
-    session_id: SessionId,
+    authority_index_and_pen: &Option<(NodeIndex, AuthorityPen)>,
+    session_id: Option<SessionId>,
```
timorl, line 35: Pass this always (it makes sense always), just don't use it sometimes.

It should also be useful for returning the session id from the method below (I'm not sure, but it might be necessary even for non-validators; it certainly makes sense even for non-validators). You might want to avoid data duplication by making the field holding the authentication into some form of `Either` (perhaps custom defined?).

finality-aleph/src/new_network/connection_manager/session.rs
```diff
@@ -94,19 +104,29 @@ impl Handler {
         if self.peers_by_node.len() + 1 == node_count {
             return Vec::new();
         }
-        (0..node_count)
-            .map(NodeIndex)
-            .filter(|node_id| *node_id != self.index() && !self.peers_by_node.contains_key(node_id))
-            .collect()
+        match self.index() {
```
timorl, line 107: This can probably be slightly simplified by using `map_or`.

finality-aleph/src/new_network/connection_manager/session.rs
```diff
@@ -94,19 +104,29 @@ impl Handler {
...
     }
 
     /// Verifies the authentication, uses it to update mappings, and returns whether we should
     /// remain connected to the multiaddresses.
     pub fn handle_authentication(&mut self, authentication: Authentication) -> bool {
         let (auth_data, signature) = authentication.clone();
-        if auth_data.session_id != self.session_id() {
-            return false;
+
+        if let Some(session_id) = self.session_id() {
```
timorl, line 124: Yeah, `session_id` definitely should return something always.

finality-aleph/src/new_network/connection_manager/session.rs
```diff
@@ -197,22 +220,24 @@ mod tests {
 
     const NUM_NODES: usize = 7;
 
-    async fn keyboxes() -> Vec<KeyBox> {
-        let num_keyboxes = NUM_NODES;
+    async fn keyboxes_components() -> Vec<(Option<(NodeIndex, AuthorityPen)>, AuthorityVerifier)> {
```
timorl, line 223: I kinda hate this name, since we don't really have any keyboxes here anymore. `crypto_basics`? I dunno, maybe you have some better ideas?

timorl: Also doesn't make sense to return always `Some` with an `Option` – I would suggest packing it into `Some` outside of this function if needed.

Also-also, this only needs to return a single `AuthorityVerifier` – they are all identical. If it doesn't implement `Clone` then it should.

finality-aleph/src/new_network/connection_manager/session.rs
```diff
@@ -265,39 +293,74 @@ mod tests {
...
+            authority_verifier,
             SessionId(43),
             local_p2p_addresses()
         )
         .await
         .is_ok());
     }
 
+    #[tokio::test]
+    async fn creates_without_node_index_nor_authority_pen() {
```
timorl, line 309: You should also test somewhere that such a handler returns `None` from `authentication()`.

timorl (changes requested):
A couple comments but in general nice and thanks a lot for this!

finality-aleph/src/new_network/connection_manager/session.rs
Yithis: Done

finality-aleph/src/new_network/connection_manager/session.rs
Yithis: Done

finality-aleph/src/new_network/connection_manager/session.rs
Yithis: Done

finality-aleph/src/new_network/connection_manager/session.rs
Yithis: Done

finality-aleph/src/new_network/connection_manager/session.rs
Yithis: Done

timorl: I cannot find this, did you forget to commit it or am I blind (we already recently saw that the latter is quite probable ;p)?

timorl (approved):
The unresolved comment is still an issue, please rebase and fix the single clippy complaint, but in general LGTM.

Edit: nvm about the clippy complaint, no idea why I saw a version with it but it doesn't seem to be present.

finality-aleph/src/new_network/connection_manager/session.rs
Yithis: This time it is my mistake, forgot to push :sweat_smile:

finality-aleph/src/new_network/connection_manager/session.rs
```diff
@@ -317,6 +317,18 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn handler_without_node_index_nor_authority_pen_returns_none_for_authentication() {
```
timorl, line 321: I know I started with the horribly long test names, but maybe here it could just be `non_validator_handler_returns...`? :P

timorl (approved):
Silly nit. ^^"

finality-aleph/src/new_network/connection_manager/session.rs
```diff
@@ -186,7 +216,7 @@ impl Handler {
 mod tests {
```
fixxxedpoint, line 216: I might have just missed it or author's intention was different (was it?) - can you upgrade a non-validator's handler to a validator's handler. Anywa, there are no test cases checking such transition (or I just missed them).

finality-aleph/src/new_network/connection_manager/session.rs
```diff
@@ -30,10 +37,10 @@ pub enum AddressError {
 }
 
 async fn construct_authentication(
```
fixxxedpoint, line 39: Name is now confusing and it also looks like a merge of two unrelated functions.

fixxxedpoint (changes requested):
I left you few comments.

finality-aleph/src/new_network/connection_manager/session.rs
```diff
@@ -151,16 +178,19 @@ impl Handler {
...
         let authentications = self.authentications.clone();
         self.authentications = HashMap::new();
         self.peers_by_node = HashMap::new();
-        self.keychain = keychain;
-        self.own_authentication = own_authentication;
+        self.authority_index_and_pen = authority_index_and_pen;
+        self.authority_verifier = authority_verifier;
+        self.session_info = session_info;
         self.own_peer_id = own_peer_id;
         for (_, (auth, maybe_auth)) in authentications {
```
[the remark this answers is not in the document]
timorl: I think this is more a comment for me (since I designed this originally) so I'll try explaining myself.

First of all, yeah, there is definitely annoying duplication here, and I would be quite happy if it could get deduplicated.

For why this is coded this way despite the above let me describe the usual path a handler takes. It gets created whenever we start caring about a session (usually a couple sessions ahead). Sometimes some of the session keys of the validators can change after creation. When this happens we would like to be able to stay connected to all the validators for which the authentications we have are still valid (often this would be almost all of them). This is why we want to recheck all the authentications we already have. It is also likely that we get an authentication signed with new keys before we are aware they are changing – that's why we keep one additional authentication and also recheck it here.

There are ways of rewriting the code keeping the above use pattern simple, most obviously by rewriting `update` to take `self` by value, but I'm not convinced they are actually better in this case. Some concrete options you think are better would be useful.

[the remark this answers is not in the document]
fixxxedpoint:
```suggestion
    pub async fn update(
        &mut self,
        authority_index_and_pen: Option<(NodeIndex, AuthorityPen)>,
        authority_verifier: AuthorityVerifier,
        addresses: Vec<Multiaddr>,
    ) -> Result<Vec<Multiaddr>, AddressError> {
        let authentications = std::mem::take(&mut self.authentications);
        *self = Handler::new(
            authority_index_and_pen,
            authority_verifier,
            self.session_id(),
            addresses,
        )
        .await?;
        for (_, (auth, maybe_auth)) in authentications {
            print!(
                "normal authentication: {:?}",
                self.handle_authentication(auth)
            );
            if let Some(auth) = maybe_auth {
                print!(
                    "alternative authentication: {:?}",
                    self.handle_authentication(auth)
                );
            }
        }
        Ok(self
            .authentications
            .values()
            .flat_map(|((auth_data, _), _)| auth_data.addresses.iter().cloned())
            .collect())
    }
```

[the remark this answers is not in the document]
timorl: Huh, cool.

finality-aleph/src/new_network/connection_manager/session.rs
```diff
@@ -96,17 +119,21 @@ impl Handler {
...
-            .filter(|node_id| *node_id != self.index() && !self.peers_by_node.contains_key(node_id))
+            .filter(|node_id| {
+                Some(*node_id) != self.index() && !self.peers_by_node.contains_key(node_id)
+            })
             .collect()
     }
 
     /// Verifies the authentication, uses it to update mappings, and returns whether we should
     /// remain connected to the multiaddresses.
     pub fn handle_authentication(&mut self, authentication: Authentication) -> bool {
```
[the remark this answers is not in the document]
Yithis: I don't agree, this alias is used many times in whole module. Changing alias here would lower readability

finality-aleph/src/new_network/connection_manager/session.rs
```diff
@@ -151,16 +178,19 @@ impl Handler {
...
+        authority_index_and_pen: Option<(NodeIndex, AuthorityPen)>,
+        authority_verifier: AuthorityVerifier,
         addresses: Vec<Multiaddr>,
     ) -> Result<Vec<Multiaddr>, AddressError> {
-        let (own_authentication, own_peer_id) =
-            construct_authentication(&keychain, self.session_id(), addresses).await?;
+        let (session_info, own_peer_id) =
+            construct_authentication(&authority_index_and_pen, self.session_id(), addresses)
+                .await?;
         let authentications = self.authentications.clone();
```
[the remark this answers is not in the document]
Yithis: I don't really like such low level calls in production code. Would prefer to not use it

finality-aleph/src/new_network/connection_manager/session.rs
Yithis: Idea for better name? @timorl what do you think since you designed it

finality-aleph/src/new_network/connection_manager/session.rs
```diff
@@ -42,50 +49,66 @@ async fn construct_authentication(
...
-    /// Returns the number of nodes we should try being connected to.
     pub fn node_count(&self) -> NodeCount {
-        self.keychain.node_count()
+        self.authority_verifier.node_count()
     }
 
     fn session_id(&self) -> SessionId {
-        self.own_authentication.0.session_id
+        match &self.session_info {
+            SessionInfo::SessionId(session_id) => *session_id,
```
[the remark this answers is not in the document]
Yithis: Done

finality-aleph/src/new_network/connection_manager/session.rs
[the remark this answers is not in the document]
Yithis: Done

finality-aleph/src/new_network/connection_manager/session.rs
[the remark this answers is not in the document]
Yithis: Done with small change (without `mem::take` call)

finality-aleph/src/new_network/connection_manager/session.rs
Yithis: Added case to forbid such change and tests that it is correct

finality-aleph/src/new_network/connection_manager/session.rs
timorl: Uh, `construct_session_info` is a pretty clear candidate, innit?

Yithis: Done

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -97,12 +97,19 @@ impl Discovery {
 
     /// Returns messages that should be sent as part of authority discovery at this moment.
     pub fn discover_authorities(&mut self, handler: &SessionHandler) -> Vec<DiscoveryCommand> {
+        let authentication = handler.authentication();
+
+        if authentication == None {
+            return Vec::new();
```
Yithis, line 103: @timorl Are you fine with changing this in other small PR? Would prefer to merge this one without targeting this case in better way. I would prepare other till Tuesday

fixxxedpoint (approved):

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: Nah, this ain't that hard:
```
let authentication = match handler.authentication() {
 Some(authentication) => authentication,
 None => return Vec::new(),
}

Comment (timorl) [reply finality-aleph/src/new_network/connection_manager/discovery.rs:103]: Same below.

File: finality-aleph/src/new_network/connection_manager/discovery.rs [hunk 14]
```diff
@@ -151,14 +158,22 @@ impl Discovery {
         authentication: Authentication,
         handler: &mut SessionHandler,
     ) -> (Vec<Multiaddr>, Vec<DiscoveryCommand>) {
+        let handler_authentication = handler.authentication();
+
+        if handler_authentication == None {
+            return (Vec::new(), Vec::new());
```
timorl, line 164: You cannot actually do that here, you need to return at least the rebroadcast.

timorl (changes requested):

finality-aleph/src/new_network/connection_manager/discovery.rs
Yithis: Your code aren't doing anything else than mine XD

timorl (approved):

finality-aleph/src/new_network/connection_manager/discovery.rs
timorl: But it's massively sexier.

finality-aleph/src/new_network/connection_manager/discovery.rs
```diff
@@ -290,35 +298,71 @@ mod tests {
...
+                    vec![address.into()],
+                )
+                .await
+                .unwrap(),
+            );
+        }
+        (Discovery::new(Duration::from_millis(MS_COOLDOWN)), handlers)
+    }
+
+    async fn build_non_validators() -> (Discovery, Vec<SessionHandler>) {
```
timorl, line 339: Just return the almost unique nonvalidator as the last element of the above function instead.

timorl (approved):
Sorry ;<

finality-aleph/src/new_network/connection_manager/discovery.rs
Yithis: Done

PR #227, opened 2021-12-06
Refactor argument handling in ChainParams and fix run_nodes.sh

h4nsu:
I was not able to reproduce the problem with network in `run_nodes.sh` you described, but I hope this fixes it :)

### DIFF
--- a/bin/node/src/chain_spec.rs
+++ b/bin/node/src/chain_spec.rs
@@ -118,23 +118,23 @@ pub struct ChainParams {
     #[structopt(long, default_value = "account_secret")]
     pub account_key_file: String,
 
-    // The length of a session
+    /// The length of a session (in seconds). Default is 900 s
     #[structopt(long)]
     pub session_period: Option<u32>,
 
-    /// Time interval (in milliseconds) between blocks
+    /// Time interval (in milliseconds) between blocks. Default is 1000 ms
     #[structopt(long)]
     pub millisecs_per_block: Option<u64>,
 
-    /// Chain name
+    /// Chain name. Default is "Aleph Zero Development"
     #[structopt(long, default_value = "Aleph Zero Development")]
     pub chain_name: String,
 
-    /// Token symbol
+    /// Token symbol. Default is DZERO
     #[structopt(long, default_value = "DZERO")]
     pub token_symbol: String,
 
-    /// AccountIds of authorities forming the committe at the genesis (comma delimited)
+    /// AccountIds of authorities forming the committee at the genesis (comma delimited)
     #[structopt(long, require_delimiter = true, parse(from_str = parse_account_id))]
     account_ids: Vec<AccountId>,
 
@@ -153,7 +153,7 @@ impl ChainParams {
     }
 
     pub fn chain_type(&self) -> ChainType {
-        match self.chain_id.as_str() {
+        match self.chain_type.as_str() {
             CHAINTYPE_DEV => ChainType::Development,
             CHAINTYPE_LOCAL => ChainType::Local,
             CHAINTYPE_LIVE => ChainType::Live,
@@ -264,7 +264,7 @@ fn deduplicate(accounts: Vec<AccountId>) -> Vec<AccountId> {
 fn genesis(
     wasm_binary: &[u8],
     authorities: Vec<AuthorityKeys>,
-    root_key: AccountId,
+    sudo_account: AccountId,
     faucet_account: Option<AccountId>,
     chain_params: ChainParams,
 ) -> GenesisConfig {
@@ -272,8 +272,8 @@ fn genesis(
     let millisecs_per_block = chain_params.millisecs_per_block();
 
     let special_accounts = match faucet_account {
-        Some(faucet_id) => vec![root_key.clone(), faucet_id],
-        None => vec![root_key.clone()],
+        Some(faucet_id) => vec![sudo_account.clone(), faucet_id],
+        None => vec![sudo_account.clone()],
     };
 
     // NOTE: some combinations of bootstrap chain arguments can potentially
@@ -303,7 +303,7 @@ fn genesis(
         },
         sudo: SudoConfig {
             // Assign network admin rights.
-            key: root_key,
+            key: sudo_account,
         },
         aleph: AlephConfig {
             authorities: vec![],
--- a/bin/node/src/chain_spec.rs
+++ b/bin/node/src/chain_spec.rs
@@ -25,6 +25,7 @@ pub const CHAINTYPE_DEV: &str = "dev";
 pub const CHAINTYPE_LOCAL: &str = "local";
 pub const CHAINTYPE_LIVE: &str = "live";
 
+// This account was used in mainnet chainspec
 pub const DEFAULT_SUDO_ACCOUNT: &str = "5F4SvwaUEQubiqkPF8YnRfcN77cLsT2DfG4vFeQmSXNjR7hD";
 
 /// Specialized `ChainSpec`. This is a specialization of the general Substrate ChainSpec type.
@@ -236,7 +237,7 @@ pub fn config(
             genesis(
                 wasm_binary,
                 authorities.clone(), // Initial PoA authorities, will receive funds
-                sudo_account.clone(), // Root account, will also be pre funded
+                sudo_account.clone(), // Sudo account, will also be pre funded
                 faucet_account.clone(), // Pre-funded faucet account
                 chain_params.clone(),
             )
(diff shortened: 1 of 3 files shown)


bin/node/src/chain_spec.rs
```diff
@@ -104,17 +95,15 @@ pub struct AuthorityKeys {
...
-    /// `dev` chain id means that a set of known accounts will be used to form a comittee
-    #[structopt(long, value_name = "CHAIN_SPEC", default_value = "a0dnet1")]
+    /// Chain ID is a short identifier of the chain
+    #[structopt(long, value_name = "ID", default_value = "a0dnet1")]
     pub chain_id: String,
 
-    #[structopt(long, default_value_if("chain-id", Some(DEVNET_ID), "true"), conflicts_with_all(&["account-ids", "sudo-account-id"]))]
-    pub dev_chain: bool,
+    /// The type of the chain. Can be "dev", "local", "live" or custom (any other string)
+    #[structopt(long, value_name = "TYPE", default_value = CHAINTYPE_LIVE)]
```
fbielejec, line 103: smells like an Enum here. Or just always run a live chain, as close to the real thing as possible?

bin/node/src/chain_spec.rs
```diff
@@ -129,37 +118,49 @@ pub struct ChainParams {
...
     pub fn chain_id(&self) -> &str {
         &self.chain_id
     }
 
+    pub fn chain_type(&self) -> ChainType {
+        match self.chain_id.as_str() {
+            CHAINTYPE_DEV => ChainType::Development,
+            CHAINTYPE_LOCAL => ChainType::Local,
+            CHAINTYPE_LIVE => ChainType::Live,
+            s => ChainType::Custom(s.to_string()),
```
fbielejec, line 160: do we know what a custom chain does?

bin/node/src/chain_spec.rs
```diff
@@ -129,37 +118,49 @@ pub struct ChainParams {
     #[structopt(long, default_value = "account_secret")]
     pub account_key_file: String,
 
+    // The length of a session
```
fbielejec, line 121: can we provide a default for these two, seems like session_period() is exactly that

bin/node/src/chain_spec.rs
```diff
@@ -129,37 +118,49 @@ pub struct ChainParams {
...
     #[structopt(long)]
     pub session_period: Option<u32>,
 
+    /// Time interval (in milliseconds) between blocks
     #[structopt(long)]
     pub millisecs_per_block: Option<u64>,
```
fbielejec, line 127: add a default?

bin/node/src/chain_spec.rs
pmikolajczyk41: I would go for an `Enum`. Then you can probably use this macro: https://docs.rs/clap/2.29.0/clap/macro.arg_enum.html to avoid implementing `FromStr` by hand.

And I really do not see the need for a custom variant. Is there a reason for that?

bin/node/src/chain_spec.rs
pmikolajczyk41, line 121: `///`

bin/node/src/chain_spec.rs
```diff
@@ -129,37 +118,49 @@ pub struct ChainParams {
...
 
+    /// Token symbol
     #[structopt(long, default_value = "DZERO")]
     pub token_symbol: String,
 
-    /// Pass the AccountIds of authorities forming the committe at the genesis
-    ///
-    /// Expects a delimited collection of AccountIds
-    #[structopt(long, require_delimiter = true, parse(from_str = parse_account_id), required_unless("dev-chain"))]
+    /// AccountIds of authorities forming the committe at the genesis (comma delimited)
```
pmikolajczyk41, line 137: you can correct the typo `committee` right away

bin/node/src/chain_spec.rs
pmikolajczyk41: +1

bin/node/src/chain_spec.rs
```diff
@@ -278,12 +265,17 @@ fn genesis(
     wasm_binary: &[u8],
     authorities: Vec<AuthorityKeys>,
     root_key: AccountId,
```
pmikolajczyk41, line 267: Can 'root account/key' have different meaning than a 'sudo account/key'? If not, then I would definitely rename this argument, since it introduces some confusion (at least for me :P)

bin/node/src/chain_spec.rs
```diff
@@ -129,37 +118,49 @@ pub struct ChainParams {
...
+    faucet_account_id: Option<AccountId>,
 }
 
 impl ChainParams {
     pub fn chain_id(&self) -> &str {
         &self.chain_id
     }
 
+    pub fn chain_type(&self) -> ChainType {
+        match self.chain_id.as_str() {
```
pmikolajczyk41, line 156: !critical typo! `check_type` - this is the reason why this still doesn't work for me :stuck_out_tongue_closed_eyes:

pmikolajczyk41 (changes requested):
chain_spec.rs - 155-162
```rust
    pub fn chain_type(&self) -> ChainType {
        match self.chain_id.as_str() {
            CHAINTYPE_DEV => ChainType::Development,
            CHAINTYPE_LOCAL => ChainType::Local,
            CHAINTYPE_LIVE => ChainType::Live,
            s => ChainType::Custom(s.to_string()),
        }
    }
```
!critical typo! (`chain_id` -> `chain_type`) - this is the reason why this still doesn't work for me :stuck_out_tongue_closed_eyes:

bin/node/src/chain_spec.rs
h4nsu: done

bin/node/src/chain_spec.rs
h4nsu: done

h4nsu: done

bin/node/src/chain_spec.rs
h4nsu: done

bin/node/src/chain_spec.rs
h4nsu: makes sense, done

bin/node/src/chain_spec.rs
h4nsu: About `enum` - I don't see any reason to define a custom `enum` when `ChainType` is already here. I could change the type of the argument from `String` to `ChainType` and provide a custom parser, but that solution doesn't really differ from what we have now (https://github.com/h4nsu/aleph-node/commit/824966b90779ec64a8e98c855983a17dd64656e2)

About `clap` macro - I'd rather avoid that solution because a) it requires defining another enum b) it mixes `clap` and `structopt` in a way I don't fully understand

pmikolajczyk41: fair point with `clap`;

I am a little bit against the "custom" type, because any typo / error with chain type leads to a custom type (like the one with the typo from another comment).

bin/node/src/chain_spec.rs
```diff
@@ -277,13 +264,18 @@ fn deduplicate(accounts: Vec<AccountId>) -> Vec<AccountId> {
...
     let session_period = chain_params.session_period();
     let millisecs_per_block = chain_params.millisecs_per_block();
 
+    let special_accounts = match faucet_account {
+        Some(faucet_id) => vec![sudo_account.clone(), faucet_id],
+        None => vec![sudo_account.clone()],
+    };
+
     // NOTE: some combinations of bootstrap chain arguments can potentially
     // lead to duplicated rich accounts, e.g. if a root account is also an authority
```
pmikolajczyk41, line 280: nit: for consistency replace "root" with "sudo" here and in line 239 (comment)

pmikolajczyk41 (approved):

bin/node/src/chain_spec.rs
h4nsu: I don't think there's really any functional difference between `Live` and `Custom` (or `Development` and `Local`), this whole chaintype seems to be more of a label for external use (modulo different network flags for `{Develompent,Local}` vs. `{Live,Custom}`)

fbielejec (approved):

bin/node/src/chain_spec.rs
h4nsu: I got rid of the custom variant, now it panics for values other than live, dev or local

mike1729 (approved):

PR #228, opened 2021-12-09
Fix/az 392

fbielejec:
closes [#AZ-392](https://cardinal-cryptography.atlassian.net/jira/software/c/projects/AZ/boards/4?modal=detail&selectedIssue=AZ-392)

This bug is caused by the fact that since we previously stored double option wrapped type (`Option<Option<u32>>`), the default query response when getting the type is `Some(None)`. New runtime tries to decode it as Option<u32> and an error is logged (error is hidden behind runtime=debug logs). This bug is not serious, since the default behavior is to return a None when it fails to decode (see https://github.com/paritytech/substrate/blob/1ec08e88ccfb436da613473c96d01ca6679d5ff8/frame/support/src/storage/unhashed.rs#L27).

We can fix this since this migration is not applied yet, by explicitly matching on the problematic pattern. 
One more pattern that is addressed is the fact that we had stored 0 as session for change at genesis:
```
<SessionForValidatorsChange<T>>::put(Some(0));
```
I doubt any one of our environments runs with a default chainspec, but this is remapped to `None`, just in case.

### DIFF
--- a/pallet/src/migrations/v0_to_v1.rs
+++ b/pallet/src/migrations/v0_to_v1.rs
@@ -4,15 +4,6 @@ use frame_support::{
     traits::{Get, GetStorageVersion, PalletInfoAccess, StorageVersion},
     weights::Weight,
 };
-use sp_std::prelude::*;
-
-frame_support::generate_storage_alias!(
-    Aleph, SessionForValidatorsChange => Value<Option<u32>>
-);
-
-frame_support::generate_storage_alias!(
-    Aleph, Validators<T: Config> => Value<Option<Vec<T::AccountId>>>
-);
 
 pub fn migrate<T: Config, P: GetStorageVersion + PalletInfoAccess>() -> Weight {
     let on_chain_storage_version = <P as GetStorageVersion>::on_chain_storage_version();
@@ -23,13 +14,15 @@ pub fn migrate<T: Config, P: GetStorageVersion + PalletInfoAccess>() -> Weight {
 
         let mut writes = 0;
 
-        match SessionForValidatorsChange::translate(|old: Option<_>| {
-            log::info!(target: "pallet_aleph", "Current storage value for SessionForValidatorsChange {:?}", old);
-            match old {
-                Some(None) | Some(Some(0u32)) | None => None,
-                _ => old,
-            }
-        }) {
+        match crate::SessionForValidatorsChange::<T>::translate(
+            |old: Option<Option<u32>>| -> Option<u32> {
+                log::info!(target: "pallet_aleph", "Current storage value for SessionForValidatorsChange {:?}", old);
+                match old {
+                    Some(Some(x)) => Some(x),
+                    _ => None,
+                }
+            },
+        ) {
             Ok(_) => {
                 writes += 1;
                 log::info!(target: "pallet_aleph", "Succesfully migrated storage for SessionForValidatorsChange");
@@ -39,7 +32,7 @@ pub fn migrate<T: Config, P: GetStorageVersion + PalletInfoAccess>() -> Weight {
             }
         };
 
-        match Validators::<T>::translate(sp_std::convert::identity) {
+        match crate::Validators::<T>::translate(sp_std::convert::identity) {
             Ok(_) => {
                 writes += 1;
                 log::info!(target: "pallet_aleph", "Succesfully migrated storage for Validators");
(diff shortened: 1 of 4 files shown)


pallet/src/migrations/v0_to_v1.rs
```diff
@@ -22,7 +22,14 @@ pub fn migrate<T: Config, P: GetStorageVersion + PalletInfoAccess>() -> Weight {
         log::info!(target: "pallet_aleph", "Running migration from STORAGE_VERSION 0 to 1");
 
         let mut writes = 0;
-        match SessionForValidatorsChange::translate(sp_std::convert::identity) {
+        match SessionForValidatorsChange::translate(|old: Option<Option<u32>>| {
+            log::info!(target: "pallet_aleph", "Current storage value for SessionForValidatorsChange {:?}", old);
+            match old {
+                Some(None) | Some(Some(0u32)) => None,
+                Some(value) => Some(value),
```
timorl, line 29: This returns `Option<Option<..>>`. Don't you want to `flatten` it (modulo the 0 thing, which is kinda annoying)?

fbielejec: it *was* a double wrapped type, now it's a single option over the value. Proof in the pudding (logs from devnet where migration was applied):

```
2-09 14:05:44 Current storage value for SessionForValidatorsChange Some(None)                                                                                                                                                                                            
2021-12-09 14:05:44 Succesfully migrated storage for SessionForValidatorsChange     
```

```
2021-12-09 14:09:16 @new_session 63533                                                                                                                                                                                                                                         
2021-12-09 14:09:16 @new_session Pallet::<T>::session_for_validators_change() None                                                                                                                                                                                             
2021-12-09 14:09:16 ✨ Imported #317660 (0xdbb4…7d02)                                                                                                                                                                                                                          
2021-12-09 14:09:17 💤 Idle (9 peers), best: #317660 (0xdbb4…7d02), finalized #317659 (0x7509…1f6d), ⬇ 5.9kiB/s ⬆ 4.5kiB/s 
```

fbielejec: I think I know what you mean, but it's a bit cumbersome to express using this translate mapping and what is stored atm:
```
        match SessionForValidatorsChange::translate(|old: Option<_>| {
            log::info!(target: "pallet_aleph", "Current storage value for SessionForValidatorsChange {:?}", old);
            match old {
                Some(None) | Some(Some(0u32)) | None => None,
                _ => old,
            }
        })
```

timorl: I really don't get how this works then – does Some(Some(X)) correctly deserialize as Some(X) and you are deliberately using that fact? Because formally you are returning `Option<Option<...>>` from this code and this is what bothers me. <_<

fbielejec: Something like this yes, it's a bit cumbersome, i know :/

We had a ValueQuery over an Option<T> type, with a default for T None, which is how we ended up with sth that decodes to Some(None) and effectively is of type Option<Option<T>>:

```
    #[pallet::type_value]
    pub fn DefaultSessionForValidatorsChange<T: Config>() -> Option<u32> {
        None
    }

    #[pallet::storage]
    #[pallet::getter(fn session_for_validators_change)]
    pub type SessionForValidatorsChange<T: Config> =
        StorageValue<_, Option<u32>, ValueQuery, DefaultSessionForValidatorsChange<T>>;
```

Now translate operates from and to Option<_>, so there is no way to double unpack and return, won't compile AFAIK, I tried. It was designed for cases like i32 -> u32.

But eventually we want to end up storing sth that decodes to an Option<T>, that the storage layer can query with OptionQuery over vanilla T type, so we want to store sth that:

- If there was a value (we are in a middle of change) decodes to Option<T>
- else decodes to None

Given that we control when we apply the migration, and can be sure now validator change was scheduled, just removing the storage item, or overwriting it with None would also work - storage layer evicts None. But to be super correct this is covered as well.

fbielejec: @timorl I completely mixed the semantics of translate and @DamianStraszak pointed it out: it takes the *new* storage type and a function that maps it from the old to new value. With this it's possible, to go from Option<Option> to Option.

DamianStraszak (approved):

pallet/src/migrations/v0_to_v1.rs
DamianStraszak: After the changes in `translate` everything makes sense to me when it comes to types. @timorl you can also take another look now

mike1729 (approved):

PR #229, opened 2021-12-09
AZ-270: Test `JustificationHandler`

pmikolajczyk41:


finality-aleph/src/party.rs
```diff
@@ -92,10 +93,20 @@ impl JustificationRequestDelay for JustificationRequestDelayImpl {
     }
 }
 
+impl<B: Block> Verifier<B> for AuthorityVerifier {
+    fn verify(&self, justification: &AlephJustification, hash: B::Hash) -> bool {
+        if !self.is_complete(&hash.encode()[..], &justification.signature) {
+            debug!(target: "afa", "Bad justification for block hash #{:?} {:?}", hash, justification);
```
DamianStraszak, line 99: This deserves the `warn` level, as something that shouldn't happen without dishonest nodes.

DamianStraszak (changes requested):
Excellent PR. I don't have any reasonable comments really, only one small request for additional test. Currently the JustificationHandler has this (perhaps suboptimal) behavior of working like a FIFO queue when it comes to justification notifications, so if two justifications come in the order of J1, J2, from sessions say `(K+1)` and `K` respectively, with session `K` being in progress, then `J1` should be dropped, but `J2` should be applied. There is already a test for ignoring a single future justification, but this one is a little bit more explicit on how to deal with ordering. This test is expected to fail once we improve JH with some kind of buffering of future justifications.

finality-aleph/src/party.rs
pmikolajczyk41: changed

DamianStraszak (approved):

maciejnems (approved):
Great PR! I have no notes or requests.

PR #230, opened 2021-12-13
AZ-383 towards stable 1k txps

fixxxedpoint:
- allows to skip accounts initialization
- allows to provide first index for the set of `integer` derived accounts, i.e. accounts derived from range x.. instead of the default 0.. range

### DIFF
--- a/flooder/src/main.rs
+++ b/flooder/src/main.rs
@@ -45,6 +45,7 @@ fn main() -> Result<(), anyhow::Error> {
     let connection = pool.get(0).unwrap();
 
     let total_users = config.transactions;
+    let first_account_in_range = config.accounts_range_start;
     let transactions_per_batch = config.throughput / rayon::current_num_threads() as u64;
     let transfer_amount = 1u128;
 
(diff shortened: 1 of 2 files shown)


fbielejec (commented):
LGTM, just one comment, seems like this single argument keeps changing it's name: 
(`accounts_range_start -> first_account_in_range -> first_account`) in different parts of the program flow, for readability might be nice to just stick with one throughout?

fbielejec (approved):

Yithis (approved):
Good PR!

flooder/src/main.rs
```diff
@@ -55,8 +57,10 @@ fn main() -> Result<(), anyhow::Error> {
     let accounts_and_nonces = derive_user_accounts(
         connection.clone(),
         account.clone(),
-        total_users as usize,
+        first_account_in_range,
+        total_users,
         transfer_amount,
+        initialize_accounts,
     );
```
kostekIV, line 64: I think this can be spliited into such a flow:
```rust
    let accounts_and_nonces = derive_user_accounts(...);
    if !skip_initialization {
        initialize_accounts(&accounts_and_nonces, ...);
    }
```

PR #234, opened 2021-12-15
AZ-369: Build excluded packages

pmikolajczyk41:


fbielejec (approved):
good work, just remember to add it to the required workflows please: https://github.com/Cardinal-Cryptography/aleph-node/settings/branch_protection_rules/21409442

fixxxedpoint (commented):
LGTM. Naming is a little bit confusing. After seeing some task like `builld something` I would expect some artifacts to download, but it is actually calling just `clippy` (which calls internally `cargo check`). It also ignores complaints outputted by clippy. Consider changing it to just `cargo check` and rename it to something like `build-check` or use clippy output somehow. What do you think?

pmikolajczyk41:
> LGTM. Naming is a little bit confusing. After seeing some task like `builld something` I would expect some artifacts to download, but it is actually calling just `clippy` (which calls internally `cargo check`). It also ignores complaints outputted by clippy. Consider changing it to just `cargo check` and renaming it to something like `build-check` or use clippy output somehow. What do you think?

Very reasonable comment, thank you. I made renaming "build" -> "check" and I use `cargo check` instead of `cargo clippy` - I think this is sufficient and completely covers our needs here.

PR #235, opened 2021-12-15
AZ-376: Produce aleph-runtime artifact

pmikolajczyk41:


.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -80,6 +78,11 @@ jobs:
           export RUSTC_WRAPPER=$HOME/.local/bin/sccache
           cargo build --release
 
+      - name: Build runtime
+        run: |
+          export RUSTC_WRAPPER=$HOME/.local/bin/sccache
+          cargo build --release -p aleph-runtime
```
timorl, line 84: Huh, this really builds the wasm runtime? Reading the `Cargo.toml` I would not have expected that – am I missing something?

timorl (approved):
Looks reasonable, one question but if I'm being silly just resolve it.

.github/workflows/e2e-tests-main-devnet.yml
pmikolajczyk41: Hmm...
by `-p aleph-runtime` we tell cargo to build only the package `aleph-runtime` and I guess is essentially the runtime, isn't it?

here you have the artifacts, including the output of this command: https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/1583175148, seems reasonable in size

pmikolajczyk41: oh, I get it now; please look at `build.rs`

PR #236, opened 2021-12-15
AZ-107: Session network and some fancy abstractions

timorl:
Really fancy abstractions that will be very useful in the next PR.

### DIFF
--- a/finality-aleph/src/new_network/bare.rs
+++ b/finality-aleph/src/new_network/bare.rs
@@ -1,0 +1,42 @@
+use crate::new_network::{DataNetwork, SendError};
+use aleph_bft::Recipient;
+use codec::Codec;
+use futures::channel::mpsc;
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+/// For sending arbitrary messages.
+pub trait Sender<D: Clone + Codec + Send>: Clone {
+    fn send(&self, data: D, recipient: Recipient) -> Result<(), SendError>;
+}
+
+/// For receiving arbitrary messages.
+#[async_trait::async_trait]
+pub trait Receiver<D: Clone + Codec + Send>: Sync + Send {
+    async fn next(&mut self) -> Option<D>;
+}
+
+/// A bare version of network components.
+pub trait Network<D: Clone + Codec + Send>: Sync {
+    type S: Sender<D>;
+    type R: Receiver<D>;
+    fn sender(&self) -> &Self::S;
+    fn receiver(&self) -> Arc<Mutex<Self::R>>;
+}
+
+#[async_trait::async_trait]
+impl<D: Clone + Codec + Send, BN: Network<D>> DataNetwork<D> for BN {
+    fn send(&self, data: D, recipient: Recipient) -> Result<(), SendError> {
+        self.sender().send(data, recipient)
+    }
+    async fn next(&self) -> Option<D> {
+        self.receiver().clone().lock_owned().await.next().await
+    }
+}
+
+#[async_trait::async_trait]
+impl<D: Clone + Codec + Send> Receiver<D> for mpsc::UnboundedReceiver<D> {
+    async fn next(&mut self) -> Option<D> {
+        self.next().await
+    }
+}
--- a/finality-aleph/src/new_network/mod.rs
+++ b/finality-aleph/src/new_network/mod.rs
@@ -1,15 +1,21 @@
 // Everything here is dead code, but I don't want to create one enormous PR.
 #![allow(dead_code)]
-use codec::{Decode, Encode};
+use aleph_bft::Recipient;
+use codec::{Codec, Decode, Encode};
 use futures::stream::Stream;
 use sc_network::{Event, Multiaddr, NotificationSender, PeerId as ScPeerId};
 use sp_api::NumberFor;
 use sp_runtime::traits::Block;
 use std::{borrow::Cow, collections::HashSet, pin::Pin};
 
+mod bare;
 mod manager;
 mod service;
+mod session;
 mod substrate;
+
+use bare::{Network as BareNetwork, Sender};
+use manager::SessionCommand;
 
 #[derive(PartialEq, Eq, Copy, Clone, Debug, Hash)]
 pub struct PeerId(pub(crate) ScPeerId);
@@ -119,4 +125,16 @@ pub enum DataCommand {
 pub enum ConnectionCommand {
     AddReserved(HashSet<Multiaddr>),
     DelReserved(HashSet<PeerId>),
+}
+
+/// What can go wrong when sending data.
+pub enum SendError {
+    SendFailed,
+}
+
+/// A generic interface for sending and receiving data.
+#[async_trait::async_trait]
+pub trait DataNetwork<D: Clone + Codec> {
+    fn send(&self, data: D, recipient: Recipient) -> Result<(), SendError>;
+    async fn next(&self) -> Option<D>;
 }
--- a/finality-aleph/src/new_network/session.rs
+++ b/finality-aleph/src/new_network/session.rs
@@ -1,0 +1,115 @@
+use crate::{
+    crypto::{AuthorityPen, AuthorityVerifier},
+    new_network::{BareNetwork, SendError, Sender as SenderTrait, SessionCommand},
+    NodeIndex, SessionId,
+};
+use aleph_bft::Recipient;
+use codec::Codec;
+use futures::channel::mpsc;
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+/// Sends data within a single session.
+#[derive(Clone)]
+pub struct Sender<D: Clone + Codec + Send> {
+    session_id: SessionId,
+    messages_for_network: mpsc::UnboundedSender<(D, SessionId, Recipient)>,
+}
+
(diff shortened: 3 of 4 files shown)


finality-aleph/src/new_network/bare.rs
```diff
@@ -0,0 +1,42 @@
+use crate::new_network::{DataNetwork, SendError};
+use aleph_bft::Recipient;
+use codec::Codec;
+use futures::channel::mpsc;
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+/// For sending arbitrary messages.
+pub trait Sender<D: Clone + Codec + Send>: Clone {
```
Yithis, line 9: Why pub? I assume there is reason why not private, but don't get why it can't be `pub(crate)`?

finality-aleph/src/new_network/bare.rs
```diff
@@ -0,0 +1,42 @@
+use crate::new_network::{DataNetwork, SendError};
```
Yithis, line 1: Please, think of a better name for this file. I don't know what to expect in the file named like this

finality-aleph/src/new_network/mod.rs
```diff
@@ -120,3 +126,15 @@ pub enum ConnectionCommand {
     AddReserved(HashSet<Multiaddr>),
     DelReserved(HashSet<PeerId>),
 }
+
+/// What can go wrong when sending data.
```
Yithis, line 130: Better comment? But you don't have to, since I really like this one :joy:

finality-aleph/src/new_network/session.rs
```diff
@@ -0,0 +1,115 @@
...
+
+/// Manages sessions for which the network should be active.
+pub struct Manager<D: Clone + Codec + Send> {
+    commands_for_service: mpsc::UnboundedSender<SessionCommand<D>>,
+    messages_for_service: mpsc::UnboundedSender<(D, SessionId, Recipient)>,
+}
+
+/// What went wrond during a session management operation.
+pub enum ManagerError {
+    SendCommand,
```
Yithis, line 52: If I would see this error message I would be at least confused

finality-aleph/src/new_network/bare.rs
Yithis: Also there are plenty of times you are using `pub`, so please think about this comment as to all of them

Yithis (changes requested):
Generally great PR, but please don't use `pub` each time, especially if it is not necessary or can be replaced with `pub(crate)`. Also some naming can be improved

finality-aleph/src/new_network/bare.rs
timorl: I keep all of this in non-`pub` modules, so it shouldn't be accessible from outside the crate anyway. And this better illustrates which things are parts of the interface of the module, and avoids the ugly `(crate)`.

I think `pub(crate)` should only be used in a module that you want to be public when a specific item shouldn't, but in general this seems like the design failed at some point already.

finality-aleph/src/new_network/mod.rs
timorl: I have no idea, I thought this was just a normal comment. <_<"

finality-aleph/src/new_network/bare.rs
timorl: Changed to `component` (as in `ComponentNetwork`, since this is just components for a proper `DataNetwork`), not sure whether this is better.

finality-aleph/src/new_network/session.rs
timorl: Changed to `CommandSendFailed`.

finality-aleph/src/new_network/mod.rs
timorl: Rephrased to possibly make it more "programmer" sounding, although I think this version was more natural. :crying_cat_face:

Yithis (approved):

finality-aleph/src/new_network/component.rs
```diff
@@ -0,0 +1,42 @@
...
+use tokio::sync::Mutex;
+
+/// For sending arbitrary messages.
+pub trait Sender<D: Clone + Codec + Send>: Clone {
+    fn send(&self, data: D, recipient: Recipient) -> Result<(), SendError>;
+}
+
+/// For receiving arbitrary messages.
+#[async_trait::async_trait]
+pub trait Receiver<D: Clone + Codec + Send>: Sync + Send {
```
kostekIV, line 15: would it be possible to auto implement `Stream<Item=D>` for the receiver?

timorl: Probably, but I would prefer not doing that as it would require directly using `Poll`s.

finality-aleph/src/new_network/component.rs
```diff
@@ -0,0 +1,42 @@
+use crate::new_network::{DataNetwork, SendError};
+use aleph_bft::Recipient;
+use codec::Codec;
+use futures::channel::mpsc;
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+/// For sending arbitrary messages.
+pub trait Sender<D: Clone + Codec + Send>: Clone {
```
KazyCC, line 9: Is similarity to std::sync::mpsc::Sender intented?

finality-aleph/src/new_network/session.rs
```diff
@@ -0,0 +1,115 @@
+use crate::{
+    crypto::{AuthorityPen, AuthorityVerifier},
+    new_network::{ComponentNetwork, SendError, Sender as SenderTrait, SessionCommand},
```
KazyCC, line 3: Isn't naming convention a little misleading? "Sender" means 3 different things depending on the context in this PR....

finality-aleph/src/new_network/mod.rs
```diff
@@ -1,16 +1,22 @@
 // Everything here is dead code, but I don't want to create one enormous PR.
 #![allow(dead_code)]
-use codec::{Decode, Encode};
+use aleph_bft::Recipient;
+use codec::{Codec, Decode, Encode};
 use futures::stream::Stream;
 use sc_network::{Event, Multiaddr, NotificationSender, PeerId as ScPeerId};
```
KazyCC, line 6: Do we really want this new module to be dependent on substrate?

finality-aleph/src/new_network/session.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+}
+
+/// Sends and receives data within a single session.
+pub struct Network<D: Clone + Codec + Send> {
+    sender: Sender<D>,
+    receiver: Arc<Mutex<mpsc::UnboundedReceiver<D>>>,
+}
+
+impl<D: Clone + Codec + Send> ComponentNetwork<D> for Network<D> {
```
KazyCC, line 33: What value brings having 3 different Networks: struct Network, trait ComponentNetwork and trait DataNetwork? Are those additional layers really worth an additional complexity? Who do we expect to implement those traits on different types than we are using right now? Kind of YAGNI for me.

KazyCC (approved):
A lot of code in this PR + a lot of digging for code in new_network/ as well. Generally looks very good, but I'm wondering if it should be more KISS.

finality-aleph/src/new_network/component.rs
timorl: Internally it will usually use `mpsc::UnboundedSender` and will be used by `aleph_bft::Network`, both influenced the design.

finality-aleph/src/new_network/mod.rs
timorl: Maciek is working on making this dependence slightly weaker (or rather this will be a side result of what he is working on).

Unfortunately at least `PeerId`s and `Multiaddr`s are necessary, since a big part of the logic here deals with keeping connections alive. These could technically be type parameters for the whole module, but the template abundance that would result from that could rival C++, not in a good way.

Eh, now that you mentioned it I _could_ be tempted to make this more generic...

finality-aleph/src/new_network/component.rs
KazyCC: Ok, so I'd recommend changing the name, because even in new_network/ we have source codes using both at once.

finality-aleph/src/new_network/session.rs
KazyCC: Definitely, since reading this code requires one to check what is named what in which file, which is not good :)

finality-aleph/src/new_network/session.rs
timorl: mickey_mouse_a_surprize_that_will_help_us_later.apng

To actually answer the question – the next PR will introduce a way of splitting networks generically, which is needed to create two interfaces, one for `aleph_bft::Network` and another for the justification component. This way has to use the components, thus the need for a `ComponentNetwork`, same for `Sender` and `Receiver`.

finality-aleph/src/new_network/component.rs
timorl: This is now reexported in `mod.rs` as `SenderComponent`.

finality-aleph/src/new_network/session.rs
timorl: As above.

PR #237, opened 2021-12-16
AZ-107: Split networks

timorl:
splits

### DIFF
--- a/finality-aleph/src/new_network/split.rs
+++ b/finality-aleph/src/new_network/split.rs
@@ -13,57 +13,70 @@ const MAXIMUM_RETRY_MS: u64 = 500;
 
 /// Used for routing data through split networks.
 #[derive(Clone, Encode, Decode)]
-pub enum Split<D1: Data, D2: Data> {
-    Left(D1),
-    Right(D2),
+pub enum Split<LeftData: Data, RightData: Data> {
+    Left(LeftData),
+    Right(RightData),
 }
 
 #[derive(Clone)]
-struct LeftSender<D1: Data, D2: Data, S: SenderComponent<Split<D1, D2>>> {
+struct LeftSender<LeftData: Data, RightData: Data, S: SenderComponent<Split<LeftData, RightData>>> {
     sender: S,
-    phantom: PhantomData<(D1, D2)>,
+    phantom: PhantomData<(LeftData, RightData)>,
 }
 
-impl<D1: Data, D2: Data, S: SenderComponent<Split<D1, D2>>> SenderComponent<D1>
-    for LeftSender<D1, D2, S>
+impl<LeftData: Data, RightData: Data, S: SenderComponent<Split<LeftData, RightData>>>
+    SenderComponent<LeftData> for LeftSender<LeftData, RightData, S>
 {
-    fn send(&self, data: D1, recipient: Recipient) -> Result<(), SendError> {
+    fn send(&self, data: LeftData, recipient: Recipient) -> Result<(), SendError> {
         self.sender.send(Split::Left(data), recipient)
     }
 }
 
 #[derive(Clone)]
-struct RightSender<D1: Data, D2: Data, S: SenderComponent<Split<D1, D2>>> {
+struct RightSender<LeftData: Data, RightData: Data, S: SenderComponent<Split<LeftData, RightData>>>
+{
     sender: S,
-    phantom: PhantomData<(D1, D2)>,
+    phantom: PhantomData<(LeftData, RightData)>,
 }
 
-impl<D1: Data, D2: Data, S: SenderComponent<Split<D1, D2>>> SenderComponent<D2>
-    for RightSender<D1, D2, S>
+impl<LeftData: Data, RightData: Data, S: SenderComponent<Split<LeftData, RightData>>>
+    SenderComponent<RightData> for RightSender<LeftData, RightData, S>
 {
-    fn send(&self, data: D2, recipient: Recipient) -> Result<(), SendError> {
+    fn send(&self, data: RightData, recipient: Recipient) -> Result<(), SendError> {
         self.sender.send(Split::Right(data), recipient)
     }
 }
 
-struct LeftReceiver<D1: Data, D2: Data, R: ReceiverComponent<Split<D1, D2>>> {
+struct LeftReceiver<
+    LeftData: Data,
+    RightData: Data,
+    R: ReceiverComponent<Split<LeftData, RightData>>,
+> {
     receiver: Arc<Mutex<R>>,
-    translated_receiver: mpsc::UnboundedReceiver<D1>,
-    left_sender: mpsc::UnboundedSender<D1>,
-    right_sender: mpsc::UnboundedSender<D2>,
+    translated_receiver: mpsc::UnboundedReceiver<LeftData>,
+    left_sender: mpsc::UnboundedSender<LeftData>,
+    right_sender: mpsc::UnboundedSender<RightData>,
 }
 
-struct RightReceiver<D1: Data, D2: Data, R: ReceiverComponent<Split<D1, D2>>> {
+struct RightReceiver<
+    LeftData: Data,
+    RightData: Data,
+    R: ReceiverComponent<Split<LeftData, RightData>>,
+> {
     receiver: Arc<Mutex<R>>,
-    translated_receiver: mpsc::UnboundedReceiver<D2>,
-    left_sender: mpsc::UnboundedSender<D1>,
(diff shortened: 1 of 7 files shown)


kostekIV (changes requested):
I don't understand what's going on here. Since we are having multiple almost the same structs/traits etc that have minimal differences (Like D1 instead of D2) it's hard for me to see connections, flow and usage of this. I would much appreciate some comments on what's going out here that would help with my understanding and with the review of the code (as of right now I could only find some nits).

finality-aleph/src/new_network/split.rs
```diff
@@ -0,0 +1,230 @@
...
+pub enum Split<D1: Data, D2: Data> {
+    Hidari(D1),
+    Migi(D2),
+}
+
+#[derive(Clone)]
+struct HidariSender<D1: Data, D2: Data, S: SenderComponent<Split<D1, D2>>> {
+    sender: S,
+    d1: PhantomData<D1>,
+    d2: PhantomData<D2>,
```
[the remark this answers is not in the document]
timorl: Welp, I was being silly. :D

finality-aleph/src/new_network/split.rs
```diff
@@ -0,0 +1,230 @@
...
+
+#[async_trait::async_trait]
+impl<D1: Data, D2: Data, R: ReceiverComponent<Split<D1, D2>>> ReceiverComponent<D1>
+    for HidariReceiver<D1, D2, R>
+{
+    async fn next(&mut self) -> Option<D1> {
+        loop {
+            tokio::select! {
+                data = self.translated_receiver.next() => return data,
+                _ = forward_or_wait(&self.receiver, &self.hidari_sender, &self.migi_sender) => (),
```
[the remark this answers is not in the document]
timorl: It will always eventually go to the first case – either `forward_or_wait` will forward something there, or it will get closed, closing the channel and making `translated_receiver` return `None`.

finality-aleph/src/new_network/split.rs
```diff
@@ -0,0 +1,230 @@
...
+    fn send(&self, data: D1, recipient: Recipient) -> Result<(), SendError> {
+        self.sender.send(Split::Hidari(data), recipient)
+    }
+}
+
+#[derive(Clone)]
+struct MigiSender<D1: Data, D2: Data, S: SenderComponent<Split<D1, D2>>> {
+    sender: S,
+    d1: PhantomData<D1>,
+    d2: PhantomData<D2>,
```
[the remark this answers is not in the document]
timorl: Dane.

finality-aleph/src/new_network/split.rs
```diff
@@ -0,0 +1,230 @@
...
+        },
+        None => {
+            interval(Duration::from_millis(MAXIMUM_RETRY_MS))
+                .tick()
+                .await;
+        }
+    }
+}
+
+const MAXIMUM_RETRY_MS: u64 = 500;
```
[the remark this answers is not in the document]
timorl: Moved to top.

finality-aleph/src/new_network/split.rs
```diff
@@ -0,0 +1,230 @@
...
+use tokio::{
+    sync::Mutex,
+    time::{interval, Duration},
+};
+
+/// Used for routing data through split networks.
+#[derive(Clone, Encode, Decode)]
+pub enum Split<D1: Data, D2: Data> {
+    Hidari(D1),
+    Migi(D2),
```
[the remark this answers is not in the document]
timorl: I have clearly been having too much fun with this. Changed.

finality-aleph/src/new_network/split.rs
```diff
@@ -0,0 +1,231 @@
...
+
+/// Used for routing data through split networks.
+#[derive(Clone, Encode, Decode)]
+pub enum Split<D1: Data, D2: Data> {
+    Left(D1),
+    Right(D2),
+}
+
+#[derive(Clone)]
+struct LeftSender<D1: Data, D2: Data, S: SenderComponent<Split<D1, D2>>> {
```
Yithis, line 22: What will be a name when we will need a third sender? :thinking:

Yithis: If I get it right for example `LeftSender<D1,D2,S> = RightSender<D2, D1, S>`, so I don't see why we should have 2 (4) different structures. I see that they have to send different `enum Split`, but maybe you can somehow work around it

Yithis (changes requested):
I really like how much simpler it is from old version, but whole `split.rs` file gives me a feeling that it is every line x2

finality-aleph/src/new_network/split.rs
timorl: Oh, I tried, believe me.. I don't think it's possible to remove much of the duplication in Rust unfortunately. :/

As for your original question – we don't need to, because trees! If we need to split into 3 we split once, and split one of the outputs again. :3 Abstraction!

finality-aleph/src/new_network/split.rs
```diff
@@ -0,0 +1,231 @@
...
+use tokio::{
+    sync::Mutex,
+    time::{interval, Duration},
+};
+
+const MAXIMUM_RETRY_MS: u64 = 500;
+
+/// Used for routing data through split networks.
+#[derive(Clone, Encode, Decode)]
+pub enum Split<D1: Data, D2: Data> {
```
kostekIV, line 16: Consider renaming `D1, D2` to smth like `LData, RData`

finality-aleph/src/new_network/split.rs
kostekIV: it probably could be deduplicated with macro usage but I don't think it is worth it (or more readable) to do it.

finality-aleph/src/new_network/split.rs
```diff
@@ -0,0 +1,231 @@
...
+                    warn!(target: "aleph-network", "Failed send despite controlling receiver, this shouldn't've happened.");
+                }
+            }
+            None => {
+                left_sender.close_channel();
+                right_sender.close_channel();
+            }
+        },
+        None => {
+            interval(Duration::from_millis(MAXIMUM_RETRY_MS))
```
kostekIV, line 86: I may be mistaken but with the usage of the async mutex we wouldn't need this arm at all

finality-aleph/src/new_network/split.rs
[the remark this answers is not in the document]
kostekIV: ok, makes sense now

finality-aleph/src/new_network/split.rs
```diff
@@ -0,0 +1,231 @@
...
+            right_sender,
+        },
+    )
+}
+
+/// Split a single component network into two separate ones. This way multiple components can send
+/// data to the same underlying session not knowing what types of data the other ones use.
+///
+/// The main example for now is creating an `aleph_bft::Network` and a separate one for accumulating
+/// signatures for justifications.
```
kostekIV, line 215: I would add a section describing how communication works inside.

Something like this (but worded a bit better): 
* Left and Right networks compete for the data produced by their parent network.
* Left can only handle LData type (Similarly Right can handle only RData)
* They can't know in advance what kind of data the sender from the parent network will send so they have to consume it anyway
* But they are actually nice and whenever they get data for the other network they are kind enough to forward it to their counterpart

Yithis (approved):

finality-aleph/src/new_network/split.rs
timorl: Renamed to `LeftData` and `RightData`.

finality-aleph/src/new_network/split.rs
timorl: We do, `try_lock()` is not `async`, so if the lock failed this would just fall through every time and do essentially a busy wait at best.

finality-aleph/src/new_network/split.rs
timorl: Added a description based on this.

finality-aleph/src/new_network/split.rs
kostekIV: I meant the usage of the https://docs.rs/async-mutex/latest/async_mutex/ (with `mutex.lock().await` instead of the try_lock)

finality-aleph/src/new_network/split.rs
kostekIV: Thanks

finality-aleph/src/new_network/split.rs
timorl: You are right! <3

kostekIV (approved):

PR #241, opened 2021-12-17
AZ-107: Splicing networks for brute force compatibility

timorl:


Yithis (approved):
lgtm since it is a temporary file

PR #242, opened 2021-12-20
AZ-107: Change peers_set_config signature

Yithis:


finality-aleph/src/lib.rs
```diff
@@ -37,16 +37,32 @@ enum Error {
...
         // Max size of alert is UNIT_SIZE * MAX_UNITS_IN_ALERT ~ 100 * 5000 = 50000 bytes
         // Max size of parents response UNIT_SIZE * N_MEMBERS ~ 100 * N_MEMBERS
         // When adding other (large) message types we need to make sure this limit is fine.
         1024 * 1024,
     );
-    config.set_config = sc_network::config::SetConfig::default();
+
+    config.set_config = match protocol {
+        Some(new_network::Protocol::Validator) => {
+            sc_network::config::SetConfig {
```
Yithis, line 57: Those values probably will be changed letter since currently there is no way to test it

finality-aleph/src/lib.rs
```diff
@@ -37,16 +37,32 @@ enum Error {
...
+                in_peers: 25,
+                out_peers: 0,
```
kostekIV, line 59: Why out peers are set to 0? it looks like it is the only difference between this config and `SetConfig::default()`

Yithis: https://www.notion.so/cardinalcryptography/Remaining-new-network-work-ba2009b5fceb439fb36e68a92e3ff67d#25d0ba3c46644d66936e5ea33e96ff63

Yithis: Maybe I will change something else letter, currently there is no way to test it, so I wanted only to set 0 outgoing slots

DamianStraszak: There is no way to test what exactly?

kostekIV: cool, this is good enough for me. I am leaving this discussion open for Damian

kostekIV (approved):

finality-aleph/src/lib.rs
Yithis: How does node is working with other protocol/config. There need to be done some additional things to enable `new_network`

DamianStraszak (approved):

PR #243, opened 2021-12-20
AZ-377: Tools for local testing of runtime updates.

h4nsu:


local-tests/chainrunner/utils.py
```diff
@@ -0,0 +1,38 @@
+import os.path as op
+import re
+import subprocess
+from collections import OrderedDict
+
+
+def Generate_keys(binary, phrases):
```
kostekIV, line 7: checkout the [bip-utils](https://pypi.org/project/bip-utils/) package, which might be useful here .

Sample usage:
```py

from bip_utils import SubstrateBip39SeedGenerator, SubstrateCoins, Substrate


phrase = "bottom drive obey lake curtain smoke basket hold race lonely fit walk"
seed_bytes = SubstrateBip39SeedGenerator(phrase).Generate()
substrate_ctx = Substrate.FromSeedAndPath(seed, '//0', SubstrateCoins.GENERIC)
print(substrate_ctx.PublicKey().ToAddress())
```

local-tests/chainrunner/node.py
```diff
@@ -0,0 +1,92 @@
+import json
+import os.path as op
+import re
+import requests
+import jsonrpcclient as rpc
+import subprocess
+
+from .utils import flags_from_dict
+
+class node:
```
pmikolajczyk41, line 10: I think that a standard way of naming classes in Python is CamelCase (and so is my strong suggestion :P)

local-tests/chainrunner/utils.py
```diff
@@ -0,0 +1,38 @@
...
+    """Turn 'flag_name' into `--flag-name`."""
+    return '--' + str(s).replace('_', '-')
+
+
+def flags_from_dict(d):
+    """Turn a dictionary of flags into a list of strings required by subprocess methods."""
+    res = []
+    for k,v in d.items():
+        res.append(flag(k))
+        if v is not True:
```
pmikolajczyk41, line 36: why not `if not v` ?

local-tests/chainrunner/utils.py
pmikolajczyk41, line 7: lowercase

local-tests/chainrunner/node.py
```diff
@@ -0,0 +1,92 @@
...
+    """A class representing a single node of a running blockchain."""
+    def __init__(self, binary, chainspec, path, logdir=None):
```
pmikolajczyk41, line 12: It is highly not obvious what `path` refers to

local-tests/chainrunner/node.py
```diff
@@ -0,0 +1,92 @@
...
+        return []
+
+
+    def highest_block(self):
+        """Find in the logs the height of the most recent block.
+        Returns two ints: highest block and highest finalized block."""
+        results = self.greplog('best: #(\d+) .+ finalized #(\d+)')
+        if results:
+            a,b = results[-1]
+            return (int(a), int(b))
```
pmikolajczyk41, line 62: a nit, but in Python you do not have to explicitly write parentheses when returning a tuple

local-tests/chainrunner/node.py
```diff
@@ -0,0 +1,92 @@
...
+        self.logfile = None
+        self.flags = {}
+        self.running = False
+
+
+    def _stdargs(self):
+        return ['--base-path', self.path, '--chain', self.chainspec]
+
+
+    def start(self, name):
```
pmikolajczyk41, line 26: a general, also controversial recommendation - I know that a big part of Python's power comes from no need of typing, but still I am a fan of type hints in public API - it makes the interface much cleaner and transparent, also some IDEs can use them to perform better analysis, and you don't have to specify type in function name (like `flags_from_dict` could be then renamed to `flags_from(d: Dict)`)

local-tests/chainrunner/node.py
```diff
@@ -0,0 +1,92 @@
...
+        """Start the node. `name` is used to name of the logfile and for --name flag."""
+        cmd = [self.binary, '--name', name] + self._stdargs() + flags_from_dict(self.flags)
+        self.logfile = op.join(self.logdir, name+'.log')
```
pmikolajczyk41, line 29: feel free to ignore this, probably we should have some linter set up, but I would wrap operators with spaces, e.g. `name + '.log'`

local-tests/chainrunner/node.py
```diff
@@ -0,0 +1,92 @@
...
+        self.process = subprocess.Popen(cmd, stderr=open(self.logfile, 'w'), stdout=subprocess.DEVNULL)
```
pmikolajczyk41, line 30: I would define `process` (as `None`) within `__init__`; adding fields ad hoc is not extremely elegant

local-tests/chainrunner/node.py
```diff
@@ -0,0 +1,92 @@
...
+        self.running = True
+
+
+    def stop(self):
+        """Stop the node by sending SIGKILL."""
+        self.process.kill()
+        self.running = False
```
pmikolajczyk41, line 37: I would ensure that you are actually running and thus `process` is not `None`

local-tests/chainrunner/node.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def purge(self):
+        """Purge chain (delete the database of the node)."""
+        cmd = [self.binary, 'purge-chain', '-y'] + self._stdargs()
+        subprocess.run(cmd, stdout=subprocess.DEVNULL).check_returncode()
+
+
+    def greplog(self, regexp):
+        """Find in the logs all occurences of the given regexp. Returns a list of matches."""
+        if self.logfile:
```
pmikolajczyk41, line 48: remove nesting by:
`if not self.logfile: return []`

local-tests/chainrunner/node.py
```diff
@@ -0,0 +1,92 @@
...
+            p = re.compile(regexp)
+            with open(self.logfile) as f:
```
pmikolajczyk41, line 50: it is a good idea to specify mode, i.e. `'r'`

local-tests/chainrunner/node.py
```diff
@@ -0,0 +1,92 @@
...
+        results = self.greplog('best: #(\d+) .+ finalized #(\d+)')
+        if results:
+            a,b = results[-1]
+            return (int(a), int(b))
+        return (-1,-1)
+
+
+    def state(self, block=None):
+        """Return a json repesentation of the chain state after the given block.
+        If `block` is `None`, the most recent state is returned.
```
pmikolajczyk41, line 68: I would appreciate an explicit info whether it is highest produced block or highest finalized one

local-tests/chainrunner/utils.py
h4nsu: Same thing, confused with Go. Fixing!

local-tests/chainrunner/node.py
h4nsu: Yeah, you right, I thought the first letter indicates if it's supposed to be "public" or not. Confused it with Go ;)

local-tests/chainrunner/utils.py
h4nsu: General Python habit of "explicit is better than implicit" ;) Here it doesn't really matter, but I burned myself multiple times in the past with `if v` or `if not v` because of type coercion, so I tend to default to `is`/`is not`.

local-tests/chainrunner/node.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+    def purge(self):
+        """Purge chain (delete the database of the node)."""
+        cmd = [self.binary, 'purge-chain', '-y'] + self._stdargs()
+        subprocess.run(cmd, stdout=subprocess.DEVNULL).check_returncode()
+
+
+    def greplog(self, regexp):
+        """Find in the logs all occurences of the given regexp. Returns a list of matches."""
```
pmikolajczyk41, line 47: typo: occurrences

local-tests/chainrunner/node.py
```diff
@@ -0,0 +1,92 @@
...
+        Returns two ints: highest block and highest finalized block."""
+        results = self.greplog('best: #(\d+) .+ finalized #(\d+)')
+        if results:
+            a,b = results[-1]
+            return (int(a), int(b))
+        return (-1,-1)
+
+
+    def state(self, block=None):
+        """Return a json repesentation of the chain state after the given block.
```
pmikolajczyk41, line 67: typo: representation

local-tests/chainrunner/node.py
```diff
@@ -0,0 +1,92 @@
...
+            with open(self.logfile) as f:
+                log = f.read()
+            return p.findall(log)
+        return []
+
+
+    def highest_block(self):
+        """Find in the logs the height of the most recent block.
+        Returns two ints: highest block and highest finalized block."""
+        results = self.greplog('best: #(\d+) .+ finalized #(\d+)')
```
pmikolajczyk41, line 59: I suggest using raw string when specifying regex, i.e. `r'best....'`

local-tests/chainrunner/node.py
```diff
@@ -0,0 +1,92 @@
...
+        self.running = False
+
+
+    def purge(self):
+        """Purge chain (delete the database of the node)."""
+        cmd = [self.binary, 'purge-chain', '-y'] + self._stdargs()
+        subprocess.run(cmd, stdout=subprocess.DEVNULL).check_returncode()
+
+
+    def greplog(self, regexp):
```
pmikolajczyk41, line 46: matter of taste, but I would pass here already a regex object (thus good to have a type hint :P) to avoid compiling arbitrary string within this method - it's responsibility is to search through logs, so it is a little bit confusing if it fails because the passed regex couldn't been compiled.

local-tests/chainrunner/chain.py
```diff
@@ -0,0 +1,108 @@
...
+        for i in idx:
+            self.nodes[i].start(name+str(i))
+
+
+    def stop(self, nodes=None):
+        """Stop the chain. Optional `nodes` argument can be used to specify which nodes are affected
+        and should be a list of integer indices (0..N-1). Affects all nodes if ommitted."""
+        idx = nodes or range(len(self.nodes))
+        for i in idx:
+            self.nodes[i].stop()
```
pmikolajczyk41, line 100: optional, but may be fun: here you have 6 methods that are almost identical, so I would suggest something like this:
```python
def _apply(self, action, nodes):
  for node_id in (nodes or range(len(self.nodes))):
    action(self.nodes[node_id])

def stop(self, nodes=None):
  self._apply(lambda node: node.stop(), nodes)
```

local-tests/run_nodes.py
```diff
@@ -0,0 +1,36 @@
...
+binary = '../target/release/aleph-node'
+
+phrases = ['//Alice','//Bob','//Cedric','//Dick', '//Ezekiel', '//Fanny', '//George', '//Hugo']
+keys_dict = Generate_keys(binary, phrases)
+keys = list(keys_dict.values())
+
+chain = Chain(workdir)
+
+print(f'Bootstraping chain for {nodes} nodes')
+chain.bootstrap(binary, keys[:nodes], chain_type='local', millisecs_per_block=2000, session_period=40)
```
pmikolajczyk41, line 22: just to avoid confusion, check whether `nodes <= len(keys)`

local-tests/run_nodes.py
```diff
@@ -0,0 +1,36 @@
...
+workdir = '.'
+binary = '../target/release/aleph-node'
+
+phrases = ['//Alice','//Bob','//Cedric','//Dick', '//Ezekiel', '//Fanny', '//George', '//Hugo']
+keys_dict = Generate_keys(binary, phrases)
+keys = list(keys_dict.values())
+
+chain = Chain(workdir)
+
+print(f'Bootstraping chain for {nodes} nodes')
```
pmikolajczyk41, line 21: typo: Bootstrapping

local-tests/chainrunner/node.py
h4nsu: Added some docstrings

local-tests/chainrunner/node.py
h4nsu: Done

local-tests/send-runtime/src/main.rs
```diff
@@ -0,0 +1,59 @@
...
+        .expect(&format!("Could not establish connection with {}", url))
+}
+
+fn read_file(filename: &String) -> Vec<u8> {
+    let mut f = fs::File::open(&filename).expect("no file found");
+    let metadata = fs::metadata(&filename).expect("unable to read metadata");
+    let mut buffer = vec![0; metadata.len() as usize];
+    f.read(&mut buffer).expect("buffer overflow");
+
+    buffer
```
pmikolajczyk41, line 54: `std::fs::read`?

local-tests/chainrunner/node.py
h4nsu: Done

local-tests/chainrunner/node.py
h4nsu: Done

local-tests/chainrunner/node.py
h4nsu: Done

local-tests/chainrunner/node.py
h4nsu: It defaults to `'r'`:
`open(file, mode='r', buffering=- 1, encoding=None, errors=None, newline=None, closefd=True, opener=None)`

local-tests/chainrunner/node.py
h4nsu: I would appreciate it too XD I have no idea what's the substrate semantics wrt that

local-tests/chainrunner/node.py
h4nsu: Done

local-tests/chainrunner/node.py
h4nsu: Done

local-tests/chainrunner/node.py
h4nsu: Done

local-tests/chainrunner/chain.py
```diff
@@ -0,0 +1,108 @@
+import os
+import os.path as op
+import subprocess
+
+from .node import node
+from .utils import flags_from_dict
+
+
+# Type alias for int
+class Seq(int): pass
```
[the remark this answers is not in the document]
h4nsu: Because then `isinstance(3,Seq) == True` and there would be no way to distinguish constant int parameters vs the ones that should be different for each node.

local-tests/chainrunner/chain.py
```diff
@@ -0,0 +1,108 @@
...
+            for n in self.nodes:
+                n.flags[k] = True
+        for k,v in kwargs.items():
+            for i,n in enumerate(self.nodes):
+                n.flags[k] = v+i if isinstance(v, Seq) else v
+
+
+    def set_binary(self, binary, nodes=None):
+        """Replace nodes' binary with `binary`. Optional `nodes` argument can be used to specify
+        which nodes are affected and should be a list of integer indices (0..N-1). Affects all nodes if ommitted."""
```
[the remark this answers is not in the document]
h4nsu: Done

local-tests/chainrunner/chain.py
```diff
@@ -0,0 +1,108 @@
...
+class Seq(int): pass
+
+
+class Chain:
+    """Chain is a class for orchestrating setting up and interaction with a local aleph-node blockchain.
+    The constructor takes only one argument - a path to a directory with the workspace.
+    All other parameters are adjusted with `bootstrap()` and `set_flags()`."""
+    def __init__(self, workdir):
+        if not op.exists(workdir):
+            os.makedirs(workdir)
```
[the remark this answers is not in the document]
h4nsu: Done

local-tests/chainrunner/chain.py
```diff
@@ -0,0 +1,108 @@
...
+        if not op.isfile(binary):
+            raise FileNotFoundError(f'file not found: {binary}')
+        cmd = [binary, 'bootstrap-chain', '--base-path', self.path, '--account-ids', ','.join(accounts), '--raw']
+        cmd += flags_from_dict(kwargs)
+
+        chainspec = op.join(self.path, 'chainspec.json')
+        with open(chainspec, 'w') as f:
+            subprocess.run(cmd, stdout=f).check_returncode()
+
+        self.nodes = []
```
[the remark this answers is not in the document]
h4nsu: Done

local-tests/chainrunner/chain.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+    def set_binary(self, binary, nodes=None):
+        """Replace nodes' binary with `binary`. Optional `nodes` argument can be used to specify
+        which nodes are affected and should be a list of integer indices (0..N-1). Affects all nodes if ommitted."""
+        if not op.isfile(binary):
+            raise FileNotFoundError(f'file not found: {binary}')
+        idx = nodes or range(len(self.nodes))
+        for i in idx:
+            self.nodes[i].binary = binary
```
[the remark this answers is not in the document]
h4nsu: Nope, `__get_item__` is meant as a convenience shortcut for the "outside" user. I don't want to rely on it internally in case the semantics of it changes in the future.

local-tests/run_nodes.py
h4nsu: Done

local-tests/run_nodes.py
h4nsu: Done

local-tests/chainrunner/node.py
h4nsu: I actually love that about Python that you can do operators both ways, with and without spaces ^^ That way you can decide which one to use to improve readability:
`y = 4 * x ^ 3 + 3 * x ^ 2 + 5 * x + 7`
`y = 4*x^3 + 3*x^2 + 5*x + 7`

local-tests/chainrunner/node.py
h4nsu: For interactive use (and I would guess this might be a common use case of that method), especially IPython console, it's much more convenient to be able to quickly grep for some string with oneliner:
`chain[2].greplog('ERROR')`
rather than:
```import re
x = re.compile('ERROR')
`chain[2].greplog(x)```

File: local-tests/chainrunner/chain.py [hunk 26]
```diff
@@ -0,0 +1,108 @@
...
+
+
+    def set_flags(self, *args, **kwargs):
+        """Set common flags for all nodes. Positional arguments are used as binary flags and should be strings.
+        Keyword arguments are translated to valued flags: `my_arg=some_val` results in `--my-arg some_val`
+        in the binary call. Seq (type alias for int) can be used to specify numerical values that should
+        be different for each node. `val=Seq(13)` results in `--val 13` for node0, `--val 14` for node1 and so on."""
+        for k in args:
+            for n in self.nodes:
+                n.flags[k] = True
```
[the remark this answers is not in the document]
h4nsu: IMO that would be overcomplicated. It would require the Node to have two types of containers (list and dict?) and some additional logic to resolve conflicts between these two. The current setup is much simpler and fits nicely into Python "polymorphic" spirit ;)

local-tests/chainrunner/chain.py
h4nsu: I thought about something similar, but going a bit further - to have `Chain` class read the `Node` deinition and automatically create, for each `Node` method, such a shortcut with additional `nodes` parameter. Needless to say, this is waaaaaaayyyy overcomplicated, so decided to just go with some code duplication. But I like your idea a lot and will implement it :)

h4nsu: Nope, that would be too beautiful. You cannot reduce `start` to that (needs a different argument for each node) and `set_chainspec` and `set_binary` are not easily expressible as lambda expressions (you cannot do `lambda x: x.attribute = value`)

local-tests/chainrunner/utils.py
h4nsu: This is awesome, thanks! :)

local-tests/send-runtime/src/main.rs
h4nsu: Done

local-tests/send-runtime/src/main.rs
```diff
@@ -0,0 +1,59 @@
...
+    connection
+        .send_extrinsic(tx.hex_encode(), XtStatus::Finalized)
+        .expect("Could not send extrinsic");
+}
+
+fn create_connection(url: String) -> Api<sr25519::Pair, WsRpcClient> {
+    let client = WsRpcClient::new(&format!("ws://{}", url));
+    Api::<sr25519::Pair, _>::new(client)
+        .expect(&format!("Could not establish connection with {}", url))
+}
```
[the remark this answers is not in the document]
h4nsu: Done

local-tests/send-runtime/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "send_runtime"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+structopt = "0.3"
+sp-core = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", branch = "master", default-features = false, features = ["full_crypto"] }
+substrate-api-client = { git = "https://github.com/scs/substrate-api-client", rev = "aa4b84f8775f972e5967aca3d96dd163619f65d5"}
```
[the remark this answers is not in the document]
h4nsu: Done

local-tests/send-runtime/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "send_runtime"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+structopt = "0.3"
```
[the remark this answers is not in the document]
h4nsu: Done

local-tests/chainrunner/chain.py
```diff
@@ -0,0 +1,108 @@
...
+        idx = nodes or range(len(self.nodes))
+        for i in idx:
+            self.nodes[i].binary = binary
+
+
+    def set_chainspec(self, chainspec, nodes=None):
+        """Replace nodes' chainspec with `chainspec`. Optional `nodes` argument can be used to specify
+        which nodes are affected and should be a list of integer indices (0..N-1). Affects all nodes if ommitted."""
+        if not op.isfile(chainspec):
+            raise FileNotFoundError(f'file not found: {chainspec}')
```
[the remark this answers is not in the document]
h4nsu: I added a utility function. Class would be an overkill imo - it would require doing the check in the constructor and then have a method for accessing wrapped string... sounds like adding more code than it saves ;)

local-tests/chainrunner/chain.py
kostekIV: You can have them in the closure/local functions, the action also may benefit from additional some `i` arg or alternatively the node should have some id inside it:
```py
    def set_binary(self, binary, nodes=None):
        ...
        def _set(node):
            node.binary = binary
        apply(lambda node, _node_id: _set(node), nodes)
```

local-tests/chainrunner/node.py
pmikolajczyk41: okay, so use `re.findall()` without explicit compiling

local-tests/chainrunner/node.py
pmikolajczyk41: look at
(substrate repo)/client/service/src/chain_ops/export_raw_state.rs, function `export_raw_state`:
```rust
/// Export the raw state at the given `block`. If `block` is `None`, the
/// best block will be used.
```

It is used (only in) client/cli/src/commands/export_state_cmd.rs, within `impl ExportStateCmd` block, which (for almost sure) you are using

local-tests/chainrunner/chain.py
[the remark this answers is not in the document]
pmikolajczyk41: fair, but then I have a problem with the term 'alias'; pls change it to wrapper or smth more relevant

local-tests/chainrunner/chain.py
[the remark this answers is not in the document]
pmikolajczyk41: anyway, for me `chain[idx]` means block at height `idx`, not a node `idx`, so I would remove it

up to you, feel free to ignore

pmikolajczyk41:
Two things are missing:
 - requirements.txt
 - README.md

local-tests/chainrunner/chain.py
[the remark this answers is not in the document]
h4nsu: Good point, done

maciejnems:
> 
About README.md, I think we should mention that `send-runtime` builds to release only if you use a nightly toolchain. I use a normal one by default and in this case it does not build because of 
```
`#![feature]` may not be used on the stable release channel
```
in substrate-api-client

local-tests/test_update.py
```diff
@@ -0,0 +1,94 @@
...
+print('Waiting a minute')
+sleep(60)
+
+check_highest(chain)
+query_runtime_version(chain)
+
+print('Stopping the chain')
+chain.stop()
+
+hf = min(node.highest_block()[1] for node in chain)
```
maciejnems, line 88: I guess it should be 
```
max(node.highest_block()[1] for node in chain)
```

local-tests/test_update.ipynb
```diff
@@ -0,0 +1,233 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "0ebd87fa-944e-4b5b-9c9e-10e401eb3272",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "hf = min(node.highest_block()[1] for node in chain)\n",
```
maciejnems, line 203: Should be max here too

maciejnems (changes requested):

local-tests/chainrunner/node.py
h4nsu: Ok

local-tests/test_update.py
pmikolajczyk41: I'm afraid that neither is 100% correct... `highest_block()`refers to the best block (see my other comment)

local-tests/chainrunner/chain.py
h4nsu: Yeah, but then we "simplify" a 3 line code-duplication by replacing it with more code AND an additional opaque step in between. This code is meant to be used as an utility tool, it's better to keep it simple and straightforward in case someone working with it wants to peek into source to quickly check something.

local-tests/test_update.py
h4nsu: The goal here is to pick a finalized block that has been seen by every node. Hence `min`

local-tests/test_update.ipynb
h4nsu: Nope :)

local-tests/chainrunner/chain.py
```diff
@@ -0,0 +1,108 @@
...
+        cmd += flags_from_dict(kwargs)
+
+        chainspec = op.join(self.path, 'chainspec.json')
+        with open(chainspec, 'w') as f:
+            subprocess.run(cmd, stdout=f).check_returncode()
+
+        self.nodes = []
+        for a in accounts:
+            n = node(binary, chainspec, op.join(self.path, a), self.path)
+            n.flags['node-key-file'] = op.join(self.path, a, 'p2p_secret')
```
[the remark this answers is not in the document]
h4nsu: Let's not add setters for the sake of having setters ;) It's Python

local-tests/chainrunner/node.py
h4nsu: Updated the comment

local-tests/chainrunner/chain.py
pmikolajczyk41: fair, resolving

local-tests/test_update.py
pmikolajczyk41: uhm, sorry, my comment is wrong - I confused it with `state`; @h4nsu is right

local-tests/chainrunner/requirements.txt
```diff
@@ -0,0 +1,2 @@
+jsonrpcclient==4.0.2
```
pmikolajczyk41, line 1: move this file to the root directory, i.e. directly in local-tests/; (it is also needed for the scripts there)

h4nsu: Done

pmikolajczyk41 (approved):
Well done! Sorry for the nit-tsunami :(

maciejnems (approved):
Very nice PR

PR #246, opened 2021-12-23
AZ-405: Python linter

pmikolajczyk41:


.github/workflows/unit_tests.yml
```diff
@@ -66,6 +66,25 @@ jobs:
...
+
+      - name: Install pylint and pycodestyle
+        run: |
+          python -m pip install --upgrade pip
+          pip install pylint pycodestyle
+
+      - name: Run pylint and pycodestyle
+        run: |
+          find . -type f -name "*.py" | xargs pylint \
+          --disable=missing-docstring,too-many-instance-attributes,duplicate-code,import-error \
```
kostekIV, line 82: can we put this in `.pylintrc` file?

.github/workflows/unit_tests.yml
```diff
@@ -66,6 +66,25 @@ jobs:
...
+          python-version: '3.8'
+
+      - name: Install pylint and pycodestyle
+        run: |
+          python -m pip install --upgrade pip
+          pip install pylint pycodestyle
+
+      - name: Run pylint and pycodestyle
+        run: |
+          find . -type f -name "*.py" | xargs pylint \
```
kostekIV, line 81: we probably want to run pylint per python project: for example `pylint ./local-tests` should be enough to run for all python files in the `local-tests` project (and is much simpler).

.github/workflows/unit_tests.yml
```diff
@@ -66,6 +66,25 @@ jobs:
...
+          --disable=missing-docstring,too-many-instance-attributes,duplicate-code,import-error \
+          --good-names-rgxs=^[_a-z][_a-z0-9]?$
+
+          find . -type f -name "*.py" | xargs pycodestyle \
```
kostekIV, line 85: We can also make a config for pycodestyle.

pmikolajczyk41: -> .pycodestyle file

.github/workflows/unit_tests.yml
pmikolajczyk41: done

.github/workflows/unit_tests.yml
pmikolajczyk41: put

local-tests/.pylintrc
```diff
@@ -0,0 +1,13 @@
+[MESSAGES CONTROL]
+
+disable=missing-docstring,
+        too-many-instance-attributes,
+        duplicate-code,
+        import-error
```
kostekIV, line 6: I would remove `import-error` from here and set the pythonpath if that is the problem this disable try to help with

pmikolajczyk41: done

kostekIV (approved):
:+1:

maciejnems (approved):

PR #248, opened 2021-12-27
AZ-362: Mock and test new network

maciejnems:


finality-aleph/src/new_network/mod.rs
```diff
@@ -85,18 +87,24 @@ impl Protocol {
...
 pub trait Network: Clone + Send + Sync + 'static {
     /// Returns a stream of events representing what happens on the network.
     fn event_stream(&self) -> NetworkEventStream;
 
-    /// A sender for sending messages to the given peer, see the substrate network docs for how to
-    /// use it. Returns None if not connected to the peer.
-    fn message_sender(
-        &self,
+    /// A method for sending data to the given peer using a given protocol. Returns Error if not connected to the peer.
+    async fn send<'a>(
```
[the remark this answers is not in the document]
maciejnems: Unfortunately because of method `ready<'a>` in `NotificationSender` depending on lifetime and returning `NotificationSenderReady<'a>` we weren't able to find any other solution with Timorl, to provide a mockable interface for Network (the issue is in creating a trait for `NotificationSender` and `NotificationSenderReady`). It was because of this lifetime and `NotificationSenderReady::send(...)` using `self` instead of reference.

finality-aleph/src/new_network/service.rs
```diff
@@ -126,20 +126,26 @@ impl<N: Network, D: Data> Service<N, D> {
...
-            if let Some(sender) = self.network.message_sender(*target, protocol.name()) {
-                return Some(sender);
+            if let Some((data, peer, protocol)) = send_queue.front() {
+                if let Ok(()) = network.send(data.encode(), *peer, protocol.name()).await {
+                    send_queue.pop_front();
+                    return Some(());
+                } else {
+                    debug!(target: "aleph-network", "Failed sending data to peer {:?}", peer);
+                }
+                send_queue.pop_front();
```
kostekIV, line 138: else branch is no needed - in Ok branch there is `return Some(());`

finality-aleph/src/new_network/service.rs
```diff
@@ -165,22 +171,425 @@ impl<N: Network, D: Data> Service<N, D> {
...
-                    match self.to_send.pop_front() {
-                        Some((data, peer, _)) => {
-                            if maybe_ready.ok().map(|ready| ready.send(data.encode()).ok()).flatten().is_none() {
-                                debug!(target: "aleph-network", "Failed sending data to peer {:?}", peer);
-                            }
-                        },
-                        None => warn!(target: "aleph-network", "Attempted to send data despite empty queue."),
-                    }
-                },
+                Some(()) =  Self::send(&self.network, &mut self.to_send) => { },
```
kostekIV, line 174: in case we want to save one character
```suggestion
                Some(_) =  Self::send(&self.network, &mut self.to_send) => { },
```

finality-aleph/src/new_network/service.rs
kostekIV, line 138: instead of peeking with `front`, the `pop_front` can be used in the if condition and both `pop_front` from branches can be removed

finality-aleph/src/new_network/service.rs
```diff
@@ -165,22 +171,425 @@ impl<N: Network, D: Data> Service<N, D> {
...
+            messages_for_user,
+            commands_from_manager,
+        };
+        let mock_io = MockIO {
+            messages_for_user: mock_messages_for_user,
+            messages_from_user: mock_messages_from_user,
+            commands_for_manager: mock_commands_for_manager,
+        };
+
+        let (oneshot_tx, event_stream_taken_oneshot_rx) = oneshot::channel();
```
kostekIV, line 227: `event_stream_taken_oneshot_rx` name suggest that it is already taken, while clearly we are waiting for that on line 241

finality-aleph/src/new_network/service.rs
```diff
@@ -165,22 +171,425 @@ impl<N: Network, D: Data> Service<N, D> {
...
+        let (service_handle, exit_tx, mut network, _mock_io) = prepare().await;
+
+        let identity = MockNetworkIdentity::new().identity();
+
+        network.emit_event(Event::SyncConnected {
+            remote: identity.1.into(),
+        });
+
+        let expected = (
+            iter::once(ScMultiaddr::empty().with(ScProtocol::P2p(identity.1 .0.into()))).collect(),
```
kostekIV, line 257:
```suggestion
            iter::once(ScMultiaddr::empty().with(ScProtocol::P2p(identity.1.0.into()))).collect(),
```

finality-aleph/src/new_network/service.rs
```diff
@@ -165,22 +171,425 @@ impl<N: Network, D: Data> Service<N, D> {
...
+                negotiated_fallback: None,
+                role: ObservedRole::Authority,
+            })
+        });
+
+        // We do this only to make sure that NotificationStreamOpened events are handled
+        network.emit_event(Event::SyncConnected {
+            remote: identities[0].1.into(),
+        });
+        let expected = ScMultiaddr::empty().with(ScProtocol::P2p(identities[0].1 .0.into()));
```
kostekIV, line 327:
```suggestion
        let expected = ScMultiaddr::empty().with(ScProtocol::P2p(identities[0].1.0.into()));
```

finality-aleph/src/new_network/service.rs
```diff
@@ -165,22 +171,425 @@ impl<N: Network, D: Data> Service<N, D> {
...
+                    protocol: Cow::Borrowed(ALEPH_PROTOCOL_NAME),
+                    remote: identity.1.into(),
+                })
+            });
+
+        // We do this only to make sure that NotificationStreamOpened events are handled
+        network.emit_event(Event::SyncConnected {
+            remote: identities[0].1.into(),
+        });
+        let expected = ScMultiaddr::empty().with(ScProtocol::P2p(identities[0].1 .0.into()));
```
kostekIV, line 407:
```suggestion
        let expected = ScMultiaddr::empty().with(ScProtocol::P2p(identities[0].1.0.into()));
```

finality-aleph/src/new_network/substrate.rs
```diff
@@ -21,19 +22,31 @@ impl<B: Block, H: ExHashT> RequestBlocks<B> for Arc<NetworkService<B, H>> {
...
+            sender
+                .ready()
+                .await
+                .ok()
+                .unwrap()
+                .send(data)
+                .map_err(|_| NetworkSendError::Closed)
+        } else {
+            Err(NetworkSendError::Closed)
+        }
```
kostekIV, line 49:
```suggestion
        if let Ok(sender) = self.notification_sender(peer_id.into(), protocol) {
            sender
                .ready()
                .await
                .ok()
                .unwrap()
                .send(data)
                .map_err(|_| NetworkSendError::Closed)
        } else {
           debug!(target: "aleph-network", "Attempted send to peer we are not connected to.");
            Err(NetworkSendError::Closed)
        }
```

finality-aleph/src/new_network/mock.rs
```diff
@@ -0,0 +1,116 @@
...
+            event_stream_taken_oneshot: Arc::new(Mutex::new(Some(oneshot_sender))),
+            network_error: Arc::new(Mutex::new(None)),
+        }
+    }
+
+    pub fn emit_event(&mut self, event: Event) {
+        for sink in &*self.event_sinks.lock() {
+            sink.unbounded_send(event.clone()).unwrap();
+        }
+    }
```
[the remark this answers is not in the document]
maciejnems: Good point. I added that

finality-aleph/src/new_network/service.rs
maciejnems: Unfortunately we can't do it, as send is asynchronous. It might happen that we pop something, try to send it, another thread finishes first and we loose the data we popped

finality-aleph/src/new_network/service.rs
maciejnems: This `rx` should receive when and event stream is taken first time. That's why I called it like that. I can just change it to `oneshot_rx`

finality-aleph/src/new_network/service.rs
maciejnems: Well, the formatter changes it to be like that :(

finality-aleph/src/new_network/service.rs
kostekIV: I think I am missing something, in the end, we pop data anyway. What the other thread you have mentioned is doing?

finality-aleph/src/new_network/service.rs
kostekIV: I get that, but from the name, it looks like it is already `taken`. This part is already quite confusing to me because in the test setup we are waiting for this stream to be taken which looks kinda fishy to me so that's why I would like this name to not be ambiguous. (as a side note `service.run` could return `Future<Future<()>>` with the outer future completing when the `service` is ready to use without any implementations details specified what it means like that the event stream is taken)

finality-aleph/src/new_network/service.rs
kostekIV: wow :O

finality-aleph/src/new_network/service.rs
maciejnems: But we pop it after the send is successful. This is used in `tokio::select` in `Service::run`.

kostekIV: what about the pop on the line 138?

maciejnems: Word successful was not a good choice here. After the send is done

finality-aleph/src/new_network/mock.rs
```diff
@@ -0,0 +1,116 @@
...
+
+type Channel<T> = (
+    Arc<Mutex<mpsc::UnboundedSender<T>>>,
+    Arc<Mutex<mpsc::UnboundedReceiver<T>>>,
+);
+
+fn channel<T>() -> Channel<T> {
+    let (tx, rx) = mpsc::unbounded();
+    (Arc::new(Mutex::new(tx)), Arc::new(Mutex::new(rx)))
+}
```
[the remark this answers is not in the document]
maciejnems: I'm not sure we should do that. In future `testing/network.rs` will be removed, so the code won't be duplicated.

finality-aleph/src/new_network/service.rs
kostekIV: discussed offline

finality-aleph/src/new_network/service.rs
maciejnems: I think this is a normal behavior that the service takes `event_stream` right before running and we do it like that in many other places. This only makes some complications when testing.  That's why I added a comment in line 238. I will ask @timorl for his opinion when he comes back from vacation.

finality-aleph/src/new_network/service.rs
```diff
@@ -126,20 +126,29 @@ impl<N: Network, D: Data> Service<N, D> {
         }
     }
 
-    fn next_sender(&mut self) -> Option<NotificationSender> {
+    async fn send(network: &N, send_queue: &mut VecDeque<(D, PeerId, Protocol)>) -> Option<()> {
```
timorl, line 129: `Option<()>` is a _really_ strange return type, and you are not even using the `?` trick from the previous version which would have justified it at least a little bit.

The most expressive variant of this I can think of would return `Option<Result<(), NetworkSendError>>`, preferably with `NetworkSendError` containing the peer id and a way of getting printed that could be useful in the`tokio::select!` for printing. This way you could make the inside of the loop something like
```
let (data, peer, protocol) = send_queue.front()?;
let result = network.send(data.encode(), *peer, protocol.name()).await;
send_queue.pop_front();
return Some(result);
```
or even better not make it a loop at all, making everything simpler. You lose the `trace` about an empty queue, but it's not that useful anyway.

finality-aleph/src/new_network/service.rs
```diff
@@ -165,22 +174,486 @@ impl<N: Network, D: Data> Service<N, D> {
...
+            messages_from_user: mock_messages_from_user,
+            commands_for_manager: mock_commands_for_manager,
+        };
+
+        let (event_stream_oneshot_tx, event_stream_oneshot_rx) = oneshot::channel();
+        let network = MockNetwork::new(event_stream_oneshot_tx);
+        let service = Service::new(network.clone(), io);
+
+        let (exit_tx, exit_rx) = oneshot::channel();
+        let task_handle = async move {
```
timorl, line 234: I would suggest rather than `run`ning the service test the methods directly. Preferably this would include rewriting the whole thing into a separate struct that just returns the correct things without sending them to any channels (like `new_network::manager::Service`), but that's probably too much of a refactor for this PR, so just using the internal methods is fine. As an additional perk this sidesteps the issue with the whole `event_stream_taken` channel, as event streams wouldn't be directly touched by these tests.

finality-aleph/src/new_network/substrate.rs
```diff
@@ -21,19 +22,30 @@ impl<B: Block, H: ExHashT> RequestBlocks<B> for Arc<NetworkService<B, H>> {
...
-        self.notification_sender(peer_id.into(), protocol)
-            .map_err(|_| debug!(target: "aleph-network", "Attempted send to peer we are not connected to."))
-            .ok()
+    ) -> Result<(), NetworkSendError> {
+        if let Ok(sender) = self.notification_sender(peer_id.into(), protocol) {
+            sender
+                .ready()
+                .await
+                .ok()
+                .unwrap()
```
timorl, line 42: Oh nononono, this will explode. You probably gotta match on this and return an error.

finality-aleph/src/new_network/substrate.rs
```diff
@@ -21,19 +22,30 @@ impl<B: Block, H: ExHashT> RequestBlocks<B> for Arc<NetworkService<B, H>> {
...
+        &'a self,
+        data: impl Into<Vec<u8>> + Send + Sync + 'static,
         peer_id: PeerId,
         protocol: Cow<'static, str>,
-    ) -> Option<NotificationSender> {
-        self.notification_sender(peer_id.into(), protocol)
-            .map_err(|_| debug!(target: "aleph-network", "Attempted send to peer we are not connected to."))
-            .ok()
+    ) -> Result<(), NetworkSendError> {
+        if let Ok(sender) = self.notification_sender(peer_id.into(), protocol) {
```
timorl, line 37: It's probably better to `map_err` this and `?` after that. Just make the error here more expressive (see other comments about that). This whole method can probably be remade into 3 calls in succession which do `map_err` into `?`. It could even be a "one liner" like `self.notification_sender(peer_id.into(), protocol).map_err(...)?.ready().await.map_err(...)?.send(data).map_err(...);`. All the logging should happen outside anyway, only the returned error should contain enough information to be useful.

timorl (changes requested):
A couple of pretty big comments, but in general very nice work, thanks!

finality-aleph/src/new_network/mock.rs
```diff
@@ -0,0 +1,148 @@
...
+
+#[derive(Clone)]
+pub struct MockNetwork<B: Block> {
```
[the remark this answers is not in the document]
maciejnems: I removed this part

finality-aleph/src/new_network/service.rs
maciejnems: Done in this way

finality-aleph/src/new_network/service.rs
maciejnems: I think that would be a huge refactor and I would prefer to keep it as it is

finality-aleph/src/new_network/substrate.rs
maciejnems: Done

finality-aleph/src/new_network/substrate.rs
maciejnems: done

finality-aleph/src/new_network/mod.rs
```diff
@@ -85,18 +88,24 @@ impl Protocol {
 
 type NetworkEventStream = Pin<Box<dyn Stream<Item = Event> + Send>>;
 
+#[derive(Copy, Clone)]
+pub enum NetworkSendError {
+    Closed,
```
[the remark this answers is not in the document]
maciejnems: I think std::error::Error is enough here.

finality-aleph/src/new_network/substrate.rs
```diff
@@ -33,19 +50,14 @@ impl<B: Block, H: ExHashT> Network for Arc<NetworkService<B, H>> {
...
-            Err(NetworkSendError::Closed)
-        }
+    ) -> Result<(), SendError> {
+        self.notification_sender(peer_id.into(), protocol)
+            .map_err(|_| SendError::NotConnectedToPeer(peer_id))?
+            .ready()
+            .await
+            .map_err(|_| SendError::NotConnectedToPeer(peer_id))?
+            .send(data)
+            .map_err(|_| SendError::NotConnectedToPeer(peer_id))
```
timorl: At least this error should be different than the two above, it means the connection has been pretty suddenly closed, which might be useful for debugging.

finality-aleph/src/new_network/substrate.rs
```diff
@@ -33,19 +50,14 @@ impl<B: Block, H: ExHashT> Network for Arc<NetworkService<B, H>> {
...
-                .map_err(|_| NetworkSendError::Closed)
-        } else {
-            debug!(target: "aleph-network", "Attempted send to peer we are not connected to.");
-            Err(NetworkSendError::Closed)
-        }
+    ) -> Result<(), SendError> {
+        self.notification_sender(peer_id.into(), protocol)
+            .map_err(|_| SendError::NotConnectedToPeer(peer_id))?
+            .ready()
+            .await
```
timorl, line 73: As you rightly pointed out this is explicitly discouraged by the docs. The simplest solution I can think of is to add some pretty tight timeout here and `tokio::select` on these two, returning a timeout error in the appropriate situation.

timorl (changes requested):
Pretty nice, I would have probably accepted it if you haven't brought up the wrong usage of `network_sender`. We need to fix that somehow, and for now I believe a timeout is fine, although maybe we should create a task for improving this in the future, e.g. using tasks spawned for specific peers.

finality-aleph/src/new_network/service.rs
kostekIV: Imho this will require changes everywhere when this behaviour will change. @timorl what is your opinion on this matter?

timorl: I suggested rewriting the tests to not use the `event_stream` at all, at least for now, but this is probably too much of a refactor for this PR. Other than that I think it's kinda alright the way this is right now, even though it's a tad too complicated.

kostekIV: Ok, I also don't like that we are depending on the event_stream in the tests but I am fine with moving it to a new issue for now.

finality-aleph/src/new_network/substrate.rs
maciejnems: I will add timeout for now and start working on a proper solution right away

maciejnems: I added a timeout for 200 ms and it's still a "one liner"

finality-aleph/src/new_network/substrate.rs
```diff
@@ -21,19 +21,74 @@ impl<B: Block, H: ExHashT> RequestBlocks<B> for Arc<NetworkService<B, H>> {
...
 
+#[derive(Debug, Copy, Clone)]
+pub enum SendError {
+    NotConnectedToPeer(PeerId),
+    LostConnectionToPeer(PeerId),
+    LostConnectionToPeerReady(PeerId),
+    SendTimeout(PeerId, u64),
+}
+
+const SEND_TO_PEER_TIMEOUT: u64 = 200;
```
timorl, line 32: Since this value does not inherently have units, I would add a `_MS` suffix to the name for clarity.

timorl (approved):
Nice! Added a tiny nit, but in general LGTM.

finality-aleph/src/new_network/substrate.rs
maciejnems: done

PR #253, opened 2022-01-04
flooder: single connection instead of connection-per-tx

fixxxedpoint:


pmikolajczyk41 (changes requested):
Ideally, I would split this PR into two smaller ones: one for resolving dependencies and one changing semantics

common/Cargo.toml
```diff
@@ -5,12 +5,15 @@ edition = "2021"
 
 # client
 [dependencies.substrate-api-client]
-git = "https://github.com/scs/substrate-api-client"
-rev = "aa4b84f8775f972e5967aca3d96dd163619f65d5"
+git = "https://github.com/Cardinal-Cryptography/substrate-api-client.git"
+rev = "8b7346a11a70f46d05de543a6bed37f3fdf6ac8f"
```
pmikolajczyk41, line 9: Unfortunately this SHA corresponds to a latest commit at branch called `single_threaded_sender.PR`, which (pardon me) definitely does not sound like a source of a stable version for 3rd-party library :P
The easiest way is to rename the branch or create a tag.

fixxxedpoint: How is renaming of a branch going to fix anything :D

fixxxedpoint: I will make a tag of from it, since pushing it to main doesn't look like a good idea.

fixxxedpoint: changed to tag

common/src/ws_rpc_client.rs
```diff
@@ -0,0 +1,251 @@
...
+            _ => Err(ApiClientError::UnsupportedXtStatus(exit_on)),
+        };
+
+        // reset the RpcClient handler used by the WebSocket's thread
+        *self
+            .next_handler
+            .lock()
+            .expect("unable to acquire a lock on RpcClient") = None;
+        result
+    }
```
pmikolajczyk41, line 124: maybe just add a method like `reset_handler()`

common/src/ws_rpc_client.rs
```diff
@@ -0,0 +1,251 @@
...
+
+    fn send_rpc_request(
+        &self,
+        jsonreq: String,
+        result_in: ThreadOut<String>,
+        on_message_fn: OnMessageFn,
+    ) -> WsResult<()> {
+        // 1 used by `on_open` of RpcClient + 1 for `close`
+        const MAGIC_SEND_CONST: usize = 2;
+        let (ws_tx, _ws_rx) = mio::channel::sync_channel(MAGIC_SEND_CONST);
```
pmikolajczyk41, line 176: `channel` is deprecated, but you can use `mio-extras`

common/src/ws_rpc_client.rs
```diff
@@ -0,0 +1,251 @@
...
+            on_extrinsic_msg_until_ready, on_get_request_msg, OnMessageFn, RpcClient,
+        },
+    },
+    ApiClientError, ApiResult, FromHexString, RpcClient as RpcClientTrait, XtStatus,
+};
+use ws::{connect, Handler, Message, Result as WsResult, Sender as WsSender};
+
+pub struct WsRpcClient {
+    mux: Mutex<()>,
+    next_handler: Arc<Mutex<Option<RpcClient>>>,
```
pmikolajczyk41, line 27: would `Option<Arc<Mutex<RpcClient>>>` work?

pmikolajczyk41 (approved):
I wasn't able to verify in-depth the mechanics, but seems ok

common/src/ws_rpc_client.rs
fixxxedpoint: I don't think I can - mismatched types. `Secret` constructor ;) https://docs.rs/ws/0.2.1/src/ws/.cargo/registry/src/github.com-1ecc6299db9ec823/ws-0.2.1/src/communication.rs.html#52

common/src/ws_rpc_client.rs
fixxxedpoint: Good idea.

mike1729 (approved):
Very nice work! If the upstream merges the changes from this PR then their average code quality will double :sunglasses:

common/src/ws_rpc_client.rs
fixxxedpoint: I am sharing this `Option` between threads (`Handler for WsHandler`) so I think I need this `Mutex` around it.

common/src/ws_rpc_client.rs
fixxxedpoint: Changed.

PR #254, opened 2022-01-04
flooder refactored

fixxxedpoint:


flooder/src/main.rs
```diff
@@ -36,50 +38,89 @@ fn main() -> Result<(), anyhow::Error> {
...
+            (
+                transactions_in_interval
+                    .try_into()
+                    .expect("`transactions_in_interval` should be within usize range"),
+                Duration::from_secs(secs),
+            )
+        },
+    );
+
+    let pool_getter = |tx_ix: usize| pool.get(tx_ix % pool.len()).unwrap();
```
kostekIV, line 111: Correct me if I am wrong but this means that different threads can get the same connection from the pool. We should have a distinct connection for each node for each thread to leverage the parallelism.

flooder/src/main.rs
```diff
@@ -104,48 +145,47 @@ fn main() -> Result<(), anyhow::Error> {
     Ok(())
 }
 
-fn flood(
-    pool: &[Api<sr25519::Pair, WsRpcClient>],
+fn flood<'a>(
+    pool: impl Fn(usize) -> &'a Api<sr25519::Pair, WsRpcClient> + Sync,
     txs: impl IndexedParallelIterator<Item = TransferTransaction>,
```
kostekIV, line 150: Not sure how it slipped but since this is ParallelIterator the rate_limiting does not do what we want it to do. It should be just a vector or normal iterator.

flooder/src/main.rs
fixxxedpoint: Please see the new version.

common/src/lib.rs
```diff
@@ -6,34 +6,40 @@ use std::{thread::sleep, time::Duration};
...
 ) -> Api<sr25519::Pair, substrate_api_client::rpc::ws_client::WsRpcClient> {
-    create_custom_connection(&address)
+    create_custom_connection(&address).expect("connection should be created")
 }
 
 pub fn create_custom_connection<Client: FromStr + RpcClient>(
     address: &str,
-) -> Api<sr25519::Pair, Client> {
-    let client = Client::from_str(&format!("ws://{}", address));
+) -> Result<Api<sr25519::Pair, Client>, <Client as FromStr>::Err> {
```
kostekIV, line 39: This function never returns `Err` it should return just `Api`

flooder/src/main.rs
```diff
@@ -36,50 +37,94 @@ fn main() -> Result<(), anyhow::Error> {
...
+            (
+                transactions_in_interval
+                    .try_into()
+                    .expect("`transactions_in_interval` should be within usize range"),
+                Duration::from_secs(secs),
+            )
+        },
+    );
+
+    // let pool_getter = |tx_ix: usize| pool.get(tx_ix % pool.len()).unwrap();
```
kostekIV, line 110: nit: remove commented code

common/src/lib.rs
fixxxedpoint: What about line :40?

flooder/src/main.rs
fixxxedpoint: removed

flooder/src/main.rs
fixxxedpoint: Please see the new version.

common/src/lib.rs
kostekIV: Im blind

kostekIV (approved):

PR #255, opened 2022-01-04
Save send-runtime binary as an artifact

pmikolajczyk41:


.github/workflows/save-send-runtime-binary.yml
```diff
@@ -0,0 +1,83 @@
...
+          toolchain: ${{ env.RUST_TOOLCHAIN_VERSION }}
+          override: true
+
+      - name: Build binary
+        run: |
+          pushd local-tests/send-runtime/
+          cargo build --release
+          popd
+
+      - name: Upload binary
```
h4nsu, line 48: Why are we uploading the "bare" binary as an artifact?

.github/workflows/save-send-runtime-binary.yml
```diff
@@ -0,0 +1,83 @@
...
+        uses: actions/upload-artifact@v2
+        with:
+          name: send-runtime
+          path: local-tests/send-runtime/target/release/send_runtime
+          if-no-files-found: error
+          retention-days: 7
+
+      - name: Build send-runtime-image
```
h4nsu, line 56: I find the name of this image somewhat confusing. How about `runtime-update-hook`? Or `runtime-update-image`?

h4nsu (approved):

.github/workflows/save-send-runtime-binary.yml
pmikolajczyk41: I took the former proposition

.github/workflows/save-send-runtime-binary.yml
pmikolajczyk41: leftover from previous approach; removed

PR #256, opened 2022-01-05
Fixed `substrate-api-client` dependencies

fixxxedpoint:


flooder/Cargo.toml
```diff
@@ -23,4 +24,4 @@ env_logger = "0.8"
 futures = { version = "0.3", features = ["alloc"] }
 hdrhistogram = "7.3.0"
 log = "0.4"
-rayon = "1.5.1"
+rayon = "1.5.1"
```
pmikolajczyk41, line 27: newline missing

flooder/Cargo.toml
```diff
@@ -1,17 +1,18 @@
 [package]
 name = "flooder"
 version = "0.1.0"
+authors = ["Cardinal Cryptography"]
```
pmikolajczyk41, line 4: in other packages we do not admit so; not a blocker, but I think we should be consistent

pmikolajczyk41 (approved):

PR #258, opened 2022-01-06
AZ-107: Restart authority tasks if they fail

timorl:
Also splits up party.rs a tad to make it more readable.

Main change is needed for a strange fail case of the new network, which shouldn't happen with proper usage when the committee actually changes, but right now would break everything. It is also actually useful in general, we want to restart if something breaks this way anyway.

### DIFF
--- a/finality-aleph/src/data_io.rs
+++ b/finality-aleph/src/data_io.rs
@@ -49,7 +49,7 @@ impl<H, N> AlephData<H, N> {
 
 pub(crate) type AlephDataFor<B> = AlephData<<B as BlockT>::Hash, NumberFor<B>>;
 
-pub(crate) trait AlephNetworkMessage<B: BlockT> {
+pub trait AlephNetworkMessage<B: BlockT> {
     fn included_blocks(&self) -> Vec<AlephDataFor<B>>;
 }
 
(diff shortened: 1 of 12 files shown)


finality-aleph/src/party/mod.rs
```diff
@@ -561,50 +454,40 @@ where
...
         loop {
             let last_finalized_number = self.client.info().finalized_number;
             debug!(target: "afa", "Highest finalized: {:?} session {:?}", last_finalized_number, session_id);
             if last_finalized_number >= last_block {
                 debug!(target: "afa", "Terminating session {:?}", session_id);
                 break;
             }
-            Delay::new(Duration::from_millis(1000)).await;
+            tokio::select! {
+                _ = Delay::new(Duration::from_millis(1000)) => (),
```
Yithis, line 477: Maybe make it a constant?

Yithis (approved):
LGTM

finality-aleph/src/data_io.rs
```diff
@@ -49,7 +49,9 @@ impl<H, N> AlephData<H, N> {
 
 pub(crate) type AlephDataFor<B> = AlephData<<B as BlockT>::Hash, NumberFor<B>>;
 
-pub(crate) trait AlephNetworkMessage<B: BlockT> {
+/// A trait allowing to check the data contained in an AlephBFT network message, for the purpose of
+/// data vailability checks.
```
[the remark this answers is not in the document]
timorl: Fixed.

finality-aleph/src/party/mod.rs
timorl: Good idea, done.

mike1729 (approved):
Hard to judge if this is the proper way of restarting without seeing the actual use case, but in general looks ok.

PR #259, opened 2022-01-06
New network: fix bug and add test

Yithis:


finality-aleph/src/new_network/component.rs
```diff
@@ -36,6 +36,59 @@ impl<D: Data, CN: Network<D>> DataNetwork<D> for CN {
...
+        type R = mpsc::UnboundedReceiver<u64>;
+        
+        fn sender(&self) -> &Self::S {
+            &self.sender
+        }
+        
+        fn receiver(&self) -> Arc<Mutex<Self::R>> {
+            self.receiver.clone()
+        }
+    }
```
timorl, line 76: Everything above this line in the test is IMO unnecessary – this is not how you would implement a `Network` anyway, and even if it was it doesn't really test anything. Just create a receiver channel and test it by calling `Receiver::next(receiver)` on it. You don't even have to implement sender for the other end of the channel in this case.

This is a nit obviously, but leaving this in might result in confusion as to what the test is actually about.

timorl (approved):
I complain about you doing too much work. xd

finality-aleph/src/new_network/component.rs
Yithis: Done, but due to `cargo fmt `now we are proposing the same changes :sweat_smile:

maciejnems (approved):

PR #262, opened 2022-01-07
AZ-408: test_update.py returns non-zero when update failed

h4nsu:
This is needed so that an automatic pipeline can detect that something went wrong with the update test

### DIFF
--- a/docker-runtime-hook/entrypoint.sh
+++ b/docker-runtime-hook/entrypoint.sh
@@ -1,11 +1,10 @@
 #/bin/bash
 
-rm -rf aleph-node runtime
 set -e
 
 NETRC_CREDS="./_netrc"
 RUNTIME_TOOL="./send_runtime"
-SUDO_PHRASE=${RUNTIME_PHRASE} #TODO take from some safe place
+SUDO_PHRASE=${RUNTIME_PHRASE}
 
 RPC_ADDR="rpc.dev.azero.dev"
 WS_ADDR="ws.dev.azero.dev"
--- a/local-tests/test_update.ipynb
+++ b/local-tests/test_update.ipynb
@@ -40,15 +40,14 @@
    "source": [
     "def query_runtime_version(nodes):\n",
     "    print('Current version:')\n",
-    "    versions = []\n",
+    "    versions = set()\n",
     "    for i, node in enumerate(nodes):\n",
     "        sysver = node.rpc('system_version').result\n",
     "        rt = node.rpc('state_getRuntimeVersion').result['specVersion']\n",
     "        print(f'  Node {i}: system: {sysver}  runtime: {rt}')\n",
-    "        versions.append(rt)\n",
-    "    versions = set(versions)\n",
+    "        versions.add(rt)\n",
     "    if len(versions) != 1:\n",
-    "        print('ERROR: nodes reported different runtime versions')\n",
+    "        print(f'ERROR: nodes reported different runtime versions: {versions}')\n",
     "    return max(versions)"
    ]
   },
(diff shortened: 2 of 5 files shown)


local-tests/test_update.py
```diff
@@ -20,10 +21,16 @@
 
 def query_runtime_version(nodes):
     print('Current version:')
+    versions = []
```
kostekIV, line 24: wdyt about making it set already instead of a list?

local-tests/test_update.py
```diff
@@ -20,10 +21,16 @@
...
     for i, node in enumerate(nodes):
-        sys = node.rpc('system_version').result
+        sysver = node.rpc('system_version').result
         rt = node.rpc('state_getRuntimeVersion').result['specVersion']
-        print(f'  Node {i}: system: {sys}  runtime: {rt}')
+        print(f'  Node {i}: system: {sysver}  runtime: {rt}')
+        versions.append(rt)
+    versions = set(versions)
+    if len(versions) != 1:
+        print('ERROR: nodes reported different runtime versions')
```
kostekIV, line 32: Add reported versions to the error msg

kostekIV, line 32: this actually won't trigger any errors in the pipeline (only an error message will be displayed). Consider changing return type to set (or dict (node -> version)) and use that information for for example trigger to `sys.exit(1)` later in the script.

local-tests/test_update.py
h4nsu: Of course, why haven't I thought about that ^^'

local-tests/test_update.py
kostekIV: It's always good to be over paranoid :) I am fine with this explanation, maybe add this "In a very unlikely ... the last query." as a comment here.

h4nsu: Done

h4nsu: Added

local-tests/test_update.ipynb
```diff
@@ -207,6 +211,17 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "9ede429e-fca2-41fe-80c7-ac7df1606ddc",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "if oldver == newver:\n",
+    "    print(\"ERROR: runtime version reported by nodes didn't change after the update\")"
+   ]
```
pmikolajczyk41, line 224: missing `sys.exit(1)`; also in the previous lines (~61 the notebook has changes that are not included in the script itself)

docker-runtime-hook/entrypoint.sh
```diff
@@ -1,11 +1,10 @@
 #/bin/bash
```
pmikolajczyk41, line 1: are these changes relevant?

pmikolajczyk41 (approved):
Thank you for spotting the problem and fixing it!

local-tests/test_update.ipynb
h4nsu: Yes, on purpose. Notebook is not supposed to do things like sleep, sys.exit etc. It's for humans to use :)

docker-runtime-hook/entrypoint.sh
h4nsu: It's a cosmetic change that somehow fell through the cracks. As I am the original author of this file and feel a bit guilty about that, I wanted to fix it. Can I, without a separate PR, pleeeeaaaase? :P

kostekIV (approved):

PR #272, opened 2022-01-12
Add some utilities to local testing

pmikolajczyk41:
Adding two utilities:
 - setting logging level for arbitrary target
 - checking hash of a finalized block (given its number)

which I found really useful while working on AZ-395. Hopefully, they will be helpful for someone else as well.


local-tests/chainrunner/node.py
```diff
@@ -24,13 +24,19 @@ def __init__(self, binary, chainspec, path, logdir=None):
         self.process = None
         self.flags = {}
         self.running = False
+        self.log_level = {}
```
h4nsu, line 27: Couldn't you just add the logging info to `flags`? Seems to me like now there are two incompatible ways of doing the same thing and no logic to disambiguate in case of conflict...

local-tests/chainrunner/node.py
```diff
@@ -65,6 +71,13 @@ def highest_block(self):
             return int(a), int(b)
         return -1, -1
 
+    def check_hash_of(self, number):
```
h4nsu, line 74: nit: I'd suggest a bit different naming here. This doesn't check anything, it's a getter. How about `get_hash_of` or (even better) `get_hash`? Also the argument would be more self-explanatory if it was called `height`

maciejzelaszczyk:
I see no obvious problems. As I am not yet familiar with our codebase, I will defer from approving at this time and wait for a parallel review by @h4nsu. On a side note, type annotations could help us get rid of the type information from the docstrings.

pmikolajczyk41:
> On a side note, type annotations could help us get rid of the type information from the docstrings.

Glad to hear that, I am also supporting type hints. However, the author of this part of code, @h4nsu, is on the dark side of the force :smile:

local-tests/chainrunner/node.py
```diff
@@ -65,6 +71,13 @@ def highest_block(self):
...
+        """Find in the logs the hash for block with number `number`."""
+        results = self.greplog(rf'Finalizing block {number} (.+)')
```
h4nsu, line 76: A small problem I see with this is that the entry you grep for is logged on a debug level. Otoh there's no info-level message that you can use to get the full hash (only something like `0x0cc8…e968`)

h4nsu (changes requested):

local-tests/chainrunner/node.py
pmikolajczyk41: well, the debug flags have the form of `-ltarget=level`, so it neither matches neither value nor non-value flags that we are using

local-tests/chainrunner/node.py
pmikolajczyk41: You are right, so what about one of the two solutions:
 - adding to docstring a warning that debug level should be on
 - checking in code whether a debug level is on, and if not, printing warning
 ?

local-tests/chainrunner/node.py
pmikolajczyk41: done, thx

local-tests/chainrunner/node.py
h4nsu: Could be done with the long form `--log target=level`

kostekIV (approved):

h4nsu (approved):

maciejzelaszczyk (approved):

PR #274, opened 2022-01-12
[Not for merging] AZ-395 test code substitute

pmikolajczyk41:


Marcin-Radecki:
Changes looks good, we might think of incorporating those tests in CI pipeline in further sprints.

PR #275, opened 2022-01-12
AZ-418: Add per peer buffering

maciejnems:


finality-aleph/src/new_network/service.rs
```diff
@@ -204,52 +226,95 @@ mod tests {
...
+        }
 
-        (service_handle, exit_tx, network, mock_io)
+        // We do this only to make sure that NotificationStreamOpened/NotificationStreamClosed events are handled
+        async fn wait_for_events_handled(&mut self) {
+            let identity = MockNetworkIdentity::new().identity();
+            self.network.emit_event(Event::SyncConnected {
+                remote: identity.1.into(),
+            });
+            let expected = ScMultiaddr::empty().with(ScProtocol::P2p(identity.1 .0.into()));
```
timorl, line 294: `.1 .0.` has an unexpected space and is a bit, uh, hard to read. Don't worry about the latter if it's hard to fix, but please remove the space.

timorl (approved):
Some smallish comments but in general very cool.

finality-aleph/src/new_network/service.rs
maciejnems: Unfortunately this is the way rust formatter wants it to be :( I had a comment about that in the previous PR too

maciejnems:
I applied changes suggested off-github, to reduce number of created NotificationSenders

finality-aleph/src/new_network/service.rs
```diff
@@ -551,17 +566,211 @@ mod tests {
...
+            .ok();
+
+        test_data
+            .mock_io
+            .messages_for_user
+            .unbounded_send((
+                message_2.clone(),
+                DataCommand::SendTo(identity.1, Protocol::Validator),
+            ))
+            .ok();
```
timorl, line 611: Those should probably be `unwrap`s rather than `ok`s.

timorl: And you have a lot of them below.

finality-aleph/src/new_network/mod.rs
```diff
@@ -88,20 +88,28 @@ impl Protocol {
...
         &'a self,
         data: impl Into<Vec<u8>> + Send + Sync + 'static,
-        peer_id: PeerId,
-        protocol: Cow<'static, str>,
-    ) -> Result<(), Self::SendError>;
+    ) -> Result<(), Self::SenderError>;
+}
+
+/// Abstraction over a network.
+pub trait Network<NS: NetworkSender>: Clone + Send + Sync + 'static {
```
timorl, line 104: Make NS a trait type, then you can drop one template arg in many places.

finality-aleph/src/new_network/service.rs
```diff
@@ -29,16 +31,28 @@ pub struct IO<D: Data> {
     commands_from_manager: mpsc::UnboundedReceiver<ConnectionCommand>,
 }
 
-async fn run_peer_sender<N: Network, D: Data>(
+async fn run_peer_sender<N: Network<NS>, NS: NetworkSender, D: Data>(
```
timorl, line 34: I think it might be more readable to put this inside of `Service` take the returned future and run that. No pressure if you don't like this, but it would eliminate one turbofish.

finality-aleph/src/new_network/mod.rs
timorl: And one `PhantomData` too!

timorl (approved):
Some minor comments, but this seems to be finally absolutely correct. :D

finality-aleph/src/new_network/service.rs
maciejnems: I agree 100%

finality-aleph/src/new_network/service.rs
```diff
@@ -38,13 +44,59 @@ impl<N: Network, D: Data> Service<N, D> {
...
+            }
+        }
+    }
+
     fn send_to_peer(&mut self, data: D, peer: PeerId, protocol: Protocol) {
-        self.to_send.push_back((data, peer, protocol));
+        if let Some(sender) = self.peer_senders.get_mut(&peer) {
+            if let Err(e) = sender.try_send((data, protocol)) {
+                if e.is_full() {
+                    debug!(target: "aleph-network", "Failed sending data to peer because buffer is full: {:?}", peer);
```
DamianStraszak, line 91: It would be great if there was some warning at the warn or info logging level indicating that we droppped a message (or multiple messages) because of buffer overflow. That's because we normally run production nodes on the default info
 level so we would not see that happening with current logs. At the same time we don't want to spam the logs with such messages (in case tons of  messages are being dropped) so this requires some thought.

maciejnems: I'm not sure if we would like to potentially spam this message here. I'm not sure how many such logs could potentially be spammed here. Also I'm not sure how useful this message would be. This could help figure out if buffer size is too small.

What do you think @timorl ?

maciejnems: Another thing is that right now we don't have any logs over level debug when it comes down to sending messages to peers.

timorl: I think this is unlikely to happen at all, but likely to be spammy conditioned on it happening. Maybe this means this can be a `warn` after all? Implementing a rate limiting system for this single log seems incredibly annoying, so I would prefer we didn't try doing that.

DamianStraszak: The general point I'm making here is that whenever we hit some issues with our node in production, we **MUST** be able to tell which part of our code is responsible for the problems, just by inspecting the logs (and the logs are on the `info` level). 
So whenever there is something that should never hapen, or we hope this will never happen, we should log when it happens.

Making it just a `warn` is probably fine. Hopefully we will not generate TBs of logs because of that :P

timorl: Still need to import it, it always turns out the log level you are trying to use is the only one not already imported. ;p

maciejnems: Yeaaaah, I forgot to check if it compiles :|

DamianStraszak (approved):

PR #276, opened 2022-01-12
AZ-107: Delete old network and connect the new one

Yithis:


finality-aleph/src/lib.rs
```diff
@@ -157,8 +161,8 @@ pub fn session_id_from_block_num<B: Block>(num: NumberFor<B>, period: SessionPer
     SessionId(num.saturated_into::<u32>() / period.0)
 }
 
-pub struct AlephConfig<B: Block, N, C, SC> {
-    pub network: N,
+pub struct AlephConfig<B: Block, H: ExHashT, C, SC> {
+    pub network: Arc<NetworkService<B, H>>,
```
timorl, line 165: Maybe rather than the specific `NetworkService` you could require an `N: network::Network + network::RequestBlocks<H>`? I believe this is what we actually want, right?

finality-aleph/src/network/manager/service.rs
```diff
@@ -419,21 +435,30 @@ impl<D: Data> IO<D> {
...
                     Some((message, session_id, recipient)) => for message in service.on_user_message(message, session_id, recipient) {
                          self.send_data(message)?;
                     },
                     None => return Err(Error::MessageChannel),
+                }
                 },
-                maybe_message = self.messages_from_network.next() => match maybe_message {
+                maybe_message = self.messages_from_network.next() => {
+                    trace!(target: "aleph-network", "Manager Loop message from network case");
+                    match maybe_message {
```
timorl, line 461: Isn't the linting broken here? Also above? The indentation seems wrong.

finality-aleph/src/network/manager/service.rs
```diff
@@ -419,21 +435,30 @@ impl<D: Data> IO<D> {
     ) -> Result<(), Error> {
         let mut maintenance = interval(Duration::from_secs(MAINTENANCE_PERIOD_SECONDS));
         loop {
+            trace!(target: "aleph-network", "Manager Loop started a next iteration");
             tokio::select! {
-                maybe_command = self.commands_from_user.next() => match maybe_command {
+                maybe_command = self.commands_from_user.next() => {
+                    trace!(target: "aleph-network", "Manager Loop command case");
```
timorl, line 441: "Manager received a command from user" would be better, same style in the other `trace`s.

finality-aleph/src/network/mod.rs
```diff
@@ -25,8 +25,13 @@ use component::{
 };
 use manager::SessionCommand;
 
-pub use aleph::NetworkData as AlephNetworkData;
+pub use aleph::{NetworkData as AlephNetworkData, NetworkWrapper};
+pub use component::SimpleNetwork;
+pub use manager::{ConnectionIO, ConnectionManager};
 pub use rmc::NetworkData as RmcNetworkData;
+pub use service::{Service, IO};
+pub use session::{Manager, ManagerError, Network as SessionNetwork};
```
timorl, line 33: If you are `pub use`ing them you need to refer to this manager as `SessionManager`, otherwise it might be confused with `ConnectionManager` above.

Do you actually need these `pub use`s tho?

finality-aleph/src/network/mod.rs
```diff
@@ -25,8 +25,13 @@ use component::{
 };
 use manager::SessionCommand;
 
-pub use aleph::NetworkData as AlephNetworkData;
+pub use aleph::{NetworkData as AlephNetworkData, NetworkWrapper};
```
timorl, line 28: At the top of this file – remove the comment and the permission for dead code. Fix everything that breaks as the result of this change ofc. :P

finality-aleph/src/network/split.rs
```diff
@@ -75,21 +75,28 @@ async fn forward_or_wait<
     receiver: &Arc<Mutex<R>>,
     left_sender: &mpsc::UnboundedSender<LeftData>,
     right_sender: &mpsc::UnboundedSender<RightData>,
-) {
+) -> bool {
+    trace!(target: "aleph-network", "forward_or_wait called");
```
timorl, line 79: I think I would remove this specific `trace`. The ones below are welcome to stay, but they need a tad better contents, e.g. "Forwarding left/right data." and "Split data channel ended.".

finality-aleph/src/network/split.rs
```diff
@@ -101,8 +108,16 @@ impl<LeftData: Data, RightData: Data, R: ReceiverComponent<Split<LeftData, Right
     async fn next(&mut self) -> Option<LeftData> {
         loop {
             tokio::select! {
-                data = self.translated_receiver.next() => return data,
-                _ = forward_or_wait(&self.receiver, &self.left_sender, &self.right_sender) => (),
+                data = self.translated_receiver.next() => {
```
timorl, line 111: Eh, I would drop both the `trace`s here (and in the right receiver too ofc).

finality-aleph/src/network/split.rs
```diff
@@ -114,9 +129,18 @@ impl<LeftData: Data, RightData: Data, R: ReceiverComponent<Split<LeftData, Right
 {
     async fn next(&mut self) -> Option<RightData> {
         loop {
+            trace!(target: "aleph-network", "Right Receiver Next Loop started a next iteration");
```
timorl, line 132: Especially this one, since it's asymmetric. :P

finality-aleph/src/party/authority.rs
```diff
@@ -65,11 +63,15 @@ impl Subtasks {
     async fn stop(self) {
         // both member and aggregator are implicitly using forwarder,
         // so we should force them to exit first to avoid any panics, i.e. `send on closed channel`
-        self.member.stop().await;
+        debug!(target: "aleph-party", "Started to stop all tasks");
+        //self.member.stop().await;
+        //trace!(target: "aleph-party", "Member stopped");
```
timorl, line 68: Uhhh, uncomment these plx. And this also explains why we have these errors all over the logs, the ones you told me not to mention. xd

finality-aleph/src/party/authority.rs
```diff
@@ -78,11 +80,12 @@ impl Subtasks {
             _ = &mut self.exit => false,
             _ = self.member.stopped() => true,
             _ = self.aggregator.stopped() => true,
-            _ = self.forwarder.stopped() => true,
             _ = self.refresher.stopped() => true,
             _ = self.data_store.stopped() => true,
         };
+        debug!(target: "aleph-party", "Something died and it was unexpected: {:?}", result);
```
timorl, line 86: I would make this log conditional on `result`, usually its not that useful.

finality-aleph/src/party/member.rs
```diff
@@ -10,11 +10,11 @@ use log::debug;
 use sp_runtime::traits::Block;
 
 /// Runs the member within a single session.
-pub fn task<B: Block>(
+pub fn task<B: Block, ADN: DataNetwork<AlephNetworkData<B>> + 'static>(
     subtask_common: AuthoritySubtaskCommon,
     multikeychain: KeyBox,
     config: Config,
-    network: AlephNetwork<B>,
+    network: NetworkWrapper<B, ADN>,
```
timorl, line 17: Ugh, I don't quite remember why this cannot be `aleph_bft::Network`, could you write it out again? I remember we spent quite some time trying to make it work, but I'm not sure if the problem wasn't with some compatibility stuff, which would potentially make this doable now.

primitives/src/lib.rs
```diff
@@ -16,6 +16,7 @@ mod app {
     use sp_application_crypto::{app_crypto, ed25519};
     app_crypto!(ed25519, crate::KEY_TYPE);
 }
+pub use app::Public;
```
timorl, line 19: Huh? Where is this used?

finality-aleph/src/party/mod.rs
```diff
@@ -1,14 +1,18 @@
...
+        AlephJustification, JustificationHandler, JustificationHandlerConfig,
+        JustificationNotification, JustificationRequestDelay, SessionInfo, SessionInfoProvider,
+        Verifier,
     },
-    last_block_of_session, network,
+    last_block_of_session,
     network::{
-        split_network, AlephNetworkData, ConsensusNetwork, DataNetwork, NetworkData, SessionManager,
+        split, AlephNetworkData, ConnectionIO, ConnectionManager, Manager, RequestBlocks,
+        RmcNetworkData, Service, SessionNetwork, Split, IO,
```
timorl, line 15: I would import the generic things in here with prefixes for readability, i.e. `NetworkService` and `NetworkIO`, perhaps more.

finality-aleph/src/party/mod.rs
```diff
@@ -174,15 +174,44 @@ where
...
+        messages_from_network,
+    );
+    let connection_manager = ConnectionManager::<_, _>::new(network.clone());
+    let session_manager = Manager::new(commands_for_service, messages_for_service);
+    let network = Service::new(
+        network.clone(),
+        IO::new(messages_from_user, messages_for_user, commands_from_io),
+    );
 
+    //TODO: we should retry
```
timorl, line 197: Uh. I actually suspect if you fail here it's better to return an error from the whole function, and perhaps retry it one level above. Maybe this should be a task though, including all the other situations in which we would actually like to return an error in here. In the meantime `expect` is... almost the same?

finality-aleph/src/party/mod.rs
```diff
@@ -174,15 +174,44 @@ where
...
+    let (messages_for_user, messages_from_network) = mpsc::unbounded();
+
+    let connection_io = ConnectionIO::new(
+        commands_for_network,
+        messages_for_network,
+        commands_from_user,
+        commands_from_manager,
+        messages_from_network,
+    );
+    let connection_manager = ConnectionManager::<_, _>::new(network.clone());
```
timorl, line 190: Empty turbofishies are redundant (even if cute), plx delete.

finality-aleph/src/party/mod.rs
```diff
@@ -174,15 +174,44 @@ where
...
+    spawn_handle.spawn("aleph/network_manager", network_manager_task);
     let network_task = async move { network.run().await };
     spawn_handle.spawn("aleph/network", network_task);
 
     debug!(target: "afa", "Consensus network has started.");
 
+    let authorities = client
+        .runtime_api()
+        .authorities(&BlockId::Number(<NumberFor<B>>::saturated_from(0u32)))
+        .unwrap();
```
timorl, line 213: The comment above is related, but also at least change this to `expect`.

finality-aleph/src/party/mod.rs
```diff
@@ -174,15 +174,44 @@ where
...
+    };
+    spawn_handle.spawn("aleph/network_manager", network_manager_task);
     let network_task = async move { network.run().await };
     spawn_handle.spawn("aleph/network", network_task);
 
     debug!(target: "afa", "Consensus network has started.");
 
+    let authorities = client
+        .runtime_api()
+        .authorities(&BlockId::Number(<NumberFor<B>>::saturated_from(0u32)))
```
timorl, line 212: Should this always be `0`?

This really is a question, I'm just not that familiar with this part of the logic. <_<"

finality-aleph/src/party/mod.rs
```diff
@@ -263,9 +293,13 @@ where
     metrics: Option<Metrics<<B::Header as Header>::Hash>>,
     authority_justification_tx: mpsc::UnboundedSender<JustificationNotification<B>>,
     unit_creation_delay: UnitCreationDelay,
+    authorities: Vec<aleph_primitives::Public>,
 }
 
 const SESSION_STATUS_CHECK_PERIOD_MS: u64 = 1000;
+const SESSION_RESTART_COOLDOWN_MS: u64 = 500;
+
+type SplitData<B> = Split<AlephNetworkData<B>, RmcNetworkData<B>>;
```
timorl, line 302: Move this slightly higher, so that it at least appears before the first usage. Perhaps to the top of the file after all. ">_>

finality-aleph/src/party/mod.rs
```diff
@@ -372,17 +402,19 @@ where
...
+        let authority_pen =
             AuthorityPen::new(authorities[node_id.0].clone(), self.keystore.clone())
                 .await
-                .expect("The keys should sign successfully"),
-        );
+                .expect("The keys should sign successfully");
+
+        let keybox = KeyBox::new(node_id, authority_verifier.clone(), authority_pen.clone());
+
+        //TODO: we should retry
```
timorl, line 413: Same comment as above about perhaps returning the error, but at least remove the todo and create a task.

timorl (changes requested):
This PR is a brilliant culmination of months of effort. I thank you for your service!

Some comments regardless. :P

finality-aleph/src/party/mod.rs
```diff
@@ -196,6 +225,7 @@ where
         spawn_handle: spawn_handle.into(),
         phantom: PhantomData,
         unit_creation_delay,
+        authorities,
```
DamianStraszak, line 228: Please add a comment on what is the intended content of this field and how it changes.

finality-aleph/src/party/mod.rs
DamianStraszak: It shouldn't for sure. 

Once I familiarize myself with how is this supposed to work I will probably leave some more constructive comments.

run_nodes.sh
```diff
@@ -84,11 +89,14 @@ for i in $(seq 0 "$(( N_VALIDATORS + N_NON_VALIDATORS - 1 ))"); do
     --ws-port $((9944 + i)) \
     --port $((30334 + i)) \
     --bootnodes $bootnodes \
+    --public-addr ${addresses[i]} \
```
DamianStraszak, line 92: This looks like a red flag to me that something went wrong... Why is this necessary? And what it does internally?

timorl: When running purely locally the substrate network does not return any own addresses ever. This flag makes it at least know one such address.

We expect everything should work without this flag in any non-local context, but we will have to test this extensively ofc.

finality-aleph/src/party/mod.rs
```diff
@@ -470,7 +482,52 @@ where
...
+                let next_session_authorities = match self
+                    .client
+                    .runtime_api()
+                    .next_session_authorities(&BlockId::Number(first_block))
+                {
+                    Ok(authorities) => authorities
+                        .expect("authorities for next session must be available at first block of current session"),
+                    Err(e) => {
+                        error!(target: "afa", "Error when getting authorities for session {:?} {:?}", next_session_id, e);
+                        return;
```
DamianStraszak, line 496: This deserves a panic. This`return` can actually lead to unexpected behaviour

DamianStraszak: The idea is that, we `return` from the session if and only if the session is finalized (last block of it)

finality-aleph/src/party/member.rs
Yithis: We just agreed that having something with enormous amount of generic arguments and constraints wouldn't be that good

primitives/src/lib.rs
Yithis: `     authorities: Vec<aleph_primitives::Public>,` in `finality-aleph/src/party/mod.rs`

finality-aleph/src/party/mod.rs
Yithis: I think it is tricky, @DamianStraszak do you have any idea what should be here?

finality-aleph/src/party/mod.rs
Yithis: Done

finality-aleph/src/party/mod.rs
Yithis: Done

finality-aleph/src/party/mod.rs
Yithis: Done

finality-aleph/src/party/mod.rs
Yithis: Done

finality-aleph/src/party/authority.rs
Yithis: Done

finality-aleph/src/party/authority.rs
Yithis: Done

finality-aleph/src/network/split.rs
Yithis: Done

finality-aleph/src/network/split.rs
Yithis: Done

finality-aleph/src/network/split.rs
Yithis: Done

finality-aleph/src/network/mod.rs
Yithis: Done

finality-aleph/src/network/mod.rs
Yithis: You need to create `SessionManager` in `party/mod.rs`. Rename done

finality-aleph/src/network/manager/service.rs
Yithis: Done

finality-aleph/src/network/manager/service.rs
Yithis: You are right, done

finality-aleph/src/lib.rs
```diff
@@ -23,8 +24,7 @@ mod hash;
 mod import;
 mod justification;
 pub mod metrics;
-mod network;
-mod new_network;
+pub mod network; //todo
```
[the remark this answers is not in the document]
Yithis: Done

run_nodes.sh
Yithis: @timorl is right, can this stay or we should change something?

Yithis:
Since this PR just broke integration tests from `finality-aleph/src/testing/network.rs` and `finality-aleph/src/testing/data_io.rs` I just deleted them and created a task for it.

finality-aleph/src/party/mod.rs
Yithis: Done for `NetworkService` and `NetworkIO`. Not sure which one should also have prefix, maybe `Split`?

finality-aleph/src/party/mod.rs
Yithis: Done

finality-aleph/src/party/mod.rs
Yithis: Done

finality-aleph/src/lib.rs
Yithis: It is not that simple :crying_cat_face:

finality-aleph/src/data_io.rs
```diff
@@ -161,10 +175,15 @@ where
         let mut import_stream = self.client.import_notification_stream();
         loop {
             tokio::select! {
-                Some(message) = &mut self.messages_rx.next() => {
+                Some(message) = &mut self.network.next() => {
                     trace!(target: "afa", "Received message at Data Store {:?}", message);
                     self.add_message(message);
                 }
+                Some((message, recipient)) = &mut self.messages_from_network.next() => {
```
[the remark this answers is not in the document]
Yithis: Task created

finality-aleph/src/party/mod.rs
Yithis: I changed it a little

finality-aleph/src/party/mod.rs
Yithis: Done

finality-aleph/src/network/split.rs
timorl: Doesn't seem to be.

finality-aleph/src/network/split.rs
timorl: Doesn't seem to be.

finality-aleph/src/party/mod.rs
timorl: I think this might be correct now, but I would prefer Damian to confirm.

timorl (changes requested):
I think you missed committing some minor things, so commenting.

finality-aleph/src/party/mod.rs
```diff
@@ -483,14 +544,17 @@ where
                         None => None,
                     } } => {
                     warn!(target: "afa", "Authority task ended prematurely, restarting.");
```
DamianStraszak, line 546: When do we expect this to happen?

finality-aleph/src/party/mod.rs
```diff
@@ -470,7 +486,52 @@ where
         };
         loop {
             let last_finalized_number = self.client.info().finalized_number;
-            debug!(target: "afa", "Highest finalized: {:?} session {:?}", last_finalized_number, session_id);
+            if last_finalized_number > first_block {
```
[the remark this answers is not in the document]
Yithis: Yes. @timorl am I right it is ok?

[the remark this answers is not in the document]
timorl: The advantage is that this might catch e.g. a public address change early and improve the chances of being connected when the time comes. So I would say yes, especially since this shouldn't be a costly operation.

finality-aleph/src/party/mod.rs
Yithis: When something will die, for example due to lack of public address

finality-aleph/src/party/mod.rs
```diff
@@ -209,9 +208,12 @@ where
...
     debug!(target: "afa", "Consensus network has started.");
 
+    let last_finalized_number = client.info().finalized_number;
+    let starting_session =
+        session_id_from_block_num::<B>(last_finalized_number, session_period);
     let authorities = client
         .runtime_api()
-        .authorities(&BlockId::Number(<NumberFor<B>>::saturated_from(0u32)))
+        .authorities(&BlockId::Number(<NumberFor<B>>::saturated_from(starting_session.0)))
         .expect("We should know authorities for the first session");
```
DamianStraszak, line 218: I would much rather like this and the starting session to be computed inside `run()` and not here. I know that you have a field `.authorities` in `ConsensusParty` that tracks authorities and that's why you must do that. I find this field `.authorities` super confusing in the current version, as it really means `.next_session_authorities`, but even that not quite as it depends on the timing. So either remove this field completely (I don't know if it's doable) or replace this field by something else. For instance there is `session_authorities` already that holds authorities by `session_id`, if you cannot reuse it, just create a separate map for that. Having a map, instead of a single field is much better because you can never make a mistake in interpreting what "current authorities" or "next authorities" mean, since the numbering is absolute and not relative to the session in progress.

If you end up making a map, you should also remember to prune it, see `prune_session_data`.

DamianStraszak: And yes, even if the map must hold max 2 elements at a time, it's not a big downside compared to the clarity it adds.

Yithis: Ok, we had discussion exactly about it with @timorl few days ago XD

run_nodes.sh
DamianStraszak: If you run a node locally, even with this flag it normally logs something like:
`
2022-01-13 12:04:35.149  INFO                 main sub-libp2p: 🔍 Discovered new external address for our node: /dns4/localhost/tcp/30335/p2p/12D3KooWBhABJ8f8LcoJPMyFtJTZnE6umqgHGQmcMhVLHSjKzoVS/p2p/12D3KooWBhABJ8f8LcoJPMyFtJTZnE6umqgHGQmcMhVLHSjKzoVS 
`
wouldn't this be exactly what we need?

If I remove this flag, then it never works, or just kicks in later?

timorl (approved):
LGTM modulo the open comments.

run_nodes.sh
timorl: Hmm, you are right, this appears, in my case `2022-01-14 16:50:30.513  INFO tokio-runtime-worker sub-libp2p: 🔍 Discovered new external address for our node: /ip4/127.0.0.1/tcp/30334/p2p/12D3KooWAYdQQNDwFw1QSWcbUSLFz2Tk6t5UWkVh2FiD8RPetYFK`. Then maybe this doesn't work after all and we have to figure out why the substrate network does not want to tell us about this address. :/

finality-aleph/src/party/mod.rs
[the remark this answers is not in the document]
DamianStraszak: But the difference between calling it once, vs a few times is only that `self.session_manager.start_validator_session` will be called once or a few times. Right?

But I would definitely expect this `start_validator_session` to be idemponent.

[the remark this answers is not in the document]
timorl: It reads the currently known external addresses and peer id, these can theoretically change (at least the former).

finality-aleph/src/party/mod.rs
```diff
@@ -470,7 +472,54 @@ where
         };
         loop {
             let last_finalized_number = self.client.info().finalized_number;
-            debug!(target: "afa", "Highest finalized: {:?} session {:?}", last_finalized_number, session_id);
+            if last_finalized_number > first_block {
+                let next_session_id = SessionId(session_id.0 + 1);
+                let next_session_authorities = match self
+                    .client
+                    .runtime_api()
+                    .next_session_authorities(&BlockId::Number(first_block))
```
DamianStraszak, line 480: Can we create a new method of ConsensusParty that fetches the committee for a given **session**, updates `session_authorities` and outputs the result (this function should attempt using `next_session_authorities`  unless the session is `0`). This way we can make this code much nicer and also call this function at the start of `run_session` instead of in `run`.

finality-aleph/src/party/mod.rs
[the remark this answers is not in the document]
DamianStraszak: Also, one tiny detail: imho this should be `last_finalized_number >= first_block`. The next_authorities should be there in the first block of the new_session already.

run_nodes.sh
timorl: Welp, found it, substrate network does find this address with the `p2p` part, but afterwards returns it without. It also returns the peer id separately, so there is a somewhat aggressive fix possible, maybe I'll make a PR deeper in the evening. <_<"

finality-aleph/src/party/mod.rs
```diff
@@ -470,7 +472,54 @@ where
...
+                match get_node_index(&authorities, self.keystore.clone()).await {
+                    Some(node_id) => {
+                        let authority_pen = AuthorityPen::new(
+                            next_session_authorities[node_id.0].clone(),
+                            self.keystore.clone(),
+                        )
+                        .await
+                        .expect("The keys should sign successfully");
+
+                        if let Err(e) = self.session_manager.start_validator_session(
```
DamianStraszak, line 501: I would expect it to be necessary to call `start_validator_session` for session `K` in order to successfully finalize session `K`, but it seems to me that it is never called for the `starting_session` (the first session this Consensus Party handles). Is it correct?

Yithis: It is called once more, inside `spawn_authority_task`, when a given session is starting. So it is called

run_nodes.sh
timorl: This flag can (and probably should) be removed now.

finality-aleph/src/party/mod.rs
timorl: I'm assuming the discussion about this is happening somewhere else now, so resolving.

finality-aleph/src/party/mod.rs
```diff
@@ -293,6 +293,33 @@ where
...
+) -> Vec<AuthorityId>
+where
+    B: Block,
+    C: sp_api::ProvideRuntimeApi<B>,
+    C::Api: aleph_primitives::AlephSessionApi<B>,
+{
+    if session_id == SessionId(0) {
+        runtime_api
+            .authorities(&BlockId::Number(<NumberFor<B>>::saturated_from(0u32)))
+            .unwrap()
```
timorl, line 309: Should also be an `expect`, even if it's extremely unlikely to fail.

finality-aleph/src/party/mod.rs
```diff
@@ -432,40 +459,32 @@ where
...
+    ) -> Option<Vec<AuthorityId>> {
+        let last_finalized_number = self.client.info().finalized_number;
+        let first_block = first_block_of_session::<B>(session_id, self.session_period);
+        if last_finalized_number < first_block {
+            return None;
         }
+
+        Some(
+            self.session_authorities
+                .lock()
```
timorl, line 474: Shouldn't this be the `await`y lock?

timorl (approved):
Still fine, but I managed to find some things I can complain about. <_<"

finality-aleph/src/party/mod.rs
Yithis: https://github.com/Cardinal-Cryptography/aleph-node/pull/276/files#diff-1abd0b64c271c22b933a2edd8813eaa74ef220d1ff5d8b8f1c36ccf2689e1fdeL436

This line suggesting that there shouldn't be, but if you think differently I will change it

Yithis: Done

finality-aleph/src/party/mod.rs
Yithis: Done

finality-aleph/src/party/mod.rs
timorl: Link doesn't point to a line. :/ Or just doesn't work.

DamianStraszak (approved):
I'm in favor of merging (after fixing clippy etc.), although it still requires extensive testing.

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -169,7 +169,7 @@ jobs:
 
       - name: Run e2e test suite
         shell: bash
-        timeout-minutes: 10
+        timeout-minutes: 20
```
Marcin-Radecki, line 172: Sorry for later commit, but not sure if would help anything - from my (brief) experience if we stucked even for one minute, waiting 19 more would not change anything - there is some problem with e2e tests that if we run them for the second time on the **same** set of aleph-node instances in docker (ie without having fresh instances every time) then we sometimes end up stuck. 

I would say we'd need to revise as separate task what happens in e2e suite so we can run them repeatedly on sme instances of aleph-node containers.

PR #278, opened 2022-01-17
Az 413 users may send transactions in batches

Marcin-Radecki:
See JIRA description for details.

### DIFF
--- a/e2e-tests/src/test/utility.rs
+++ b/e2e-tests/src/test/utility.rs
@@ -3,7 +3,6 @@ use log::info;
 use substrate_api_client::{compose_call, compose_extrinsic, GenericAddress, XtStatus};
 use sp_core::Pair;
 use crate::transfer::{setup_for_transfer};
-use std::fmt::Write;
 
 use codec::{Compact};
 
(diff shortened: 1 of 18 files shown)


e2e-tests/src/test/utility.rs
```diff
@@ -0,0 +1,53 @@
...
+        transactions
+    );
+
+    let finalized_block_hash = connection
+        .send_extrinsic(extrinsic.hex_encode(), XtStatus::Finalized)
+        .expect("Could not send extrinsc")
+        .expect("Could not get tx hash");
+    info!("[+] A batch of {} transactions was included in {} block.", NUMBER_OF_TRANSACTIONS, finalized_block_hash);
+
+    let block_with_batch_tx = connection.get_block::<aleph_runtime::Block>(Some(finalized_block_hash)).
```
fbielejec, line 42: YMMV, but not sure if it's worth pulling WASM, and entire runtime just to have the Block. There should be some generic Block definition in `sp_runtime::{generic}` ?

Marcin-Radecki: Yeah, this is heavy I agree to add it. Let me explore one more time if I can find some other solution which does not involve including whole runtime.

Marcin-Radecki: I've reverted dependency on `aleph-runtime` in `e2e-tests` package. Decision was made on the fact build of e2e workflow varied from 45 to 50 mins, not exactly sure it was due to that change but still a long time. Test is good enough since `send_extrinsic` is a blocking call, and it waits for a finalized block in chain with that transaction.

fbielejec (approved):

maciejnems (approved):

PR #281, opened 2022-01-20
AZ-455 use ComponentNetwork in data_io

Marcin-Radecki:
This is a small refactor simplifying `select!` in `DataStore::run`.

### DIFF
--- a/finality-aleph/src/data_io.rs
+++ b/finality-aleph/src/data_io.rs
@@ -1,20 +1,19 @@
 use crate::{
     metrics::Checkpoint,
-    network::{DataNetwork, RequestBlocks, SimpleNetwork},
+    network::{DataNetwork, RequestBlocks, SimpleNetwork, ComponentNetwork, ReceiverComponent},
     Metrics,
 };
-use aleph_bft::Recipient;
 use async_trait::async_trait;
 use codec::{Decode, Encode};
 use futures::{
     channel::{
-        mpsc::{self, UnboundedReceiver, UnboundedSender},
+        mpsc::{self, UnboundedSender},
         oneshot,
     },
     StreamExt,
 };
 use futures_timer::Delay;
-use log::{debug, error, trace, warn};
+use log::{debug, error, trace};
 use lru::LruCache;
 use parking_lot::Mutex;
 use sc_client_api::backend::Backend;
@@ -98,7 +97,7 @@ impl Default for DataStoreConfig {
 
 /// This component is used for filtering available data for Aleph Network.
 /// It needs to be started by calling the run method.
-pub struct DataStore<B, C, BE, RB, Message, N>
+pub struct DataStore<B, C, BE, RB, Message, R>
 where
     B: BlockT,
     C: crate::ClientForAleph<B, BE> + Send + Sync + 'static,
@@ -106,7 +105,7 @@ where
     RB: RequestBlocks<B> + 'static,
     Message:
         AlephNetworkMessage<B> + std::fmt::Debug + Send + Sync + Clone + codec::Codec + 'static,
-    N: DataNetwork<Message>,
+    R: ReceiverComponent<Message>,
 {
     next_message_id: MessageId,
     missing_blocks: HashMap<AlephDataFor<B>, MissingBlockInfo>,
@@ -117,12 +116,11 @@ where
     block_requester: RB,
     config: DataStoreConfig,
     _phantom: PhantomData<BE>,
-    network: N,
+    messages_from_network: Arc<tokio::sync::Mutex<R>>,
     messages_for_network: UnboundedSender<Message>,
-    messages_from_network: UnboundedReceiver<(Message, Recipient)>,
 }
 
-impl<B, C, BE, RB, Message, N> DataStore<B, C, BE, RB, Message, N>
+impl<B, C, BE, RB, Message, R> DataStore<B, C, BE, RB, Message, R>
 where
     B: BlockT,
     C: crate::ClientForAleph<B, BE> + Send + Sync + 'static,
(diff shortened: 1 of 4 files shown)


finality-aleph/src/data_io.rs
```diff
@@ -117,30 +116,30 @@ where
...
         client: Arc<C>,
         block_requester: RB,
         config: DataStoreConfig,
-        network: N,
+        component_network: N,
     ) -> (Self, impl DataNetwork<Message>) {
         let (messages_for_network, messages_from_data_store) = mpsc::unbounded();
-        let (messages_to_data_store, messages_from_network) = mpsc::unbounded();
+        let messages_to_data_store = component_network.sender().clone();
+        let messages_from_network = component_network.receiver();
```
timorl, line 142: The names here are a bit confusing. I would say either use "network" to refer to the aleph network and, uh, "world"(?) to refer to the interface with the world, or use "aleph" to refer to the aleph network and "network" to refer to the network.

After I wrote this maybe the actual answer is using "aleph" and "world"... Anyway the channels would then be called (in the order they appear in the three lines above this comment):
1. `messages_for_aleph`
2. `messages_from_data_store`
3. `messages_for_world`
4. `messages_from_world`
If you pick any other of the naming schemes then change the names accordingly ofc. I just don't like using "network" in two meanings here, and implying that the messages that are going straight through the sender are "to data store".

timorl (approved):
Just one extensive nit about naming, other than that nice.

finality-aleph/src/data_io.rs
```diff
@@ -117,30 +116,30 @@ where
     block_requester: RB,
     config: DataStoreConfig,
     _phantom: PhantomData<BE>,
-    network: N,
+    messages_from_network: Arc<tokio::sync::Mutex<R>>,
```
Yithis, line 119: Nit: Please import Mutex, rather than right in this way

finality-aleph/src/data_io.rs
```diff
@@ -178,15 +176,13 @@ where
...
             tokio::select! {
-                Some(message) = &mut self.network.next() => {
+                Some(message) = async {
+                    let mut lock = self.messages_from_network.lock().await;
+                    lock.next().await
+                } => {
                     trace!(target: "afa", "Received message at Data Store {:?}", message);
                     self.add_message(message);
                 }
-                Some((message, recipient)) = &mut self.messages_from_network.next() => {
```
Yithis, line 185: Update comment above the function

Yithis (approved):
Great job! Two small comments

finality-aleph/src/data_io.rs
```diff
@@ -178,15 +176,13 @@ where
         let mut import_stream = self.client.import_notification_stream();
         loop {
             tokio::select! {
-                Some(message) = &mut self.network.next() => {
+                Some(message) = async {
+                    let mut lock = self.messages_from_network.lock().await;
+                    lock.next().await
```
deuszx, line 181: The component is `Arc`, so it can be shared in multiple places, which means that there may be another place in the code that calls `lock.next()` - that's a race condition for the message.

Marcin-Radecki: Isn't it that indeed many threads can run line 180 at the same time, but only one will be granted lock so line 181 is in critical section?

finality-aleph/src/data_io.rs
Marcin-Radecki: Good catch, will do

finality-aleph/src/data_io.rs
Marcin-Radecki: There's another `Mutex` name imported in this file (from `parking_lot`) so either I'd need to rename that entry or I can import tokio Mutex as sth like `TokioMutex`, and I think latter is less readable than current version. For the moment I'll leave it as it is then, I understand there's a convention to import and use first-level-names mostly, but on the other hand we do use symbols like `std::fmt::Debug` even in this file.

finality-aleph/src/data_io.rs
Marcin-Radecki: I like convention to use `aleph`for filtered network and just use `network` to refer to non-filtered one. Will align code to this,

finality-aleph/src/data_io.rs
timorl: Blah, we really need to refactor this... Also remove the `parking_lot` mutexes, they are dangerous in async code.

Not in this PR ofc, the specific changes require more though.

finality-aleph/src/data_io.rs
deuszx: I meant that you pass in the `ComponentNetwork` instance to the constructor, and create an instance of `recipient` (`messages_from_network`). If that same instance of `ComponentNetwork` is passed somewhere else, that place also can call `component_network.receiver()` and create an instance of a `receiver` which will compete for messages with yours.

Marcin-Radecki: But `component_network` is passed with ownership  to `DataStore::new`, hence there cannot be other place in the code that uses `component_network` reference after that. It must have been `component_network.clone()`, assuming `ComponentNetwork` can be `clone()`d, and it that case we should be good?

timorl: Actually you can have multiple `Arc`es of the receiver, since you could've gotten some before handing away the `ComponentNetwork`. But they receiver itself is behind a `Mutex` within the `Arc`, so cloning the `Arc` does not matter. In fact this is the usual way in which you use mutexes in Rust.

deuszx: > But they receiver itself is behind a `Mutex` within the `Arc`, so cloning the `Arc` does not matter.

But that `Mutex` will be dropped at some point and the other thread could acquire an access to it.

timorl: Ohh, I see, you mean some other thread might get a specific message instead of this one? Yeah, that is correct, depending on what we are doing this might even be intended (see `network/split.rs` for an example). In this case we don't want that to happen, but we have to ensure that ourselves.

deuszx: Yes, that's exactly what I meant :)

Yithis (approved):

PR #283, opened 2022-01-21
Make the validator peerset more restrictive

timorl:
Also remove some ancient print debugging that somehow slipped by
reviewers. <_<"

### DIFF
--- a/finality-aleph/src/lib.rs
+++ b/finality-aleph/src/lib.rs
@@ -54,10 +54,10 @@ pub fn peers_set_config(protocol: Protocol) -> sc_network::config::NonDefaultSet
 
     config.set_config = match protocol {
         Protocol::Validator => sc_network::config::SetConfig {
-            in_peers: 25,
+            in_peers: 0,
             out_peers: 0,
             reserved_nodes: Vec::new(),
-            non_reserved_mode: sc_network::config::NonReservedPeerMode::Accept,
+            non_reserved_mode: sc_network::config::NonReservedPeerMode::Deny,
         },
         Protocol::Generic => sc_network::config::SetConfig::default(),
     };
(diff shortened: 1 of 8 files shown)


finality-aleph/src/lib.rs
```diff
@@ -54,10 +54,10 @@ pub fn peers_set_config(protocol: Protocol) -> sc_network::config::NonDefaultSet
 
     config.set_config = match protocol {
         Protocol::Validator => sc_network::config::SetConfig {
-            in_peers: 25,
+            in_peers: 0,
             out_peers: 0,
             reserved_nodes: Vec::new(),
-            non_reserved_mode: sc_network::config::NonReservedPeerMode::Accept,
+            non_reserved_mode: sc_network::config::NonReservedPeerMode::Deny,
         },
```
kostekIV, line 61: would be cool to add a comment here on what it does. At first glance, it looks like we don't allow anyone to connect to us at all.

kostekIV (approved):

finality-aleph/src/lib.rs
timorl: And that is absolutely correct. We only connect based on authentications we get through the other protocol, which we use to make peers reserved.

timorl:
Whoopsie, this might not actually work, I just noticed the way I tested this might have masked any issues.

timorl:
Nvm this actually works as expected the python scripts/AWS were just acting up.

Yithis (approved):
Looks good to me

kostekIV (approved):

finality-aleph/src/lib.rs
timorl: Added comment.

PR #285, opened 2022-01-25
AZ-453: Add Data Store tests

maciejnems:


finality-aleph/src/testing/data_io.rs
```diff
@@ -0,0 +1,385 @@
...
+            mut network,
+            block_requests_rx: _,
+            network_tx,
+            exit_data_store_tx,
+        },
+    ) = prepare_data_store();
+    let data_store_handle = tokio::spawn(task_handle);
+
+    // Import 4 blocks
+    let blocks = import_blocks(&mut client, 4, false).await;
```
Marcin-Radecki, line 317: I can see those comments were now present in the previous file version, hence wanted to ask if they are required? E.g. this one reader can see clearly that we import 4 blocks here (ie name of function `import blocks` fits behavior here). Not really a big deal, but wanted to clarify whether there are any testing code convention I'm not aware of.

This is a general rule I learned in the Robert C Martin videos in the past - if I need the comment I usually extract method so method name says what the code does etc., instead of putting comment.

finality-aleph/src/testing/data_io.rs
```diff
@@ -0,0 +1,385 @@
...
+        .build()
+        .unwrap()
+        .block;
+    let data = AlephData::new(block.header.hash(), block.header.number);
+
+    // Send new block
+    network_tx.unbounded_send(vec![data]).unwrap();
+
+    // New block should be requested
+    let requested_block = block_requests_rx.next().await.unwrap();
```
Marcin-Radecki, line 380: Nit: I'm comparing previous version to this one, and IMHO `expect` is more readable - this clearly states intention to a reader and on top of that in the possible scenario that this test fails it states it in stacktrace. 
```rust
 let requested_block = block_requests_rx
        .next()
        .await
        .expect("Block should be requested");
```

finality-aleph/src/testing/data_io.rs
Marcin-Radecki: This comment applies to other places as well, but leaving it into Author's consideration :)

Marcin-Radecki (approved):
Change looks good, just minor comments on naming.

Good to have that tests back :)

finality-aleph/src/testing/data_io.rs
```diff
@@ -0,0 +1,385 @@
...
+    fn sender(&self) -> &Self::S {
+        &self.sender_tx
+    }
+
+    fn receiver(&self) -> Arc<Mutex<Self::R>> {
+        self.next_rx.clone()
+    }
+}
+
+struct DataStoreChannels {
```
timorl, line 114: Suggestion – rename this to `DataStoreIO`, remove the `network` field and add methods like `block_request()`, `send_data(data)`, and `exit()`. This should make the code below more readable (you are always calling methods on the `io` object when you want to interact with the data store), remove the `Box<dyn ...>` (just return `impl DataNetwork<...>` from the function), and remove the unpacking boilerplate from the test cases.

timorl: On the other hand maybe `TestClient` should be included here and `import_blocks` could be a method too. There would need to be some block building methods too, but I'm pretty sure they should be functions anyway (see some comments below), so that is only an advantage.

finality-aleph/src/testing/data_io.rs
```diff
@@ -0,0 +1,385 @@
...
+        DataStoreChannels {
+            mut network,
+            block_requests_rx: _,
+            network_tx,
+            exit_data_store_tx,
+        },
+    ) = prepare_data_store();
+    let data_store_handle = tokio::spawn(task_handle);
+
+    // Build block
```
timorl, line 231: You repeat this quite a lot, maybe abstract it away into a function.

finality-aleph/src/testing/data_io.rs
```diff
@@ -0,0 +1,385 @@
...
+}
+
+#[tokio::test]
+async fn sends_messages_with_imported_blocks() {
+    let (
+        task_handle,
+        mut client,
+        DataStoreChannels {
+            mut network,
+            block_requests_rx: _,
```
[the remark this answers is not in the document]
timorl: Unless you follow the suggesstion above ofc.

timorl (changes requested):
Very nice tests, mostly cosmetic comments from me. The one about the timeout is pretty important though.

finality-aleph/src/testing/data_io.rs
maciejnems: The comment about importing is definitely unnecessary. I added all those comments, do describe the scenarios tested as code might be a little bit convolutet in some places. Tbh the names of tests are descriptive enough, so I can just remove them.

finality-aleph/src/testing/data_io.rs
maciejnems: Well, the thing is that when the block is not requested there will be no error and the test will just run forever. The only case when this except prints something is when all sending endpoints are dropped. Which is very unlikely to happen. That's why I removed it.

Marcin-Radecki: Right, I just mean that this snippet:
```rust
// New block should be requested 
let requested_block = block_requests_rx.next().await.unwrap();
```
or this one-liner
```rust
 let requested_block = block_requests_rx.next().await.expect("Block should be requested");
```

I would say the bottom is more readable to a reader, without even knowing details of code around, since we state our intent explicitly what an exceptional scenario is (and we don't need comment now).

finality-aleph/src/testing/data_io.rs
maciejnems: Done

finality-aleph/src/testing/data_io.rs
maciejnems: Done. I moved all this into one struct and named it `TestHandler`. It has all necessary methods to carry out the tests.

finality-aleph/src/testing/data_io.rs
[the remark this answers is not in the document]
maciejnems: Suggestion above/below.

finality-aleph/src/testing/data_io.rs
```diff
@@ -0,0 +1,385 @@
...
+            exit_data_store_tx,
+        },
+    ) = prepare_data_store();
+    let data_store_handle = tokio::spawn(task_handle);
+
+    let blocks = import_blocks(&mut client, 4, false).await;
+
+    network_tx.unbounded_send(blocks.clone()).unwrap();
+
+    let message = network.next().await.unwrap();
```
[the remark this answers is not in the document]
maciejnems: I slapped a timeout on all receives, The timeout is quite long.

finality-aleph/src/testing/data_io.rs
```diff
@@ -0,0 +1,385 @@
...
+}
+
+#[derive(Clone)]
+struct TestNetwork {
+    sender_tx: UnboundedSender<(TestData, Recipient)>,
+    next_rx: Arc<Mutex<UnboundedReceiver<TestData>>>,
+}
+
+impl TestNetwork {
+    fn new() -> (
```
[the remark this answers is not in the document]
maciejnems: Removed `TestNetwork`

finality-aleph/src/testing/data_io.rs
maciejnems: Not that there is timeout there is a panic in case receive fails.

finality-aleph/src/testing/data_io.rs
maciejnems: Removed all comments as now everything is less spaghettier code.

finality-aleph/src/testing/data_io.rs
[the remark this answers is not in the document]
timorl: Might be better to do it inside the `TestHandler` methods, preferably following @Marcin-Radecki 's suggestions to use `expect` rather than `unwrap` – then you can panic there with a message appropriate either to a timeout or an empty response.

timorl (approved):
One tiny comment left from the previous review, but in general beautiful. <3

finality-aleph/src/testing/data_io.rs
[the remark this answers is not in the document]
maciejnems: I'm not sure about that. Would you like me to move network to `Test`Handler`?  And right now there is no unwrap.

Marcin-Radecki (approved):
Rework looks good.

finality-aleph/src/testing/data_io.rs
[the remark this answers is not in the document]
maciejnems: done

timorl (approved):

PR #286, opened 2022-01-26
substarte++

fbielejec:


bin/runtime/src/lib.rs
```diff
@@ -118,6 +119,9 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     impl_version: 1,
     apis: RUNTIME_API_VERSIONS,
     transaction_version: 1,
+    /// Version of the state implementation used by this runtime.
+    /// Use of an incorrect version is consensus breaking.    
+    state_version: 1,
```
Marcin-Radecki, line 124: We can spawn a task for it, but I guess update procedure to testnet/mainnet needs to be careful thought.

bin/runtime/src/lib.rs
```diff
@@ -470,6 +494,8 @@ pub const TREASURY_BURN: u32 = 0;
 pub const TREASURY_PROPOSAL_BOND: u32 = 0;
 // The proposer should deposit max{`TREASURY_PROPOSAL_BOND`% of the proposal value, $10}.
 pub const TREASURY_MINIMUM_BOND: Balance = 1000 * CENTS;
+// NOTE: I pulled this value from the air
```
h4nsu, line 497: Maybe add here that this value needs to be understood and possibly adjusted before landing on {test,main}net? Probably with some caps and exclamation marks ;)

bin/runtime/src/lib.rs
```diff
@@ -343,7 +349,25 @@ impl pallet_scheduler::Config for Runtime {
...
+
+// impl pallet_preimage::Config for Runtime {
+//     type WeightInfo = pallet_preimage::weights::SubstrateWeight<Runtime>;
+//     type Event = Event;
+//     type Currency = Balances;
+//     type ManagerOrigin = EnsureRoot<AccountId>;
+//     type MaxSize = PreimageMaxSize;
+//     type BaseDeposit = PreimageBaseDeposit;
+//     type ByteDeposit = PreimageByteDeposit;
+// }
```
h4nsu, line 370: Kick out?

h4nsu (approved):
LGTM. Comments I left are about comments, feel free to ignore and resolve

bin/runtime/src/lib.rs
fbielejec: done

PR #287, opened 2022-01-26
Fix discovery when number of validators is a multiple of 3

timorl:


finality-aleph/src/network/manager/discovery.rs
```diff
@@ -348,6 +352,38 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn broadcasts_when_clueless_modulo_3() {
+        let (mut discovery, mut handlers, _) = build_number(3).await;
```
kostekIV, line 357: could check here also other multiplies of 3 (just 2 or 3 would be enough)

finality-aleph/src/network/manager/discovery.rs
```diff
@@ -348,6 +352,38 @@ mod tests {
...
+            message,
+            (
+                DiscoveryMessage::AuthenticationBroadcast(handler.authentication().unwrap()),
+                DataCommand::Broadcast
+            )
+        );
+    }
+
+    #[tokio::test]
+    async fn broadcasts_when_clueless_just_2() {
```
kostekIV, line 372: Pls add one more test when it is not multiple of 3 and is bigger than 3

finality-aleph/src/network/manager/discovery.rs
```diff
@@ -92,7 +92,7 @@ impl Discovery {
 
     fn should_broadcast(missing_authorities_num: usize, total_node_count: NodeCount) -> bool {
         // If we are not sure we know of at least one honest node.
```
kostekIV, line 94: Pls add explanation to the formula used here (with assumptions etc) so we will not change it by accident later

finality-aleph/src/network/manager/discovery.rs
timorl: Refactored to check more cases.

finality-aleph/src/network/manager/discovery.rs
timorl: As above, although this already existed.

finality-aleph/src/network/manager/discovery.rs
timorl: Added long comment and made the formula actually BFT.

DamianStraszak (approved):

kostekIV (approved):

PR #288, opened 2022-01-27
AZ-338: Allow archivists to get finalization more often and start without "validator" flag

Yithis:


finality-aleph/src/justification/mod.rs
```diff
@@ -113,73 +102,22 @@ where
...
-            }
+            block_requester: BlockRequester::new(
+                block_requester,
+                client,
+                finalizer,
+                justification_request_delay,
+                metrics,
+            ),
+            verifier_timeout,
+            notification_timeout,
```
timorl, line 120: I think the `Config` struct might have been useful here after all, although not quite in the way it was used before. You could pass all the things that are now `const`ants (in addition to the already configurable things you pass now) and only make them `Default`s in an implementation. You would need to deconstruct the `Config` struct  in the constructor and pass some of them to the requester, and keep the rest.

In this setup I would only pass more or less numeric values in the config – `JustificationRequestDelay` and metrics should be passed as separate arguments.

finality-aleph/src/justification/requester.rs
```diff
@@ -0,0 +1,150 @@
...
+
+        let num = if num > self.client.info().best_number
+            && self.client.info().best_number > MIN_ALLOWED_DELAY.into()
+        {
+            self.client.info().best_number - MIN_ALLOWED_DELAY.into()
+        } else {
+            num
+        };
+
+        if self.justification_request_delay.can_request_now() {
```
timorl, line 135: The comment higher up in this file sorta overrides this problem but worth mentioning – note that by calling `maintain_requests` outside of this `if` you are increasing the attempt counter even without attempts.

timorl (changes requested):
The cleanup logic added here competes with another piece of code over control over justification requests, making reasoning about the behaviour tricky. Sorry for not noticing this issue earlier when we talked. :/

Other than that (big) comment mostly some cosmetic stuff.

I really like the splitting up of `justification.rs` into multiple files tho.

finality-aleph/src/justification/backward.rs
```diff
@@ -0,0 +1,44 @@
+use crate::{
```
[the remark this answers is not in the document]
DamianStraszak: I'm afraid this needs to stay. When a new node enters the network it will get justifications of "ancient" blocks in the `V1` format, hence we need that.

finality-aleph/src/network/mod.rs
```diff
@@ -138,6 +138,9 @@ pub trait RequestBlocks<B: Block>: Clone + Send + Sync + 'static {
 
     /// Request the given block -- this is supposed to be used only for "old forks".
     fn request_stale_block(&self, hash: B::Hash, number: NumberFor<B>);
+
+    /// Clear all pending justification requests.
```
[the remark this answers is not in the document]
Yithis: Done

finality-aleph/src/justification/requester.rs
```diff
@@ -0,0 +1,150 @@
...
+};
+use aleph_primitives::ALEPH_ENGINE_ID;
+use codec::Encode;
+use log::{debug, error, warn};
+use sc_client_api::HeaderBackend;
+use sp_api::{BlockId, BlockT, NumberFor};
+use sp_runtime::traits::Header;
+use std::{marker::PhantomData, sync::Arc, time::Instant};
+
+pub(crate) struct BlockRequester<B, RB, C, D, F, V>
```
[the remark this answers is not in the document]
Yithis: Done

finality-aleph/src/justification/mod.rs
```diff
@@ -91,11 +80,11 @@ where
...
-    block_requester: RB,
-    client: Arc<C>,
-    finalizer: F,
-    config: JustificationHandlerConfig<B, D>,
-    phantom: PhantomData<V>,
+    block_requester: BlockRequester<B, RB, C, D, F, V>,
+    /// How long should we wait when the session verifier is not yet available.
+    pub(crate) verifier_timeout: Duration,
+    /// How long should we wait for any notification.
+    pub(crate) notification_timeout: Duration,
```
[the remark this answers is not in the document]
Yithis: Done

finality-aleph/src/justification/requester.rs
Yithis: Done

finality-aleph/src/justification/mod.rs
```diff
@@ -69,111 +58,46 @@ pub struct JustificationNotification<Block: BlockT> {
...
+            verifier_timeout: Duration::from_millis(500),
+            notification_timeout: Duration::from_millis(1000),
+            max_attemps: 5,
+            min_allowed_delay: 3u32.into(),
+        }
+    }
+}
+
+impl<B: BlockT> JustificationHandlerConfig<B> {
+    #[allow(dead_code)] //for tests purpose
```
timorl, line 85: Then mark it as test-only code.

finality-aleph/src/justification/requester.rs
```diff
@@ -78,73 +75,66 @@ where
...
-            self.attempt = 0;
-        }
-    }
+    pub fn request_justification(&mut self, num: NumberFor<B>) {
+        match self.justification_request_scheduler.schedule_action() {
+            SchedulerActions::Request => {
+                let num = if num > self.client.info().best_number
+                    && self.client.info().best_number > self.min_allowed_delay.into()
+                {
+                    self.client.info().best_number - self.min_allowed_delay.into()
```
timorl, line 113: The only uses of `min_allowed_delay` call `into` on it – you should probably define it as being of the appropriate type in the first place?

finality-aleph/src/justification/scheduler.rs
```diff
@@ -0,0 +1,89 @@
...
+
+impl JustificationRequestScheduler for JustificationRequestSchedulerImpl {
+    fn schedule_action(&mut self) -> SchedulerActions {
+        let now = Instant::now();
+        if self.enough_time_elapsed() {
+            if now - self.last_finalization_time > now - self.last_request_time {
+                self.attempt += 1;
+
+                if self.attempt == self.max_attemps
+                    && now - self.last_finalization_time > self.max_attemps * self.delay
```
timorl, line 67: I think I get the logic here, but seems very convoluted. Maybe you could just clear `self.attempt` in the `on_block_finalized` method? The whole logic here would be "if attempts too high clear". with no need for all the timing conditions.

finality-aleph/src/party/mod.rs
```diff
@@ -160,18 +120,20 @@ where
...
     let handler = JustificationHandler::new(
         get_session_info_provider(session_authorities.clone(), session_period),
         block_requester.clone(),
         client.clone(),
         AlephFinalizer::new(client.clone()),
-        JustificationRequestDelayImpl::new(&session_period, &millisecs_per_block),
+        JustificationRequestSchedulerImpl::new(
+            &session_period,
+            &millisecs_per_block,
+            justification_handler_config.max_attemps(),
```
timorl, line 133: Uhh, this is really strange. It might be better to construct this struct internally and pass only the config? This will be strange with testing though... maybe don't make `max_attempts` part of the config?

timorl (changes requested):

finality-aleph/src/justification/mod.rs
Yithis: Done

finality-aleph/src/justification/requester.rs
Yithis: Actually it is appropriate type, I just forgot to delete `into` s

Yithis: Anyway - Done.

finality-aleph/src/party/mod.rs
Yithis: I move it, but now there is a random constant in `party/mod.rs`. Neither of them is good for me, but I don't know how to make it better :disappointed:

finality-aleph/src/justification/scheduler.rs
Yithis: Then you would end up with clearing a queue with every call of `schedule_action` after `self.attempt` reached some level. I'm not sure if I would like to do such logic and I don't know how to make current less convoluted. Maybe some comments would help?

timorl: Ah, I meant this would be "attempts since last finalization or queue clearing", so you also reset the attempts when cleaning the queue ofc. That would work, wouldn't it?

finality-aleph/src/party/mod.rs
```diff
@@ -157,20 +124,20 @@ where
     let session_authorities = Arc::new(Mutex::new(HashMap::new()));
     let block_requester = network.clone();
 
+    let justification_handler_config: JustificationHandlerConfig<_> = Default::default();
```
timorl, line 127: The typing here shouldn't be required anymore.

timorl (approved):
Only one nit and the unresolved comment left. Seems cool otherwise. :3

finality-aleph/src/party/mod.rs
Yithis: Done

finality-aleph/src/justification/compatibility.rs
```diff
@@ -0,0 +1,44 @@
...
+            signature: just_drop_id,
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum JustificationDecoding {
+    V1(AlephJustificationV1),
+    V2(AlephJustification),
+    Err,
```
fbielejec, line 33: could we make the Err variant a bit more descriptive (UnknownJustificationFormatError or something to that effect?)

finality-aleph/src/justification/compatibility.rs
```diff
@@ -0,0 +1,44 @@
...
+use codec::{Decode, DecodeAll, Encode};
+
+/// Old format of justifications, needed for backwards compatibility.
+#[derive(Clone, Encode, Decode, Debug, PartialEq)]
+pub struct AlephJustificationV1 {
+    pub signature: SignatureSet<SignatureV1>,
+}
+
+impl From<AlephJustificationV1> for AlephJustification {
+    fn from(just_v1: AlephJustificationV1) -> AlephJustification {
```
fbielejec, line 15: nit: just call the argument `justification`, type annotation is self-explanatory?

finality-aleph/src/justification/requester.rs
```diff
@@ -0,0 +1,140 @@
...
+        min_allowed_delay: NumberFor<B>,
+    ) -> Self {
+        BlockRequester {
+            block_requester,
+            client,
+            finalizer,
+            justification_request_scheduler,
+            metrics,
+            min_allowed_delay,
+            phantom: PhantomData,
```
fbielejec, line 60: _phantom ?

finality-aleph/src/justification/scheduler.rs
```diff
@@ -0,0 +1,85 @@
...
+impl JustificationRequestSchedulerImpl {
+    pub fn new(
+        session_period: &SessionPeriod,
+        millisecs_per_block: &MillisecsPerBlock,
+        max_attemps: u32,
+    ) -> Self {
+        Self {
+            last_request_time: Instant::now(),
+            last_finalization_time: Instant::now(),
+            delay: Duration::from_millis(min(
```
fbielejec, line 40: a comment for why these constants would be nice here

finality-aleph/src/justification/scheduler.rs
```diff
@@ -0,0 +1,85 @@
...
+                millisecs_per_block.0 * 2,
+                millisecs_per_block.0 * session_period.0 as u64 / 10,
+            )),
+            attempt: 0,
+            max_attemps,
+        }
+    }
+}
+
+impl JustificationRequestSchedulerImpl {
```
fbielejec, line 50: nit: this could be in the `impl` block above?

finality-aleph/src/testing/data_io.rs
```diff
@@ -60,6 +60,10 @@ impl<B: BlockT> RequestBlocks<B> for TestBlockRequester<B> {
             .unbounded_send(AlephData { hash, number })
             .unwrap();
     }
+
+    fn clear_justification_requests(&self) {
+        panic!("`clear_justification_requests` not implemented!")
```
fbielejec, line 65: there's already `unimplemented!()` macro?

finality-aleph/src/testing/justification.rs
```diff
@@ -167,18 +166,20 @@ fn prepare_env(
     let info_provider = SessionInfoProviderImpl::new(SESSION_PERIOD, verification_policy);
     let finalizer = MockedBlockFinalizer::new();
     let requester = MockedBlockRequester::new();
-    let config = JustificationHandlerConfig::new(request_policy);
-    let jrd = config.justification_request_delay.clone();
+    let config = JustificationHandlerConfig::test();
+    let jrs = JustificationRequestSchedulerImpl::new(request_policy);
```
fbielejec, line 170: nit, but not a fan of these abbreviated names, they are hard to read / follow - imho code is for humans, bytes are cheap - just spell it out verbose `justification_request_scheduler` etc (it all ends up as bytecode in the end)

finality-aleph/src/testing/mocks/block_request.rs
```diff
@@ -39,4 +39,9 @@ impl RequestBlocks<TBlock> for MockedBlockRequester {
     fn request_stale_block(&self, _hash: THash, _number: TNumber) {
         panic!("`request_stale_block` not implemented!")
     }
+
+    /// Clear all pending justification requests.
+    fn clear_justification_requests(&self) {
+        panic!("`clear_justification_requests` not implemented!")
```
fbielejec, line 45: `unimplemented!()` ?

Yithis: I know, but I wanted to be consistent with rest of the repo (we never use this macro and write this in this way)

finality-aleph/src/testing/data_io.rs
Yithis: I know, but I wanted to be consistent with rest of the repo (we never use this macro and write this in this way)

finality-aleph/src/testing/justification.rs
Yithis: Done

finality-aleph/src/justification/scheduler.rs
Yithis: Done

finality-aleph/src/justification/requester.rs
Yithis: Done

finality-aleph/src/justification/scheduler.rs
Yithis: Those are values that @DamianStraszak come out with a few months ago, I didn't change anything here. In other words I don't really know why these constants were chosen xd

finality-aleph/src/justification/compatibility.rs
Yithis: I don't really know what to put there, it can really be either unknown format or error while decoding. The name `JustificationDecoding::Err` is enough descriptive in my opinion

finality-aleph/src/justification/compatibility.rs
Yithis: Done

fbielejec (approved):

finality-aleph/src/justification/scheduler.rs
DamianStraszak: There is not much I can say :) You must have some delays here so that
1) you don't check too often
2) you check often enough so that the latency for non-validators is small.

Checking every `2` blocks sounds right to me. 

The min is in case someone does some peculiar tests with session_period like `2`. If the other branch was not there (in the `min`) then the non-validator nodes would lag behind.

PR #289, opened 2022-01-27
AZ-517 smartnet tokenomics (runtime weights for pallet_contracts)

fbielejec:


bin/runtime/src/lib.rs
```diff
@@ -170,9 +170,25 @@ pub const MAX_BLOCK_SIZE: u32 = 5 * 1024 * 1024;
...
-    pub BlockWeights: frame_system::limits::BlockWeights = frame_system::limits::BlockWeights
-        ::with_sensible_defaults(MAX_BLOCK_WEIGHT, NORMAL_DISPATCH_RATIO);
-    pub BlockLength: frame_system::limits::BlockLength = frame_system::limits::BlockLength
+    pub RuntimeBlockWeights: BlockWeights = BlockWeights::builder()
+        .base_block(BlockExecutionWeight::get())
+        .for_class(DispatchClass::all(), |weights| {
+            weights.base_extrinsic = ExtrinsicBaseWeight::get();
+        })
+        .for_class(DispatchClass::Normal, |weights| {
+            weights.max_total = Some(NORMAL_DISPATCH_RATIO * MAX_BLOCK_WEIGHT);
```
pmikolajczyk41, line 179: actually, you could extract normal_weight value as `NORMAL_DISPATCH_RATIO * MAX_BLOCK_WEIGHT` to clean it a little

bin/runtime/src/lib.rs
```diff
@@ -170,9 +170,25 @@ pub const MAX_BLOCK_SIZE: u32 = 5 * 1024 * 1024;
 parameter_types! {
     pub const Version: RuntimeVersion = VERSION;
     pub const BlockHashCount: BlockNumber = 2400;
-    pub BlockWeights: frame_system::limits::BlockWeights = frame_system::limits::BlockWeights
-        ::with_sensible_defaults(MAX_BLOCK_WEIGHT, NORMAL_DISPATCH_RATIO);
-    pub BlockLength: frame_system::limits::BlockLength = frame_system::limits::BlockLength
+    pub RuntimeBlockWeights: BlockWeights = BlockWeights::builder()
+        .base_block(BlockExecutionWeight::get())
```
pmikolajczyk41, line 174: is this needed?

pmikolajczyk41 (commented):
as far as i can see, the only semantic thing was that we used two different average block initialization weight, am I right?
if so, then why don't we stick to the 10% like Substrate?

bin/runtime/src/lib.rs
fbielejec: it's verbose, but at the same time lays out all the weights

bin/runtime/src/lib.rs
fbielejec: to my horror Rust does not atm support evaluation at compile-time :/ So to implement this something like [lazy_static](https://crates.io/crates/lazy_static) is needed. 
If this start to look like too much I can bring back the `sensible_defaults` implementation?

fbielejec:
> as far as i can see, the only semantic thing was that we used two different average block initialization weight, am I right?
> if so, then why don't we stick to the 10% like Substrate?

Done

h4nsu (approved):

bin/runtime/src/lib.rs
```diff
@@ -532,10 +554,10 @@ impl pallet_utility::Config for Runtime {
 
 parameter_types! {
     pub const DepositPerItem: Balance = deposit(1, 0);
-    pub const DepositPerByte: Balance = deposit(0, 1);
+    pub const DepositPerByte: Balance = 1_000_000_000; // 1e-3 alepha (1 milli aleph)
```
pmikolajczyk41, line 557: why? `deposit(0,1)` is exactly the price for a byte storage which we use e.g. for multisig. I agree, that probably it is not good having such universal, single cost calculation method, but we use it also one line above for `DepositPerItem`, so it is kinda confusing :|

also, Polish comment :stuck_out_tongue_closed_eyes:

fbielejec: well, there's a discussion to be had here: 

- with the default settings we had something like 1e11 / byte
- flipper (release optimized) is ~1.8KB (unoptimized is like 50KB), 1.8e14 pico-salpeh's total
- a shot from faucet is 9001 saleph's, so it's what? ~50 flippers.
- Since we want to promote experimentation this is raising it 100x 

Flipper is an ultra basic contract, I have no idea how WASM size increases with code complexity, but it won't be linear I'd imagine...

pmikolajczyk41 (approved):

PR #290, opened 2022-01-27
AZ-502: Restarts are bad after all

timorl:


Yithis (approved):
LGTM

finality-aleph/src/party/mod.rs
```diff
@@ -542,12 +546,16 @@ where
...
 
-                        if let Err(e) = self.session_manager.start_validator_session(
-                            next_session_id,
-                            authority_verifier,
-                            node_id,
-                            authority_pen,
-                        ) {
+                        if let Err(e) = self
+                            .session_manager
+                            .start_validator_session(
```
DamianStraszak, line 551: If I understand correctly this (along with `get_node_index` `AuthorityPen::new(` etc.) is executed in every iteration of the loop (so every 1sec). If this is necessary then add a comment explaining why, otherwise try to make it fire only once...

timorl: Good point, I'm rewriting this to make it more explicit what we are waiting for where.

PR #291, opened 2022-01-31
AZ-490: Remove parking_lot mutex from non test code

maciejnems:


finality-aleph/src/data_io.rs
```diff
@@ -377,8 +377,10 @@ pub(crate) async fn refresh_best_chain<B, BE, SC, C>(
     // uses is that once the block in `proposed_block` reaches the height of `max_block_num`, and the just queried
     // `best_block` is a `descendant` of the previous query, then we don't need to update proposed_block, as it is
     // already correct.
-    let mut prev_best_hash: B::Hash = proposed_block.lock().hash;
-    let mut prev_best_number: NumberFor<B> = proposed_block.lock().number;
+    let block = proposed_block.lock().await;
+    let mut prev_best_hash: B::Hash = block.hash;
+    let mut prev_best_number: NumberFor<B> = block.number;
+    drop(block);
```
timorl, line 383: Rather than explicitly `drop`ping it might be better to enclose this into a section – I think something like this would work:
```
let (prev_best_hash, prev_best_numbere) = async {
 let block = proposed_block.lock().await;
 (block.hash, block.number)
}.await;
```
Or perhaps just `.clone()` the whole thing and extract things from that.

BIg alternative solution – keep data about the previous block separately from the mutex and use it for updating, just modifying the mutex by copying the current state into it as the last step of the whole computation process.

finality-aleph/src/network/mock.rs
```diff
@@ -10,14 +13,30 @@ use std::{
     sync::Arc,
 };
 
-type Channel<T> = (
-    Arc<Mutex<mpsc::UnboundedSender<T>>>,
-    Arc<Mutex<mpsc::UnboundedReceiver<T>>>,
+#[derive(Clone)]
+pub struct Channel<T>(
+    pub mpsc::UnboundedSender<T>,
+    pub Arc<tokio::sync::Mutex<mpsc::UnboundedReceiver<T>>>,
```
timorl, line 19: Huh, why is the other mutex still here? Could you remove it?

timorl (approved):
Minor quibbles. Thanks for doing this!

timorl (changes requested):
Wait, scratch that, you haven't removed `parking_lot` from the deps, and that was, like, a big part of this whole exercise!

finality-aleph/src/network/mock.rs
maciejnems: The methods here are not async (in trait that is being mocked) but still need mutexes in tests.

finality-aleph/src/data_io.rs
maciejnems: I added an async block (I think it looks beautiful :3 ). The alternative solution is not in scope of this PR and I guess @DamianStraszak is working on it right now?

maciejnems:
> Wait, scratch that, you haven't removed `parking_lot` from the deps, and that was, like, a big part of this whole exercise!

Unfortunately `parking_lot` cannot be removed, as it is used in metrics. The use of `parking_lot::mutex` is fine there.

timorl (approved):
:<

finality-aleph/src/data_io.rs
```diff
@@ -388,18 +391,19 @@ pub(crate) async fn refresh_best_chain<B, BE, SC, C>(
                     .await
                     .expect("No best chain");
                 if *new_best_header.number() <= max_block_num {
-                    *proposed_block.lock() = AlephData::new(new_best_header.hash(), *new_best_header.number());
+                    *proposed_block.lock().await = AlephData::new(new_best_header.hash(), *new_best_header.number());
                 } else {
+                    let mut block = proposed_block.lock().await;
```
DamianStraszak, line 396: You are holding the lock over a call to `reduce_header_to_num`, which is an expensive operation -- this must be avoided.

DamianStraszak: There are also locks inside `reduce_header_to_num` (client), so I'm not even sure this cannot lead to deadlocks.

maciejnems: Well, the thing is that if statement is dependent on lock, and `reduce_header_to_num` happens between if statement and change in locked value. So Im not sure we can avoid it here. Unless there is no problem if we change this value after if statement state changes.

DamianStraszak: The `if condition` depends on `block.number`, but then the next line doesn't.

DamianStraszak: And the `if condition` is evaluated prior to evaluating the `if body`.

DamianStraszak: Ok, I think I see what your concern is. This is the only thread that writes to this variable, so it is safe.

maciejnems: The concern is that `block.number` can change during execution of `if body`. The value of `proposed_block` could change (so that other branch of `if` should be selected), and we would override `proposed_block` with a wrong value. This is a bug from a concurrent programming point of view.

That said you are right, that now `proposed_block` is only changed in this place. If we had condition that it can be changed only here, only by this thread then this part is correct. But it comes from `spawn_authority_subtasks` function, so it could be changed in other places in the future and the person changing something here would have to remember that we have a strange lock and in other places we can do only a read lock.
 
But then I looked at the code and it would be really strange if it was changed somewhere else. What do you think about it @timorl ?

finality-aleph/src/data_io.rs
```diff
@@ -388,18 +391,19 @@ pub(crate) async fn refresh_best_chain<B, BE, SC, C>(
                     .await
                     .expect("No best chain");
                 if *new_best_header.number() <= max_block_num {
-                    *proposed_block.lock() = AlephData::new(new_best_header.hash(), *new_best_header.number());
+                    *proposed_block.lock().await = AlephData::new(new_best_header.hash(), *new_best_header.number());
```
Marcin-Radecki, line 394: First time see this concept that we assign to `lock()` on the left-hand side of assignment - so just double checking if this line is an equivalent to 
```rust
{
   let mut block = proposed_block.lock().await;                    
   *block = AlephData::new(new_best_header.hash(), *new_best_header.number());
}
```
as I'm not sure entirelely what's the lifetime of `MutexGourd` temporary object `*proposed_block.lock().await `.

maciejnems: Yes, I'm pretty sure this is how it works. I mean I'm not sure about exact equivalence, as I'm not sure what this block compiles to. Anyway, lifetime of temporary object ends in this line and guard is released. And value is set.

finality-aleph/src/data_io.rs
timorl: Eh, I'm not convinced the whole optimization happening in these `if` is such a good idea anyway, so maybe drop at least the external `if`? Then we wouldn't need the (b)lock for anything else than assigning.

maciejnems: But the problem is with internal `if`, not with external. Which `if` do you mean by external?

maciejnems: Do you mean this optimization?
```
let reduced_header = reduce_header_to_num(client.clone(), new_best_header.clone(), prev_best_number);
```

timorl: Yeah, the `prev_best_number` part requires some thought, but should be changeable to something reasonable.

maciejnems: Ok, so I guess this can be changed to something like that:
```
if *new_best_header.number() <= max_block_num {
    *proposed_block.lock().await = AlephData::new(new_best_header.hash(), *new_best_header.number());
} else {
    // we check if prev_best_header is an ancestor of new_best_header:
    if prev_best_number < max_block_num {
        let reduced_header = reduce_header_to_num(client.clone(), new_best_header.clone(), max_block_num);
        *proposed_block.lock().await = AlephData::new(reduced_header.hash(), *reduced_header.number());
    } else {
        let reduced_header = reduce_header_to_num(client.clone(), new_best_header.clone(), prev_best_number);
        if reduced_header.hash() != prev_best_hash {
            // the new best block is not a descendant of previous best
            let reduced_header = reduce_header_to_num(client.clone(), new_best_header.clone(), max_block_num);
            *proposed_block.lock().await = AlephData::new(reduced_header.hash(), *reduced_header.number());
        }
    }
}
```
And now there is no problem with locks

DamianStraszak: Yes, this version looks correct to me.

DamianStraszak (approved):

Marcin-Radecki (approved):
Changes look good, the e2e failed but looks like a false positive - I've checked locally and they passed. So perhaps a re-run of this workflow would do the trick.

PR #297, opened 2022-02-11
contract flood

kostekIV:


Marcin-Radecki:
Please fix linter https://github.com/Cardinal-Cryptography/aleph-node/runs/5154290395?check_suite_focus=true

local-tests/test_flood.py
```diff
@@ -0,0 +1,63 @@
...
+import sys
+from os.path import abspath, join
+from time import sleep, time
+
+from chainrunner import Chain, Seq, generate_keys
+
+# Path to working directory, where chainspec, logs and nodes' dbs are written:
+workdir = abspath(os.getenv('WORKDIR', '/tmp/workdir'))
+# Path to the pre-update aleph-node binary:
+bin = abspath(os.getenv('BINARY', join(workdir, 'aleph-node-old')))
```
Marcin-Radecki, line 13: Technically we want to have the same behavior for this script in the case of `BINARY` or `FLOODER` envs missing, ie exiting with non-zero exit code? ie as `test_flood.sh` does.

local-tests/test_flood.py
```diff
@@ -0,0 +1,63 @@
...
+                port=Seq(30334),
+                ws_port=Seq(9944),
+                rpc_port=Seq(9933),
+                unit_creation_delay=200,
+                execution='Native')
+
+chain.start('aleph')
+
+print('Waiting a minute')
+sleep(15)
```
Marcin-Radecki, line 50: Technically we need to have 60 here?

local-tests/test_flood.py
```diff
@@ -0,0 +1,63 @@
...
+
+start = time()
+old = check_highest(chain)
+
+subprocess.run([flooder]).check_returncode()
+new = check_highest(chain)
+
+expected = int((time() - start) * 1000) // MILLISECS_PER_BLOCK
```
Marcin-Radecki, line 58: Nit: we can name this variable as `expected_milis` to express the unit. Also, we can use `MILLISECS_PER_BLOCK` variable declared in the same script.

.github/workflows/flood-smartnet.yml
```diff
@@ -0,0 +1,59 @@
...
+    uses: Cardinal-Cryptography/aleph-node/.github/workflows/build-node-and-runtime.yml@smartnet
+    with:
+      rust-toolchain-version: 'nightly-2021-10-24'
+
+  test-contracts-flooding:
+    name: Test contracts flooding
+    runs-on: ubuntu-latest
+    needs: [ build-node ]
+    env:
+      RUST_TOOLCHAIN_VERSION: nightly-2021-10-24
```
Marcin-Radecki, line 33: After this is merged https://github.com/Cardinal-Cryptography/aleph-node/pull/298, we don't need to state `RUST_TOOLCHAIN_VERSION` explicetely?

local-tests/test_flood.py
kostekIV: the file exist check will throw an error in such a case

local-tests/test_flood.py
kostekIV: changed the printing value to avoid confusion here

local-tests/test_flood.py
kostekIV: renamed the variable to `expected_blocks`, I think we are already using `MILLISECS_PER_BLOCK` here?

.github/workflows/flood-smartnet.yml
kostekIV: yes

PR #300, opened 2022-02-14
Az 477 add staking

mike1729:
This PR adds pallet staking to our runtime.

You can find more on pallet staking [here](https://www.notion.so/cardinalcryptography/Staking-7a88dc9fb6dc44e085cbc7bf4e309193) and an intro to the UI [here](https://www.notion.so/cardinalcryptography/UI-473a142e03f74f3ab83bcbadef1b985a).


bin/node/src/chain_spec.rs
```diff
@@ -282,23 +316,29 @@ fn genesis(
...
-            .map(|auth| &auth.account_id)
-            .cloned()
+        to_account_ids(&authorities)
             .chain(special_accounts)
+            .chain(stakers.iter().cloned())
             .collect(),
     );
 
+    const ENDOWMENT: u128 = 1 << 60;
+    const MIN_VALIDATOR_BOND: u128 = 25_000 * 1_000_000_000_000;
```
DamianStraszak, line 328: Would be good to have `const AZERO = ....` and then represent other values in terms of `AZERO`.

pallets/aleph/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    #[pallet::pallet]
+    #[pallet::storage_version(STORAGE_VERSION)]
+    pub struct Pallet<T>(_);
+
+    #[pallet::hooks]
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        fn on_runtime_upgrade() -> frame_support::weights::Weight {
+            migrations::v1_to_v2::migrate::<T, Self>()
```
DamianStraszak, line 52: I'm pretty certain this is wrong. The `v0_to_v1` migration is not on chain yet. Why do we even consider removing migrations?

mike1729: Since it doesn't compile and I don't know how to force it. There is this tricky macro `generate_storage_alias!` that I used in `v1_to_v2` to remove fields of pallet aleph that doesn't exist in the new version, but here it's different since field `Validators` is of type `Vec<T::AccountId>` and I cannot use `T` in this macro.

We should then apply updates one by one.

DamianStraszak: Applying updates one by one is of course an option, but we should have a discussion about it, as that would me a major change of our update strategy and essentially everything that is being done now considering devnet, featurnets, etc. 

I don't have experience in migrations, but you might ask Filip if he knows how to do that.

pallets/elections/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+                        Support {
+                            total: 0,
+                            voters: Vec::new(),
+                        },
+                    )
+                })
+                .collect();
+
+            for (voter, vote, targets) in voters {
+                assert!(
```
DamianStraszak, line 152: I think there is a policy that runtime should never panic. So I would suggest removing this assert.

pallets/elections/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+                    targets.len() == 1,
+                    "There should be only one target in one vote"
+                );
+                let member = &targets[0];
+                if !members.contains(member) {
```
DamianStraszak, line 157: I think this is unnecessary -- just use the output of `supports.get_mut(member)` to check it.

DamianStraszak: Also, rule of thumb: using `contains` on a `Vec` should ALWAYS raise a red flag :) This does a linear search over the whole vector.

pallets/elections/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+    pub enum Error {
+        DataProvider(&'static str),
+    }
+
+    impl<T: Config> ElectionProvider<T::AccountId, BlockNumberFor<T>> for Pallet<T> {
+        type Error = Error;
+        type DataProvider = T::DataProvider;
+
+        // We use authority list so the supports don't influance the result
+        // but we still have to calculate them for the sake of eras payouts
```
DamianStraszak, line 134: This comment is rather confusing. It should say something along the lines of:
"As of now the election results do not influence the list of validators, which is based on a sudo-managed authority list. However, we still need to calculate the `Supports` for the sake of era payouts."

pallets/elections/src/lib.rs
mike1729: Sure!

pallets/elections/src/lib.rs
mike1729: Good idea :)

pallets/elections/src/lib.rs
mike1729: But your version is not quite right. For every new era pallet staking uses this `elect` method to get new validators, so the election results do influence the list of validators. What is missing in this scheme is some `vote` extrinsic. What we have is just `change_members` sudo call.
Which part is confusing? I can try to improve it.

pallets/aleph/src/migrations/v1_to_v2.rs
```diff
@@ -0,0 +1,38 @@
...
+pub fn migrate<T: Config, P: GetStorageVersion + PalletInfoAccess>() -> Weight {
+    let on_chain_storage_version = <P as GetStorageVersion>::on_chain_storage_version();
+    let current_storage_version = <P as GetStorageVersion>::current_storage_version();
+
+    if on_chain_storage_version == 1 && current_storage_version == 2 {
+        log::info!(target: "pallet_aleph", "Running migration from STORAGE_VERSION 1 to 2");
+
+        SessionForValidatorsChange::kill();
+        MillisecsPerBlock::kill();
+        SessionPeriod::kill();
```
DamianStraszak, line 22: I would love to see a sanity check here of whether the keys exist, before killing them. Because how else will we know if the migration actually does anything? (Unless `kill()` already does that?)

pallets/elections/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+                );
+                let member = &targets[0];
+                if !members.contains(member) {
+                    continue;
+                }
+
+                let support = supports
+                    .get_mut(member)
+                    .expect("Was initialized with all members; qed.");
+                support.total += vote as u128;
```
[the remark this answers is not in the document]
mike1729: https://paritytech.github.io/substrate/master/frame_support/traits/struct.U128CurrencyToVote.html
This is how stake gets converted to vote.

bin/node/src/chain_spec.rs
mike1729: Fixed.

bin/runtime/src/lib.rs
```diff
@@ -383,12 +391,82 @@ impl<I: From<u32>> ::frame_support::traits::Get<I> for SessionPeriod {
...
+impl pallet_session::historical::Config for Runtime {
+    type FullIdentification = pallet_staking::Exposure<AccountId, Balance>;
+    type FullIdentificationOf = pallet_staking::ExposureOf<Runtime>;
+}
+
+parameter_types! {
+    pub const BondingDuration: EraIndex = 14;
+    pub const SlashDeferDuration: EraIndex = 13;
+    pub const MaxNominatorRewardedPerValidator: u32 = 512;
+    pub const OffendingValidatorsThreshold: Perbill = Perbill::from_percent(33);
```
DamianStraszak, line 412: What is the consequence of setting this to 33? What will happen if this threshold is reached? (And what really counts towards this threshold?)

pallets/elections/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+    impl<T: Config> Default for GenesisConfig<T> {
+        fn default() -> Self {
+            Self {
+                members: Vec::new(),
+                millisecs_per_block: DEFAULT_MILLISECS_PER_BLOCK,
+                session_period: DEFAULT_SESSION_PERIOD,
+                sessions_per_era: DEFAULT_SESSIONS_PER_ERA,
+            }
+        }
+    }
```
DamianStraszak, line 110: This pallet has some genesis config (where storage items are initialized at genesis). We add it to the runtime **not at genesis**. Are these storage items then initialized? If so, what is the mechanism here that does it?

bin/runtime/src/lib.rs
```diff
@@ -107,7 +108,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("aleph-node"),
     impl_name: create_runtime_str!("aleph-node"),
     authoring_version: 1,
-    spec_version: 7,
+    spec_version: 9,
     impl_version: 1,
     apis: RUNTIME_API_VERSIONS,
     transaction_version: 3,
```
DamianStraszak, line 114: This must be bumped as well

bin/runtime/src/lib.rs
```diff
@@ -383,12 +391,82 @@ impl<I: From<u32>> ::frame_support::traits::Get<I> for SessionPeriod {
 impl pallet_session::Config for Runtime {
     type Event = Event;
     type ValidatorId = <Self as frame_system::Config>::AccountId;
-    type ValidatorIdOf = ConvertInto;
+    type ValidatorIdOf = pallet_staking::StashOf<Self>;
     type ShouldEndSession = pallet_session::PeriodicSessions<SessionPeriod, Offset>;
     type NextSessionRotation = pallet_session::PeriodicSessions<SessionPeriod, Offset>;
-    type SessionManager = pallet_aleph::AlephSessionManager<Self>;
+    type SessionManager = pallet_session::historical::NoteHistoricalRoot<Self, Staking>;
```
DamianStraszak, line 397: What does this change mean to us?

pallets/elections/src/lib.rs
mike1729: The pallets starts from the next block form where it got added so there is no initialization. Similarly is for pallet staking and it starts in mode `NotForcing` for new era and it's not possible for the pallet to start a new one. So we first have to set members via `change_members` and then force new era in pallet staking.

bin/runtime/src/lib.rs
```diff
@@ -655,21 +742,23 @@ impl_runtime_apis! {
...
-        }
-
         fn authorities() -> Vec<AlephId> {
             Aleph::authorities()
         }
 
-        fn session_period() -> u32 {
-            Aleph::session_period()
+        fn millisecs_per_block() -> u64 {
+            Elections::millisecs_per_block()
```
DamianStraszak, line 750: Why did we move it from Aleph to Elections? Not saying it is wrong, but just would like to know...

pallets/aleph/src/migrations/v1_to_v2.rs
mike1729: If there is no storage then kill does nothing.

bin/runtime/src/lib.rs
mike1729: That now pallet staking is responsible for triggering new sessions and not pallet aleph. Pallet historical is just a wrapper that stores previous data.

bin/runtime/src/lib.rs
mike1729: It's more natural for `session_period` and `sessions_per_era` to be in pallet elections, so `millisecs_per_block` kind of joined them on the way but there is also a possibility that this parameter will be needed in connection to eras.

bin/runtime/src/lib.rs
mike1729: Bumped.

pallets/elections/src/lib.rs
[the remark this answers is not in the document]
DamianStraszak: Ok, then this is fine.

pallets/elections/src/lib.rs
DamianStraszak: OK, so apparently I did not understand how it works. But what confuses me in this comment is saying that `supports` don't influence (also note typo) the `result`. Which result? If this is the result of the function, then well, the result is `supports` :P

pallets/elections/src/lib.rs
DamianStraszak: I don't quite understand. If it is not initialized then how will the `AlephSessionApi` work? If they were not initialized right after the update then would `session_period` etc. be initialized ever?
I'm pretty sure these items will get initialized in the first block after the update, but I'm wondering how would that happen.

pallets/elections/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+    #[pallet::config]
+    pub trait Config: frame_system::Config {
+        type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+        type DataProvider: ElectionDataProvider<Self::AccountId, Self::BlockNumber>;
+    }
+
+    #[pallet::event]
+    #[pallet::generate_deposit(pub(super) fn deposit_event)]
+    pub enum Event<T: Config> {
+        ChangeMembers(Vec<T::AccountId>),
```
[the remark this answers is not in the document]
DamianStraszak: Also, a concrete and important case study: that happens if I send this extrinsic in the last session of an era?

pallets/elections/src/lib.rs
mike1729: I meant the results of the elections :)

pallets/elections/src/lib.rs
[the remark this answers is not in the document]
mike1729: In the previous version, pallet-aleph returned a list of authorities directly in the implementation of SessionManager::new_session. Here pallet staking implements this method and inside it it calls elect instead of reading directly.
The only difference is that we previously had also a way of delaying elections with `session_for_validators_change`. Now elections are for an era `n` which starts in session `k` happen in `on_initialize` hook of last block of session `k-2`.

I suppose that it's now obvious, that in your example the change triggered by the extrinsic will take place in the era after the era that comes after this last session you send the extrinsic.

bin/runtime/src/lib.rs
mike1729: A new era will be forced after the session in which the threshold has been reached. (Slaching)

pallets/elections/src/lib.rs
mike1729: Those constants are set in bin/runtime/src/lib.rs and overwritten in genesis, so here everything is ok with `AlephSessionApi`

pallets/elections/src/lib.rs
Marcin-Radecki: Btw isn't it that for small vector sizes (<100) linear search in vec would be faster than logarithmic for tree structures based on key comparison?

DamianStraszak: Absolutely correct, but not sure where the threshold lies -- especially since this is in wasm. Anyway, we want `members` to grow way beyond `10` :)

mike1729: But then, the implementation of the elections will be different ;)

DamianStraszak: Yes. Anyway, this access to the map is necessary, so I think the current version is OK no matter whether membership check for vectors or for BTrees is faster :)

pallets/elections/src/lib.rs
DamianStraszak: I think the nomenclature here is that `Supports` is the result of elections.

bin/runtime/src/lib.rs
DamianStraszak: Yes, what does it mean that a new era will be forced?

pallets/elections/src/lib.rs
mike1729: You're right. How about:
// The elections are PoA so only the nodes listed in the Members will be elected as validators.
// We calculate the supporst for them for the sake of eras payouts.

bin/runtime/src/lib.rs
mike1729: That a new era with new validators that were elected for it will start in the next session and not in possible further session as planned before.

pallets/elections/src/lib.rs
DamianStraszak: So in my eyes, the `session_period` call from `AlephSessionApi` reads and outputs the `session_period` storage item from pallet `Elections`. My question is why would that item exist just after the update?

pallets/elections/src/lib.rs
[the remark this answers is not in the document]
DamianStraszak: Ok, so definitely a comment is required that explain exactly what will be the effect of sending this extrinsic. 

But also I think this sentence "The only difference is that ..." is not quite correct, because we move from the granularity of sessions to eras. This is a huge difference that gives us way less freedom when reacting to emergencies.

bin/runtime/src/lib.rs
DamianStraszak: So historical wraps `Staking`? Why do we need that?

pallets/elections/src/lib.rs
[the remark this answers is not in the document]
mike1729: We can sent this xt and the [force_new_era](https://paritytech.github.io/substrate/master/pallet_staking/pallet/enum.Call.html#variant.force_new_era) to have the same functionality.

bin/runtime/src/lib.rs
DamianStraszak: Ok, why would we want that? I mean, it's not like forcing a new era will help with anything if finality is stalled. And we don't want the eras to randomly change because something unexpected happened, or at least not till we control the validators.

One general question is: is modifying these constants possible in an update after we launch the pallet?

pallets/elections/src/lib.rs
DamianStraszak: Great. A small typo only: `supporst`

pallets/aleph/src/migrations/v1_to_v2.rs
DamianStraszak: Ok, so I think we really need some logs here that can make us confident that the migration went **EXACTLY AS EXPECTED**, i.e. that every key that we expected was present etc. This is mainly for the purpose of testing -- if we leave it as it is then we might be killing non-existing keys and stuff like that.

bin/runtime/src/lib.rs
mike1729: It's possible to modify in an update.

Eras won't randomly change since for a validator to get kicked out of the validators list a sudo call is needed.

e2e-tests/src/test/treasury.rs
```diff
@@ -38,11 +61,16 @@ pub fn channeling_fee(config: Config) -> anyhow::Result<()> {
...
-    assert!(
-        issuance_after <= issuance_before,
-        "Unexpectedly {} was minted",
-        issuance_after - issuance_before
-    );
+    if issuance_after > issuance_before {
+        let difference = issuance_after - issuance_before;
+        assert_eq!(
+            difference % possibly_treasury_gain_from_staking,
+            0,
```
timorl, line 68: Uh? Shouldn't you just compare the two above, rather than taking modulo?

bin/runtime/src/lib.rs
mike1729: This is used for equivocations in block production and finality. It provides a basic implementation of https://paritytech.github.io/substrate/master/frame_support/traits/trait.KeyOwnerProofSystem.html which is then extended in those modules.

pallets/aleph/src/migrations/v1_to_v2.rs
mike1729: So simple check if all of them exist or reading values as well? I don't see how they cannot exit but logging doesn't hurt.

pallets/aleph/src/lib.rs
```diff
@@ -0,0 +1,106 @@
+//! Aleph session manager.
+//!
+//! This pallet manages the changes in the committee responsible for establishing consensus.
+//! Currently, it's PoA where the validators are set by the root account. In the future, a new
+//! pallet for PoS elections will replace this one.
+//!
+//! For full integration with Aleph finality gadget, the `primitives::AlephSessionApi` should be implemented.
```
timorl, line 7: This whole comment is at least partially outdated.

bin/runtime/src/lib.rs
DamianStraszak: How do we modify these constants? Some kind of custom storage migration?

DamianStraszak: Oh, or are these contants only in the runtime and not in the storage?

bin/runtime/src/lib.rs
DamianStraszak: Ok, but is this functionality used in the current code, or only in the future?

pallets/elections/src/lib.rs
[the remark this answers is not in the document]
DamianStraszak: Nice, I think documenting all that in a comment would make a lot of sense.

pallets/elections/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "pallet-elections"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
timorl, line 6: I think this is unnecessary. :>

pallets/aleph/src/migrations/v1_to_v2.rs
DamianStraszak: They can not exists because we might be doing the migration wrong, which is exactly what happened in one of the previous PRs that a few people (including me btw.) read, tested and accepted.

pallets/aleph/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+use sp_std::prelude::*;
+
+use frame_support::{
+    sp_runtime::BoundToRuntimeAppPublic,
+    traits::{OneSessionHandler, StorageVersion},
+    Parameter,
+};
+pub use pallet::*;
+
+/// The current storage version.
```
timorl, line 27: This comment brought to you by department of redundancy department.

pallets/elections/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+
+#[cfg(test)]
+mod mock;
+#[cfg(test)]
+mod tests;
+
+use frame_support::traits::StorageVersion;
+pub use pallet::*;
+
+/// The current storage version.
```
timorl, line 11: I would repeat the same terrible joke as above, but that would be too meta.

pallets/elections/src/mock.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+impl Config for Test {
+    type Event = Event;
+    type DataProvider = StakingMock;
+}
+
+pub struct StakingMock;
+impl ElectionDataProvider<AccountId, u64> for StakingMock {
+    const MAXIMUM_VOTES_PER_VOTER: u32 = 16;
```
timorl, line 100: Huh, I might be misunderstanding this terribly, but shouldn't this be 1?

pallets/elections/src/tests.rs
```diff
@@ -0,0 +1,20 @@
...
+
+#[test]
+fn test_elect() {
+    new_test_ext(vec![1, 2]).execute_with(|| {
+        let elected = <Elections as ElectionProvider<AccountId, u64>>::elect();
+        assert!(elected.is_ok());
+
+        let supp = Support {
+            total: 0,
+            voters: Vec::new(),
```
timorl, line 15: Could you add a test that outputs something nontrivial?

primitives/src/lib.rs
```diff
@@ -45,3 +48,21 @@ sp_api::decl_runtime_apis! {
         fn millisecs_per_block() -> u64;
     }
 }
+
+pub mod staking {
+    use super::Balance;
+    use sp_runtime::Perbill;
+
+    pub fn era_payout(miliseconds_per_era: u64) -> (Balance, Balance) {
+        const YEARLY_INFLATION: Balance = 30 * 1_000_000 * 1_000_000_000_000;
```
timorl, line 57: Why is this written as a product?

bin/runtime/src/lib.rs
```diff
@@ -655,21 +742,23 @@ impl_runtime_apis! {
...
-            Aleph::millisecs_per_block()
+        fn session_period() -> u32 {
+            Elections::session_period()
         }
 
+        fn next_session_authorities() -> Result<Vec<AlephId>, AlephApiError> {
+                Session::queued_keys()
+                .iter()
+                .map(|(_, key)| key.get(AlephId::ID).ok_or(AlephApiError::DecodeKey))
+                .collect::<Result<Vec<AlephId>, AlephApiError>>()
```
DamianStraszak, line 761: Why did we move the implementation from pallet Aleph to here?

timorl (changes requested):
This could really use some documentation about the choices made – not even in the code, but somewhere.

bin/runtime/src/lib.rs
```diff
@@ -219,10 +218,21 @@ impl pallet_aura::Config for Runtime {
     type DisabledValidators = ();
 }
 
+parameter_types! {
+    pub const UncleGenerations: BlockNumber = 5;
```
DamianStraszak, line 222: Wait, what the hell is that :P

timorl: I just finished my review in which I had a similarly perplexed reaction. :P

bin/runtime/src/lib.rs
mike1729: The original reason was that the main functionality is provided by pallet session and all the types are from primitives, so there is no dependence on pallet aleph. Explicitly, ofc, since this works because pallet aleph implements `OneSessionHanlder`.

bin/node/src/chain_spec.rs
```diff
@@ -122,13 +129,17 @@ pub struct ChainParams {
...
+    /// Time interval (in milliseconds) between blocks. Default is 1000ms
+    #[structopt(long)]
+    millisecs_per_block: Option<u64>,
+
+    /// The length of a session (in seconds). Default is 900s
     #[structopt(long)]
     session_period: Option<u32>,
 
-    /// Time interval (in milliseconds) between blocks. Default is 1000 ms
+    /// The length of an era (in Sessions). Default is 4*24=96
```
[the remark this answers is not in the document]
mike1729: Sure.

bin/node/src/chain_spec.rs
```diff
@@ -122,13 +129,17 @@ pub struct ChainParams {
...
     #[structopt(long)]
-    millisecs_per_block: Option<u64>,
+    sessions_per_era: Option<SessionIndex>,
```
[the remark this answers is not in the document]
mike1729: Unfortunately, this is what substrate uses :( https://paritytech.github.io/substrate/master/pallet_staking/trait.Config.html#associatedtype.SessionsPerEra

bin/node/src/chain_spec.rs
```diff
@@ -282,23 +316,29 @@ fn genesis(
     // NOTE: some combinations of bootstrap chain arguments can potentially
     // lead to duplicated rich accounts, e.g. if a sudo account is also an authority
     // which is why we remove the duplicates if any here
+    // endow as well stash accounts, if any
```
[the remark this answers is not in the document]
mike1729: this last sentence is just unnecessary

bin/node/src/chain_spec.rs
```diff
@@ -282,23 +316,29 @@ fn genesis(
...
-            .iter()
-            .map(|auth| &auth.account_id)
-            .cloned()
+        to_account_ids(&authorities)
             .chain(special_accounts)
+            .chain(stakers.iter().cloned())
             .collect(),
     );
 
+    const ENDOWMENT: u128 = 1 << 60;
```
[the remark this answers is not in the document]
mike1729: lets make it consistent.

bin/node/src/chain_spec.rs
```diff
@@ -307,26 +347,48 @@ fn genesis(
...
                 })
                 .collect(),
         },
+        staking: StakingConfig {
+            force_era: Forcing::NotForcing,
+            validator_count: authorities.len() as u32,
+            minimum_validator_count: authorities.len() as u32,
+            invulnerables: to_account_ids(&authorities).collect(),
+            slash_reward_fraction: Perbill::from_percent(10),
+            stakers: stakers
```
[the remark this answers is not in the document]
mike1729: This is used to generate a chainspec for devnet that is easy to test, similarily as we endow accounts. Chainspec for testnet and mainnet are already generated so this method will always be used for devnet only in the future. As such, I would leave stakers here.

bin/node/src/chain_spec.rs
```diff
@@ -307,26 +347,48 @@ fn genesis(
...
                         SessionKeys {
                             aura: auth.aura_key.clone(),
-                            aleph: auth.aleph_key,
+                            aleph: auth.aleph_key.clone(),
                         },
                     )
                 })
                 .collect(),
         },
+        staking: StakingConfig {
```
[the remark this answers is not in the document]
mike1729: This configuration is for devnet only. I believe the parameters defined in runtime deserve documentation if any.

bin/runtime/src/lib.rs
```diff
@@ -206,8 +207,6 @@ impl frame_system::Config for Runtime {
     type OnSetCode = ();
 }
 
-impl pallet_randomness_collective_flip::Config for Runtime {}
```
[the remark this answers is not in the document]
mike1729: Right and it doesn't define anything, so nothing wrong here.

bin/node/src/chain_spec.rs
```diff
@@ -217,9 +232,25 @@ fn system_properties(token_symbol: String) -> serde_json::map::Map<String, Value
     .collect()
 }
 
+pub fn devnet_config(
```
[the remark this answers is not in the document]
mike1729: That's a good idea, but I'd leave it to implement for someone who would like to write some tests using staking.

bin/runtime/src/lib.rs
```diff
@@ -219,10 +218,21 @@ impl pallet_aura::Config for Runtime {
...
 
+parameter_types! {
+    pub const UncleGenerations: BlockNumber = 5;
+}
+
+impl pallet_authorship::Config for Runtime {
+    type FindAuthor = pallet_session::FindAccountFromAuthorIndex<Self, Aura>;
+    type UncleGenerations = UncleGenerations;
+    type FilterUncle = ();
+    type EventHandler = (Staking,); // consider adding ImOnline
```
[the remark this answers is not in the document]
mike1729: Already a jira ticket, so I remove it.

bin/runtime/src/lib.rs
```diff
@@ -383,12 +391,82 @@ impl<I: From<u32>> ::frame_support::traits::Get<I> for SessionPeriod {
...
+
+impl pallet_session::historical::Config for Runtime {
+    type FullIdentification = pallet_staking::Exposure<AccountId, Balance>;
+    type FullIdentificationOf = pallet_staking::ExposureOf<Runtime>;
+}
+
+parameter_types! {
+    pub const BondingDuration: EraIndex = 14;
+    pub const SlashDeferDuration: EraIndex = 13;
+    pub const MaxNominatorRewardedPerValidator: u32 = 512;
```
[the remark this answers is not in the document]
mike1729: Actually we would like to increase it, as there are approximately 7k account that could nominate if the bottom limit is 1k AZERO. The q for communication is probably out of the scope ;) Happy to talk elsewhere.

bin/runtime/src/lib.rs
```diff
@@ -383,12 +391,82 @@ impl<I: From<u32>> ::frame_support::traits::Get<I> for SessionPeriod {
...
+    }
+}
+
+pub struct UniformEraPayout {}
+
+impl pallet_staking::EraPayout<Balance> for UniformEraPayout {
+    fn era_payout(_: Balance, _: Balance, _: u64) -> (Balance, Balance) {
+        let miliseconds_per_era = Elections::millisecs_per_block()
+            * Elections::session_period() as u64
+            * Elections::sessions_per_era() as u64;
```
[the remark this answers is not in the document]
mike1729: Just for this one use?

bin/runtime/src/lib.rs
DamianStraszak: Ok, fine, but just to clarify -- why do you think there is a connection between this working, and Aleph implementing `OneSessionHandler`. I think this works just because there is a dedicated Aleph key (with id `AlephId::ID`) in `SessionKeys`

bin/node/src/chain_spec.rs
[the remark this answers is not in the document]
Marcin-Radecki: My goal here was that since we've added line just below
```rust
     .chain(stakers.iter().cloned())
```
Then it supposed to mean that for some scenarios (for devnet only) `stakers` is not empty and we endow some more account, but I agree the comment might sound  misleading

bin/runtime/src/lib.rs
```diff
@@ -655,21 +742,23 @@ impl_runtime_apis! {
...
         }
 
-        fn millisecs_per_block() -> u64 {
-            Aleph::millisecs_per_block()
+        fn session_period() -> u32 {
+            Elections::session_period()
         }
 
+        fn next_session_authorities() -> Result<Vec<AlephId>, AlephApiError> {
+                Session::queued_keys()
```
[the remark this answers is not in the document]
mike1729: This is in macro, fmt is helpless :(

e2e-tests/src/accounts.rs
```diff
@@ -5,26 +5,29 @@ use substrate_api_client::Balance;
...
-        vec![
-            "//Damian".into(),
-            "//Tomasz".into(),
-            "//Zbyszko".into(),
-            "//Hansu".into(),
-        ]
-    });
-    seeds.into_iter().map(keypair_from_string).collect()
+pub fn accounts_from_seeds(seeds: Option<&Vec<String>>) -> Vec<KeyPair> {
+    if let Some(seeds) = seeds {
```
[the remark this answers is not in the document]
mike1729: KK

e2e-tests/src/accounts.rs
```diff
@@ -5,26 +5,29 @@ use substrate_api_client::Balance;
...
+        seeds
+            .iter()
+            .map(String::as_str)
+            .map(keypair_from_string)
+            .collect()
+    } else {
+        ["//Damian", "//Hansu", "//Tomasz", "//Zbyszko"]
```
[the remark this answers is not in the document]
mike1729: Good point :)

pallets/aleph/src/lib.rs
mike1729: Heavily! I'll change this and pallet elections doc.

pallets/elections/Cargo.toml
mike1729: cargo new leftover...

pallets/elections/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    #[pallet::type_value]
+    pub(super) fn DefaultForMillisecsPerBlock() -> u64 {
+        DEFAULT_MILLISECS_PER_BLOCK
+    }
+
+    #[pallet::storage]
+    #[pallet::getter(fn millisecs_per_block)]
+    pub(super) type MillisecsPerBlock<T: Config> =
+        StorageValue<_, u64, ValueQuery, DefaultForMillisecsPerBlock>;
```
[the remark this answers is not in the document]
DamianStraszak: This is also related to the other comment I had about initializing the storage items for `millisecs_per_block` etc. Essentially, if we were able to do it like that then we would not need these storage items anymore (only `members`) and the trouble would be gone.

pallets/elections/src/lib.rs
mike1729: _

pallets/elections/src/mock.rs
mike1729: You're right, forgot to change here.

primitives/src/lib.rs
mike1729: Changed.

pallets/elections/src/lib.rs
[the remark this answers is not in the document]
mike1729: They are strorage items so that they may be changed via command line option. I'm all for removing them.

pallets/elections/src/lib.rs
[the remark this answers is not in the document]
mike1729: Maybe a notion doc would be a better place to describe it?

bin/runtime/src/lib.rs
mike1729: Don't remember now but I've seen some convoluted macro magic that allows keys in opaque_keys macro only if this trait is implemented. I'll try to find it.

pallets/aleph/src/lib.rs
mike1729: This is fixed now.

e2e-tests/src/test/treasury.rs
Marcin-Radecki: Author (myself) reasoning was that potentially (although unlikely) we can span this test among more than one era in e2e tests, so we can be different in terms of fee comparison or total issuance different by `K*possibly_treasury_gain_from_staking`, where `K` is era span difference.

bin/runtime/src/lib.rs
```diff
@@ -504,20 +582,29 @@ construct_runtime!(
...
-        Session: pallet_session::{Pallet, Call, Storage, Event, Config<T>},
-        Scheduler: pallet_scheduler::{Pallet, Call, Storage, Event<T>},
-        Treasury: pallet_treasury::{Pallet, Call, Storage, Config, Event<T>},
-        Vesting: pallet_vesting::{Pallet, Call, Storage, Event<T>, Config<T>},
-        Multisig: pallet_multisig::{Pallet, Call, Storage, Event<T>},
-        Utility: pallet_utility::{Pallet, Call, Storage, Event},
+        System: frame_system::{Pallet, Call, Config, Storage, Event<T>} = 0,
+        RandomnessCollectiveFlip: pallet_randomness_collective_flip::{Pallet, Storage} = 1,
+        Scheduler: pallet_scheduler::{Pallet, Call, Storage, Event<T>} = 2,
+
```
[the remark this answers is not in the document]
timorl: I'm still unclear as to why there are whitespaces between them.

bin/runtime/src/lib.rs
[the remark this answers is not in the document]
timorl: Still wrong, all lines but the first should be indented one deeper. Dx

e2e-tests/src/accounts.rs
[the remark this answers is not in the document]
timorl: Ah, you can make `default_accounts` return `Vec<KeyPair>` directly, and then again the code below becomes even simpler.

e2e-tests/src/accounts.rs
```diff
@@ -5,26 +5,36 @@ use substrate_api_client::Balance;
...
+            .map(keypair_from_string)
+            .collect(),
+    }
+}
+
+pub fn get_sudo(config: &Config) -> KeyPair {
+    match &config.sudo {
         Some(seed) => keypair_from_string(seed),
-        None => accounts_from_seeds(config.seeds)[0].to_owned(),
+        None => accounts_from_seeds(Some(default_accounts()).as_ref())[0].clone(),
```
timorl, line 37: Huh, why is there an `as_ref` here suddenly?

pallets/elections/src/tests.rs
timorl: I would expect so, but if it would be lots of work then feel free to resolve this instead.

pallets/elections/src/lib.rs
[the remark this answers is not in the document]
DamianStraszak: Hmm, yes, it would certainly be more convenient to read. But then it would still be good to somehow indicate in the source where is this documented... (Not sure a link would make sense though, as github is public and notion is private...)

bin/runtime/src/lib.rs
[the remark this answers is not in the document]
DamianStraszak: But how does it actually work? How would the runtime know about forks?

pallets/aleph/src/migrations/v1_to_v2.rs
mike1729: Added logs.

pallets/elections/src/lib.rs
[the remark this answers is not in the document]
mike1729: I don't see why it should be noted here. Even stronger, for me such comments don't fit here. I would document how to achieve some goals by combination of various call in our docs not in here.

pallets/elections/src/lib.rs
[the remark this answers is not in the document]
mike1729: Right.

e2e-tests/src/accounts.rs
mike1729: Since `accounts_from_seed` require `Option<&Vec<_>_`.

pallets/elections/src/lib.rs
mike1729: Because they are instantiated in runtime during compilation.

mike1729: Because they are instantiated in runtime during compilation.

e2e-tests/src/accounts.rs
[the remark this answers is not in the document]
timorl: Then `default_account_seeds` is a slightly better name imo.

Also (related to the comment below that I resolved and don't want to risk unresolving, 'cause RAM), I think you could make this function take an option of a vector, rather than a reference to a vector, since afaict in all uses you have to go out of your way to change a vector into a reference.

e2e-tests/src/test/treasury.rs
timorl: I like this cautious approach, and I think it would be made better by making it more explicit. Could you create a function that takes all the needed data, asserts this, but then also logs the somewhat unexpected case when this is unequal, including how many eras have apparently passed (the `K`)? It would mostly make the code here much clearer, and sometimes it might end up being more useful by letting us know something unexpected happened.

bin/runtime/src/lib.rs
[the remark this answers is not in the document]
timorl: Does that happen with this version of the code? Do we really want to have this info in blocks? And what exactly does this number mean – how long the offshoot branch can be before we start disregarding blocks?

bin/node/src/chain_spec.rs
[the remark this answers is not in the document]
timorl: I resolved the other comments, but I think all this should really be tackled in some task. Right now we have essentially test code very entangled with production code, and that is always worrying.

bin/runtime/src/lib.rs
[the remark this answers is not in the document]
mike1729: > Does that happen with this version of the code?

Yes it does.

> Do we really want to have this info in blocks?

Ethereum has it and it's well received.

> And what exactly does this number mean – how long the offshoot branch can be before we start disregarding blocks?

It means that `N`th block we accept in that block uncles of numbers greater than `N-UncleGenerations`.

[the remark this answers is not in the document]
timorl: > > Do we really want to have this info in blocks?
> 
> Ethereum has it and it's well received.

Ethereum is PoW and we have PoS committees, I think most of the reasons for this making sense in Ethereum don't apply to us.
> 
> > And what exactly does this number mean – how long the offshoot branch can be before we start disregarding blocks?
> 
> It means that `N`th block we accept in that block uncles of numbers greater than `N-UncleGenerations`.

So if I create an arbitrary length parallel branch all the blocks in there can be considered good, they just have to be included in many blocks?

pallets/aleph/src/lib.rs
mike1729: Changed.

bin/runtime/src/lib.rs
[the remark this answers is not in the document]
timorl: > But the one reason that there can be some poor network conditions or some block take more time to process are still valid. I don't have strong opinion here, though and can set it to 0.

I'm not sure, this is one of the things that should've been cleared earlier. >.<
> 
> No, the parent of an uncle has to be in the db.
Oh, so they are at most 1 away from the chain? That makes me less worried, although I still feel this should be an official decision where we actually considered the consequences.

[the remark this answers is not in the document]
mike1729: Ok, so setting it to `0` and preparing an ADR.

e2e-tests/src/test/treasury.rs
Marcin-Radecki: Sure will do.

bin/node/src/chain_spec.rs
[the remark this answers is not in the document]
mike1729: chain_spec.rs contains now only test code, as we already generated chainspecs for testnet and mainnet.

pallets/elections/src/lib.rs
DamianStraszak: I think there is a misunderstanding here. How in your opinion is `Elections::session_period()` executed?

pallets/elections/src/lib.rs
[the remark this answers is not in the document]
DamianStraszak: Fine, makes sense. The only thing I would like to make sure is that we have a precise documentation what this call does, so that in case of emergency we can look it up.

bin/runtime/src/lib.rs
DamianStraszak: Yeah, I saw it once and I don't with anyone to look at it again :P I'm resolving :)

pallets/elections/src/lib.rs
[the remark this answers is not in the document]
DamianStraszak: Can you make a jira task for it? We must just leave it for later, I guess :/

pallets/elections/src/lib.rs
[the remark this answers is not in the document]
mike1729: Added a notion doc.

pallets/elections/src/lib.rs
mike1729: Changed to constants.

pallets/elections/src/lib.rs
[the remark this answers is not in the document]
mike1729: Changed to constants.

bin/node/src/chain_spec.rs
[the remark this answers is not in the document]
mike1729: Added a jira task for a cleanup.

scripts/change_consts.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -e
+
+mpb="DEFAULT_MILLISECS_PER_BLOCK: u64 ="
```
DamianStraszak, line 5: I would remove the prefix `DEFAULT_` since it is not configurable other than replacing this value in the code :) 

Also there is a problem that once you run `./run_nodes.sh` then the constants in this file change, and the change is not reverted. It's not that I know a solution to this, just wanted to point out that this might be problematic ;/

DamianStraszak: Generally I think, there should be a follow-up task to figure out how to properly build this for test scenarios... (So It might be ok to leave it as it is).

bin/runtime/src/lib.rs
```diff
@@ -345,50 +335,89 @@ impl_opaque_keys! {
 }
 
 parameter_types! {
-    pub const Offset: u32 = 0;
+    pub const MillisecsPerBlock: u64 = DEFAULT_MILLISECS_PER_BLOCK;
+    pub const SessionPeriod: u32 = DEFAULT_SESSION_PERIOD;
+}
+
+impl pallet_elections::Config for Runtime {
+    type SessionPeriod = SessionPeriod;
```
DamianStraszak, line 343: Why is this here and MillisecsPerBlock is not? It's probably fine, but I would like to understand myself...

bin/runtime/src/lib.rs
```diff
@@ -345,50 +335,89 @@ impl_opaque_keys! {
 }
 
 parameter_types! {
-    pub const Offset: u32 = 0;
+    pub const MillisecsPerBlock: u64 = DEFAULT_MILLISECS_PER_BLOCK;
```
DamianStraszak, line 338: I think there should be some kind of a warning in the code, that: 
Once you start a chain with a runtime that has given values of these two parameters, then they must stay so forever. If you update to a runtime with different values, then I think they will be ignored, but I'm not super confident -- something unexpected might as well happen.

DamianStraszak: I don't think this is a new thing because of this change, but this deserved a warning a long time ago already...

bin/runtime/src/lib.rs
mike1729: MillisecsPerBlock may be read of from pallet_timestamp but SessionPeriod cannot be read from pallet_session

bin/runtime/src/lib.rs
mike1729: I believe that after an upgrade the new version would take place. Why do you think otherwise?

PR #301, opened 2022-02-14
Improve performance of staking e2e test

mike1729:


e2e-tests/src/main.rs
```diff
@@ -5,19 +5,20 @@ use clap::Parser;
...
     let config: Config = Config::parse();
 
-    run(test::finalization, "finalization", &config)?;
-    run(test::change_validators, "validators change", &config)?;
-    run(test::fee_calculation, "fee calculation", &config)?;
-    run(test::token_transfer, "token transfer", &config)?;
-    run(test::channeling_fee, "channeling fee", &config)?;
-    run(test::treasury_access, "treasury access", &config)?;
-    run(test::batch_transactions, "batch_transactions", &config)?;
+    // run(test::finalization, "finalization", &config)?;
```
Marcin-Radecki, line 15: Please uncomment before merge

e2e-tests/src/test/staking.rs
```diff
@@ -172,35 +189,42 @@ fn get_key_pairs() -> (Vec<KeyPair>, Vec<KeyPair>) {
     (stashes_accounts_key_pairs, validator_accounts_key_pairs)
 }
 
+// 1. endow stash accounts balances, controller accounts are already endowed in chainspec
```
Marcin-Radecki, line 192: This is not strictly true as we're passing `a0dnet1-e2e` when generating chainspec , thus we start from the scratch (ie no bond() and validate() calls in the first place).

Marcin-Radecki: Nvm, I did not ready comment properly, sorry

Marcin-Radecki: Btw proper idea to add those comment here

e2e-tests/src/test/staking.rs
```diff
@@ -172,47 +149,64 @@ fn get_key_pairs() -> (Vec<KeyPair>, Vec<KeyPair>) {
...
+        .get_header::<Header>(None)
+        .unwrap()
+        .unwrap()
+        .number;
+    info!(
+        "Current block number is {}, waiting till it finalizes",
+        block_number,
+    );
+
+    wait_for_finalized_block(&connection, block_number);
```
Marcin-Radecki, line 209: This line raises a warning , please consider fixing it
```
warning: unused `Result` that must be used
   --> src/test/staking.rs:209:5
    |
209 |     wait_for_finalized_block(&connection, block_number);
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_must_use)]` on by default
    = note: this `Result` may be an `Err` variant, which should be handled
```

mike1729: fixed

e2e-tests/src/test/staking.rs
```diff
@@ -1,158 +1,125 @@
...
+fn wait_for_next_era(connection: &Connection) -> anyhow::Result<BlockNumber> {
+    let sessions_per_era: u32 = connection
+        .get_storage_value("Elections", "SessionsPerEra", None)
+        .unwrap()
+        .unwrap();
+    let current_era: u32 = connection
+        .get_storage_value("Staking", "ActiveEra", None)
+        .unwrap()
+        .unwrap();
+    let next_era = current_era + 1;
```
Marcin-Radecki, line 99: I think we need to wait one session more, `let next_era = current_era + 2`. This is because 
* in the `current era`, we set nominators and validators but settings are not yet in place, ie they are waiting to be elected but not yet participating in reward mechanism
* in `current_era + 1`, nominators and validators are working on rewards, and we need full era for that,
* in `current_era + 2`, we can schedule rewards for `current_era + 1`.
Please note how it was in previous version (https://github.com/Cardinal-Cryptography/aleph-node/commit/5fde521d183fdbdf27a9bd120008961b2d4dfb85#diff-ea9a4b230f3d7675b407201681c67f5708ab08d60192844a3c8935d9c4330c39R122) , yet this version us much better as previous one was obfuscated.

mike1729: I agree, I simplified too much ;) fixed

Marcin-Radecki (approved):
This is a superb rework. The code of tests looks very much readable now.

PR #302, opened 2022-02-15
new build process

fixxxedpoint:
Please read BUILD.md for details on how to use this new approach to our build process.
Tested on Ubuntu 20.04 LTS.


BUILD.md
```diff
@@ -0,0 +1,42 @@
+## TL;DR
+```
+docker build -t aleph-node/build -f docker/Dockerfile_build . && \
+docker run -ti --volume=$(pwd):/node/build aleph-node/build
```
DamianStraszak, line 4: Doesn't work for me without sudo.

bartoszjedrzejewski: Add your user to docker group and relog: https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user

DamianStraszak: The point is that this is a guide for users, and it should be completely idiot-proof. If that's a necessary step, then it should be part of the guide.

fixxxedpoint: also without `docker` ;)

fixxxedpoint: I am not sure with this. On one hand I don't want to encourage people to use docker via `sudo` and without `user-namaspaces`, on the other I don't how much detailed should be the dependencies described here.

DamianStraszak: Also `/node/nix-build.sh: line 28: patchelf: command not found`

bartoszjedrzejewski: Well, the issue is you can use docker as you want. In rootless mode, with your user added to the group or as a root. If we want to be clear, we can link to the doc that I commented above and let user decide.

fixxxedpoint: I don't want to make it to become `a guide on how to install docker`, so perhaps something like: requirements - being able to run `docker` locally (I don't care whether you are using root account).

DamianStraszak: Ideally the TL-DR should also specify where I can find the binary afterwards.

fixxxedpoint: `patchelf` thing should be fixed now - sorry for `force` push. I just realized last moment that it shouldn't be part of the `nix.shell` script, which should just describe how to build it on nix (that patchelf there was making it broken).

DamianStraszak: I can tell you with high certainty that what a user wants here is a list of commands to run and the location of the binary afterwards. That's it. They just want it to work.

An explanation what the commands do and stuff is a nice bonus, but most users will ignore this part unless something is not working properly. Familiarity with docker might vary.

fixxxedpoint: Sounds cool, but also little naive. There is also nothing mentioned about which shell we are using throughout this guide, and for sure it want work under `fish` due to how we call `pwd`. Any idea what should be some sane requirements. I can just add `sudo su` at the start, but it won't look good.

DamianStraszak: Just take a look at other repos of comparable projects, like solana, near, polkadot to see how they do it. They have lots of users and their guides already got feedback and were refined a few times, so they are worth inspecting.

fixxxedpoint: I did :) https://github.com/paritytech/polkadot#build-from-source-with-docker

fixxxedpoint: `TL;DR` should now have a comment about where binary is stored afterwards.

fixxxedpoint: Actually, solana's and polkadot's build guides are really bad and totally not idion-proof.

DamianStraszak: OK, so I will not insist to add `sudo` here, because as you say it's stupid, but mark me: normally when building they give you at least 2 alternatives: either you build with `cargo`, or with `docker`. If I don't like docker, I go ahead with cargo or whatever other alternative. Here I'm forced to use docker, and the commands, as given don't work out of the box :)

DamianStraszak: After building
`damian@damian-HP:~/(...)/aleph-node$ ./aleph-node --help`
`./aleph-node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./aleph-node)`

DamianStraszak: I would expect that `BUILD.md` comprehensively explains how to build, and not fall back to the building section of `README.md`, unless I misunderstood the point of this guide.

fixxxedpoint: `We also prepared a configuration for the `nix` package manager that automates this process. For details please consult the [BUILD][BUILD] guide.` So, it's not a comprehensive, all types of preferences and systems, type of guide.

DamianStraszak: Ok, fine. I think I misunderstood what this PR is supposed to do, sorry for that. I thought it solves `AZ-524`, but now I see that not quite.  You can ignore my comments -- they would make sense for a general build guide, not a guide for the nix build process.

fixxxedpoint: I am aware of this problem with `glibc` - I will try to figure it out somehow. Since it seems it is our only runtime dependency, we should make it as requirement (can be also the nix way), try to statically build with it, or allow user to link with its local glibc.

fixxxedpoint: I think your interpretation of `AZ-524` is just different than my. That's why I also made that document on notion describing it, to see what are others opinions. It's not wrong or anything and I will try to enhance this document.

DamianStraszak: What I'm saying is that this dockerized nix build method is great, and I think many people will use it, but there will also be a high percentage of users who will want to just run `cargo build` without docker. So ideally the build guide should have two sections: 1) using cargo, 2) using nix/docker. And while 2) is done, I think 1) requires some work, because atm the instructions from 1) link to a not quite up-to-date external guide. We should ideally include complete steps (installing rust and dependencies) in our readme (or build guide, whatever works) for 1).

fixxxedpoint: Please see the new version of BUILD.md

docker/nix-build.sh
```diff
@@ -0,0 +1,30 @@
...
+    echo "Usage:
+      ./nix-build.sh [-s - spawn nix-shell]"
+}
+
+if [ $SPAWN_SHELL = true ]
+then
+    nix-shell --pure $SHELL_NIX_FILE
+else
+    nix-build $SHELL_NIX_FILE
+    patchelf --set-interpreter /lib64/ld-linux-x86-64.so.2 ./result/bin/aleph-node
```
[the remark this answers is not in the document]
fixxxedpoint: Added a comment for it.

default.nix
```diff
@@ -0,0 +1,99 @@
...
+    });
+  nixpkgs = import (builtins.fetchTarball {
+    url = "https://github.com/NixOS/nixpkgs/archive/refs/tags/21.11.tar.gz";
+    sha256 = "162dywda2dvfj1248afxc45kcrg83appjd0nmdb541hl7rnncf02";
+  }) { overlays = [ rustOverlay ]; };
+  rust-nightly = with nixpkgs; ((rustChannelOf { date = "2021-10-24"; channel = "nightly"; }).rust.override {
+    extensions = [ "rust-src" ];
+    targets = [ "x86_64-unknown-linux-gnu" "wasm32-unknown-unknown" ];
+  });
+  nixpkgs-old = import (builtins.fetchTarball {
```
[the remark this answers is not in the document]
fixxxedpoint: Current version is not using this anymore - versions were incompatible.

default.nix
```diff
@@ -25,15 +26,16 @@ let
     };
     patches = [];
   });
-  llvm = nixpkgs.llvmPackages_13;
-  llvmVersionString = "13.0.0";
+
+  # declares a build environment where C and C++ compilers are delivered by the llvm/clang project
+  # in this version build process should rely only on clang, without access to gcc
```
timorl, line 30: Please add some short explanation why this is important, something like "some dependencies default to gcc if it is present at all, ignoring any settings to the contrary, and we want to avoid that"

default.nix
```diff
@@ -1,22 +1,23 @@
 let
+  # this overlay allows us to use a specified version of the rust toolchain
   rustOverlay =
     import (builtins.fetchGit {
       url = "https://github.com/mozilla/nixpkgs-mozilla.git";
       rev = "f233fdc4ff6ba2ffeb1e3e3cd6d63bb1297d6996";
     });
+
+  # pinned version of nix packages
```
timorl, line 9: An explanation as to why we are using this version (not the newest at time of writing) would be useful (points about compatibility with Ubuntu LTS; does that even work now?)

timorl (changes requested):
Thanks! `BUILD.md` is now much better, and you can tell by the fact that I have very specific criticisms of it. ;>

BUILD.md
DamianStraszak: That looks pretty good, but still the `manual way` lacks concrete steps. I'm finding it hard to understand why would you choose to write a longish explanation why things make not work instead of just giving concrete steps that work 90 out of 100 times and 100 out of 100 times on a popular linux distribution like ubuntu. What we need here, is 1) how to install rust/cargo and git, 2) something like "Compiling and Running a Node without Docker" here https://docs.near.org/docs/develop/node/validator/running-a-node .

timorl: I think there are good reasons for putting the disclaimers there – we don't really want to support this, even if itusually works. Although I do agree on adding clearer steps.

Perhaps leave the longish explanation in place and add clearer steps after?

DamianStraszak: Yeah, the explanation itself makes sense, but not if the steps are not there :)

timorl (approved):
Very cool.

A follow up task about making our testing infra use this build system would be nice.

BUILD.md
fixxxedpoint: Please see the new version of BUILD.md

bartoszjedrzejewski (approved):

PR #306, opened 2022-02-18
AZ-593: Fix validators connections in a large networks

Yithis:


finality-aleph/src/network/manager/discovery.rs
```diff
@@ -3,40 +3,32 @@ use crate::{
...
-/// How many nodes we should query about unknown authorities in one go.
-const NODES_TO_QUERY: usize = 2;
-
 /// Messages used for discovery and authentication.
 #[derive(Clone, Debug, PartialEq, Encode, Decode)]
 pub enum DiscoveryMessage {
     AuthenticationBroadcast(Authentication),
-    // Requests always contain own authentication, to avoid asymmetric trust.
-    Request(Vec<NodeIndex>, Authentication),
     // Always assumed to contain only authentications for one session.
```
Marcin-Radecki, line 19: This comment needs to be adapted afaik , as it was more suitable for `Vec`

finality-aleph/src/network/manager/discovery.rs
```diff
@@ -259,12 +140,10 @@ impl Discovery {
             AuthenticationBroadcast(authentication) => {
                 self.handle_broadcast(authentication, handler)
             }
-            Request(node_ids, authentication) => {
-                self.handle_request(node_ids, authentication, handler)
-            }
-            Authentications(authentications) => {
-                (self.handle_response(authentications, handler), Vec::new())
-            }
+            Authentication(authentication) => (
```
Marcin-Radecki, line 143: It took me 10 secs to realize that we're returning tuple from here, maybe below formatting is more readable?
``` rust
 match message {
            AuthenticationBroadcast(authentication) => {
                self.handle_broadcast(authentication, handler)
            }
            Authentication(authentication) =>
                (self.handle_authentication(authentication, handler), Vec::new()),
           }
}
```

finality-aleph/src/network/manager/discovery.rs
```diff
@@ -472,9 +318,9 @@ mod tests {
         );
         assert_eq!(commands.len(), 1);
         assert!(matches!(&commands[0], (
-                DiscoveryMessage::Authentications(authentications),
+                DiscoveryMessage::Authentication(authentication),
                 DataCommand::SendTo(_, _),
-            ) if authentications == &vec![handler.authentication().unwrap()]));
+            ) if *authentication == handler.authentication().unwrap()));
```
Marcin-Radecki, line 323: I was wondering, do we need to dereference `authentication`? Its type is `Authentication`, which is a tuple

finality-aleph/src/network/manager/discovery.rs
```diff
@@ -498,181 +344,32 @@ mod tests {
             ) if rebroadcast_authentication == &authentication)));
     }
 
-    #[tokio::test]
-    async fn responds_to_correct_request_when_can() {
```
Marcin-Radecki, line 502: Btw why do we remove some tests and adapt others? Test coverage is the same right, we just remove tests that covered functionalities no longer exist.

finality-aleph/src/network/manager/service.rs
```diff
@@ -42,13 +42,34 @@ struct Session<D: Data> {
...
-struct PreSession {
+struct PreValidatorSession {
     session_id: SessionId,
     verifier: AuthorityVerifier,
     node_id: NodeIndex,
     pen: AuthorityPen,
 }
 
+#[derive(Clone)]
+struct PreNonvalidatorSession {
```
Marcin-Radecki, line 53: Not sure what's good practice in Rust in this case, but in other programming languages, I worked in the past which has type inheritance, I'd consider making some base class with fields `session_id` and `verifier` as we can see both `PreValidatorSession` and `PreNonValidatorSession` have those in common - nevertheless inheritance as such is not a popular concept nowadays. In Rust, I understand we can use composition (like make base type `Validator` which have those two fields and make `Validator` a field of  `PreValidatorSession`), but it might be too much hustle for this little duplication.

finality-aleph/src/network/service.rs
```diff
@@ -145,24 +173,44 @@ impl<N: Network, D: Data> Service<N, D> {
...
+                    let (tx, rx) = mpsc::channel(PEER_BUFFER_SIZE);
+                    self.spawn_handle.spawn(
+                        "aleph/network/peer_sender",
+                        None,
+                        self.peer_sender(remote.into(), rx),
+                    );
+                    self.validator_connected_peers.insert(remote.into());
+                    self.validator_peer_senders.insert(remote.into(), tx);
+                }
+                Err(_) => {}
```
Marcin-Radecki, line 199: We don't need to at least log in that situations right (to avoid silent errors); or are those really unreachable paths

finality-aleph/src/network/service.rs
```diff
@@ -843,6 +896,312 @@ mod tests {
         test_data.cleanup().await
     }
 
+    #[tokio::test]
+    async fn test_generic_data_command_send_to() {
```
Marcin-Radecki, line 900: Do you think those new tests can be grouped in pairs with existing ones to extract some base methods and reuse them in those two flavors? E.g. extract this test and `test_validator_data_command_send_to` to come common either parametrized or generic base method

Marcin-Radecki (approved):
Just minor comments

finality-aleph/src/network/service.rs
```diff
@@ -25,8 +26,10 @@ pub struct Service<N: Network, D: Data> {
     messages_from_user: mpsc::UnboundedReceiver<(D, DataCommand)>,
     messages_for_user: mpsc::UnboundedSender<D>,
     commands_from_manager: mpsc::UnboundedReceiver<ConnectionCommand>,
-    connected_peers: HashSet<PeerId>,
-    peer_senders: HashMap<PeerId, mpsc::Sender<(D, Protocol)>>,
+    generic_connected_peers: HashSet<PeerId>,
+    validator_connected_peers: HashSet<PeerId>,
+    generic_peer_senders: HashMap<PeerId, mpsc::Sender<(D, Protocol)>>,
+    validator_peer_senders: HashMap<PeerId, mpsc::Sender<(D, Protocol)>>,
```
timorl, line 32: Since you now start them per protocol, you don't need to pass the protocol to them. This requires some changes to `peer_sender`, but you can remove `Protocol` from the pair here.

finality-aleph/src/network/service.rs
```diff
@@ -145,24 +173,44 @@ impl<N: Network, D: Data> Service<N, D> {
...
+                        trace!(target: "aleph-network", "NotificationStreamClosed event for peer {:?}", remote);
+                        self.generic_connected_peers.remove(&remote.into());
+                        self.generic_peer_senders.remove(&remote.into());
+                    }
+                    Ok(Protocol::Validator) => {
+                        trace!(target: "aleph-network", "NotificationStreamClosed event for peer {:?}", remote);
+                        self.validator_connected_peers.remove(&remote.into());
+                        self.validator_peer_senders.remove(&remote.into());
+                    }
+                    Err(_) => {}
```
timorl, line 213: Same as in the above thread started by Marcin – a comment about why we don't care about other protocols would be useful.

finality-aleph/src/network/service.rs
```diff
@@ -843,6 +896,312 @@ mod tests {
...
+                .send_message
+                .1
+                .lock()
+                .await
+                .by_ref()
+                .take(all_authorities_n - closed_authorities_n)
+                .collect::<Vec<_>>()
+                .await
+                .iter()
+                .cloned(),
```
timorl, line 1059: Shouldn't the last two lines here be `into_iter` instead?

timorl (changes requested):
Generally looks good, some minor comments.

finality-aleph/src/network/manager/service.rs
timorl: I think the duplication here is more coincidental than meaningful – even if we created a common substruct we couldn't use the commonality for anything in the rest of the code, because we do very different things with these two cases.

Or at least it would require a major refactor to actually use the commonality – if you have an idea for such broader restructuring I would be quite interested. It might be very viable after this round of changes, since now I'm no longer confident the design avoids major redundancies.

finality-aleph/src/lib.rs
```diff
@@ -60,7 +60,12 @@ pub fn peers_set_config(protocol: Protocol) -> sc_network::config::NonDefaultSet
             reserved_nodes: Vec::new(),
             non_reserved_mode: sc_network::config::NonReservedPeerMode::Deny,
         },
-        Protocol::Generic => sc_network::config::SetConfig::default(),
+        Protocol::Generic => sc_network::config::SetConfig {
+            in_peers: 4,
+            out_peers: 4,
```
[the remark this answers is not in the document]
Yithis: Done

finality-aleph/src/network/manager/discovery.rs
Yithis: I don't know if any comment is necessary now, since there are no assumptions for single authentication

finality-aleph/src/network/manager/discovery.rs
```diff
@@ -3,40 +3,32 @@ use crate::{
...
 #[derive(Clone, Debug, PartialEq, Encode, Decode)]
 pub enum DiscoveryMessage {
     AuthenticationBroadcast(Authentication),
-    // Requests always contain own authentication, to avoid asymmetric trust.
-    Request(Vec<NodeIndex>, Authentication),
     // Always assumed to contain only authentications for one session.
     // Only authentications from the same session as the first present are guaranteed to be
     // processed.
-    Authentications(Vec<Authentication>),
+    Authentication(Authentication),
```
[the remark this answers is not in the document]
Yithis: Look at Marcin comment above

finality-aleph/src/network/manager/discovery.rs
Yithis: Done

finality-aleph/src/network/manager/discovery.rs
Yithis: It is exactly what I did. I had to delete every test that was connected to`DiscoveryMessage::Request` and adapt those connected to `DiscoveryMessage::Authentication(s)`

finality-aleph/src/network/service.rs
Yithis: Done

finality-aleph/src/network/service.rs
```diff
@@ -103,9 +125,10 @@ impl<N: Network, D: Data> Service<N, D> {
         }
     }
 
-    fn send_to_peer(&mut self, data: D, peer: PeerId, protocol: Protocol) {
-        if let Some(sender) = self.peer_senders.get_mut(&peer) {
+    fn send_to_peer(&mut self, data: D, peer: PeerId, protocol: Protocol) -> Result<(), SendError> {
+        if let Some(sender) = self.get_sender(&peer, protocol.clone()) {
             if let Err(e) = sender.try_send((data, protocol)) {
+                println!("{}", e);
```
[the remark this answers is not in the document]
Yithis: Done

finality-aleph/src/network/service.rs
```diff
@@ -145,24 +173,44 @@ impl<N: Network, D: Data> Service<N, D> {
...
+                Err(_) => {}
+            },
             Event::NotificationStreamClosed { remote, protocol } => {
-                if protocol == ALEPH_PROTOCOL_NAME {
-                    trace!(target: "aleph-network", "NotificationStreamClosed event for peer {:?}", remote);
-                    self.connected_peers.remove(&remote.into());
-                    self.peer_senders.remove(&remote.into());
+                match protocol.as_ref().try_into() {
+                    Ok(Protocol::Generic) => {
+                        trace!(target: "aleph-network", "NotificationStreamClosed event for peer {:?}", remote);
```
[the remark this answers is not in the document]
Yithis: Done

finality-aleph/src/network/service.rs
Yithis: Done

finality-aleph/src/network/service.rs
Yithis: I can do it, but I don't think that such version is better. When current tests fail you already get a few functions to check, I don't want to add more layers

finality-aleph/src/network/service.rs
Yithis: I don't know, you accepted it in other pr xd

Yithis: Done

timorl (approved):
<3

Marcin-Radecki (approved):

PR #307, opened 2022-02-18
AZ-568: Dynamic transaction fees

pmikolajczyk41:
This PR introduces dynamic transaction fees. We make a use of `TargetedFeeAdjustment` (already present in Substrate). Parameters that we agreed upon:
 - minimum multiplier: 1.0
 - target saturation level: 25%
 - variability: 0.067


mike1729 (approved):

bin/runtime/src/lib.rs
```diff
@@ -302,11 +280,21 @@ impl OnUnbalanced<NegativeImbalance> for EverythingToTheTreasury {
     }
 }
 
+parameter_types! {
+    // We expect that on average 25% of the normal capacity will be occupied with normal txs.
+    pub TargetSaturationLevel: Perquintill = Perquintill::from_percent(25);
+    // During 20 blocks the fee may not change more than by 100%.
```
h4nsu, line 286: Nit: 67/1000 implying the fee doubling over 20 blocks sounds totally mysterious. Adding here a bit of explanation (formula? link?) would be a favor to our future selves ;)

h4nsu (approved):

bin/runtime/src/lib.rs
pmikolajczyk41: sure, I referred to the source formulas

PR #310, opened 2022-02-22
AZ-620: Extract managing session map from party

kostekIV:
This extract the logic of updating the mapping of the session to its authorities from the party code. This is needed because this map will also be needed by node that does not take a part in the consensus.


finality-aleph/src/party/mod.rs
```diff
@@ -501,6 +435,20 @@ where
                 }
             }
         }
+
+        loop {
```
Yithis, line 439: Maybe a comment here why it is not an infinite loop would be good. But also as far as I understand now it is possible to start a session without knowing an authorities for it and such behavior won't end up as `panic`

Yithis (commented):
Honestly I'm not a fan of creating traits like `FinalityNotificator` and `AuthorityProvider`, since imo they are complicated ways to call `client` functions. Personaly they are making code less readable for me, but I am aware it can only be for me. If you really want to make SessionMapUpdater easier to test, I would probably prefer solution closer to something like this for example:
Since  `finality_notificator.notification_stream()` is called only once it can be either a parameter of `run` as struct or a parameter of `run` as function, that will be called inside `run`.

finality-aleph/src/session_map.rs
```diff
@@ -0,0 +1,465 @@
...
+                // we have finalized first block of some session, now we can put the next known authorities into session map
+                self.handle_first_block_of_session(*last_finalized, session_id)
+                    .await;
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
kostekIV: The only things used from the mock module are type definitions as far as I can see. Other than that the `TestingClient` from substrate is used to produce blocks as data for testing not as a part of the component under the test.

finality-aleph/src/party/mod.rs
kostekIV: I will add a comment here, only the first ever session can have such behaviour. Also this loop is here to  prevents from advancing to session start without knowing the authorities

kostekIV:
> Honestly I'm not a fan of creating traits like `FinalityNotificator` and `AuthorityProvider`, since imo they are complicated ways to call `client` functions. Personaly they are making code less readable for me, but I am aware it can only be for me. If you really want to make SessionMapUpdater easier to test, I would probably prefer solution closer to something like this for example: Since `finality_notificator.notification_stream()` is called only once it can be either a parameter of `run` as struct or a parameter of `run` as function, that will be called inside `run`.

I am actually a fan of them, they clearly show what they are doing, make testing much easier as we don't need to deal with clients at all in the unit tests to mock them. The solution with passing the stream would not be enough since we also need to have a way to mock the authorities client interface

finality-aleph/src/session_map.rs
```diff
@@ -0,0 +1,465 @@
...
+
+    if first_block_of_session::<B>(session, period) == num {
+        Some(session)
+    } else {
+        None
+    }
+}
+
+/// Struct responsible for updating session map
+pub struct SessionMapUpdater<AP, FN, B>
```
Marcin-Radecki, line 200: I was thinking if logic in this struct can be a part of `SessionMap`? This is because normally we encapsulate a class behavior in the class itself, not in a separate class. I understand we have a distinction for `SharedSessionMap` and `ReadOnlySessionMap`, hence the need for another class. I'm just thinking of justification for having three separate structs. Alternative would be to 
* rename `SessionMapUpdater` to `SessionMap`, and incorporate current `SharedSessionMap` logic into it,
* do not distinguish on `SharedSessionMap` and `ReadOnlySessionMap`, have only one type `SessionMap` and have all the mutating/non-mutating logic encapsulated in renamed `SessionMap` struct

Marcin-Radecki:
> > Honestly I'm not a fan of creating traits like `FinalityNotificator` and `AuthorityProvider`, since imo they are complicated ways to call `client` functions. Personaly they are making code less readable for me, but I am aware it can only be for me. If you really want to make SessionMapUpdater easier to test, I would probably prefer solution closer to something like this for example: Since `finality_notificator.notification_stream()` is called only once it can be either a parameter of `run` as struct or a parameter of `run` as function, that will be called inside `run`.
> 
> I am actually a fan of them, they clearly show what they are doing, making testing much easier as we don't need to deal with clients at all in the unit tests to mock them. The solution with passing the stream would not be enough since we also need to have a way to mock the authorities client interface

I think it's standard way of doing UTs - make interfaces (traits in Rust nomenclature) and mock them in UTs.

Marcin-Radecki (approved):
I think this one comment is optional, this is solid rework.

finality-aleph/src/party/mod.rs
kostekIV: Added

finality-aleph/src/session_map.rs
kostekIV: Now with the additional subscriptions to insertions to the map, I think it makes sense to have such a struct (so the updater does not need to deal with it and only with the 'updating' logic) Wdyt?

finality-aleph/src/party/mod.rs
kostekIV: actually we no longer have a loop here, now we are waiting for a notification here

finality-aleph/src/session_map.rs
```diff
@@ -0,0 +1,528 @@
...
+    oneshot::{Receiver as OneShotReceiver, Sender as OneShotSender},
+    Mutex,
+};
+
+type SessionMap = HashMap<SessionId, Vec<AuthorityId>>;
+type SessionSubscribers = HashMap<SessionId, Vec<OneShotSender<()>>>;
+
+pub trait AuthorityProvider<B> {
+    /// returns authorities for block `b`
+    fn authorities(&self, b: B) -> Option<Vec<AuthorityId>>;
```
fbielejec, line 25: can we call the argument explicitly `block`?

finality-aleph/src/session_map.rs
```diff
@@ -0,0 +1,528 @@
...
+    /// returns next session authorities where current session is for block `b`
+    fn next_authorities(&self, b: B) -> Option<Vec<AuthorityId>>;
```
fbielejec, line 27: same

finality-aleph/src/session_map.rs
kostekIV: done

finality-aleph/src/session_map.rs
kostekIV: done

finality-aleph/src/session_map.rs
```diff
@@ -0,0 +1,528 @@
...
+
+    /// returns an end of the oneshot channel that fires a message if either authorities are already
+    /// known for the session with id = `id` or when the authorities are inserted for this session.
+    pub async fn subscribe_to_insertion(&self, id: SessionId) -> OneShotReceiver<()> {
+        let (sender, receiver) = tokio::sync::oneshot::channel();
+
+        let mut guard = self.inner.lock().await;
+
+        if guard.0.contains_key(&id) {
+            // if the value is already present notify immidiatelly
```
fbielejec, line 192:
```suggestion
            // if the value is already present notify immediately
```

finality-aleph/src/session_map.rs
```diff
@@ -0,0 +1,528 @@
...
+
+#[derive(Clone)]
+/// Wrapper around Mapping from sessionId to Vec of AuthorityIds allowing mutation
+/// and hiding locking details
+struct SharedSessionMap(Arc<Mutex<(SessionMap, SessionSubscribers)>>);
+
+#[derive(Clone)]
+/// Wrapper around Mapping from sessionId to Vec of AuthorityIds allowing only reads
+pub struct ReadOnlySessionMap {
+    inner: Arc<Mutex<(SessionMap, SessionSubscribers)>>,
```
fbielejec, line 138: maybe a good candidate for a use of RwLock, if you see some read concurrency ever happening?

kostekIV: done

finality-aleph/src/session_map.rs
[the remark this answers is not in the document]
Yithis: So please export them, but don't make whole mock module public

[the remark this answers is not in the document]
kostekIV: changed to `pub(crate)`

fbielejec (approved):

finality-aleph/src/session_map.rs
Marcin-Radecki: Yup, looks good, thanks.

PR #312, opened 2022-02-24
Basic staking

mike1729:


bin/runtime/Cargo.toml
```diff
@@ -16,17 +16,21 @@ serde = { version = "1.0.101", optional = true, features = ["derive"] }
...
 frame-system-rpc-runtime-api = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13"}
 
 pallet-aura = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13"}
+pallet-authorship = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13"}
 pallet-balances = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13"}
 pallet-randomness-collective-flip = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13"}
 pallet-session = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13"}
 pallet-scheduler = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13"}
+pallet-staking = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13"}
+pallet-staking-reward-curve = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13"}
```
DamianStraszak, line 33: If I'm not mistaken, this dependency is not used.

DamianStraszak (approved):

e2e-tests/src/test/treasury.rs
```diff
@@ -38,11 +58,16 @@ pub fn channeling_fee(config: Config) -> anyhow::Result<()> {
...
-        issuance_after - issuance_before
-    );
+    if issuance_after > issuance_before {
+        let difference = issuance_after - issuance_before;
+        assert_eq!(
+            difference % possibly_treasury_gain_from_staking,
+            0,
+            "Unexpectedly {} was minted, and it's not related to staking treasury reward which is {}",
+            difference, possibly_treasury_gain_from_staking
+        );
```
timorl, line 68: Shouldn't this also be checked using `check_treasury_balance`? Or a similar function if this one is not appropriate?

bin/runtime/src/lib.rs
```diff
@@ -384,15 +352,105 @@ impl<I: From<u32>> ::frame_support::traits::Get<I> for SessionPeriod {
...
+        let miliseconds_per_era =
+            MillisecsPerBlock::get() * SessionPeriod::get() as u64 * SessionsPerEra::get() as u64;
+        let portion = Perbill::from_rational(miliseconds_per_era, MILLISECONDS_PER_YEAR);
+        let total_payout = portion * YEARLY_INFLATION;
+        let validators_payout = Perbill::from_percent(90) * total_payout;
+        let rest = total_payout - validators_payout;
+
+        (validators_payout, rest)
+    }
+}
```
timorl, line 410: I think you might've wanted to use what got defined in `primitives` here? At least up to the two `const`s above.

timorl (approved):
Code wise seems ok, some minor comments.

bin/runtime/src/lib.rs
mike1729: This is a split glitch.

.github/scripts/run_consensus.sh
```diff
@@ -8,7 +8,10 @@ mkdir -p docker/data/
 source docker/env
 
 # Generate chainspec and populate comittee keystores
-docker run -v $(pwd)/docker/data:/data --entrypoint "/bin/sh" -e DAMIAN -e TOMASZ -e ZBYSZKO -e HANSU -e RUST_LOG=info aleph-node:latest -c "aleph-node bootstrap-chain --base-path /data --chain-id a0dnet1 --millisecs-per-block 2000 --session-period 10 --account-ids $DAMIAN,$TOMASZ,$ZBYSZKO,$HANSU --sudo-account-id $DAMIAN > /data/chainspec.json"
+# hidden assumption: --chain-id  must be different from a0dnet1, otherwise default chainspec would generate some
```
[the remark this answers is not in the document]
Marcin-Radecki: Yes, it's a hack really I agree. Do you mean the current comment needs to be adapted somehow? It already has `must be different than default chain-id` meaning.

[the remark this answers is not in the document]
mike1729: Added.

bin/runtime/src/lib.rs
mike1729: Fixed.

.github/scripts/run_consensus.sh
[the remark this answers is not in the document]
timorl: If I tried running the command without `--chain-id` at all I would expect that this is different than running it with `--chain-id "a0dnet"`, but it is actually the same. So this comment needs to explicitly mention that the flag has to be present _and_ set to something different.

e2e-tests/src/test/treasury.rs
Marcin-Radecki: Let me extract method for this one too.

mike1729:
Merging of this PR is postponed until #293 gets merged.

PR #313, opened 2022-02-25
Slim pallet aleph

mike1729:
Remove storage from pallet aleph that was used for tests and caused problems on production.


.github/workflows/build-node-and-runtime.yml
```diff
@@ -44,15 +44,37 @@ jobs:
...
+      - name: Upload test runtime
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ inputs.artifact-prefix }}aleph-test-runtime
+          path: target/release/wbuild/aleph-runtime/aleph_runtime.compact.wasm
+          if-no-files-found: error
+          retention-days: 7
+
+      - name: Stop cache
+        uses: ./.github/actions/post-cache
```
pmikolajczyk41, line 80: you want to remove previous step in line 41 which already stops the cache

.github/workflows/build-node-and-runtime.yml
```diff
@@ -44,15 +44,37 @@ jobs:
       - name: Upload binary
```
pmikolajczyk41, line 44: just for consistency: `Upload release binary`

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -19,34 +19,33 @@ concurrency:
 jobs:
   build-new-node:
     name: Build node and runtime artifacts (PR version)
-    uses: Cardinal-Cryptography/aleph-node/.github/workflows/build-node-and-runtime.yml@main
-
+    uses: ./.github/workflows/build-node-and-runtime.yml
```
pmikolajczyk41, line 22: +1

I'm quite sure that at the time of introducing reusable workflows to our repo, the only way of pointing workflow was through full git path :(

bin/node/Cargo.toml
```diff
@@ -68,3 +68,7 @@ substrate-build-script-utils = { git = "https://github.com/paritytech/substrate.
 
 [features]
 default = []
+short_session = [
+"aleph-runtime/short_session",
+"aleph-primitives/short_session"
```
pmikolajczyk41, line 73: add here some indent please

bin/runtime/src/lib.rs
```diff
@@ -104,19 +107,21 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("aleph-node"),
     impl_name: create_runtime_str!("aleph-node"),
     authoring_version: 1,
-    spec_version: 8,
+    spec_version: 9,
```
pmikolajczyk41, line 110: most probably you will be merging after #307 so remember to bump this to `10`

docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:focal-20210827
+FROM ubuntu:latest
```
pmikolajczyk41, line 1: uhm, using `latest` sometimes may lead to some unpleasant surprises; if you do not have strong incentive, I would either leave the previous 'sealed' one or choose newer, but still, 'sealed'

pallets/aleph/Cargo.toml
```diff
@@ -2,21 +2,21 @@
...
 serde = "1.0"
-primitives = { path = "../primitives", default-features = false}
-pallet-session = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13" }
 sp-io = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13" }
 scale-info = { version = "1.0", default-features = false, features = ["derive"] }
+primitives = { path = "../../primitives", default-features = false}
+pallet-balances = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13" }
+pallet-session = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13" }
 
 [dev-dependencies]
```
pmikolajczyk41, line 19: does this section with substrate dependencies only makes sense here?

pallets/aleph/src/mock.rs
```diff
@@ -33,7 +31,7 @@ construct_runtime!(
         System: frame_system::{Pallet, Call, Config, Storage, Event<T>},
         Balances: pallet_balances::{Pallet, Call, Storage, Config<T>, Event<T>},
         Session: pallet_session::{Pallet, Call, Storage, Event, Config<T>},
-        Aleph: pallet_aleph::{Pallet, Call, Config<T>, Storage, Event<T>},
+        Aleph: pallet_aleph::{Pallet,  Storage},
```
pmikolajczyk41, line 34: one space should be enough :)

pmikolajczyk41 (changes requested):
There are a lot of changes that (in my opinion) should be placed in a separate PR:
 - I would isolate all the rust-related changes (version, edition, dependencies)
 - I would isolate 'short-session' feature
 - what remains should form a more consistent PR

A general nit: until now we usually tried to keep the blank line at the end of every file (kinda unnecessary rule, but anyway followed). We should either drop it and announce that or (preferably) add the blank lines to all the files in this PR that are missing it (there are a couple of them).

(I'm still reviewing the core changes)

pallets/aleph/src/migrations/v1_to_v2.rs
```diff
@@ -0,0 +1,49 @@
...
+            log::info!(target: "pallet_aleph", "Storage item SessionPeriod does not exist!");
+        }
+        SessionPeriod::kill();
+
+        // store new version
+        StorageVersion::new(2).put::<P>();
+
+        T::DbWeight::get().reads(1) + T::DbWeight::get().writes(1)
+    } else {
+        log::warn!(
```
pmikolajczyk41, line 40: why `warn` here?

pallets/aleph/src/mock.rs
```diff
@@ -173,18 +174,26 @@ pub fn new_test_ext(authorities: &[(u64, u64)]) -> sp_io::TestExternalities {
...
-pub(crate) fn run_session(n: u64) {
-    while System::block_number() < n {
+pub(crate) fn run_session(n: u32) {
+    for i in Session::current_index()..n {
         Session::on_finalize(System::block_number());
         Aleph::on_finalize(System::block_number());
         System::on_finalize(System::block_number());
 
+        let parent_hash = if System::block_number() > 1 {
+            System::finalize().hash()
```
pmikolajczyk41, line 184: what are the reasons behind introducing `System::finalize().hash()` and `set_block_number` (below)?

pallets/elections/src/lib.rs
```diff
@@ -20,7 +20,7 @@ pub mod pallet {
     use frame_election_provider_support::{
         ElectionDataProvider, ElectionProvider, Support, Supports,
     };
-    use frame_support::pallet_prelude::*;
+    use frame_support::{pallet_prelude::*, traits::Get};
```
pmikolajczyk41, line 23: do you need this import?

pallets/aleph/src/tests.rs
```diff
@@ -0,0 +1,64 @@
...
+
+        Aleph::update_authorities(to_authorities(&[2, 3, 4]).as_slice());
+
+        assert_eq!(Aleph::authorities(), to_authorities(&[2, 3, 4]));
+
+        run_session(2);
+
+        Aleph::update_authorities(to_authorities(&[1, 2, 3]).as_slice());
+
+        assert_eq!(Aleph::authorities(), to_authorities(&[1, 2, 3]),);
```
pmikolajczyk41, line 49: remove trailing comma

pallets/aleph/src/tests.rs
```diff
@@ -0,0 +1,64 @@
+#![cfg(test)]
```
pmikolajczyk41, line 1: why do we drop migration tests?

pallets/aleph/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+        fn on_genesis_session<'a, I: 'a>(validators: I)
+        where
+            I: Iterator<Item = (&'a T::AccountId, T::AuthorityId)>,
+            T::AccountId: 'a,
+        {
+            let (_, authorities): (Vec<_>, Vec<_>) = validators.unzip();
+            Self::initialize_authorities(authorities.as_slice());
+        }
+
+        fn on_new_session<'a, I: 'a>(_changed: bool, validators: I, _queued_validators: I)
```
pmikolajczyk41, line 89: why don't we take advantage of `changed` flag?

pallets/aleph/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+        fn on_runtime_upgrade() -> frame_support::weights::Weight {
+            migrations::v0_to_v1::migrate::<T, Self>() + migrations::v1_to_v2::migrate::<T, Self>()
+        }
+    }
+
+    #[pallet::storage]
+    #[pallet::getter(fn authorities)]
+    pub(super) type Authorities<T: Config> = StorageValue<_, Vec<T::AuthorityId>, ValueQuery>;
+
+    impl<T: Config> Pallet<T> {
```
pmikolajczyk41, line 57: do I understand correctly, that the newest version of this pallet is only to serve `authorities` set internally to the AlephBFT (which itself is computed elsewhere)? if so, then it seems as a massive overhead for something that seems achievable without dedicated pallet

pallets/aleph/src/migrations/v1_to_v2.rs
```diff
@@ -0,0 +1,49 @@
...
+
+        if !SessionPeriod::exists() {
+            log::info!(target: "pallet_aleph", "Storage item SessionPeriod does not exist!");
+        }
+        SessionPeriod::kill();
+
+        // store new version
+        StorageVersion::new(2).put::<P>();
+
+        T::DbWeight::get().reads(1) + T::DbWeight::get().writes(1)
```
pmikolajczyk41, line 38: don't we use DB when calling `exists` and `kill`?

pmikolajczyk41 (changes requested):

pallets/aleph/src/lib.rs
mike1729: Talked through :)

.github/workflows/build-node-and-runtime.yml
mike1729: Done.

.github/workflows/build-node-and-runtime.yml
mike1729: Changed.

bin/node/Cargo.toml
mike1729: Done.

bin/runtime/src/lib.rs
mike1729: bumped.

docker/Dockerfile
mike1729: Changed.

pallets/aleph/Cargo.toml
mike1729: Unfortunately, without it, cargo doesn't compile with `-p`.

pallets/aleph/src/mock.rs
mike1729: Removed.

pallets/aleph/src/migrations/v1_to_v2.rs
mike1729: We don't want to see it by default on info level.

pallets/aleph/src/mock.rs
mike1729: We don't need to run all blocks from scratch on every call of this method.

pallets/elections/src/lib.rs
mike1729: Get? Yes, it's needed for `SessionPerdiod`.

pallets/aleph/src/tests.rs
mike1729: Done.

pallets/aleph/src/tests.rs
mike1729: It would check nothing as `StorageVersion!=default`.

pallets/aleph/src/lib.rs
mike1729: Good point, changed.

pallets/aleph/src/migrations/v1_to_v2.rs
mike1729: We do, changed.

pallets/aleph/src/lib.rs
mike1729: This is the way.

pallets/aleph/src/mock.rs
```diff
@@ -173,18 +174,26 @@ pub fn new_test_ext(authorities: &[(u64, u64)]) -> sp_io::TestExternalities {
     t.into()
 }
 
-pub(crate) fn run_session(n: u64) {
-    while System::block_number() < n {
+pub(crate) fn run_session(n: u32) {
+    for i in Session::current_index()..n {
         Session::on_finalize(System::block_number());
         Aleph::on_finalize(System::block_number());
         System::on_finalize(System::block_number());
```
pmikolajczyk41, line 181: not sure how this is relevant, but as far as I understand, the hook `on_finalize` is called on pallets in the reversed order than the one defined in `bin/runtime/src/lib.rs`;
however, we ordered Aleph after Session, so in order to mock the real behavior we should keep the true succession

Yithis (approved):

pallets/aleph/src/lib.rs
```diff
@@ -46,7 +47,27 @@ pub mod pallet {
...
     impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
         fn on_runtime_upgrade() -> frame_support::weights::Weight {
-            migrations::v0_to_v1::migrate::<T, Self>() + migrations::v1_to_v2::migrate::<T, Self>()
+            let on_chain = <Pallet<T> as GetStorageVersion>::on_chain_storage_version();
+            let current = 2;
+            let prev = 1;
+            let genesis = 0;
+            T::DbWeight::get().reads(1)
+                + match on_chain {
+                    _ if on_chain == current => 0,
```
fbielejec, line 56: perhaps just use explicit versions, w-out storing them in a variable?

pallets/aleph/src/lib.rs
```diff
@@ -46,7 +47,27 @@ pub mod pallet {
...
+                    _ if on_chain == prev => migrations::v0_to_v1::migrate::<T, Self>(),
```
fbielejec, line 57: `v1_to_v2`

pallets/aleph/src/mock.rs
mike1729: Good catch, changed.

pallets/aleph/src/lib.rs
mike1729: Changed.

pallets/aleph/src/lib.rs
mike1729: Fixed.

pallets/aleph/src/mock.rs
pmikolajczyk41: almost :) the `on_initialize` block below is out-of-order :stuck_out_tongue:

pmikolajczyk41 (approved):

PR #314, opened 2022-02-25
remove active waiting in party

kostekIV:


finality-aleph/src/party/mod.rs
```diff
@@ -521,7 +527,8 @@ where
                         }
                         start_next_session_network = None;
                     } else {
-                        start_next_session_network = Some(Delay::new(NEXT_SESSION_NETWORK_START_ATTEMPT_PERIOD));
+                        // should not happen
```
timorl, line 530: Then log a warning.

finality-aleph/src/party/mod.rs
```diff
@@ -478,14 +481,17 @@ where
                 },
                 Some(_) = async {
                     match &mut start_next_session_network {
-                        Some(start_next_session_network) => {
-                            start_next_session_network.await;
+                        Some(notification) => {
+                            if let Err(e) = notification.await {
+                                error!(target: "aleph-party", "Error with subscription {:?}", e);
```
timorl, line 486: This should only be a `warn`, something went wrong, but we can go on.

finality-aleph/src/party/mod.rs
```diff
@@ -478,14 +481,17 @@ where
                 },
                 Some(_) = async {
                     match &mut start_next_session_network {
-                        Some(start_next_session_network) => {
-                            start_next_session_network.await;
+                        Some(notification) => {
+                            if let Err(e) = notification.await {
```
timorl, line 485: Perhaps return `next_session_authorities` here already.

timorl (changes requested):
I really like this improvement, nice.

Marcin-Radecki (approved):
I have nothing more to add other than @timorl's comments. Nice rework.

finality-aleph/src/party/mod.rs
kostekIV: else is no longer there

finality-aleph/src/party/mod.rs
kostekIV: done

finality-aleph/src/party/mod.rs
kostekIV: done

finality-aleph/src/party/mod.rs
```diff
@@ -39,12 +39,15 @@ use sp_runtime::traits::{Block, Header};
 use std::{collections::HashSet, default::Default, marker::PhantomData, sync::Arc, time::Duration};
 
 mod task;
+
 use task::{Handle, Task};
+
```
timorl, line 44: In other modules we put all these `use`s below a block of `mod`s, the sudden empty lines here are kinda ugly, so I would prefer if you did that.

The empty line below, separating the `mod`s from the `use`s is a fine addition ofc.

timorl (approved):
Nice, tiny whitespace nit, but other than that LGTM.

finality-aleph/src/party/mod.rs
kostekIV: I think fixed, could you check if this is what you meant?

PR #315, opened 2022-02-25
AZ-601 add e2e test for new validator

Marcin-Radecki:
Added one new e2e test and reworked code a little bit to extract common e2e functions.


bin/node/src/chain_spec.rs
```diff
@@ -372,7 +372,7 @@ fn generate_genesis_config(
         staking: StakingConfig {
             force_era: Forcing::NotForcing,
             validator_count: authorities.len() as u32,
-            minimum_validator_count: authorities.len() as u32,
+            minimum_validator_count: 3,
```
timorl, line 375: Could this perhaps be 4? I keep insisting that values below 4 don't really make sense for BFT algorithms. <_<"

Not a blocker though, especially if there are good reasons for "3".

kostekIV (approved):

e2e-tests/src/staking.rs
```diff
@@ -0,0 +1,96 @@
...
+    connection
+        .get_storage_value("Staking", "ActiveEra", None)
+        .unwrap()
+        .unwrap()
+}
+
+pub fn wait_for_full_era_completion(connection: &Connection) -> anyhow::Result<BlockNumber> {
+    let current_era: u32 = get_current_era(connection);
+    info!("Current era is {}", current_era);
+    wait_for_era_completion(connection, current_era + 2)
```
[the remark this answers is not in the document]
Marcin-Radecki: The reason is that staking works in such a way, that when we request a controller to be a validator in era N, then the changes are applied in the era N+1 (so the new validator is receiving points in N+1), so that we need N+1 to finish in order to claim the reward in era N+2 for the N+1 era. I'll add a comment which clarifies this.

bin/node/src/chain_spec.rs
Marcin-Radecki: The reason is pragmatic that we have a default setup in e2e tests so that 4 validators are starting, so in order to have to remove one from the validator lists in the newly added e2e test. Better solution would be not to change this line ofc, and add   new 5fth validator for example.

timorl: Or you could set this to 4 and start with 5 validators (if removing one is simpler than adding).

e2e-tests/src/test/staking.rs
```diff
@@ -195,3 +101,94 @@ pub fn staking_test(config: &Config) -> anyhow::Result<()> {
...
+    let _ = wait_for_session(&connection, current_session + 2)?;
+
+    let current_era = wait_for_full_era_completion(&connection)?;
+    info!(
+        "Era {} started, claiming rewards for era {}",
+        current_era,
+        current_era - 1
+    );
+
+    payout_stakers(node, stash, current_era - 1);
```
mike1729, line 192: Maybe we should check if there was some payout? Not exactly, just check if stash increased.

mike1729 (changes requested):

e2e-tests/src/test/staking.rs
```diff
@@ -195,3 +101,94 @@ pub fn staking_test(config: &Config) -> anyhow::Result<()> {
...
+pub fn staking_new_validator(config: &Config) -> anyhow::Result<()> {
+    let controller = keypair_from_string("//Controller");
+    let controller_account = AccountId::from(controller.public());
+    let stash_seed = "//Stash";
+    let stash = keypair_from_string(stash_seed);
+    let stash_account = AccountId::from(stash.public());
+    let (_, mut validator_accounts) = get_key_pairs();
+    let node = &config.node;
+    let sender = validator_accounts.remove(0);
+    let connection = create_connection(node).set_signer(sender);
```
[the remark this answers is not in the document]
Marcin-Radecki: Good catch will do.

e2e-tests/src/test/staking.rs
Marcin-Radecki: Right, I was thinking the same. Let me do it.

e2e-tests/src/test/staking.rs
```diff
@@ -133,35 +36,38 @@ fn get_key_pairs() -> (Vec<KeyPair>, Vec<KeyPair>) {
...
 
     validator_accounts
         .par_iter()
-        .for_each(|account| validate(node, account));
+        .for_each(|account| validate(node, account, XtStatus::InBlock));
 
     stashes_accounts
         .par_iter()
-        .for_each(|nominator| bond(node, NOMINATOR_STAKE, nominator));
+        .for_each(|nominator| bond(node, NOMINATOR_STAKE, &nominator.clone(), nominator));
```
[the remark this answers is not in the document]
Marcin-Radecki: I think I did, but let me recheck.

e2e-tests/src/test/staking.rs
```diff
@@ -133,35 +36,38 @@ fn get_key_pairs() -> (Vec<KeyPair>, Vec<KeyPair>) {
...
     let node = &config.node;
     let sender = validator_accounts[0].clone();
     let connection = create_connection(node).set_signer(sender);
 
-    endow_stash_balances(&connection, &stashes_accounts, VALIDATOR_STAKE);
+    batch_endow_account_balances(&connection, &stashes_accounts, VALIDATOR_STAKE);
 
     validator_accounts.par_iter().for_each(|account| {
-        bond(node, VALIDATOR_STAKE, account);
+        bond(node, VALIDATOR_STAKE, &account.clone(), account);
```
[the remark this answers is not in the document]
Marcin-Radecki: Let me fix it

e2e-tests/src/rpc.rs
```diff
@@ -0,0 +1,29 @@
...
+}
+
+pub fn author_rotate_keys() -> Value {
+    json_req("author_rotateKeys", Value::Null, 1)
+}
+
+pub fn get_author_rotate_keys(connection: &Connection) -> ApiResult<Option<TestSessionKeys>> {
+    let jsonreq = author_rotate_keys();
+    let p = connection.get_request(jsonreq)?;
+    match p {
```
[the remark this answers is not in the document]
Marcin-Radecki: Okay, will change

e2e-tests/src/staking.rs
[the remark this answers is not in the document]
kostekIV: thanks

bin/node/src/chain_spec.rs
Marcin-Radecki: I think not, so either we do changes to `docker-compose.yml` to spawn a new validator during tests or we add comments above that line with reasoning. Let me try if I can add Marcin node easily :)

e2e-tests/src/test/staking.rs
```diff
@@ -195,3 +102,126 @@ pub fn staking_test(config: &Config) -> anyhow::Result<()> {
```
mike1729, line 102: Maybe remove the above sanity check, as it would be better suited for your test, but maybe even better would be just to move finalization test in main.rs as the last one?

bin/node/src/chain_spec.rs
Marcin-Radecki: It's possible though it's not possible to set minimum validator count via extrinsic, hence we need to decrease it in either chain_spec.rs. Let me try this version with added new validator.

mike1729: https://paritytech.github.io/substrate/master/pallet_staking/pallet/enum.Call.html#variant.set_staking_configs

Marcin-Radecki: Done

e2e-tests/src/test/staking.rs
Marcin-Radecki: Done

e2e-tests/src/test/staking.rs
```diff
@@ -176,22 +82,142 @@ pub fn staking_test(config: &Config) -> anyhow::Result<()> {
...
+pub fn staking_new_validator(config: &Config) -> anyhow::Result<()> {
+    let controller = keypair_from_string("//Controller");
+    let controller_account = AccountId::from(controller.public());
+    let stash_seed = "//Stash";
+    let stash = keypair_from_string(stash_seed);
+    let stash_account = AccountId::from(stash.public());
+    let (_, mut validator_accounts) = get_key_pairs();
+    let node = &config.node;
+    let sender = validator_accounts.remove(0);
+    // skigner of this connection is sudothe same node which in this test is used as the new one
```
fbielejec, line 110:
```suggestion
    // signer of this connection is sudo, the same node which in this test is used as the new one
```

fork-off/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2021"
 # client
 [dependencies.substrate-api-client]
 git = "https://github.com/Cardinal-Cryptography/substrate-api-client.git"
-tag = "fixed_dependencies"
+tag = "polkadot-v0.9.13.b2bfc0c3"
```
fbielejec, line 9: best confirm if these changes are backwards compat with encoding on current testnet (can the tool still communicate with it?), or confirm with @bartoszjedrzejewski this does not hinder with their work on devnet

Marcin-Radecki: I'll confirm with @bartoszjedrzejewski

Marcin-Radecki: Actually, I confirmed running locally, and it works.

PR #316, opened 2022-02-27
AZ-528: Add network integration tests

maciejnems:


Yithis (approved):

lesniak43 (approved):

finality-aleph/src/testing/network.rs
```diff
@@ -321,3 +489,227 @@ async fn test_connects_to_others() {
...
+            .await
+            .ok()
+            .flatten(),
+        Some((
+            HashSet::from_iter(test_data.authorities.iter().skip(1).map(|a| a.peer_id())),
+            Protocol::Validator.name()
+        ))
+    );
+
+    // This assert should be before cleanup
```
fixxxedpoint, line 546: Why?

fixxxedpoint (approved):
lgtm

finality-aleph/src/testing/network.rs
maciejnems: We want to check whether  `session_manager.stop_session(...)` drops the sender. After cleanup all network tasks end and senders will be dropped, so if assert was after cleanup we wouldn't know whether `data_network` receiver is droopped because of `.stop_session(...)` or because of cleanup.

finality-aleph/src/testing/network.rs
```diff
@@ -297,11 +298,178 @@ impl TestData {
...
+            .ok()
+            .flatten()
+            .expect("Should add reserved nodes"),
+        (
+            sending_peer.addresses().into_iter().collect(),
+            Protocol::Validator.name()
+        ),
+    );
+
+    loop {
```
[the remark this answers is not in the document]
maciejnems: I guess we could add an upper bound on how many other messages it receives. I think something big like 20 other messages should be fine. Also I guess I could add an upper bound like that in other places where loop might run infinitely. WDYT?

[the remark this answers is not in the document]
maciejnems: Ok, I have a better idea. I created and async function that will return only upon receiving wanted data, so in case we don't receive it the timeout will end.

finality-aleph/src/testing/network.rs
fixxxedpoint: Consider extending that comment with this explanation.

PR #322, opened 2022-03-03
AZ-551: Refactor some stuff into common

timorl:
Preparation for a CLI tool for key and validator rotation


common/src/waiting.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use crate::Connection;
+
+pub fn wait_for_event<E: Decode + Clone, P: Fn(E) -> bool>(
+    connection: &Connection,
+    event: (&str, &str),
+    predicate: P,
+) -> anyhow::Result<E> {
+    let (module, variant) = event;
+    info!("[+] Creating event subscription {}/{}", module, variant);
```
Marcin-Radecki, line 15: What about we give up on the convention of adding `[+]` to each log line? It originates from `substrate-api-client` code and made more sense when this code was in the `e2e-client` crate. Besides now we can run locally e2e tests with `./scripts/run_e2e.sh` which sets `RUST_LOG=aleph-e2e-client=info`, that effectively strips out this `[+]` log lines from `substrate-api-client`, so one don't need to grep on `[+]` anymore.

e2e-tests/src/lib.rs
```diff
@@ -1,24 +1,15 @@
...
 mod fee;
-pub mod rpc;
-pub mod session;
 mod staking;
 pub mod test;
 mod transfer;
 mod waiting;
 
-type BlockNumber = u32;
-type Header = generic::Header<BlockNumber, BlakeTwo256>;
```
Marcin-Radecki, line 19: Much better, so we're using now the _real_ `Header`, `KeyPair` types from `runtime/src.lib.rs`? That should be good, as the types below are really sneaked from runtime anyway.

e2e-tests/src/test/validators_change.rs
```diff
@@ -28,7 +27,7 @@ pub fn change_validators(config: &Config) -> anyhow::Result<()> {
 
     accounts.remove(0);
     let new_members: Vec<AccountId> = accounts.iter().map(|pair| pair.public().into()).collect();
-    send_change_members(&connection, new_members.clone());
+    change_members(&connection, new_members.clone(), XtStatus::InBlock);
```
Marcin-Radecki, line 30: Nice, passing everywhere now `XtStatus`, so it's clear for the developer what's the block waiting status. Good standard in e2e test code.

e2e-tests/src/waiting.rs
```diff
@@ -1,31 +1,6 @@
...
-            Ok(_) => (),
-            Err(why) => error!("Error {:?}", why),
-        }
-    }
-}
+use log::info;
+use common::{Connection, Header};
 
 /// blocks the main thread waiting for a block with a number at least `block_number`
 pub fn wait_for_finalized_block(connection: &Connection, block_number: u32) -> anyhow::Result<u32> {
```
Marcin-Radecki, line 6: Should this function be moved to `common` as well? Or we just doing part of refactoring in this PR only. I'm just thinking out loud what should be in `common` anyway.

Marcin-Radecki (approved):
Nice rework, minor comments.

e2e-tests/src/waiting.rs
timorl: I've been trying to make this refactor pretty minimal, since I'm not sure what actually should be in the common crate and I didn't want to spend too much time on this. So I'm afraid this stays here.

e2e-tests/src/lib.rs
timorl: Uhhh, I have no idea what you mean, I just moved these definitions into `common`. >.<"

common/src/waiting.rs
timorl: Good idea, I'll try also adding a `target:` in this module.

e2e-tests/src/lib.rs
Marcin-Radecki: There's sometimes a challenge when writing e2e tests such that we need a concrete type from the runtime, but we don't want to add a runtime dependency to e2e Cargo.toml. In such case, one "reverse-engineers" runtime, and copy-paste that `type` to e2e tests, essentially increasing code duplication counted globally in `aleph-node`. Having one `type` in `common` accessible by the runtime and e2e decreases that duplication. 

A side note is there's an ongoing discussion whether we need to add `../bin/runtime` dependency to e2e-client Cargo.toml. but I think since we're moving some e2e code into `common`, then we've found a solution to this .
CC @fbielejec

common/src/waiting.rs
timorl: Done.

e2e-tests/src/lib.rs
timorl: I see, I think this is out of scope for this change, but I understand why this would be useful.

I would be worried if we wanted to mainline `common` in its current shape, I thought of it more as of "common client things". I would argue that if we want to expose this kind of stuff we should make a separate "runtime types" crate and refer to that both in `runtime` and `common`. I was kind of under the impression that `primitives` was supposed to be this, but apparently not quite.

Anyway, I think this requires some designing to get right.

Yithis:
Maybe we should rename `common` into `aleph-client`? It will be very confusing when we will start to work on `Common`. What do you think @mike1729 ?

PR #323, opened 2022-03-03
AZ-551: Add CLI tool for key and validator rotation

timorl:


bin/cliain/src/validators.rs
```diff
@@ -0,0 +1,17 @@
+use log::info;
+use aleph_client::{change_members, create_connection, KeyPair};
+use substrate_api_client::{AccountId, XtStatus};
+use sp_core::crypto::Ss58Codec;
+
+/// Change validators to the provided list by calling the provided node.
+/// The keypair has to be capable of signing sudo calls.
+pub fn change(validators: Vec<String>, node: String, key: KeyPair) {
```
Marcin-Radecki, line 8: We can change the name `key` to `sudo_key` to reflect this function usage better.

Marcin-Radecki, line 8: We can change the name `key` to `sudo_key` to reflect this function usage better.

timorl: Done.

Marcin-Radecki (approved):

deuszx (commented):
Do you think it's worth adding a README or some kind of descriptive `--help` message to the new command?

maciejnems (approved):
LGTM, only one stupid nit

bin/cliain/rustfmt.toml
```diff
@@ -0,0 +1 @@
+edition = "2021"
```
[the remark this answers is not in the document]
timorl: Done.

timorl:
Also added some minimal documentation to the command that was lacking it @deuszx . Probably insufficient for informed external usage, but I believe we will keep developing this tool and we should build up a better documentation on the way when more commands appear and the interface becomes more standardized.

PR #324, opened 2022-03-03
az-471-terra-light-client

fbielejec:
The pallet is considered WIP / POC, therefore this PR omits a couple of important features:
- ~~Benchmarks are not ready, left at the stage where they compile (just so)~~
- There are mappings from Tendermint types to storage ready types in our pallet but not into other direction

It's mainly opened so that the overall architecture / design (especially around storage) can be reviewed. 
Storage revolves around the idea of maintaining last N (where N is configurable) headers, ordered by their insertion time, with the oldest entry evicted.

Another important feature would be to make the pallet into a client factory - i.e. it would support any number of bridged chains, and maintain a mapping between their unique id and options plus headers storage.


fbielejec:
> Let's cut a feature branch from `main` and target PRs at that.

You mean for the entire bridge?

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,308 @@
...
+            log::info!(target: "runtime::tendermint-lc", "Light client initialized");
+            Self::deposit_event(Event::LightClientInitialized);
+
+            Ok(())
+        }
+
+        // TODO : benchmark & adjust weights
+        /// Verify a block header against a known state.        
+        #[pallet::weight((T::DbWeight::get().reads_writes(1, 1), DispatchClass::Operational))]
+        pub fn submit_finality_proof(
```
deuszx, line 177: Can we change it to `update_client`?

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,308 @@
...
+                    fail!(<Error<T>>::InvalidBlock)
+                }
+            }
+        }
+
+        /// Halt or resume all light client operations
+        ///
+        /// Can only be called by root
+        #[pallet::weight((T::DbWeight::get().reads_writes(1, 1), DispatchClass::Operational))]
+        pub fn set_halted(origin: OriginFor<T>, halted: bool) -> DispatchResult {
```
deuszx, line 240: This method should be called by the pallet itself if it detects a fork.

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,308 @@
...
+
+    fn verify_light_block(
+        verifier: ProdVerifier,
+        untrusted_block: LightBlockStorage,
+        trusted_block: LightBlockStorage,
+        options: &Options,
+        now: Time,
+    ) -> tendermint_light_client_verifier::Verdict {
+        let untrusted_block: LightBlock = untrusted_block
+            .clone()
```
deuszx, line 264: Why do we need to clone it? Its ownership is moved to the method.

fbielejec: fixed

pallets/tendermint-light-client/src/types.rs
```diff
@@ -0,0 +1,753 @@
...
+    /// no vote was received from a validator.
+    BlockIdFlagAbsent,
+    /// voted for the Commit.BlockID.
+    BlockIdFlagCommit {
+        /// Validator address
+        validator_address: TendermintAccountId,
+        /// Timestamp of vote
+        timestamp: TimestampStorage,
+        /// Signature of vote
+        signature: TendermintVoteSignature,
```
deuszx, line 326: I've noticed that while working on the mappings the other way around - in Tendermint this field is optional. It should be optiona here as well.

pallets/tendermint-light-client/src/types.rs
```diff
@@ -0,0 +1,753 @@
...
+    },
+    /// voted for nil.
+    BlockIdFlagNil {
+        /// Validator address
+        validator_address: TendermintAccountId,
+        /// Timestamp of vote
+        timestamp: TimestampStorage,
+        /// Signature of vote
+        signature: TendermintVoteSignature,
```
deuszx, line 335: Same as with the other variant.

pallets/tendermint-light-client/src/types.rs
```diff
@@ -0,0 +1,753 @@
...
+
+                    let timestamp = timestamp_from_rfc3339(
+                        value.get("timestamp").and_then(Value::as_str).unwrap(),
+                    )
+                    .unwrap();
+
+                    let signature = base64string_as_h512(
+                        value.get("signature").and_then(Value::as_str).unwrap(),
+                    )
+                    .unwrap();
```
deuszx, line 369: This, and the other variant's mapping, becomes
```rust
                let signature = signature.and_then(CommitSignatureStorage::signature);

```

pallets/tendermint-light-client/src/types.rs
```diff
@@ -0,0 +1,753 @@
...
+        timestamp: TimestampStorage,
+        /// Signature of vote
+        signature: TendermintVoteSignature,
+    },
+}
+
+impl CommitSignatureStorage {
+    fn signature(signature: TendermintVoteSignature) -> Option<tendermint::Signature> {
+        tendermint::Signature::try_from(signature.as_bytes()).ok()
+    }
```
deuszx, line 342: This function name and its location is confusing. It doesn't map Tendermint's type to our but the other way around. Why not implement it as `Try`/`From`?

deuszx (changes requested):
Values that were optional in Tendermint should remain optional.

pallets/tendermint-light-client/src/types.rs
fbielejec: it's just a helper fn, it wraps try_from - perhaps it belongs to utils better

pallets/tendermint-light-client/src/lib.rs
fbielejec: done

pallets/tendermint-light-client/src/types.rs
fbielejec: done

pallets/tendermint-light-client/src/lib.rs
fbielejec: added as TODO item

DamianStraszak (changes requested):
This is a great piece of work (the very foundations of the bridge :) ). I left a few small comments, but generally I think it is still WIP, so there is no point in being too nitpicky. 

What I think still requires work is figuring out what kind of interface we would expect from this light client in order to support the functioning of the bridge. Right now this is in a PoC stage, i.e., we want it to somehow accept terra blocks without thinking too much what to do with them. To achieve the end goal (bridge) we will need to specify more concretely what kind of queries the client should be able to support, and that I believe might also require to refine the storage design.

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+    // https://substrate.recipes/ringbuffer.html
+
+    /// Hash of the last imported header from the bridged chain
+    #[pallet::storage]
+    #[pallet::getter(fn get_last_imported_hash)]
+    pub type LastImportedHash<T: Config> = StorageValue<_, BridgedBlockHash, ValueQuery>;
+
+    /// A ring buffer of imported hashes "ordered" by their insertion time
+    #[pallet::storage]
+    pub type ImportedHashes<T: Config> = StorageMap<_, Identity, u32, BridgedBlockHash>;
```
mike1729, line 95: Specify a getter and use it instead of `ImportedHashes::<T>::get`.

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+
+    /// Current ring buffer position
+    #[pallet::storage]
+    pub(super) type ImportedHashesPointer<T: Config> = StorageValue<_, u32, ValueQuery>;
```
mike1729, line 99: Specify a getter and use it instead of `ImportedHashesPointer::<T>::get`.

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+
+    /// Bridged chain Headers which have been imported by the client
+    /// Client keeps HeadersToKeep number of these at any time
+    #[pallet::storage]
+    pub(super) type ImportedBlocks<T: Config> =
```
mike1729, line 104: Specify a getter and use it instead of `ImportedBlocks::<T>::get`.

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+    // - header storage should be a ring buffer (i.e. we keep n last headers, ordered by the insertion time)
+    // - we keep a pointer to the last finalized header (to avoid deserializing the whole buffer)
+    // - insertion moves the pointer and updates the buffer
+    // or?
+    // https://substrate.recipes/ringbuffer.html
+
+    /// Hash of the last imported header from the bridged chain
+    #[pallet::storage]
+    #[pallet::getter(fn get_last_imported_hash)]
+    pub type LastImportedHash<T: Config> = StorageValue<_, BridgedBlockHash, ValueQuery>;
```
mike1729, line 91: Hash is a little big vague, could you use more specific name like `LastImportedBlockHash`?

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+        verifier.verify(
+            untrusted_block.as_untrusted_state(),
+            trusted_block.as_trusted_state(),
+            options,
+            now,
+        )
+    }
+
+    /// update light client storage
+    /// should only be called by a trusted origin, *after* performing a verification
```
mike1729, line 268: You call `insert_light_block` in `update_client` which only ensures that the xt wast signed. I think this is fine so only this comment needs to be corrected.

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+    fn insert_light_block<T: Config>(hash: BridgedBlockHash, light_block: LightBlockStorage) {
+        let index = <ImportedHashesPointer<T>>::get();
+        let pruning = <ImportedHashes<T>>::try_get(index);
```
mike1729, line 271: You should use `mutate` instead of reading and writing for better performance.

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+
+    /// Bridged chain Headers which have been imported by the client
+    /// Client keeps HeadersToKeep number of these at any time
+    #[pallet::storage]
+    pub(super) type ImportedBlocks<T: Config> =
+        StorageMap<_, Identity, BridgedBlockHash, LightBlockStorage, OptionQuery>;
+
+    // TODO : expose in runtime API and nodes RPC
+    impl<T: Config> Pallet<T> {
+        pub fn get_last_imported_block() -> Option<LightBlockStorage> {
```
mike1729, line 109: You could store `LastImportedBlock` instead of `LastImportedHash`, it would simplify the logic in many places and be cheaper in number of storage reads/writes as obv block contains hash. Block takes more storage and is slower to decode but I believe that it's negligible comparing to fetch costs.

mike1729 (changes requested):
Great work! And a lot of it...

pallets/tendermint-light-client/src/lib.rs
DamianStraszak: Actually the headers contain the whole committee, lost of signatures and stuff, so hash is likely to be 100-500 times smaller than a header :/

mike1729: Uhh...so maybe if we implementing those types ourselves, then we could change them and for example don't store everything after the block got validated?

DamianStraszak: I think we still need at least the most recent block with full information :/ (because it seems that the tendermint verifier needs it?). But generally I think one can make multiple optimizations here, it's just at this stage we don't know yet what API will we require from this pallet to build the bridge, so I'm not suggesting any such changes yet...

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+            let ptr = ImportedHashesPointer::<T>::get()
+                .checked_sub(1)
+                .expect("unexpected failure when subtracting");
```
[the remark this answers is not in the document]
fbielejec: fixed

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+    #[pallet::storage]
+    pub(super) type ImportedHashesPointer<T: Config> = StorageValue<_, u32, ValueQuery>;
+
+    /// Bridged chain Headers which have been imported by the client
+    /// Client keeps HeadersToKeep number of these at any time
+    #[pallet::storage]
+    pub(super) type ImportedBlocks<T: Config> =
+        StorageMap<_, Identity, BridgedBlockHash, LightBlockStorage, OptionQuery>;
+
+    // TODO : expose in runtime API and nodes RPC
```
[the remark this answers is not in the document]
fbielejec: YMMV, but sometimes it just *feels* right to leave a TODO comment as a constant reminder for self (I look at it way more often than at Jira) - and definitely beats not leaving any comment at all and forgetting. On the other hand in this case it will eventually surface, as other pallets in the bridge ecosystem will eventually need the query, so I opted for just removing the comment.

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+
+            let verdict = verify_light_block(
+                verifier,
+                untrusted_block.clone(),
+                most_recent_trusted_block,
+                &options,
+                now,
+            );
+
+            match verdict {
```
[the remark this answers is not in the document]
fbielejec: done

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+    fn verify_light_block(
+        verifier: ProdVerifier,
+        untrusted_block: LightBlockStorage,
+        trusted_block: LightBlockStorage,
+        options: &Options,
+        now: Time,
+    ) -> tendermint_light_client_verifier::Verdict {
+        let untrusted_block: LightBlock = untrusted_block
+            .try_into()
+            .expect("Unexpected failure when casting untrusted block as tendermint::LightBlock");
```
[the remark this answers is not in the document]
fbielejec: a bug mostly really.

There is explicit handling in place now, returns a typed failure in the call itself.

pallets/tendermint-light-client/src/utils.rs
```diff
@@ -0,0 +1,98 @@
...
+pub fn sha256_from_bytes(bytes: &[u8]) -> Hash {
+    Hash::from_bytes(hash::Algorithm::Sha256, bytes).expect("Can't produce Hash from raw bytes")
+}
+
+pub fn account_id_from_bytes(bytes: [u8; 20]) -> account::Id {
+    account::Id::new(bytes)
+}
+
+/// create Vec of known size filled with empty bytes
+pub fn empty_bytes(size: i32) -> Vec<u8> {
```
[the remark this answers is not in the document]
fbielejec: correct, nonetheless removed anyway, generator is not a part of this PR

pallets/tendermint-light-client/src/types.rs
```diff
@@ -0,0 +1,854 @@
...
+        Self {
+            signed_header,
+            validators,
+            next_validators,
+            provider,
+        }
+    }
+
+    pub fn create(
+        chain_id_length: i32,
```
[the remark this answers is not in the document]
fbielejec: removed

pallets/tendermint-light-client/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+sp-core = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13", default-features = false }
+sp-io = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13", default-features = false, features = ["disable_oom", "disable_panic_handler"] }
+sp-runtime = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13", default-features = false }
+sp-std = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13", default-features = false }
+subtle-encoding = { version = "0.5", default-features = false, features = ["bech32-preview"] }
+tendermint = { version = "0.24.0-pre.1", default-features = false, features = ["secp256k1"] }
+tendermint-light-client-verifier = { version = "0.24.0-pre.1", default-features = false }
+time = {version = "0.3.5", default-features = false, features = [ "parsing"] }
+
+# [dev-dependencies]
```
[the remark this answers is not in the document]
fbielejec: cleaned

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,293 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+//! This pallet is an on-chain light-client for tendermint (Cosmos) based chains
+//! It verifies headers submitted to it via on-chain transactions, performed by a so-called relayer
+//! It is a part of the Aleph0 <-> Terra bridge
```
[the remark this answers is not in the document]
fbielejec: done

pallets/tendermint-light-client/src/types.rs
```diff
@@ -0,0 +1,854 @@
...
+    ) -> Self {
+        Self {
+            signed_header,
+            validators,
+            next_validators,
+            provider,
+        }
+    }
+
+    pub fn create(
```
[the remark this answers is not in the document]
fbielejec: yes - heavily refactored and not part of this PR anyway

pallets/tendermint-light-client/src/lib.rs
fbielejec: done

pallets/tendermint-light-client/src/lib.rs
fbielejec: done

pallets/tendermint-light-client/src/lib.rs
fbielejec: done

pallets/tendermint-light-client/src/lib.rs
fbielejec: changed

pallets/tendermint-light-client/src/lib.rs
fbielejec: not sure I understand the comment, you want to see the second comments line removed?

pallets/tendermint-light-client/src/lib.rs
fbielejec: done

pallets/tendermint-light-client/src/types.rs
```diff
@@ -0,0 +1,1101 @@
...
+        Self {
+            trust_threshold: TrustThresholdStorage {
+                numerator: 1,
+                denominator: 3,
+            },
+            trusting_period: 1210000, // 2 weeks
+            clock_drift: 5,
+        }
+    }
+}
```
deuszx, line 160: What do we need the defaults for? Testing? If so, can we mark this with a proper `#[cfg(test)]` annotation?

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,337 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+//! This pallet is an on-chain light-client for tendermint (Cosmos) based chains
```
deuszx, line 3:
```suggestion
//! This pallet is an on-chain light-client for Tendermint (Cosmos) based chains
```

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+//! It verifies headers submitted to it via on-chain transactions, performed by a so-called relayer
+//! It is a part of the Aleph Zero <-> Terra bridge
```
deuszx, line 5: Not sure if we should mention Terra explicitly here since this client should be useable by any Tendermint-based chain.

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+    #[pallet::generate_deposit(pub(super) fn deposit_event)]
+    pub enum Event<T: Config> {
+        /// Pallet is halted
+        LightClientHalted,
+        /// Pallet operations are resumed        
+        LightClientResumed,
+        /// Light client is initialized
+        LightClientInitialized,
+        /// New block has been verified and imported into storage \[relayer_address, imported_block_hash\]
+        ImportedLightBlock(T::AccountId, TendermintBlockHash),
```
deuszx, line 68: Let's add a block height to this since Tendermint chain uses that to identify blocks, rather than block hashes.

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+        /// Light client is currently halted
+        Halted,
+        /// The minimum voting power threshold is not reached, the block cannot be trusted yet
+        NotEnoughTrust,
+        /// Verification failed, the block is invalid        
+        InvalidBlock,
+        /// Initial block is invalid
+        InvalidInitialBlock,
+        /// General error during client operations
+        Other,
```
deuszx, line 86: Can we enrich this error variant with a message? If not, do we log the error? Trying to think how the consumer of it can figure out what went wrong.

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+        /// Pallet operations are resumed        
+        LightClientResumed,
+        /// Light client is initialized
+        LightClientInitialized,
+        /// New block has been verified and imported into storage \[relayer_address, imported_block_hash\]
+        ImportedLightBlock(T::AccountId, TendermintBlockHash),
+    }
+
+    #[pallet::error]
+    pub enum Error<T> {
```
deuszx, line 72: `T` is not used in any of the variants, can we remove it?

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+        #[pallet::constant]
+        type MaxVotesCount: Get<u32>;
+
+        /// time provider type, used to gauge whether blocks are within the trusting period
+        type TimeProvider: UnixTime;
+    }
+
+    #[pallet::pallet]
+    #[pallet::storage_version(STORAGE_VERSION)]
+    pub struct Pallet<T>(_);
```
deuszx, line 56: `T` is not used, can we remove it?

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+
+    /// Current ring buffer position
+    #[pallet::storage]
+    #[pallet::getter(fn get_imported_hashes_pointer)]
+    pub(super) type ImportedHashesPointer<T: Config> = StorageValue<_, u32, ValueQuery>;
+
+    /// Bridged chain Headers which have been imported by the client
+    /// Client keeps HeadersToKeep number of these at any time
+    #[pallet::storage]
+    #[pallet::getter(fn get_imported_block)]
```
deuszx, line 115:
```suggestion
    #[pallet::getter(fn get_imported_blocks)]
```

Should be plural, correct?

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+        }
+
+        // TODO: This method will need to be called by the pallet itself if it detects a fork.
+        /// Halt or resume all light client operations
+        ///
+        /// Can only be called by root
+        #[pallet::weight((T::DbWeight::get().reads_writes(1, 1), DispatchClass::Operational))]
+        pub fn set_halted(origin: OriginFor<T>, halted: bool) -> DispatchResult {
+            ensure_root(origin)?;
+            <IsHalted<T>>::put(halted);
```
deuszx, line 281: Let's check first if the current `IsHalted` isn't already equal to `halted`. In such case, this should be a no-op rather than logging and creating new events.

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+        #[pallet::weight((T::DbWeight::get().reads_writes(1, 1), DispatchClass::Operational))]
+        pub fn update_client(
+            origin: OriginFor<T>,
+            untrusted_block: LightBlockStorage,
+        ) -> DispatchResult {
+            ensure_not_halted::<T>()?;
+            let who = ensure_signed(origin)?;
+            let hash = untrusted_block.signed_header.commit.block_id.hash;
+
+            log::debug!(target: "runtime::tendermint-lc", "Verifying light block {:#?}", &hash);
```
deuszx, line 188: Let's log height as well - like previously, Tendermint uses heights more than hashes. It will be easier to correlate things if we have both informations.

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+
+        // TODO : benchmark & adjust weights
+        /// Verify a block header against a known state.        
+        #[pallet::weight((T::DbWeight::get().reads_writes(1, 1), DispatchClass::Operational))]
+        pub fn update_client(
+            origin: OriginFor<T>,
+            untrusted_block: LightBlockStorage,
+        ) -> DispatchResult {
+            ensure_not_halted::<T>()?;
+            let who = ensure_signed(origin)?;
```
deuszx, line 185: Wouldn't the system take care of verifying that transactions are signed by its sender? Or we really have to do that manually?

DamianStraszak (approved):

deuszx (commented):
As discussed, we should allow for multiple instances of the light client to be created - this pallet is currently implemented in such a way that it can create only once instance and from then on would return errors. Let's create tickets for that. And similarly - it should be possible to update each of these different light client instances.

pallets/tendermint-light-client/src/lib.rs
deuszx: > Actually the headers contain the whole committee, lost of signatures and stuff, so hash is likely to be 100-500 times smaller than a header :/

Is that a problem? We will have to load the last block every time regardless for the purpose of verification. I think that the current implementation has to be more expensive actually since it has to read from multiple storages (pointers, hashes, blocks) and then map to the expected type (from our own `LightBlockStorage` to `tendermint_light_client_verifier::types::LightBlock`).

> I think we still need at least the most recent block with full information :/ (because it seems that the tendermint verifier needs it?)

It seems like we can improve things a bit here. The [`ProdVerifier::verify`](https://docs.rs/tendermint-light-client-verifier/latest/tendermint_light_client_verifier/trait.Verifier.html#tymethod.verify) expects [`TrustedBlockState`](https://docs.rs/tendermint-light-client-verifier/latest/tendermint_light_client_verifier/types/struct.TrustedBlockState.html) which does not contain the signatures - this type is created with [`LightBlock::as_trusted_state(&self)`](https://docs.rs/tendermint-light-client-verifier/latest/src/tendermint_light_client_verifier/types.rs.html#147-154). And we're already doing that. This means, we can have a special storage where we would put the latest trusted state and read that instead.


@fbielejec, can you make the tickets for that work? It doesn't have to hold up this PR.

deuszx:
There's one more approach that could decrease the cost of using the pallet: instead of mapping between `*Storage` types and the Tendermint ones we can wrap them with a `struct` that we control:
```rust
struct LightBlockWrapper(tendermint_light_client_verifier::types::LightBlock);
```
and write `Encode`/`Decode` for `LightBlockWrapper`. Now, when we read from storage we no longer have to call expensive `try_into()` as now it's just a wrapper. Additionally, there's no _trying_ anymore since the inner type is exactly what we need.

deuszx (approved):
No blockers but there are some things that can be improved.

One last note, in my local tests I was not able to initialize a new client using `//Damian` account. Even though he's sudoer. I had to remove `ensure_root(origin)` to make it work. It might be some hiccup in my local setup so we need to check.

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
+        // TODO : benchmark & adjust weights
+        #[pallet::weight((T::DbWeight::get().reads_writes(1, 1), DispatchClass::Operational))]
+        pub fn initialize_client(
+            origin: OriginFor<T>,
+            options: LightClientOptionsStorage,
+            initial_block: LightBlockStorage,
+        ) -> DispatchResult {
+            ensure_root(origin)?;
```
deuszx, line 149: This will have to be replaced with some other approach before releasing. We cannot require the `origin` to be a root/sudo since that would mean our relayer has to be given sudo keys. Which is not safe.

Let's make a ticket for it as well.

pallets/tendermint-light-client/src/lib.rs
fbielejec: https://cardinal-cryptography.atlassian.net/browse/A0-845

We can't  just store Tendermint type here OC - but we could store something that can be converted into Tendermints TrustedState, sure.

pallets/tendermint-light-client/src/types.rs
fbielejec: done

pallets/tendermint-light-client/src/lib.rs
fbielejec: done

pallets/tendermint-light-client/src/lib.rs
fbielejec: done

pallets/tendermint-light-client/src/lib.rs
fbielejec: It would be great to add message String here, but the pallet::error must be of Unit type :/ in Substarte. 

We do log runtime errors however, and the ConversionError from types mod preserves all context, so we will be able to sift through logs and see what went wrong and where.

pallets/tendermint-light-client/src/lib.rs
fbielejec: not really, because substrate macros and stuff :/

pallets/tendermint-light-client/src/lib.rs
fbielejec: it is used, just obscured by the attribute macros

pallets/tendermint-light-client/src/lib.rs
fbielejec: no - the storage is a map, so the auto-generated getter returns one block (value) by it's hash (key).

deuszx: Ah, ok. Thanks.

pallets/tendermint-light-client/src/lib.rs
deuszx: We will, I was wondering what would other users of the chain do. They won't have access to the chain.

I understand that we can't add message "because Substrate".

pallets/tendermint-light-client/src/lib.rs
fbielejec: I introduced a pattern match and log whether there was or was not any operation. TODO : weights need ot be adjusted but that's generally true for all calls, so I will include it in the benchmarks PR

pallets/tendermint-light-client/src/lib.rs
fbielejec: done

pallets/tendermint-light-client/src/lib.rs
deuszx: It's not only about the cost but I think it might be confusing to people when they see an event about halting the light client. 

Not a blocker of course, we can leave it as is if you prefer.

pallets/tendermint-light-client/src/lib.rs
fbielejec: there are three types of origin in Substrate: Root, Signed and None. 
I'm blurry on the details but yeah, there can be a call with no signature, for some reason. Maybe if there is some other mechanism for validating it I guess

pallets/tendermint-light-client/src/lib.rs
fbielejec: but why does a relayer need a sudo key? Someone in the foundation creates the Client instance, and from that point on relayer can start submitting consecutive blocks (update call needs just a regular signature, can be called by any account).

pallets/tendermint-light-client/src/lib.rs
fbielejec: > There's one more approach that could decrease the cost of using the pallet: instead of mapping between `*Storage` types and the Tendermint ones we can wrap them with a `struct` that we control:
> 
> ```rust
> struct LightBlockWrapper(tendermint_light_client_verifier::types::LightBlock);
> ```
> 
> and write `Encode`/`Decode` for `LightBlockWrapper`. Now, when we read from storage we no longer have to call expensive `try_into()` as now it's just a wrapper. Additionally, there's no _trying_ anymore since the inner type is exactly what we need.

This won't work IMHO - if the inner type is not Encode/Decode the outer cannot be either. And we cannot impl for Tendermint types becasue Rust, orphaned implementations and all that...

pallets/tendermint-light-client/src/lib.rs
fbielejec: I return Error<T>::ConversionError in addition to Other now

pallets/tendermint-light-client/src/lib.rs
```diff
@@ -272,20 +286,29 @@ pub mod pallet {
...
-            if halted {
-                log::warn!(target: "runtime::tendermint-lc", "Halting light client operations");
-                Self::deposit_event(Event::LightClientHalted);
-            } else {
-                log::warn!(target: "runtime::tendermint-lc", "Resuming light client operations");
-                Self::deposit_event(Event::LightClientResumed);
+
+            let current_status = Self::is_halted();
+
+            match [halted, current_status].as_slice() {
```
deuszx, line 299: I would just do
```rust

let current_status = Self::is_halted();

if current_status == halted {
  // exit early
  return
}

// old code
```

Seems simpler and I don't think we need to emit any `NoOp` event.

pallets/tendermint-light-client/src/lib.rs
deuszx: >  And we cannot impl for Tendermint external types (not from our crate) because Rust, orphaned implementations and all that...

We can implement that manually though for the outer one, without having separate impls for the inner ones:
```rust
struct LightBlockWrapper(tendermint_light_client_verifier::types::LightBlock);

impl Encode for LightBlockWrapper {
	// Here, we would encode the whole inner `LightBlock` without having separate impls for header, commit, etc.
}
```

pallets/tendermint-light-client/src/lib.rs
fbielejec: no problem, will add these changes in next PRs

PR #327, opened 2022-03-08
AZ-645: Remove AlephBFT dependency from Aggregator

piotrMocz:
This PR introduces the `aggregation::multicast::Multicast` trait that abstracts away the behaviour of `aleph_bft::ReliableMulticast` and a corresponding `Multisigned` trait that does the same for `aleph_bft::Multisigned`.

It also introduces a slightly improved directory structure, wherein the aggregator and its auxiliary traits are moved to a separate `aggregation` module.

Additionally, the aggregator (both `src/aggregation/aggregator` and `src/party/aggregator`) are broken down into a few functions to avoid gigantic loops.


finality-aleph/src/network/rmc.rs
```diff
@@ -1,6 +1,99 @@
...
+
+/// Anything that exposes the same interface as `aleph_bft::Multisigned` struct.
+/// In production code this will be the `MultisignedWrapper`, containing an instance
+/// of `aleph_bft::Multisigned` (here imported as `MultisignedStruct` for clarity).
+pub trait Multisigned<'a, H: Hash, MK: MultiKeychain> {
+    fn as_signable(&self) -> &SignableHash<H>;
+    fn into_unchecked(self) -> UncheckedSigned<SignableHash<H>, MK::PartialMultisignature>;
+}
+
+pub struct MultisignedWrapper<'a, H: Hash, MK: MultiKeychain> {
```
timorl, line 37: I don't think you need the wrapper (and the other one either), is there a reason why you cannot implement this trait directly for `MultisignedStruct`?

finality-aleph/src/network/rmc.rs
```diff
@@ -1,6 +1,99 @@
...
+#[async_trait::async_trait]
+pub trait Multicast<H: Hash>: Send + Sync {
+    type Signed;
+    async fn start_multicast(&mut self, hash: SignableHash<H>);
+    fn get_multisigned(&self, hash: &SignableHash<H>) -> Option<Self::Signed>;
+    async fn next_multisigned_hash(&mut self) -> Self::Signed;
+}
+
+/// A wrapper for `aleph_bft::ReliableMulticast` that forwards all calls to its underlying instance.
+pub struct RMCWrapper<'a, H: Hash, MK: MultiKeychain> {
```
timorl, line 73: Nit: the Rust naming convention for acronyms in names is to still have only the first letter capitalized, i.e. in this case `RmcWrapper`.

timorl: But my other comment about this being likely unneeded applies. <_<

finality-aleph/src/party/aggregator.rs
```diff
@@ -23,17 +24,71 @@ pub struct IO<B: Block> {
     pub justifications_for_chain: mpsc::UnboundedSender<JustificationNotification<B>>,
 }
 
-async fn run_aggregator<B, C, N>(
-    mut aggregator: BlockSignatureAggregator<'_, B, N, KeyBox>,
+type Rmc<'a, B> = RMCWrapper<'a, <B as Block>::Hash, KeyBox>;
+
+async fn process_new_block_data<'a, B, N>(
```
timorl, line 29: The addition of two functions here is somewhat worrying as the `party::foo` files for tasks were supposed to be very minimal things with close to no logic. But I guess you only dispelled the illusion in the case of aggregator. :/

Anyway, it would be nice to move some of this logic out of here, but I suspect it's out of scope for this PR – what do you think?

timorl (changes requested):
A couple comments, but in general looks like a nice improvement in abstraction.

finality-aleph/src/party/aggregator.rs
piotrMocz: As per our discussion: aggregator seems to care a bit too much about the networking layer, while it could be a simple piece of _almost_ pure logic. Even though this PR gets simpler after addressing the comments above, I still think this refactor deserves a subsequent PR.

finality-aleph/src/aggregator.rs
```diff
@@ -82,28 +83,59 @@ impl<
...
     }
 
     pub(crate) fn notify_last_hash(&mut self) {
         self.last_hash_placed = true;
     }
 
-    pub(crate) async fn next_multisigned_hash(
-        &mut self,
-    ) -> Option<(B::Hash, MK::PartialMultisignature)> {
+    pub(crate) async fn process_network_messages(&mut self) -> NetworkResult {
```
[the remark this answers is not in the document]
piotrMocz: Very nice suggestion, moving the loop inside the function really does make the whole thing clearer. I wanted to extract as much logic as possible out of the loops to make the testing easier but I guess in this case it would mean sacrificing code quality :)

finality-aleph/src/network/rmc.rs
piotrMocz: Two reasons: 1) I was told it was the correct way of doing it :joy:  2) I wasn't sure how Rust would be able to dispatch to the right method. Removed the wrappers and it works fine, thanks! :)

finality-aleph/src/aggregator/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+    collections::{HashMap, HashSet, VecDeque},
+    fmt::Debug,
+    marker::PhantomData,
+};
+
+mod multicast;
+
+pub use multicast::{Hash, Multicast, Multisigned, NetworkData as RmcNetworkData, SignableHash};
+
+/// A type encapsulating three different results of the `process_network_messages` method
```
pmikolajczyk41, line 16: this comment seems outdated

finality-aleph/src/aggregator/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+        self.multicast
+            .start_multicast(SignableHash::new(hash))
+            .await;
+    }
+
+    pub(crate) fn notify_last_hash(&mut self) {
+        self.last_hash_placed = true;
+    }
+
+    pub(crate) async fn wait_for_next_signature(&mut self) -> Result<(), AggregatorError> {
```
pmikolajczyk41, line 93: this one doesn't have to be exposed with `pub(crate)`, does it? or are you going to test it directly outside this module?

finality-aleph/src/aggregator/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+
+pub use multicast::{Hash, Multicast, Multisigned, NetworkData as RmcNetworkData, SignableHash};
+
+/// A type encapsulating three different results of the `process_network_messages` method
+pub(crate) enum AggregatorError {
+    NetworkChannelClosed,
+}
+
+/// A wrapper around an `rmc::Multicast` returning the signed hashes in the order of the [`Multicast::start_multicast`] calls.
+pub(crate) struct BlockSignatureAggregator<
```
pmikolajczyk41, line 22: personally, I do prefer light `mod.rs` files, thus, I would suggest placing this structure in a dedicated file and reexport it from here; for me, rather the content of `multicast.rs` should go here, but actually, it can stay where it is now as well

finality-aleph/src/aggregator/multicast.rs
```diff
@@ -0,0 +1,93 @@
...
+    rmc::Message, MultiKeychain, Multisigned as MultisignedStruct, Signable, SignatureSet,
+    UncheckedSigned,
+};
+use codec::{Codec, Decode, Encode};
+use sp_runtime::traits::Block;
+use std::{fmt::Debug, hash::Hash as StdHash};
+
+/// A wrapper allowing block hashes to be signed.
+#[derive(PartialEq, Eq, StdHash, Clone, Debug, Default, Encode, Decode)]
+pub struct SignableHash<H: Codec + Send + Sync> {
```
pmikolajczyk41, line 20: I do not feel so comfortable with this design, that here we require from `H` some properties that seems to be unnecessary (`Signable` expects `H` to be only `AsRef<[u8]>`) and nearby we define a trait `Hash` with a superset of these properties; I would either require minimal set here, or just use this `Hash` everywhere

Also, the trait `Hash` seems it is needed only here and in `BlockSignatureAggregator`, so probably it can be moved to `mod.rs` and stay private.

finality-aleph/src/aggregator/multicast.rs
```diff
@@ -0,0 +1,93 @@
...
+    }
+}
+
+/// A convenience trait for gathering all of the desired hash characteristics.
+pub trait Hash: AsRef<[u8]> + StdHash + Eq + Clone + Codec + Debug + Send + Sync {}
+
+impl<T: AsRef<[u8]> + StdHash + Eq + Clone + Codec + Debug + Send + Sync> Hash for T {}
+
+pub type NetworkData<B> =
+    Message<SignableHash<<B as Block>::Hash>, Signature, SignatureSet<Signature>>;
```
pmikolajczyk41, line 43: what about moving it to `mod.rs`? this is not used here

finality-aleph/src/aggregator/multicast.rs
```diff
@@ -0,0 +1,93 @@
...
+
+/// Anything that exposes the same interface as `aleph_bft::Multisigned` struct.
+pub trait Multisigned<'a, H: Hash, MK: MultiKeychain> {
+    fn as_signable(&self) -> &SignableHash<H>;
+    fn into_unchecked(self) -> UncheckedSigned<SignableHash<H>, MK::PartialMultisignature>;
```
pmikolajczyk41, line 48: why don't we use `Signable` here?

finality-aleph/src/party/aggregator.rs
pmikolajczyk41: I agree with both of you - most of the code from here that is not responsible for creation and launching should be moved to `aggregator` module, but I admit that it could be done in another PR

finality-aleph/src/aggregator/mod.rs
piotrMocz: I thought about that but it comes down to the hardest problem of them all: naming the new module :joy:
Any suggestions?

finality-aleph/src/aggregator/mod.rs
piotrMocz: At least for now I'm not, so I'll change it, thanks

finality-aleph/src/aggregator/multicast.rs
piotrMocz: I went for using the `Hash` everywhere here, so for now I can't make it private as it would leak as part of the public `SignableHash` definition. In a subsequent refactor I might strip down the type bounds to the bare minimum and then this refactor would be possible.

What do you think?

finality-aleph/src/aggregator/multicast.rs
piotrMocz: Truth be told, I don't know how to do it without resorting to `dyn`... I'd appreciate some pointers ;)

finality-aleph/src/aggregator/multicast.rs
piotrMocz: I went with the following structure:
```
+ aggregation/
|- aggregator.rs
|- mod.rs
|- multicast.rs
```

finality-aleph/src/aggregator/multicast.rs
pmikolajczyk41: what about bounded type parameter? like 
```rust
impl<'a, MK: MultiKeychain, S: Signable> Multisigned<'a, H, MK>
    for MultisignedStruct<'a, S, MK>
{
    fn as_signable(&self) -> &S {
        self.as_signable()
    }

    fn into_unchecked(self) -> UncheckedSigned<S, MK::PartialMultisignature> {
        self.into_unchecked()
    }
}
```

finality-aleph/src/aggregator/multicast.rs
pmikolajczyk41: > I went for using the Hash everywhere here...
for me it is completely fine; further refactor with limiting bounds is not so urgent as long as we do not intend to mock it or implement it somewhere outside
> I went with the following structure...
great, beautiful! :smiley:

finality-aleph/src/aggregator/mod.rs
pmikolajczyk41: I see that this is resolved in a good way :)

finality-aleph/src/aggregation/aggregator.rs
```diff
@@ -0,0 +1,146 @@
+use crate::aggregation::{Hash, Multicast, Multisigned, SignableHash};
+use crate::{metrics::Checkpoint, network::DataNetwork, Metrics};
```
pmikolajczyk41, line 2: it is strange that even though we have `imports_granularity = "Crate"` in our `rustfmt.toml` it doesn't merge these

pmikolajczyk41 (approved):

finality-aleph/src/aggregation/mod.rs
```diff
@@ -0,0 +1,12 @@
+use crate::crypto::Signature;
+use aleph_bft::{rmc::Message, SignatureSet};
+use sp_runtime::traits::Block;
+
+mod aggregator;
+mod multicast;
+
+pub use aggregator::BlockSignatureAggregator;
+pub use multicast::{Hash, Multicast, Multisigned, SignableHash};
```
timorl, line 9: Are those (under `multicast`) used outside of this module? If not they shouldn't be `pub` here.

timorl (approved):
A couple comments left, but in general LGTM.

finality-aleph/src/party/aggregator.rs
```diff
@@ -1,13 +1,15 @@
...
     justification::{AlephJustification, JustificationNotification},
     metrics::Checkpoint,
-    network::{DataNetwork, RmcNetworkData},
+    network::DataNetwork,
     party::{AuthoritySubtaskCommon, Task},
     BlockHashNum, Metrics, SessionBoundaries,
 };
-use aleph_bft::SpawnHandle;
+use aleph_bft::rmc::{DoublingDelayScheduler, ReliableMulticast};
+use aleph_bft::{KeyBox as BftKeyBox, SignatureSet, SpawnHandle};
```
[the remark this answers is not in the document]
piotrMocz: CLion's autoimport is playing tricks on me :(

finality-aleph/src/aggregation/multicast.rs
```diff
@@ -59,15 +59,17 @@ impl<'a, H: Hash, MK: MultiKeychain> Multisigned<'a, H, MK>
 /// The trait defines an associated type: `Signed`. For `ReliableMulticast`, this is simply
 /// `aleph_bft::Multisigned` but the mocks are free to use the simplest matching implementation.
 #[async_trait::async_trait]
-pub trait Multicast<H: Hash>: Send + Sync {
+pub trait Multicast<H: Hash, S: Signable>: Send + Sync {
     type Signed;
-    async fn start_multicast(&mut self, hash: SignableHash<H>);
-    fn get_multisigned(&self, hash: &SignableHash<H>) -> Option<Self::Signed>;
+    async fn start_multicast(&mut self, hash: S);
```
pmikolajczyk41, line 64: nit: I would rename the argument here and there to `signable`

pmikolajczyk41 (approved):

finality-aleph/src/party/aggregator.rs
piotrMocz: https://cardinal-cryptography.atlassian.net/browse/AZ-645

finality-aleph/src/aggregation/aggregator.rs
```diff
@@ -98,20 +102,27 @@ where
...
                 }
                 message_from_network = self.network.next() => {
-                    if let Some(message_from_network) = message_from_network {
-                        trace!(target: "aleph-aggregator", "Received message for rmc: {:?}", message_from_network);
-                        self.messages_for_rmc.unbounded_send(message_from_network).expect("sending message to rmc failed");
-                    } else {
-                        warn!(target: "aleph-aggregator", "the network channel closed");
-                        // In case the network is down we can terminate (?).
-                        return Err(AggregatorError::NetworkChannelClosed);
+                    match message_from_network {
```
timorl, line 116: Nit: The external braces are unneeded now.

timorl (approved):
I think there are still some instances of `aleph-party` rather than `aleph-aggregator` here and there, right?

PR #329, opened 2022-03-09
Adjust path in script

kostekIV:


scripts/run_local_e2e_pipeline.sh
```diff
@@ -10,7 +10,7 @@ docker build --tag aleph-node:latest -f ./docker/Dockerfile .
 # run the chain and the tests in two separate tmux windows
 tmux new-session -d -s aleph0 './.github/scripts/run_consensus.sh';
 tmux new-window -t "aleph0:1";
-tmux send-keys -t "aleph0:1" './run_e2e.sh' Enter;
+tmux send-keys -t "aleph0:1" './scripts/run_e2e.sh' Enter;
```
Marcin-Radecki, line 13: So this fix helps if we run `scripts/run_local_e2e_pipeline.sh` from root of the repo right? What about when we are in `scripts` directory, does it work?

kostekIV: no, but there is also a cmd from line 11 which also needs to be run at the root of the repo.

Marcin-Radecki: Okay, thanks.

Marcin-Radecki (approved):

PR #330, opened 2022-03-09
Added log reloading to entrypoint.

DamianStraszak:


docker/docker_entrypoint.sh
```diff
@@ -46,6 +46,7 @@ ARGS=(
   --no-mdns
   --ws-max-connections "${WS_MAX_CONNECTIONS}"
   --unsafe-ws-external --unsafe-rpc-external
+  --enable-log-reloading
```
Marcin-Radecki, line 49: Should we add it to `run_nodes-sh` as well?

Marcin-Radecki: Right, I just thought about consistency. I think we don't need it here then.

Marcin-Radecki (approved):

fbielejec (approved):

PR #332, opened 2022-03-11
Improve ong logging

Yithis:


finality-aleph/src/network/manager/service.rs
```diff
@@ -632,7 +632,7 @@ impl<D: Data> IO<D> {
                     match maybe_message {
                         Some(message) => if let Err(e) = self.on_network_message(&mut service, message) {
                             match e {
-                                Error::UserSend => warn!(target: "aleph-network", "Failed to send to user in session."),
+                                Error::UserSend => trace!(target: "aleph-network", "Failed to send to user in session."),
```
timorl, line 635: Maybe also change the one below, getting messages from unknown sessions is actually also expected to happen sometimes during normal operation.

DamianStraszak: Yes, agreed. And indeed these messages happen quite often.

DamianStraszak (approved):

PR #336, opened 2022-03-15
AZ-690 Fix a too high finalized block bug in chain tracker

DamianStraszak:


finality-aleph/src/data_io/data_provider.rs
```diff
@@ -148,7 +148,15 @@ where
         // of `data_to_propose` unaffected.
 
         let client_info = self.client.info();
-        let finalized_block = (client_info.finalized_hash, client_info.finalized_number).into();
+        let finalized_block: BlockHashNum<B> =
```
pmikolajczyk41, line 151: is ascription needed now?

finality-aleph/src/data_io/data_provider.rs
```diff
@@ -148,7 +148,15 @@ where
...
+            (client_info.finalized_hash, client_info.finalized_number).into();
+
+        if finalized_block.num >= self.session_boundaries.last_block() {
+            // This session is already finished, but this instance of ChainTracker has not been terminated yet.
+            // We go with the default -- empty proposal, this does not have any significance.
+            *self.data_to_propose.lock().await = AlephData::Empty;
```
pmikolajczyk41, line 157: i do not want to be seen as mental, but it seems that a small testcase would be kinda appropriate - just to expose this case, which occurred to be easily overlooked

mike1729 (approved):

finality-aleph/src/data_io/data_provider.rs
DamianStraszak: Yes :/

finality-aleph/src/data_io/data_provider.rs
DamianStraszak: Added a test case.

finality-aleph/src/data_io/data_provider.rs
pmikolajczyk41: okay, I see that it does not compile without it, what makes me really confused:
 - before it was not necessary (e.g. by usages like `prev.highest_finalized == finalized_block` the type could have been inferred)
 - it seems that because now the first usage (`finalized_block.num >= ...`) does not allow for a full inference the compiler starts shouting
 
strange :|

pmikolajczyk41 (approved):

PR #341, opened 2022-03-17
Enable wss in aleph-client

pmikolajczyk41:
Here we introduce two changes to `create_connection` and `create_custom_connection` methods:

1. In `create_custom_connection` we use loop instead of tail recursion, as recently we have observed that when connection problems persist, the program fails with stack overflow.
2. We can pass a flag `ssl` specifying whether we aim for `wss` or `ws` protocol. While in all tools within this repo we need only `ws`, there are other tools like `rest-endpoints` or `traffic-maker` which use `aleph-client` crate, but require secured connections.

### DIFF
--- a/aleph-client/src/lib.rs
+++ b/aleph-client/src/lib.rs
@@ -36,15 +36,30 @@ pub type Header = GenericHeader<BlockNumber, BlakeTwo256>;
 pub type KeyPair = sr25519::Pair;
 pub type Connection = Api<KeyPair, WsRpcClient>;
 
-pub fn create_connection(address: &str, ssl: bool) -> Connection {
-    create_custom_connection(address, ssl).expect("connection should be created")
+#[derive(Copy, Clone, Debug)]
+pub enum Protocol {
+    WS,
+    WSS,
+}
+
+impl Default for Protocol {
+    fn default() -> Self {
+        Protocol::WS
+    }
+}
+
+pub fn create_connection(address: &str, protocol: Protocol) -> Connection {
+    create_custom_connection(address, protocol).expect("connection should be created")
 }
 
 pub fn create_custom_connection<Client: FromStr + RpcClient>(
     address: &str,
-    ssl: bool,
+    protocol: Protocol,
 ) -> Result<Api<sr25519::Pair, Client>, <Client as FromStr>::Err> {
-    let protocol = if ssl { "wss" } else { "ws" };
+    let protocol = match protocol {
+        Protocol::WS => "ws",
+        Protocol::WSS => "wss",
+    };
     loop {
         let client = Client::from_str(&format!("{}://{}", protocol, address))?;
         match Api::<sr25519::Pair, _>::new(client) {
--- a/aleph-client/src/lib.rs
+++ b/aleph-client/src/lib.rs
@@ -36,15 +36,17 @@ pub type Header = GenericHeader<BlockNumber, BlakeTwo256>;
 pub type KeyPair = sr25519::Pair;
 pub type Connection = Api<KeyPair, WsRpcClient>;
 
-pub fn create_connection(address: &str) -> Connection {
-    create_custom_connection(address).expect("connection should be created")
+pub fn create_connection(address: &str, ssl: bool) -> Connection {
+    create_custom_connection(address, ssl).expect("connection should be created")
 }
 
 pub fn create_custom_connection<Client: FromStr + RpcClient>(
     address: &str,
+    ssl: bool,
 ) -> Result<Api<sr25519::Pair, Client>, <Client as FromStr>::Err> {
+    let protocol = if ssl { "wss" } else { "ws" };
     loop {
-        let client = Client::from_str(&format!("ws://{}", address))?;
+        let client = Client::from_str(&format!("{}://{}", protocol, address))?;
         match Api::<sr25519::Pair, _>::new(client) {
             Ok(api) => return Ok(api),
             Err(why) => {
--- a/aleph-client/src/lib.rs
+++ b/aleph-client/src/lib.rs
@@ -43,16 +43,17 @@ pub fn create_connection(address: &str) -> Connection {
 pub fn create_custom_connection<Client: FromStr + RpcClient>(
     address: &str,
 ) -> Result<Api<sr25519::Pair, Client>, <Client as FromStr>::Err> {
-    let client = Client::from_str(&format!("ws://{}", address))?;
-    match Api::<sr25519::Pair, _>::new(client) {
-        Ok(api) => Ok(api),
-        Err(why) => {
-            warn!(
-                "[+] Can't create_connection because {:?}, will try again in 1s",
-                why
-            );
-            sleep(Duration::from_millis(1000));
-            create_custom_connection(address)
+    loop {
+        let client = Client::from_str(&format!("ws://{}", address))?;
+        match Api::<sr25519::Pair, _>::new(client) {
+            Ok(api) => return Ok(api),
+            Err(why) => {
+                warn!(
+                    "[+] Can't create_connection because {:?}, will try again in 1s",
+                    why
+                );
+                sleep(Duration::from_millis(1000));
+            }
         }
     }
 }
--- a/aleph-client/src/lib.rs
+++ b/aleph-client/src/lib.rs
@@ -48,6 +48,13 @@ impl Default for Protocol {
     }
 }
 
+pub fn from(use_ssl: bool) -> Protocol {
+    match use_ssl {
+        true => Protocol::WSS,
+        false => Protocol::WS,
+    }
+}
+
 pub fn create_connection(address: &str, protocol: Protocol) -> Connection {
     create_custom_connection(address, protocol).expect("connection should be created")
 }
--- a/benches/payout-stakers/src/main.rs
+++ b/benches/payout-stakers/src/main.rs
@@ -1,4 +1,5 @@
 use aleph_client::create_connection;
+use aleph_client::staking_bond;
 use e2e::{
     accounts::derive_user_account,
     staking::{
@@ -17,7 +18,6 @@ use sp_core::sr25519::Pair;
 use sp_keyring::AccountKeyring;
 use std::iter;
 use substrate_api_client::{rpc::WsRpcClient, Api, XtStatus};
-use aleph_client::staking_bond;
 
 // testcase parameters
 const NOMINATOR_COUNT: u64 = 1024;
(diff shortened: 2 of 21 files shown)


bin/cliain/src/main.rs
```diff
@@ -12,6 +12,10 @@ struct Config {
     #[clap(long, default_value = "127.0.0.1:9944")]
     pub node: String,
 
+    /// Whether to use `ws` or `wss` protocol
+    #[clap(long)]
```
kostekIV, line 16: add default here

aleph-client/src/lib.rs
```diff
@@ -36,23 +36,26 @@ pub type Header = GenericHeader<BlockNumber, BlakeTwo256>;
 pub type KeyPair = sr25519::Pair;
 pub type Connection = Api<KeyPair, WsRpcClient>;
 
-pub fn create_connection(address: &str) -> Connection {
-    create_custom_connection(address).expect("connection should be created")
+pub fn create_connection(address: &str, ssl: bool) -> Connection {
```
Marcin-Radecki, line 39: We can consider not using bool arguments in public API functions, this is because it's A) easy to make swap error as a caller when they are more than one bool params (not this case here but making a general point) and B) without peeking into docs harder to guess what those bool args are about). E.g. consider `pub fn hedge_assets(native_currency: bool, buy: bool` as caller `hedge_assets(true, false) -without peeking into docs. The alternative is to expose public API functions with proper naming e.g.
```rust
pub fn create_connection_secured(address: &str) -> Connection {
     create_connection(address, true)
}

pub fn create_connection_not_secured(address: &str) -> Connection {
     create_connection(address, false)
}

fn create_connection(address: &str, ssl: bool) -> Connection {
// ....
}
```

benches/payout-stakers/src/main.rs
```diff
@@ -1,4 +1,5 @@
 use aleph_client::create_connection;
+use aleph_client::staking_bond;
```
Marcin-Radecki, line 2: Could you please collapse `use`s

local-tests/send-runtime/src/main.rs
```diff
@@ -20,6 +20,10 @@ struct Args {
     #[structopt(long, short, name = "ADDRESS")]
     url: String,
 
+    /// Whether to use `wss` or `ws` protocol
+    #[structopt(long)]
+    ssl: bool,
```
kostekIV, line 25: add default here (does test_update works without it?)

aleph-client/src/lib.rs
kostekIV, line 39: Maybe enum instead of the bool:
```rust
enum Protocol {
  WS,
  WSS
}
```

local-tests/send-runtime/src/main.rs
pmikolajczyk41: compiler shouts about meaningless value - check for yourself

bin/cliain/src/main.rs
pmikolajczyk41: see the other comment

aleph-client/src/lib.rs
pmikolajczyk41: yes, I was thinking about it, but gave up this because then, there is a need of putting this ugly conditional in many places:
```
if config.ssl { create_connection_secured } else { create_connection_not_secured }
```
and this imho is worse, because we have this semantics in methods, that should not care much; for now we are just passing this argument from the user to the end, so it is not so perilous

anyway, maybe using `enum` as @kostekIV suggested is a compromise?

benches/payout-stakers/src/main.rs
```diff
@@ -30,23 +30,25 @@ const NOMINATE_CALL_BATCH_LIMIT: usize = 128;
...
 fn main() -> Result<(), anyhow::Error> {
     let address = "127.0.0.1:9944";
+    let ssl = false;
     let sudoer = AccountKeyring::Alice.pair();
 
     env_logger::init();
     info!("Starting benchmark with config ");
 
-    let connection = create_connection(address).set_signer(sudoer);
+    let connection = create_connection(address, ssl).set_signer(sudoer);
```
Marcin-Radecki, line 39: Please consider using named function e.g. `create_connection_ws` or `create_connection_not_secured` here instead of bool param. Or name this param `not_secured` so it's more readable to see what type of connection we're creating here.

benches/payout-stakers/src/main.rs
```diff
@@ -30,23 +30,25 @@ const NOMINATE_CALL_BATCH_LIMIT: usize = 128;
...
 
     let accounts = generate_1024_accounts(&connection);
-    let validators = set_validators(address);
+    let validators = set_validators(address, ssl);
```
Marcin-Radecki, line 42: Ideally, we should not be passing `ssl` here but move to refactor `set_validator` API to have `Connection&` as an argument. I know it's more work as we'd need to update all depndess of `set_validator`, but then we don't need to pass through all crates this bool flag. Even more ideally we should move `set_validators` to `aleph-client`, please see other functions there as we're trying to establish some kind of convention that all `aleph-client` functions should have `Connection&`.

benches/payout-stakers/src/main.rs
```diff
@@ -30,23 +30,25 @@ const NOMINATE_CALL_BATCH_LIMIT: usize = 128;
...
     let nominee = nominate_validator_0(&connection, accounts, &validators);
-    wait_for_10_eras(address, &connection, nominee)?;
+    wait_for_10_eras(address, ssl, &connection, nominee)?;
```
Marcin-Radecki, line 44: Same comment as per `set_validator` above.

bin/cliain/src/keys.rs
```diff
@@ -1,10 +1,12 @@
-use aleph_client::{staking_bond, set_keys, rotate_keys, create_connection, KeyPair};
+use aleph_client::{create_connection, rotate_keys, set_keys, staking_bond, KeyPair};
 use primitives::staking::MIN_VALIDATOR_BOND;
 use substrate_api_client::XtStatus;
 
-pub fn prepare(node: String, key: KeyPair) {
-    let connection = create_connection(&node).set_signer(key.clone());
+pub fn prepare(node: String, ssl: bool, key: KeyPair) {
+    let connection = create_connection(&node, ssl).set_signer(key.clone());
```
Marcin-Radecki, line 6: Is it possible to make a `connection` object not in the command functions, but in `main` of `cliain` create? And pass the `connection` object to those methods. Btw this PR adds way more command: https://github.com/Cardinal-Cryptography/aleph-node/pull/338 , so I'd like to establish a common convention

bin/cliain/src/main.rs
Marcin-Radecki, line 16: Can we make a default `false` here? Or it is what it's going to be when one does not pass `--ssl` arg. There're some scripts in https://github.com/Cardinal-Cryptography/aleph-node/pull/338 which do not use this `cliain` version yet.

e2e-tests/src/config.rs
```diff
@@ -7,6 +7,10 @@ pub struct Config {
     #[clap(long, default_value = "127.0.0.1:9943")]
     pub node: String,
 
+    /// Whether to use `ws` or `wss` protocol
+    #[clap(long)]
+    pub ssl: bool,
```
Marcin-Radecki, line 12: If `--ssl` is mandatory arg, do we need to update any `scripts` or workflows?

bin/cliain/src/main.rs
pmikolajczyk41: afaik it is `false` by default (actually I cannot set `default-value` for `bool`)

flooder/src/main.rs
```diff
@@ -505,7 +505,8 @@ fn create_connection_pool(
         .map(|urls| {
             urls.iter()
                 .map(|url| {
-                    create_custom_connection(url).expect("it should return initialized connection")
+                    create_custom_connection(url, false)
```
Marcin-Radecki, line 508: Please consider naming this to variable or constant like `let not_secured = false`

e2e-tests/src/config.rs
pmikolajczyk41: tests are passing, so it is not mandatory iiuc; if you add `--ssl` you are just passing `true`

Marcin-Radecki (changes requested):
Nice contriibution. some general questions to design

bin/cliain/src/main.rs
Marcin-Radecki: Yes, it makes sense to be `false` by default, just wanted to confirm, thanks.

e2e-tests/src/config.rs
Marcin-Radecki: Oks

aleph-client/src/lib.rs
Marcin-Radecki: Afaik in `e2e-client` we can leave by default `create_connection_not_secured` in all code, so does in `benches/payout_stakers`.  In `cliain` we'd need to pass it but just once, see the other comment to pass `Connection&` to `cliain` command methods.

Marcin-Radecki: And yes , either enums or named functions are good for me.

fbielejec: imho with clear naming the problem dissappears, e.g `use_ssl: true/false`

fbielejec (approved):
no major comments, maybe use_ssl / is_ssl would be a clearer naming for a bool arg, but YMMV

benches/payout-stakers/src/main.rs
fbielejec: Maybe helpful, with this in fmt config:
```
edition = "2021"
imports_granularity = "Crate"
use_field_init_shorthand = true
reorder_imports = true
reorder_modules = true

```
I was able to add rust fmt to the on-save hook (in emacs, should be possible with other IDEs too)

aleph-client/src/lib.rs
pmikolajczyk41: I changed to `enum`. The only inconvenience is that I need to implement some kind of parsing for each cli tool (I cannot derive `ArgEnum` on this `enum` as it would require adding `clap` dependency to `aleph-client`)

benches/payout-stakers/src/main.rs
pmikolajczyk41: done, but had to do it manually

aleph-client/src/lib.rs
pmikolajczyk41: thx, done

benches/payout-stakers/src/main.rs
pmikolajczyk41: solved in other thread

benches/payout-stakers/src/main.rs
pmikolajczyk41: done

benches/payout-stakers/src/main.rs
pmikolajczyk41: done

flooder/src/main.rs
pmikolajczyk41: solved with `enum Protocol`

Marcin-Radecki (approved):

flooder/src/config.rs
```diff
@@ -69,3 +74,10 @@ pub fn read_phrase(phrase: String) -> String {
         phrase
     }
 }
+
+fn parse_to_protocol(use_ssl: bool) -> Protocol {
+    match use_ssl {
+        true => Protocol::WSS,
+        false => Protocol::WS,
+    }
+}
```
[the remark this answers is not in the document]
pmikolajczyk41: done

PR #343, opened 2022-03-18
AZ-676: pure aggregator

piotrMocz:
Splits the `BlockSignatureAggregator` into two parts:
* the _pure_ part, only containing the core logic (still called `BlockSignatureAggregator`)
* the _impure_/networking part, containing the channels and very simple sending logic (called `IO` but imported as `AggregatorIO`).

The `IO` methods take the aggregator as an argument: making the aggregator into one of `IO`'s fields makes the code more succint, although at the cost of slightly breaking the semantics of `IO` -- I'm open to discussion ;)


finality-aleph/src/aggregation/aggregator.rs
```diff
@@ -14,29 +14,112 @@ use std::{
...
+    'a,
+    H: Hash + Copy,
+    MK: MultiKeychain,
+    MS: Multisigned<'a, SignableHash<H>, MK>,
+> {
+    signatures: HashMap<H, MK::PartialMultisignature>,
+    hash_queue: VecDeque<H>,
+    last_hash_placed: bool,
+    started_hashes: HashSet<H>,
+    metrics: Option<Metrics<H>>,
```
deuszx, line 42: Isn't inclusion of metrics here making this _technically_ impure again?

deuszx:
Did you consider making the _pure_ part of the aggregator a member of its IO counterpart? You would still get the advantage of being able to easily unit-test it. 

On the other hand, is there a change that we may have multiple different `Aggregator`s but single `AggregatorIO`?

piotrMocz:
Yes, I actually started out that way: it makes the code easier to read but
strictly speaking the aggregator is not IO, so this is a trade off between
readability and good semantics 🙂

W dniu pt., 18.03.2022 o 10:06 deuszx ***@***.***> napisał(a):

deuszx:
> Yes, I actually started out that way: it makes the code easier to read but strictly speaking the aggregator is not IO, so this is a trade off between readability and good semantics 🙂 

But the `AggregatorIO` kind of is, right? I think it's not only about good semantics though. The way it's encoded right now would allow for multiple, different aggregators to be passed to the same `AggregatorIO` but would it be correct? There is a set of channels there that are fed with some data. Each element can be consumed only once. What if someone makes a mistake and uses the same `AggregatorIO` on multiple `Aggregator`s? That would probably get both into an invalid state since elements from the channels are supposed to be passed on to the same, single `Aggregator`.

piotrMocz:
> > Yes, I actually started out that way: it makes the code easier to read but strictly speaking the aggregator is not IO, so this is a trade off between readability and good semantics slightly_smiling_face
> 
> But the `AggregatorIO` kind of is, right? I think it's not only about good semantics though. The way it's encoded right now would allow for multiple, different aggregators to be passed to the same `AggregatorIO` but would it be correct? There (in the `AggregatorIO`) is a set of channels there that are fed with some data. Each element can be consumed only once. What if someone makes a mistake and uses the same `AggregatorIO` on multiple `Aggregator`s? That would probably get both into an invalid state since elements from the channels are supposed to be passed on to the same, single `Aggregator`.

That's a very valid point and I agree -- I'll be happy to make the change provided that @timorl agrees as well ;)

deuszx:
@piotrMocz , looks like @timorl gave a silent approval :P

piotrMocz:
> @piotrMocz , looks like @timorl gave a silent approval - that's how I'm interpreting the thumbs-up see_no_evil

@deuszx yes, he even confirmed that in a meeting, so we're good to go -- I made the change :slightly_smiling_face:

finality-aleph/src/aggregation/aggregator.rs
```diff
@@ -14,29 +14,112 @@ use std::{
...
+        self.hash_queue.push_back(hash);
+
+        Ok(())
+    }
+
+    pub(crate) fn notify_last_hash(&mut self) {
+        self.last_hash_placed = true;
+    }
+
+    fn on_multisigned_hash(&mut self, multisigned_hash: MS) -> AggregatorResult {
```
timorl, line 76: This shouldn't be returning a result if it cannot ever return an error – I thought this was a mistake in a logging message about network failing below, but the problem actually is here.

finality-aleph/src/aggregation/aggregator.rs
```diff
@@ -14,29 +14,112 @@ use std::{
...
-pub(crate) enum AggregatorError {
+pub enum AggregatorError {
+    LastHashPlaced,
+    NoHashFound,
+    DuplicateHash,
+}
+
+pub enum IOError {
     NetworkChannelClosed,
+    AggregatorError(AggregatorError),
```
timorl, line 25: I don't think you need this, you actually handle all aggregator errors internally to `IO`, see a comment below for the single usage and why it should be removed.

finality-aleph/src/aggregation/aggregator.rs
```diff
@@ -131,23 +202,20 @@ where
...
-                        return None;
-                    }
+                Err(AggregatorError::LastHashPlaced) => {
+                    debug!(
+                        target: "aleph-aggregator",
+                        "Terminating next_multisigned_hash because the last hash has been signed."
+                    );
+                    return None;
                 }
+                Err(_) => {}
```
timorl, line 216: This looks mysterious, perhaps spell out the other kind of error you expect here (`NoHashFound`) and ignore it, while logging any unexpected error as problematic (although you can likely ignore it in the logic, it'll never happen anyway).

timorl (changes requested):
Nice, some minor comments.

timorl (approved):

PR #347, opened 2022-03-21
AZ-596 towards new docker images

fixxxedpoint:
It uses crate2nix instead of naersk for building aleph-node. It gives us possibility to use nix's cache mechanism or build docker images containing most of the rust dependencies `precompiled` (which should speed up CI as well). It also defines a minimalistic docker image capable of running aleph-node. Easiest way to interact with it is to follow the docker based guide from BUILD.md.


default.nix
```diff
@@ -1,130 +1,18 @@
...
-  installPhase = ''
-    mkdir -p $out/bin
-    mv target/x86_64-unknown-linux-gnu/release/aleph-node $out/bin/
-  '';
-
-  fixupPhase = ''
-    find $out -type f -exec patchelf --shrink-rpath '{}' \; -exec strip '{}' \; 2>/dev/null
-  '';
-}
+if allCrates then
```
piotrMocz, line 15: I think this flag's logic would be more apparent when flipped, i.e.
```
if alephNodeOnly then
    workspaceMembers."aleph-node"
else
    allWorkspaceMembers
```

fixxxedpoint: actually I would prefer to have something like `crates = [ "aleph-node" ]` as parameter, where for all crates there would be a special value `[ ]`

BUILD.md
```diff
@@ -51,6 +51,7 @@ aleph-node>`.
 ## Manual
 These are build dependencies we use in our linux images for `aleph-node`:
 ```
+rust-nightly-2021-10-24
```
Marcin-Radecki, line 54: Not related to this change, but are there any significant changes to the dependencies version in the current production build? Ie, what below list looks like for a `aleph-node` binary that's running in our mainnet atm.

nix/versions.nix
```diff
@@ -0,0 +1,113 @@
...
+
+  gitignoreSource = (import fetchGitignoreSource { inherit (nixpkgs) lib; }).gitignoreSource;
+
+  fetchRocksdb = builtins.fetchGit {
+    url = "https://github.com/facebook/rocksdb.git";
+    ref = "refs/tags/v${rocksDBVersion}";
+  };
+
+  # we use a newer version of rocksdb than the one provided by nixpkgs
+  # we disable all compression algorithms and force it to use SSE 4.2 cpu instruction set
```
Marcin-Radecki, line 82: Just to confirm, those are not some braking changes to RocksDB driver crate, and they are fully compatible with the format we have on production

Marcin-Radecki:
This is decent work but as well a lot of new tech knowledge; it'd be good to have some knowledge sharing on this , like nix shenanigans or just nix syntax

BUILD.md
fixxxedpoint: Current `production` build is based on github's ubuntu-latest, so we kind of don't know which versions we are using. This PR will force build procedure to use fixed versions - if we ever want to use newer version of some package, we need to manually update `versions.nix` file.

nix/versions.nix
fixxxedpoint: Enabling SSE 4.2 support shouldn't break anything (but who knows; Substrate doesn't say anything about it?), snappy support might be problematic, since I think kvdb-rocksdb kind of expects snappy to be available. Local and aws tests were green. Anyway, it should be tested.

default.nix
piotrMocz: Oh sorry, forgot to reply since I agree with you :joy:

PR #348, opened 2022-03-21
add test for catch up mechanism

kostekIV:


timorl:
Does this fix [AZ-475](https://cardinal-cryptography.atlassian.net/browse/AZ-475)?

kostekIV:
> Does this fix [AZ-475](https://cardinal-cryptography.atlassian.net/browse/AZ-475)?

maybe as a side effect. Though imho the task requirements seem more rigorous with the state checking from what we have here.

Marcin-Radecki:
Should new pipeline member be where it is now? Just double-checking , as e2e test suite is run later 

image

kostekIV:
> Should new pipeline member be where it is now? Just double-checking , as e2e test suite is run later
> 
> image

for this test we just need a binary with aleph-node nothing else.

.github/scripts/test_catch_up.sh
```diff
@@ -0,0 +1,20 @@
...
+  exit 1
+fi
+
+echo 'Preparing environment'
+chmod +x $BINARY
+
+pip install -r requirements.txt
+
+echo 'Running test'
+./test_catch_up.py
```
Marcin-Radecki, line 18: Minor: we can use `exec ./test_catch_up.py` to decrease proces child-parent stack (really for debugging purposes in case of some dangling test process etc)

local-tests/chainrunner/chain.py
```diff
@@ -28,39 +29,41 @@ def __getitem__(self, i):
     def __iter__(self):
         return iter(self.nodes)
 
-    def bootstrap(self, binary, accounts, **kwargs):
-        """Bootstrap the chain. `accounts` should be a list of strings.
+    def bootstrap(self, binary, validator_accounts, accounts=None, **kwargs):
```
Marcin-Radecki, line 32: Where are other testcases which pass `accounts` ? Maybe we can rework them so they use `validator_accounts` instead. I understand now we support both

local-tests/chainrunner/chain.py
```diff
@@ -28,39 +29,41 @@ def __getitem__(self, i):
...
         self.nodes = []
-        for a in accounts:
+        for a in chain(validator_accounts, accounts):
             n = Node(binary, chainspec, op.join(self.path, a), self.path)
             n.flags['node-key-file'] = op.join(self.path, a, 'p2p_secret')
             self.nodes.append(n)
 
-    def set_flags(self, *args, **kwargs):
-        """Set common flags for all nodes.
+    def set_flags(self, *args, predicate=lambda n, i: True, **kwargs):
```
Marcin-Radecki, line 53: I am thinking whether  `predicate` argument can be part of a `bootstrap`, as this is where `self.nodes` are constucted

Marcin-Radecki (changes requested):
Nice usage of `local-tests` framework, minor comments

local-tests/chainrunner/chain.py
kostekIV: here I wanted something that could set different flags to different nodes but maybe it makes more sense to divide nodes as you have mentioned in the bootstrap and have 3 functions (set_flags, set_flags_for_validators, set_flags_for_nonvalidators). wdyt?

.github/scripts/test_catch_up.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -eu
+
+pushd local-tests/
+
+if [ ! -f "$BINARY" ]; then
```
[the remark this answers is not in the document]
kostekIV: done

.github/scripts/test_catch_up.sh
```diff
@@ -0,0 +1,20 @@
...
+  echo "Binary $BINARY does not exist."
+  exit 1
+fi
+
+echo 'Preparing environment'
+chmod +x $BINARY
```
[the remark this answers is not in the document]
kostekIV: done

local-tests/chainrunner/chain.py
Marcin-Radecki: Yes please good idea.

local-tests/chainrunner/chain.py
kostekIV: in this aleph-node repo

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -296,6 +296,30 @@ jobs:
...
+
+      - name: Download all artifacts
+        uses: actions/download-artifact@v2
+        with:
+          path: local-tests/
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
```
h4nsu, line 315: Do you think it would possible to somehow reuse these steps in `test-runtime-update` step to avoid doing the same setup twice?

local-tests/chainrunner/chain.py
```diff
@@ -21,32 +22,48 @@ def __init__(self, workdir):
...
 
     def __getitem__(self, i):
         return self.nodes[i]
 
     def __iter__(self):
         return iter(self.nodes)
 
-    def bootstrap(self, binary, accounts, **kwargs):
-        """Bootstrap the chain. `accounts` should be a list of strings.
+    def bootstrap(self, binary, validator_accounts, accounts=None, **kwargs):
```
h4nsu, line 34: Maybe naming these arguments `validators` and `nonvalidators` would be more descriptive?

local-tests/chainrunner/chain.py
```diff
@@ -21,32 +22,48 @@ def __init__(self, workdir):
...
-            self.nodes.append(n)
+        def account_to_node(account):
+            n = Node(binary, chainspec, op.join(self.path, account), self.path)
+            n.flags['node-key-file'] = op.join(self.path, account, 'p2p_secret')
+            return n
+
+        self.validator_nodes = [account_to_node(a) for a in validator_accounts]
+        self.nonvalidator_nodes = [account_to_node(a) for a in accounts]
+
+        self.nodes = list(chain(self.nonvalidator_nodes, self.validator_nodes))
```
h4nsu, line 57: Why not `self.nodes = self.nonvalidator_nodes + self.validator_nodes`? Not really an issue, just curious

local-tests/test_catch_up.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def check_finalized(nodes):
+    results = [node.highest_block() for node in nodes]
+    highest, finalized = zip(*results)
+    print('Blocks seen by nodes:')
+    print('  Highest:   ', *highest)
+    print('  Finalized: ', *finalized)
+
+    return finalized
```
h4nsu, line 22: This is almost 1:1 duplicate from the other test. Maybe moving it to `chainrunner/utils.py` is not a bad idea?

local-tests/test_catch_up.py
```diff
@@ -0,0 +1,75 @@
...
+                rpc_port=Seq(9933),
+                unit_creation_delay=200,
+                execution='Native')
+
+chain.set_flags_validator('validator')
+
+print('Starting the chain')
+chain.start('aleph')
+
+print('Waiting 30s')
```
h4nsu, line 47: Recently while running similar small tests locally I noticed that sometimes, at the very beginning of the chain life, the finalization is kinda lazy and lags at 0 until around block 100. Do you think that could be a problem here?

h4nsu (approved):

Marcin-Radecki (approved):

local-tests/test_catch_up.py
kostekIV: I don't think the test will fail because of that but probably will not test fully what we want (catching up from some already finalized state)

.github/workflows/e2e-tests-main-devnet.yml
kostekIV: not sure how to do it cleanly but at the same time spent here is negligible compared to other steps in the pipeline.
Screenshot from 2022-03-23 09-52-45

h4nsu (approved):

local-tests/chainrunner/chain.py
kostekIV: done

local-tests/chainrunner/chain.py
kostekIV: :| changed

local-tests/test_catch_up.py
kostekIV: done

.github/workflows/e2e-tests-main-devnet.yml
h4nsu: Nevermind then

local-tests/test_catch_up.py
h4nsu: Then maybe some bigger delay (90s?) is worth considring here?

kostekIV: set to 90s

PR #349, opened 2022-03-22
embedded chainspecs

fixxxedpoint:
- both mainnet and testnet chainspecs in raw format embedded into aleph-node binary
- mainnet uses conditional compilation - feature mainnet
- backward compability - calling with `chain=<filepath>` loads a chainspec from a file
- default value = testnet (mainnet when compiled with feature mainnet)


bin/node/src/cli.rs
```diff
@@ -44,11 +44,17 @@ impl SubstrateCli for Cli {
     }
 
     fn load_spec(&self, id: &str) -> Result<Box<dyn sc_service::ChainSpec>, String> {
+        #[cfg(not(feature = "mainnet"))]
+        let default_chain = "testnet";
+        #[cfg(feature = "mainnet")]
+        let default_chain = "mainnet";
```
artemis-cc, line 50:
```suggestion
        let default_chain = if cfg!(feature = "mainnet") { "mainnet" } else { "testnet" };
```

fixxxedpoint:
> I'm wondering if it would be better for the crate feature to turn on _support_ for connecting to the mainnet, but leave the _default_ set to the _testnet_ (still requiring `--chain=mainnet` on the command line to use the mainnet). Maybe I'm worrying too much, but I'm just imagining someone building with `--feature=mainnet` or even just `--all-features` without thinking, and suddenly they have a node on the mainnet by accident…

I think you're right and this conditional compilation thing is just useless here. I will change it to default = testnet.

bin/node/src/cli.rs
fixxxedpoint: Removed and using `testnet` as default value.

mike1729 (approved):

artemis-cc (approved):
:+1: Fairly simple and looks good to me.

PR #353, opened 2022-03-24
Move functionality from e2e-tests to aleph-client

pmikolajczyk41:
We move most of the functionality from `e2e/src` directory to `aleph-client`. There are no semantic changes, however, in some places minor refactor has been done. Finally, `benches/payout-stakers` is no longer dependent on `e2e`

Once #351 is merged, the target branch will be automatically changed and pipelines should trigger. Also, some unrelated changes from main will disappear.


bin/cliain/src/staking.rs
```diff
@@ -38,5 +38,5 @@ pub fn set_staking_limits_command(
 }
 
 pub fn force_new_era_command(root_connection: Connection) {
```
fbielejec, line 40: this function seems to be just adding unnecessary indirection?

local-tests/send-runtime/src/main.rs
```diff
@@ -44,6 +44,6 @@ fn main() {
         .expect("Could not send extrinsic");
 }
 
-fn keypair_from_string(seed: &String) -> sr25519::Pair {
-    sr25519::Pair::from_string(&seed, None).expect("Can't create pair from seed value")
+fn keypair_from_string(seed: &str) -> sr25519::Pair {
```
fbielejec, line 47:
```suggestion
fn keypair_from_str(seed: &str) -> sr25519::Pair {
```

fbielejec (commented):
minor comment

bin/cliain/src/staking.rs
Marcin-Radecki: Right, it was done (by me) to comply with `*_command` function name convention but I guess we don't need that additional level of indirection. I can volunter to remove it in my next PR.

Marcin-Radecki (approved):
This is a great work, finally we got rid of that dependency of `payput_stakers` on `e2e-client`.

bin/cliain/src/staking.rs
pmikolajczyk41: there was an issue raised today by @h4nsu that actually `cliain` should be just a binary interface for `aleph-client`

Marcin-Radecki: Yes, we can back up this and push this idea of binary interface. Btw how does exactly the solution looks like? We'd have one package `aleph-client` which will have `lib.rs` and `main.rs`? This will work, however one would not be able to depend on that package since that package would be a binary one.

Marcin-Radecki: Ok we spoke on the side and turns out we can make `cliain` and `aleph-node` distinct rust crates in one `Cargo.toml` in the same folder on FS. That would work for me.

fbielejec (approved):

PR #354, opened 2022-03-25
Whole state is downloaded and the new spec is applied on top of it.

DamianStraszak:
Changes to the fork off tool:
-- One block is used as a reference point to download the state
-- The whole state is downloaded and only certain prefixes are replaced from the provided spec
-- The downloaded state is dumped to a file
-- There are workers that allow to parallelize downloading values.

### DIFF
--- a/fork-off/src/main.rs
+++ b/fork-off/src/main.rs
@@ -48,7 +48,7 @@ pub struct Config {
         value_delimiter = ',',
         default_value = "Aura,Aleph,Sudo,Staking,Session,Elections"
     )]
-    pub pallets: Vec<String>,
+    pub pallets_keep_state: Vec<String>,
 }
 
 const KEYS_BATCH_SIZE: u32 = 1000;
@@ -125,21 +125,22 @@ impl StateFetcher {
         input: Arc<Mutex<Vec<StorageKey>>>,
         output: Arc<Mutex<Storage>>,
     ) {
+        const LOG_PROGRESS_FREQUENCY: usize = 500;
         let fetcher = StateFetcher::new(http_rpc_endpoint);
         loop {
-            let maybe_key: Option<StorageKey> = {
+            let maybe_key = {
                 let mut keys = input.lock();
                 keys.pop()
             };
-            if let Some(key) = maybe_key {
-                let value = fetcher.get_value(key.clone(), block.clone()).await;
-                let mut output_guard = output.lock();
-                output_guard.insert(key, value);
-                if output_guard.len() % 500 == 0 {
-                    info!("Fetched {} values", output_guard.len());
-                }
-            } else {
-                break;
+            let key = match maybe_key {
+                Some(key) => key,
+                None => break,
+            };
+            let value = fetcher.get_value(key.clone(), block.clone()).await;
+            let mut output_guard = output.lock();
+            output_guard.insert(key, value);
+            if output_guard.len() % LOG_PROGRESS_FREQUENCY == 0 {
+                info!("Fetched {} values", output_guard.len());
             }
         }
         info!("Worker {:?} finished", id);
@@ -184,9 +185,8 @@ impl StateFetcher {
             );
             if new_keys.len() < KEYS_BATCH_SIZE as usize {
                 break;
-            } else {
-                last_key_fetched = Some(new_keys.last().unwrap().clone());
             }
+            last_key_fetched = Some(new_keys.last().unwrap().clone());
         }
         all_keys
     }
@@ -223,10 +223,23 @@ impl StateFetcher {
         std::mem::take(&mut guard)
     }
 
-    pub async fn get_full_state(&self, num_workers: u32) -> Storage {
-        let best_block = self.get_most_recent_block().await;
-        let keys = self.get_all_keys(best_block.clone()).await;
-        self.get_values(keys, best_block, num_workers).await
+    pub async fn get_full_state_at(
+        &self,
+        num_workers: u32,
+        fetch_block: Option<BlockHash>,
+    ) -> Storage {
+        let block = if let Some(block) = fetch_block {
+            block
+        } else {
+            self.get_most_recent_block().await
+        };
+        info!("Fetching state at block {:?}", block);
+        let keys = self.get_all_keys(block.clone()).await;
+        self.get_values(keys, block, num_workers).await
+    }
+
+    pub async fn get_full_state_at_best_block(&self, num_workers: u32) -> Storage {
+        self.get_full_state_at(num_workers, None).await
     }
 }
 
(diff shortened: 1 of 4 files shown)


fork-off/src/main.rs
```diff
@@ -12,168 +19,336 @@ pub struct Config {
...
     #[clap(
         long,
         multiple_occurrences = true,
         takes_value = true,
         value_delimiter = ',',
-        default_value = "Aura,Aleph"
+        default_value = "Aura,Aleph,Sudo,Staking,Session,Elections"
     )]
-    pub prefixes: Vec<String>,
+    pub pallets: Vec<String>,
```
fbielejec, line 51: Maybe a rename of the argument is in order:
`--keep-state-for-pallets`,  `--idempotent-pallets` (to get fancy) or something to that effect?

fork-off/src/main.rs
```diff
@@ -12,168 +19,336 @@ pub struct Config {
...
+    )?;
+
+    assert_ne!(
+        initial_spec["genesis"]["raw"],
+        Value::Null,
+        "The initial provided chainspec must be raw! Make sure you use --raw when generating it."
+    );
+    if !use_snapshot_file {
+        let fetcher = StateFetcher::new(http_rpc_endpoint);
+        let block = fetcher.get_most_recent_block().await;
```
fbielejec, line 333: you fetch that hash bu tthan it's not used anywhere (past logging it?)

fork-off/src/main.rs
```diff
@@ -12,168 +19,336 @@ pub struct Config {
...
-    if !result.is_empty() {
-        return Some(
-            result
-                .first()
-                .expect("No key in result")
-                .as_str()
-                .expect("Not a string")
-                .to_owned(),
-        );
+    pub async fn get_full_state(&self, num_workers: u32) -> Storage {
```
fbielejec, line 226: a) this could be called `get_full_state_at` and take a block hash argument or take an Option<Hash> and default to the  most recent finalized one?

fork-off/src/main.rs
```diff
@@ -12,168 +19,336 @@ pub struct Config {
...
-                debug!(target: "fork", "hashed prefix: {}, next key: {:?}", &prefix, &first_key);
+fn combine_states(mut state: Storage, initial_state: Storage, pallets: Vec<String>) -> Storage {
+    let pallets_prefixes: Vec<(String, String)> = pallets
+        .iter()
+        .map(|pallet| {
+            let hash = format!("0x{}", prefix_as_hex(pallet));
+            (pallet.clone(), hash)
+        })
+        .collect();
+    let mut cnt_removed_per_pallet: HashMap<String, usize> = pallets_prefixes
```
fbielejec, line 263: can we just call it with full name `removed_per_pallet_count` / `added_per_pallet_count` ?

fbielejec (commented):
just a few nitpicks, this will be eons faster now, nice!

fork-off/src/main.rs
```diff
@@ -12,168 +19,336 @@ pub struct Config {
...
+        .await
+        .expect("Could not deserialize response as JSON");
+    let result = response["result"].take();
+    serde_json::from_value(result).expect("Incompatible type of the result")
+}
 
-    let mut fork_spec: Value = serde_json::from_str(
-        &fs::read_to_string(&fork_spec_path).expect("Could not read chainspec file"),
-    )?;
+fn construct_json_body(method_name: &str, params: Value) -> Value {
```
kostekIV, line 79: this function is already in the aleph-client/rpc.rs file but is not yet made public. Let's use it here

fork-off/src/main.rs
```diff
@@ -12,168 +19,336 @@ pub struct Config {
...
+            http_rpc_endpoint,
+        }
+    }
 
-    info!("Done!");
-    Ok(())
-}
+    async fn value_fetching_worker(
+        http_rpc_endpoint: String,
+        id: usize,
```
kostekIV, line 123: consider using some Id(usize) type to denote the id (and feel free to resolve this comment)

fork-off/src/main.rs
```diff
@@ -12,168 +19,336 @@ pub struct Config {
...
+        block: BlockHash,
+        input: Arc<Mutex<Vec<StorageKey>>>,
+        output: Arc<Mutex<Storage>>,
+    ) {
+        let fetcher = StateFetcher::new(http_rpc_endpoint);
+        loop {
+            let maybe_key: Option<StorageKey> = {
```
kostekIV, line 130: nit: type annotation is not needed here

fork-off/src/main.rs
```diff
@@ -12,168 +19,336 @@ pub struct Config {
...
+                let mut keys = input.lock();
+                keys.pop()
+            };
+            if let Some(key) = maybe_key {
+                let value = fetcher.get_value(key.clone(), block.clone()).await;
```
kostekIV, line 135: nit: to reduce nesting:
```suggestion
            let key = match maybe_key {
                Some(key) => key,
                None => break;
            };
```

fork-off/src/main.rs
```diff
@@ -12,168 +19,336 @@ pub struct Config {
...
+                let mut output_guard = output.lock();
+                output_guard.insert(key, value);
+                if output_guard.len() % 500 == 0 {
```
kostekIV, line 138: consider moving 500 into function local constant

fork-off/src/main.rs
```diff
@@ -12,168 +19,336 @@ pub struct Config {
...
+            info!(
+                "Got {} new keys and have {} in total",
+                new_keys.len(),
+                all_keys.len()
+            );
+            if new_keys.len() < KEYS_BATCH_SIZE as usize {
+                break;
+            } else {
+                last_key_fetched = Some(new_keys.last().unwrap().clone());
+            }
```
kostekIV, line 189: doesn't need to be in the else block

fork-off/src/main.rs
DamianStraszak: Renamed by `pallets_keep_state`

fork-off/src/main.rs
DamianStraszak: Oh, this is only some leftover from debugging :/ Removed it and added logging for best block at a different place.

fork-off/src/main.rs
DamianStraszak: Sure, added `get_full_state_at` which takes an option and `get_full_state_at_best_block` that takes no argument.

fork-off/src/main.rs
DamianStraszak: For you, done :)

fork-off/src/main.rs
DamianStraszak: Removed

fork-off/src/main.rs
DamianStraszak: Reduced :)

fork-off/src/main.rs
DamianStraszak: Added `LOG_PROGRESS_FREQUENCY`

fork-off/src/main.rs
DamianStraszak: Removed

fork-off/src/main.rs
DamianStraszak: Ok, I see, but this is controversial as long as we are still planning to break the repo into smaller pieces. 

Also by depending upon `aleph_client` I automatically start depending on lots of substrate crates, which I wanted to avoid.

fbielejec (approved):

fork-off/src/main.rs
kostekIV: Ok, Maybe as an improvement we can introduction features to aleph-client to not depend on substrate when we dont need it (not in this pr tho).

kostekIV (approved):

PR #355, opened 2022-03-25
Print debug storage information if pipeline fails

kostekIV:
This adds printing some storage info after e2e tests fails (As can be seen here: https://github.com/Cardinal-Cryptography/aleph-node/runs/5716993692?check_suite_focus=true)


aleph-client/Cargo.toml
```diff
@@ -14,6 +14,10 @@ features = ["staking-xt"]
 sp-core = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13", features = ["full_crypto"] }
 sp-runtime = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13", default-features = false }
 pallet-staking = {git = "https://github.com/paritytech/substrate.git",branch = "polkadot-v0.9.13", default-features = false }
+pallet-treasury = {git = "https://github.com/paritytech/substrate.git",branch = "polkadot-v0.9.13", default-features = false }
+pallet-aleph = {path = "../pallets/aleph", default-features = false}
+primitives = { path = "../primitives", default-features = false }
+aleph-runtime = { path = "../bin/runtime"}
```
Marcin-Radecki, line 20: I remember @fbielejec  raising a point we should be not doing that, as it increases the compilation time of `e2e` crate. Personally, I was okay with that, technically e2e are those which tests the runtime, but maybe @fbielejec wants to make similar point here? The code of `e2e` tests can be simplified (e.g. we don't need to create our own `Header` or `Block` types, we can just re-use types runtime creates etc)

Marcin-Radecki: Ah now seen it is `aleph-client` crate, not `e2e`. I think `aleph-client` should not depend on the runtime as it should be a lightweight library wrapping over the `substrate-api-client` ? Maybe we can move this dep to `e2e` create?
CC: @timorl

kostekIV: I am using it here only because of one type definition I think. Maybe it would be a good idea to extract such things from the runtime into primitives? Then depending on primitive is not such a burden as depending on the runtime, wdyt?

Marcin-Radecki: What type is it? Ideally, we'd like to keep `aleph-client` lightweight, (in terms of compile-time for example). Seem we have two options
* we can move `aleph-runtime` dep to `e2e` crate as it is where it is used,
* we can add this one type which is required to `aleph-client`
Is any of them okay for u?

e2e-tests/src/config.rs
```diff
@@ -20,4 +20,8 @@ pub struct Config {
     /// seed value of sudo account
     #[clap(long)]
     pub sudo: Option<String>,
+
+    /// whether to print debug storage info
+    #[clap(long)]
+    pub storage_debug: bool,
```
pmikolajczyk41, line 26: this is really confusing, because you can either run tests xor debug storage; so it should be rather a subcommand than just a flag which completely changes behavior and invalidates more of the other flags

however, what is even more confusing for me is that binary for running tests has an additional role of debugging storage, which definitely should be a part of our `cliain` tool; this would require including `cliain` for the e2e pipeline, but we would gain better organization imho

also, I would like to call out to @Marcin-Radecki for his opinion

.github/scripts/debug_e2e_storage.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -e
+
+docker run -v $(pwd)/docker/data:/data --network container:damian -e NODE_URL=127.0.0.1:9943 -e STORAGE_DEBUG=yes aleph-e2e-client:latest
```
pmikolajczyk41, line 5: imho it would be much better to set environment variables in the pipeline itself then create different scripts for different sets

pmikolajczyk41 (changes requested):

pmikolajczyk41:
And please, make the PR title less enigmatic :)

.github/scripts/debug_e2e_storage.sh
Marcin-Radecki: Or use the same script but pass env as command line param?

pmikolajczyk41: whatever

aleph-client/src/debug/mod.rs
```diff
@@ -0,0 +1,23 @@
+use crate::Connection;
+
+mod aleph;
+mod elections;
+mod treasury;
+
+fn pallet_prompt(name: &'static str) -> String {
+    format!("-----------{}-----------", name)
```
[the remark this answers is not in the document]
Marcin-Radecki: On the other hand keeping `format!` here might cause this code evolving more into builder battern, in which we can have different formats. Test results artifacts can be written in some format like XML or JSON, and benefit of that is better parsing (imagine we'll add more debug data, e.g. all staking pallet which has 10+ `StorageMaps` etc) 

Now after making this comment, I think we should lean towards that in the future, ie return all this data as JSON. `jq` is super friendly :) not in this PR ofc

aleph-client/src/debug/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+fn entry_prompt(name: &'static str) -> String {
+    format!("----{}", name)
+}
+
+fn element_prompt(out: String) -> String {
+    format!("\t{}", out)
+}
+
+pub fn print_storages(connection: &Connection) {
+    treasury::print_storage(&connection);
```
Marcin-Radecki, line 20: We don't need `&` here as `connection` is already a ref?

e2e-tests/src/config.rs
Marcin-Radecki: Yes, I think using `cliain` here would be a better fit. All the calls from this PR can be embedded into `cliain`, but it's more work for sure. For now, we can add a `Subcommand` `mode` for example which have two values `storage_debug` and `run_tests`, with the latter as default. This is ofc just a suggestion :)

Marcin-Radecki (changes requested):
Great feature, minor comments

e2e-tests/src/main.rs
```diff
@@ -1,19 +1,30 @@
...
 use clap::Parser;
-use log::info;
 
+use aleph_client::{create_connection, print_storages};
 use aleph_e2e_client::{
     test_batch_transactions, test_change_validators, test_channeling_fee, test_fee_calculation,
     test_finalization, test_staking_era_payouts, test_staking_new_validator, test_token_transfer,
     test_treasury_access, Config,
 };
+use log::info;
```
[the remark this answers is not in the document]
kostekIV: done by `cargo fmt`

aleph-client/src/debug/elections.rs
```diff
@@ -0,0 +1,22 @@
+use crate::{
+    debug::{element_prompt, entry_prompt, pallet_prompt},
```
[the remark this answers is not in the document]
kostekIV: I prefer crate

aleph-client/src/debug/mod.rs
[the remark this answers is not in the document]
kostekIV: Iam keeping format for now

e2e-tests/src/config.rs
kostekIV: added this to cliain

.github/scripts/debug_e2e_storage.sh
kostekIV: not here anymore

aleph-client/src/debug/aleph.rs
```diff
@@ -0,0 +1,22 @@
+use crate::{
+    debug::{element_prompt, entry_prompt, pallet_prompt},
+    Connection,
+};
+use primitives::AuthorityId;
+
+pub fn print_storage(connection: &Connection) {
+    let authorities: Vec<AuthorityId> = connection
+        .get_storage_value("Aleph", "Authorities", None)
+        .unwrap()
```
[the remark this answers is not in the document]
kostekIV: done

aleph-client/src/debug/mod.rs
kostekIV: you are right, done

aleph-client/Cargo.toml
```diff
@@ -14,6 +14,9 @@ features = ["staking-xt"]
 sp-core = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13", features = ["full_crypto"] }
 sp-runtime = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13", default-features = false }
 pallet-staking = {git = "https://github.com/paritytech/substrate.git",branch = "polkadot-v0.9.13", default-features = false }
+pallet-treasury = {git = "https://github.com/paritytech/substrate.git",branch = "polkadot-v0.9.13", default-features = false }
+pallet-aleph = {path = "../pallets/aleph", default-features = false}
```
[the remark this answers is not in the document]
kostekIV: fixed

aleph-client/src/debug/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+
+fn pallet_prompt(name: &'static str) -> String {
+    format!("-----------{}-----------", name)
+}
+
+fn entry_prompt(name: &'static str) -> String {
+    format!("----{}", name)
+}
+
+fn element_prompt(out: String) -> String {
```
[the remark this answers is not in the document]
kostekIV: -> `el`

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -158,6 +158,12 @@ jobs:
         timeout-minutes: 15
         run: ./.github/scripts/run_e2e_tests.sh
 
+      - name: Print debug if failed
+        if: ${{ failure() }}
+        run: |
+          cd bin/cliain
+          cargo run -- --seed //Alice debug-storage
```
Marcin-Radecki, line 165: In general, I'm not a fan of putting any seeds anywhere, but here is justified as it is a known developer account for a  Substrate chain run in dev mode :)

pmikolajczyk41, line 165: just a nit: I prefer subcommand before flags, but feel free to ignore; resolve at your will

aleph-client/Cargo.toml
```diff
@@ -8,22 +8,29 @@ edition = "2021"
 # Substrate dependencies
 sp-core = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13", features = ["full_crypto"] }
 sp-runtime = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13", default-features = false }
-pallet-staking = { git = "https://github.com/paritytech/substrate.git",branch = "polkadot-v0.9.13", default-features = false }
-pallet-balances = { git = "https://github.com/paritytech/substrate.git",branch = "polkadot-v0.9.13", default-features = false }
+pallet-staking = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13", default-features = false }
+pallet-treasury = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13", default-features = false }
+pallet-aleph = { path = "../pallets/aleph", default-features = false }
+primitives = { path = "../primitives", default-features = false }
+pallet-balances = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13", default-features = false }
```
pmikolajczyk41, line 15: nit: reorder deps

Marcin-Radecki (approved):
Solid rework

bin/cliain/src/main.rs
```diff
@@ -102,6 +102,9 @@ enum Command {
         #[clap(long)]
         commission_percentage: u8,
     },
+
+    /// print debug info of storage
```
pmikolajczyk41, line 106: king of the nits: start with capital letter, as other docstrings above

pmikolajczyk41 (approved):
I really like it now :+1: 

I left few nits - ignore and resolve at your will

.github/workflows/e2e-tests-main-devnet.yml
kostekIV: wont work in that order

PR #357, opened 2022-03-29
Remove Python linter

pmikolajczyk41:
After 3 months it seems that linting an auxiliary tool guarantees only frustration. No real gain.

If some bigger project in Python is coming up, we can think of linting it, but for a few scripts it does not make sense.


h4nsu (approved):
With fire kill it!

fbielejec (approved):

kostekIV (approved):

PR #358, opened 2022-03-30
Make protocol implicit

pmikolajczyk41:
After some suggestions, we make passing protocol (either `ws` or `wss`) implicit. This means that if you want to connect without ssl, you can either pass a bare address like `127.0.0.1:9943` or `ws://127.0.0.1:9943`. In order to use ssl, pass `wss://127.0.0.1:9943`.

### DIFF
--- a/aleph-client/src/lib.rs
+++ b/aleph-client/src/lib.rs
@@ -55,12 +55,34 @@ pub fn create_connection(address: &str) -> Connection {
     create_custom_connection(address).expect("Connection should be created")
 }
 
+enum Protocol {
+    WS,
+    WSS,
+}
+
+impl Default for Protocol {
+    fn default() -> Self {
+        Protocol::WS
+    }
+}
+
+impl ToString for Protocol {
+    fn to_string(&self) -> String {
+        match self {
+            Protocol::WS => String::from("ws://"),
+            Protocol::WSS => String::from("wss://"),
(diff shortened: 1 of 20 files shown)


aleph-client/src/lib.rs
```diff
@@ -51,39 +51,23 @@ pub type Header = GenericHeader<BlockNumber, BlakeTwo256>;
...
-pub fn from(use_ssl: bool) -> Protocol {
-    match use_ssl {
-        true => Protocol::WSS,
-        false => Protocol::WS,
-    }
-}
-
-pub fn create_connection(address: &str, protocol: Protocol) -> Connection {
-    create_custom_connection(address, protocol).expect("connection should be created")
+    format!("ws://{}", address)
```
fbielejec, line 63: what about `impl Display for Protocol` and `Protocol::WS.to_string();`

pmikolajczyk41: You mean reviving `Protocol` enum and use its string representation instead of in-place strings?

fbielejec: yes

pmikolajczyk41: done

Marcin-Radecki (approved):
Very nice :)

fbielejec (approved):

PR #360, opened 2022-03-30
AZ-733  Support for getStorageAt in aleph-client

Marcin-Radecki:
Code changes:
* Added support for `queryStorageAt` to aleph-client. 
* Extended payout_stakers bench so all validators are nominated - this gives `N*MaxNominatorsPerValidator` total nominators in `payout_stakers` bench, where N is a number of validators 
* Checking locked balance in one call - this is required for AWS tests as locally a few thousand` getStorage` calls are fine performance-wise

To be done for this task to completed: AWS tests


aleph-client/src/account.rs
```diff
@@ -13,30 +14,34 @@ pub fn get_free_balance(connection: &Connection, account: &AccountId) -> Balance
...
+        Ok(storage_entries) => storage_entries
+            .into_iter()
+            .map(|storage_entry| {
+                let entry_bytes = Vec::from_hex(storage_entry.expect("Storage entry is null!"))
+                    .expect("Cannot parse hex string!");
+                let balance_lock: Vec<pallet_balances::BalanceLock<Balance>> =
+                    Decode::decode(&mut entry_bytes.as_slice())
+                        .expect("Failed to decode locked balances!");
+                balance_lock
+            })
```
kostekIV, line 40: nit: consider extracting this to local function

aleph-client/src/account.rs
```diff
@@ -13,30 +14,34 @@ pub fn get_free_balance(connection: &Connection, account: &AccountId) -> Balance
...
+            .collect::<Vec<Vec<_>>>(),
```
kostekIV, line 41: is the type annotation needed here?

aleph-client/src/account.rs
```diff
@@ -13,30 +14,34 @@ pub fn get_free_balance(connection: &Connection, account: &AccountId) -> Balance
...
+            connection
+                .metadata
+                .storage_map_key("Balances", "Locks", account)
+                .expect(&format!(
+                    "Cannot create storage key for account {}!",
+                    account
+                ))
+        })
+        .collect::<Vec<_>>();
+    let storage_entries = match state_query_storage_at(&connection, storage_keys) {
```
kostekIV, line 30: nit: we can return immediately here without bounding the variable

aleph-client/src/account.rs
```diff
@@ -13,30 +14,34 @@ pub fn get_free_balance(connection: &Connection, account: &AccountId) -> Balance
...
+        .map(|account| {
+            connection
+                .metadata
+                .storage_map_key("Balances", "Locks", account)
+                .expect(&format!(
+                    "Cannot create storage key for account {}!",
+                    account
+                ))
+        })
+        .collect::<Vec<_>>();
```
kostekIV, line 29: is the type annotation needed here? We are passing the storage_keys into a function that expects the input to be of type vec

aleph-client/src/rpc.rs
```diff
@@ -10,18 +11,95 @@ fn json_req(method: &str, params: Value, id: u32) -> Value {
...
+            ),
+            Value::Null,
+        ]),
+        1,
+    )
+}
+
+fn parse_query_storage_at_result(
+    maybe_json_result: Option<String>,
+) -> Result<Vec<Option<String>>, String> {
```
kostekIV, line 36: from what I saw we expect this Vec to be full of `Somes` and if it has None it means something went wrong. Lets just return here Result<Vec<String>, String> and treat the None as an error already here

benches/payout-stakers/src/main.rs
```diff
@@ -33,53 +32,68 @@ fn main() -> Result<(), anyhow::Error> {
...
 
     let connection = create_connection(address).set_signer(sudoer);
-
-    let nominator_accounts = generate_nominator_accounts_with_minimal_bond(&connection);
     let validators = set_validators(address);
-    let nominee = &validators[0];
-    nominate_validator_0(&connection, nominator_accounts.clone(), nominee);
-    wait_for_10_eras(address, &connection, nominee, nominator_accounts)?;
+    let mut validators_and_its_nominators = Vec::new();
+    validators
```
kostekIV, line 37: nit: validators_and_its_nominators can be created from validators as a transformation rather than a mutable push. 

not nit: Extract this to function for example called `test_validators_and_its_nominators()` this might be useful in the future.

benches/payout-stakers/src/main.rs
```diff
@@ -33,53 +32,68 @@ fn main() -> Result<(), anyhow::Error> {
...
+            nominate_validator(&connection, nominator_accounts.clone(), nominee_account);
+            validators_and_its_nominators.push((validator_pair.clone(), nominator_accounts))
+        });
+    wait_for_10_eras(address, &connection, validators_and_its_nominators)?;
 
     Ok(())
 }
 
 pub fn derive_user_account(seed: u64) -> KeyPair {
+    debug!("Generating account from numeric seed {}", seed);
```
kostekIV, line 54: consider using trace, I think this is called a lot of time

kostekIV (changes requested):

aleph-client/src/account.rs
Marcin-Radecki: Done

aleph-client/src/account.rs
Marcin-Radecki: Done

aleph-client/src/account.rs
Marcin-Radecki: Right, removed

aleph-client/src/rpc.rs
Marcin-Radecki: You're right but this is really up to the user of `parse_query_storage_at`. I mean that `null` returned for a given storage key from `parse_query_storage_at` it's not a fatal error. So far we have one user of `parse_query_storage_at` which is `locks`, and that function treats `None` as error. One can add at some point different function that depends on generic `parse_query_storage_at` and I did not want to close that door.

kostekIV (approved):

Marcin-Radecki:
> Please bump `aleph-client` version

Good point, done.

aleph-client/src/rpc.rs
```diff
@@ -10,18 +11,95 @@ fn json_req(method: &str, params: Value, id: u32) -> Value {
...
+                    }
+                };
+                if storage_key_and_entry.len() != 2 {
+                    return Err(String::from(&format!(
+                        "Expected {:?} to be length of 2!",
+                        storage_key_and_entry
+                    )));
+                }
+
+                let maybe_entry_in_hex_string = match storage_key_and_entry[1].as_str() {
```
piotrMocz, line 71: Nit: a pattern like that is nicely expressed by using `Option::map` (i.e. when you want to transform the value inside the `Some` and leave `None` as is.

aleph-client/src/rpc.rs
```diff
@@ -10,18 +11,95 @@ fn json_req(method: &str, params: Value, id: u32) -> Value {
...
+            Value::Null,
+        ]),
+        1,
+    )
+}
+
+fn parse_query_storage_at_result(
+    maybe_json_result: Option<String>,
+) -> Result<Vec<Option<String>>, String> {
+    match maybe_json_result {
```
piotrMocz, line 37: To be honest, I'm not a fan of this function's signature. Building on what @kostekIV mentioned, I'm going to add that it's a bit suspicious that it takes a single optional value and throws an error if it's `None`. This seems like something better handled on use-site, especially given the argument above about deferring the missing-value case to the caller.

aleph-client/src/rpc.rs
```diff
@@ -10,18 +11,95 @@ fn json_req(method: &str, params: Value, id: u32) -> Value {
...
+        None => Err(String::from(format!("Returned result was null!",))),
```
piotrMocz, line 38: I believe this `format!` is not needed?

piotrMocz (approved):
Very nice PR, just some minor stylistic comments which you can freely choose to ignore :)

pmikolajczyk41:
Sorry for uninvited comments, but I have two kind requests:
 - please bump the version again after rebasing (I've just merged other changes to `aleph-client` API
 - please expand a little the PR title, so one does not have to check Jira or PR description after merging

aleph-client/src/rpc.rs
Marcin-Radecki: Great comments, thanks for that :) simplified code

aleph-client/src/rpc.rs
Marcin-Radecki: Right, but those are all private functions - nobody from this module will be calling them, so I'd say it's okay. I blindly extracted function as my IDE suggested :) I can rework it though if you'd like to

aleph-client/src/rpc.rs
```diff
@@ -10,18 +11,94 @@ fn json_req(method: &str, params: Value, id: u32) -> Value {
...
+) -> Result<Vec<Option<String>>, String> {
+    match maybe_json_result {
+        None => Err(String::from("Returned result was null!")),
+        Some(result) => {
+            let mut storage_entries_in_hex_string = vec![];
+
+            let parsed_json_value: Value = serde_json::from_str(&result[..]).map_err(|_| {
+                String::from(&format!("Failed to parse result {:?} into JSON", result))
+            })?;
+            let storage_entries = match parsed_json_value[0]["changes"].as_array() {
```
DamianStraszak, line 45: As mentioned in slack:
1) I would be careful with just taking `[0]` here without fully understanding what it means. The fact that this call outputs a vector has some meaning.
2) The proper way to parse this should be something like:
`let parsed: Vec<StorageChangeSet<Hash>> = serde_json::from_str(&result[..]).unwrap();` -- where `StorageChangeSet<Hash>` you can probably even copy from `substrate/client/rpc-api/src/state/mod.rs` or import from substrate

Marcin-Radecki: Great comment, done.

aleph-client/src/rpc.rs
Marcin-Radecki: Yes, removed.

Marcin-Radecki:
> Sorry for uninvited comments, but I have two kind requests:
> 
>     * please bump the version again after rebasing (I've just merged other changes to `aleph-client` API
> 
>     * please expand a little the PR title, so one does not have to check Jira or PR description after merging

Done

aleph-client/src/rpc.rs
piotrMocz: No worries, it’s okay, just something that I found a bit strange-looking 🙂

aleph-client/src/account.rs
```diff
@@ -13,30 +15,46 @@ pub fn get_free_balance(connection: &Connection, account: &AccountId) -> Balance
...
+) -> Vec<Vec<BalanceLock<Balance>>> {
+    match state_query_storage_at(&connection, storage_keys) {
+        Ok(storage_entries) => storage_entries
+            .into_iter()
+            .map(|storage_entry| {
+                let entry_bytes = storage_entry.expect("Storage entry is null!").0;
+                let balance_lock: Vec<pallet_balances::BalanceLock<Balance>> =
+                    Decode::decode(&mut entry_bytes.as_slice())
+                        .expect("Failed to decode locked balances!");
+                balance_lock
```
kostekIV, line 53: nit: should work:
```suggestion
                Decode::decode(&mut entry_bytes.as_slice())
                    .expect("Failed to decode locked balances!")
```

kostekIV (approved):
Looks great :)

aleph-client/src/account.rs
Marcin-Radecki: Good point, done.

PR #364, opened 2022-03-31
Make sending transaction recoverable

pmikolajczyk41:
In some tools where we use `aleph-client` it is extremely uncomfortable to panic when some error happens during sending an extrinsic. Thus we add `try_send_xt` method which returns `substrate_api_client::ApiResult`.

### DIFF
--- a/aleph-client/src/lib.rs
+++ b/aleph-client/src/lib.rs
@@ -1,10 +1,12 @@
+use codec::Encode;
 use std::{thread::sleep, time::Duration};
 
 use log::{info, warn};
-use sp_core::{sr25519, Pair};
+use sp_core::{sr25519, Pair, H256};
 use sp_runtime::{generic::Header as GenericHeader, traits::BlakeTwo256};
 use substrate_api_client::{
-    rpc::ws_client::WsRpcClient, std::error::Error, Api, ApiResult, RpcClient, XtStatus,
+    rpc::ws_client::WsRpcClient, std::error::Error, Api, ApiResult, RpcClient,
+    UncheckedExtrinsicV4, XtStatus,
 };
 
 pub use account::{get_free_balance, get_locked_balance, locks};
@@ -107,64 +109,38 @@ pub fn create_custom_connection<Client: FromStr + RpcClient>(
     }
 }
 
-/// Sends transaction `xt` using `connection`. If `tx_status` is either `Finalized` or `InBlock`
-/// additionally fetches corresponding block number, which is then returned.
-/// Can panic.
-pub fn send_xt(
+/// `panic`able utility wrapper for `try_send_xt`.
+pub fn send_xt<T: Encode>(
     connection: &Connection,
-    xt: String,
-    xt_name: &'static str,
-    tx_status: XtStatus,
-) -> Option<BlockNumber> {
-    let hash = connection
-        .send_extrinsic(xt, tx_status)
-        .expect("Could not send extrinsic")
-        .expect("Could not get tx hash");
-
-    match tx_status {
-        XtStatus::Finalized | XtStatus::InBlock => {
-            let block_number = connection
-                .get_header::<Header>(Some(hash))
-                .expect("Could not fetch header")
-                .expect("Block exists; qed")
-                .number;
-            info!(
-                target: "aleph-client",
-                "Transaction {} was included in block {}.",
-                xt_name, block_number
-            );
-            Some(block_number)
-        }
-        _ => None,
-    }
+    xt: UncheckedExtrinsicV4<T>,
+    xt_name: Option<&'static str>,
+    xt_status: XtStatus,
+) -> Option<H256> {
+    try_send_xt(connection, xt, xt_name, xt_status).expect("Should manage to send extrinsic")
 }
 
 /// Sends transaction `xt` using `connection`. If `tx_status` is either `Finalized` or `InBlock`
-/// fetches corresponding block number, which is then returned.
+/// additionally returns hash of the containing block. `xt_name` is used only for logging purposes.
 /// Recoverable.
-pub fn try_send_xt(
+pub fn try_send_xt<T: Encode>(
     connection: &Connection,
-    xt: String,
-    xt_name: &'static str,
-    tx_status: XtStatus,
-) -> ApiResult<Option<BlockNumber>> {
+    xt: UncheckedExtrinsicV4<T>,
+    xt_name: Option<&'static str>,
+    xt_status: XtStatus,
+) -> ApiResult<Option<H256>> {
     let hash = connection
-        .send_extrinsic(xt, tx_status)?
-        .ok_or_else(|| Error::Other(String::from("Could not get block hash").into()))?;
+        .send_extrinsic(xt.hex_encode(), xt_status)?
+        .ok_or_else(|| Error::Other(String::from("Could not get tx/block hash").into()))?;
 
-    match tx_status {
+    match xt_status {
         XtStatus::Finalized | XtStatus::InBlock => {
-            let block_number = connection
-                .get_header::<Header>(Some(hash))?
-                .expect("Block exists; qed")
-                .number;
-            info!(
-                target: "aleph-client",
-                "Transaction {} was included in block {}.",
-                xt_name, block_number
-            );
-            Ok(Some(block_number))
+            info!(target: "aleph-client",
+                "Transaction `{}` was included in block with hash {}.",
+                xt_name.unwrap_or_default(), hash);
+            Ok(Some(hash))
         }
+        // Other variants either do not return (see https://github.com/scs/substrate-api-client/issues/175)
+        // or return xt hash, which is kinda useless here.
         _ => Ok(None),
     }
 }
--- a/aleph-client/src/lib.rs
+++ b/aleph-client/src/lib.rs
@@ -107,42 +107,66 @@ pub fn create_custom_connection<Client: FromStr + RpcClient>(
     }
 }
 
-pub fn send_xt(connection: &Connection, xt: String, xt_name: &'static str, tx_status: XtStatus) {
-    let block_hash = connection
+/// Sends transaction `xt` using `connection`. If `tx_status` is either `Finalized` or `InBlock`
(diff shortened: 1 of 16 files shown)


aleph-client/src/lib.rs
```diff
@@ -122,6 +124,27 @@ pub fn send_xt(connection: &Connection, xt: String, xt_name: &'static str, tx_st
...
+    xt_name: &'static str,
+    tx_status: XtStatus,
+) -> ApiResult<()> {
+    let block_hash = connection
+        .send_extrinsic(xt, tx_status)?
+        .ok_or_else(|| Error::Other(String::from("Could not get block hash").into()))?;
+    let block_number = connection
+        .get_header::<Header>(Some(block_hash))?
+        .expect("Block exists; qed")
+        .number;
```
deuszx, line 139: I think this isn't necessary. It's making one more RPC call to the node and not all clients of this API might want that. If they do though, they can do it themselves.

aleph-client/src/lib.rs
```diff
@@ -122,6 +124,27 @@ pub fn send_xt(connection: &Connection, xt: String, xt_name: &'static str, tx_st
     );
 }
 
+pub fn try_send_xt(
```
deuszx, line 127: It's a public API so it's always a good habit to add comments.

aleph-client/src/lib.rs
```diff
@@ -122,6 +124,27 @@ pub fn send_xt(connection: &Connection, xt: String, xt_name: &'static str, tx_st
...
+    connection: &Connection,
+    xt: String,
+    xt_name: &'static str,
+    tx_status: XtStatus,
+) -> ApiResult<()> {
+    let block_hash = connection
+        .send_extrinsic(xt, tx_status)?
```
deuszx, line 134: This blocks until the transaction's status becomes `tx_status`, correct? Is there any timeout on that?

pmikolajczyk41: I want to add async api here soon

aleph-client/src/lib.rs
```diff
@@ -122,6 +124,27 @@ pub fn send_xt(connection: &Connection, xt: String, xt_name: &'static str, tx_st
...
+        .ok_or_else(|| Error::Other(String::from("Could not get block hash").into()))?;
+    let block_number = connection
+        .get_header::<Header>(Some(block_hash))?
+        .expect("Block exists; qed")
+        .number;
+    info!(
```
fbielejec, line 140: not neccessarily, what if you just dispatch into mempool and do not wait for receipt ?

pmikolajczyk41: yeah, the original method has some defects, but I wanted to keep it consistent :|

pmikolajczyk41: I guess I should 'fix' both methods

aleph-client/src/lib.rs
kostekIV: maybe let's return the block hash from the function instead and then the consumer of this API will decide what to do with it

aleph-client/src/lib.rs
```diff
@@ -122,6 +124,27 @@ pub fn send_xt(connection: &Connection, xt: String, xt_name: &'static str, tx_st
     );
 }
 
+pub fn try_send_xt(
+    connection: &Connection,
+    xt: String,
```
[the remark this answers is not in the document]
deuszx: You probably shouldn't change the original method as that would break other people's code. Unless that's used mostly internally in which case I think you can make a commit that fixes all of that - advantages of a monorepo ;P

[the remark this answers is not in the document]
pmikolajczyk41: actually the target method `send_extrinsic` expects `String` anyway, so I'm not sure what do we gain here

aleph-client/src/lib.rs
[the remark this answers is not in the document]
pmikolajczyk41: We loose the transparency which `expect` failed, thus I'm not so sure if that is the good way

aleph-client/src/lib.rs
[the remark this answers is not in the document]
pmikolajczyk41: What do you understand by 'type-safety' here then? That when `xt` is only `String` then it can be just rubbish?

aleph-client/src/lib.rs
fbielejec: if you send to mempool you won't get even that..

pmikolajczyk41: Indeed, now I'm returning block number, but after waiting for it, so in case of mempool we do not gain anything, but in a single-call we get the number for free if needed

aleph-client/src/lib.rs
```diff
@@ -105,21 +107,66 @@ pub fn create_custom_connection<Client: FromStr + RpcClient>(
...
+
+/// Sends transaction `xt` using `connection`. If `tx_status` is either `Finalized` or `InBlock`
+/// fetches corresponding block number, which is then returned.
+/// Recoverable.
+pub fn try_send_xt(
+    connection: &Connection,
+    xt: String,
+    xt_name: &'static str,
+    tx_status: XtStatus,
+) -> ApiResult<Option<BlockNumber>> {
```
deuszx, line 150: I would keep returning the block's hash. It's less ambiguous. In theory, there may be multiple blocks under same block height (equivocations) but block hash is unique. Usually, APIs will expect the block hash rather than height when querying for data as well.

aleph-client/src/lib.rs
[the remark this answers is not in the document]
deuszx: Yes.

aleph-client/src/lib.rs
```diff
@@ -105,21 +109,40 @@ pub fn create_custom_connection<Client: FromStr + RpcClient>(
...
-        "Transaction {} was included in block {}.",
-        xt_name, block_number
-    );
+/// `panic`able utility wrapper for `try_send_xt`.
+pub fn send_xt<T: Encode>(
+    connection: &Connection,
+    xt: UncheckedExtrinsicV4<T>,
+    xt_name: Option<&'static str>,
+    xt_status: XtStatus,
+) -> Option<H256> {
```
deuszx, line 118: If we're changing API in a non-backwards compatible way we should bump minor version of the client.

aleph-client/src/lib.rs
```diff
@@ -105,21 +109,40 @@ pub fn create_custom_connection<Client: FromStr + RpcClient>(
...
+    match xt_status {
+        XtStatus::Finalized | XtStatus::InBlock => {
+            info!(target: "aleph-client",
+                "Transaction `{}` was included in block with hash {}.",
+                xt_name.unwrap_or_default(), hash);
+            Ok(Some(hash))
+        }
+        // Other variants either do not return (see https://github.com/scs/substrate-api-client/issues/175)
+        // or return xt hash, which is kinda useless here.
+        _ => Ok(None),
```
deuszx, line 144: Why is it useless? Even if user didn't specify that he wants to wait for specific xt_status I think it's still useful to return the block hash the txn was included in. Why throw away that information?

pmikolajczyk41: Hash of the extrinsic is the thing that you already had before calling this method. Note, that only `InBlock` and `Finalize` variants provide block hash.

aleph-client/src/lib.rs
pmikolajczyk41: true, I bumped patch part before and forgot to adjust version to the newest changes; fixed

aleph-client/src/lib.rs
pmikolajczyk41: changed

aleph-client/src/lib.rs
```diff
@@ -105,21 +107,66 @@ pub fn create_custom_connection<Client: FromStr + RpcClient>(
...
+    let hash = connection
+        .send_extrinsic(xt, tx_status)?
+        .ok_or_else(|| Error::Other(String::from("Could not get block hash").into()))?;
+
+    match tx_status {
+        XtStatus::Finalized | XtStatus::InBlock => {
+            let block_number = connection
+                .get_header::<Header>(Some(hash))?
+                .expect("Block exists; qed")
+                .number;
```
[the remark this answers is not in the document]
pmikolajczyk41: given up

aleph-client/src/lib.rs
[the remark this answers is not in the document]
pmikolajczyk41: now, after changes, there would be only one such `expect`, so followed the advice

aleph-client/src/lib.rs
pmikolajczyk41: done, (narrowed to only some variants)

aleph-client/src/lib.rs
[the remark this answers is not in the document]
pmikolajczyk41: used `UncheckedExtrinsicV4<T>` instead, as discussed offline

aleph-client/src/lib.rs
pmikolajczyk41: and what is funnier, there is no documentation in `substrate-api-client` on it

and what is even funnier, in different examples in their repository there are different interpretations xD

but I checked and it's indeed block hash

deuszx: so, if it is a block hash, why not return it in all cases?

pmikolajczyk41: no no, it is a block hash only for `InBlock` and `Finalize`; for other it is extrinsic hash

aleph-client/src/lib.rs
```diff
@@ -58,29 +62,29 @@ pub fn create_connection(address: &str) -> Connection {
 }
 
 enum Protocol {
-    WS,
-    WSS,
+    Ws,
+    Wss,
```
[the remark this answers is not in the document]
pmikolajczyk41: not necessary, but making special PR for silencing clippy for these two variants seems overkill with our review/pipeline overhead

deuszx (approved):

fbielejec (approved):

PR #367, opened 2022-04-01
fix caching for unit tests workflow

fbielejec:
# Description

Fixing `Unable to reserve cache with key X, another job may be creating this cache.`
Use one, versioned cache. This will hopefully fix any potential race conditions between different jobs and makes it easy to invalidate cache.

## Type of change

- [X ] Bug fix (non-breaking change which fixes an issue)

## Before
before

## After

after

### DIFF
--- a/.github/actions/restore-cache/action.yml
+++ b/.github/actions/restore-cache/action.yml
@@ -33,7 +33,7 @@ runs:
         echo "SCCACHE_PATH=${{ inputs.sccache-path }}" >> $GITHUB_ENV
 
     - name: Restore cargo cache
-      uses: actions/cache@v2
+      uses: actions/cache@v3
       with:
         path: |
           ~/.cargo/bin/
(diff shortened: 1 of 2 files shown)


.github/actions/restore-cache/action.yml
```diff
@@ -56,11 +57,11 @@ runs:
...
-    - name: Restore sccache
-      uses: actions/cache@v2
-      with:
-        path: ${{ inputs.sccache-path }}
-        key: ${{ runner.os }}-sccache-${{ hashFiles('**/Cargo.lock') }}
+    # - name: Restore sccache
+    #   uses: actions/cache@v3
+    #   with:
+    #     path: ${{ inputs.sccache-path }}
+    #     key: ${{ runner.os }}-sccache-${{ hashFiles('**/Cargo.lock') }}
```
deuszx, line 64: Why do we keep it around commented out?

deuszx (approved):
Approving since it seems to fix the problem.

.github/actions/restore-cache/action.yml
fbielejec: cleaned

bartoszjedrzejewski (approved):

PR #368, opened 2022-04-04
[AZ-638]: Implement new update strategy to devnet

bartoszjedrzejewski:
# Description

## Type of change

# Checklist:


.github/workflows/deploy-to-devnet.yml
```diff
@@ -60,21 +50,73 @@ jobs:
...
+          COMMIT_ID=${TMP_COMMIT_ID: -7}
+          echo $COMMIT_ID
+
+          if aws s3api head-object --bucket alephzero-devnet-eu-central-1-keys-bucket --key forks/chainspec-${COMMIT_ID}.json; then
+            echo "Chainspec exists"
+            IF_CHAINSPEC_EXISTS=true
+            aws s3 cp s3://alephzero-devnet-eu-central-1-keys-bucket/forks/chainspec-${COMMIT_ID}.json chainspec.skeleton.json
+          else
+            echo "Chainspec doesn't exist"
+            IF_CHAINSPEC_EXISTS=false
```
fbielejec, line 79: why set this if not used anywhere?

.github/workflows/deploy-to-devnet.yml
```diff
@@ -60,21 +50,73 @@ jobs:
...
+
+            docker run -it --env RUST_BACKTRACE=1 --entrypoint "/usr/local/bin/aleph-node" public.ecr.aws/p6e8q1z1/aleph-node:${COMMIT_ID} bootstrap-chain --raw --base-path /data --chain-id a0dnet1 --account-ids 5D34dL5prEUaGNQtPPZ3yN5Y6BnkfXunKXXz6fo7ZJbLwRRH,5GBNeWRhZc2jXu7D55rBimKYDk8PGk8itRYFTPfC8RJLKG5o,5Dfis6XL8J2P6JHUnUtArnFWndn62SydeP8ee8sG2ky9nfm9,5F4H97f7nQovyrbiq4ZetaaviNwThSVcFobcA5aGab6167dK,5DiDShBWa1fQx6gLzpf3SFBhMinCoyvHM1BWjPNsmXS8hkrW,5EFb84yH9tpcFuiKUcsmdoF7xeeY3ajG1ZLQimxQoFt9HMKR,5DZLHESsfGrJ5YzT3HuRPXsSNb589xQ4Unubh1mYLodzKdVY,5GHJzqvG6tXnngCpG7B12qjUvbo5e4e9z8Xjidk3CQZHxTPZ,5CUnSsgAyLND3bxxnfNhgWXSe9Wn676JzLpGLgyJv858qhoX,5CVKn7HAZW1Ky4r7Vkgsr7VEW88C2sHgUNDiwHY9Ct2hjU8q --sudo-account-id 5F4SvwaUEQubiqkPF8YnRfcN77cLsT2DfG4vFeQmSXNjR7hD > chainspec.skeleton.json
```
fbielejec, line 81: maybe you could leave a comment here that these accounts are:
```
Alice Bob Charlie Dave Eve Ferdie Alice//stash Bob//stash Charlie//stash Dave//stash Eve//stash Ferdie//stash
```

.github/workflows/deploy-to-devnet.yml
```diff
@@ -60,21 +50,73 @@ jobs:
...
 
+            aws s3 cp chainspec.skeleton.json s3://alephzero-devnet-eu-central-1-keys-bucket/forks/chainspec-${COMMIT_ID}.json
+          fi
+
+          docker run -it -v $(pwd):/app public.ecr.aws/p6e8q1z1/fork-off:latest --http-rpc-endpoint=https://rpc.test.azero.dev --initial-spec-path=chainspec.skeleton.json --combined-spec-path=chainspec.json
```
fbielejec, line 86: Add:
--pallets-keep-state=Aura,Aleph,Sudo,Staking,Session,Elections

These are the defaults but it never hurts to be verbose, and chances are there will be soon more pallets joining, with their own state

fbielejec (commented):
Two minor comments

It is important to note, that with this setup the forked devnet will always use the same genesis state as the starting point (because it pulls it from S3). 

I'm wondering if there might be a situation when this is **not** a wanted behavior. What comes to mind is if there was a change enacted in one of the "pallets-keeps-state" - i.e. sudo account was changed, or a set of validators, or staking parameters etc. Than this change will not  get carried over to the fork.

.github/workflows/deploy-to-devnet.yml
bartoszjedrzejewski: If there will be more pallets joining I'd rather do it by updating fork-off, which would automatically update this workflow(it uses latest image of fork-off tool). So leaving it as default is beneficial.

.github/workflows/deploy-to-devnet.yml
bartoszjedrzejewski: leftover, thanks

bartoszjedrzejewski:
> Two minor comments
> 
> It is important to note, that with this setup the forked devnet will always use the same genesis state as the starting point (because it pulls it from S3).
> 
> I'm wondering if there might be a situation when this is **not** a wanted behavior. What comes to mind is if there was a change enacted in one of the "pallets-keeps-state" - i.e. sudo account was changed, or a set of validators, or staking parameters etc. Than this change will not get carried over to the fork.


> Two minor comments
> 
> It is important to note, that with this setup the forked devnet will always use the same genesis state as the starting point (because it pulls it from S3).
> 
> I'm wondering if there might be a situation when this is **not** a wanted behavior. What comes to mind is if there was a change enacted in one of the "pallets-keeps-state" - i.e. sudo account was changed, or a set of validators, or staking parameters etc. Than this change will not get carried over to the fork.

You are right. But if they were to change they would have to be changed in the workflow anyway, so we can wait with this change if it will be needed(maybe not :) ).

skrolikowski-10c (approved):

PR #369, opened 2022-04-05
Everybody wins

mike1729:
# Description

## Type of change

# Checklist:

### DIFF
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -322,18 +322,62 @@ impl_opaque_keys! {
 
 parameter_types! {
     pub const SessionPeriod: u32 = DEFAULT_SESSION_PERIOD;
+    pub const MembersPerSession: u32 = 4;
 }
 
 impl pallet_elections::Config for Runtime {
     type Event = Event;
     type DataProvider = Staking;
     type SessionPeriod = SessionPeriod;
+    type MembersPerSession = MembersPerSession;
 }
 
 impl pallet_randomness_collective_flip::Config for Runtime {}
 
 parameter_types! {
     pub const Offset: u32 = 0;
+}
+
+fn subsample() -> Option<Vec<AccountId>> {
+    let current_era = match Staking::active_era() {
+        Some(ae) if ae.index > 0 => ae.index,
+        _ => return None,
+    };
+    let all_validators: Vec<AccountId> =
+        pallet_staking::ErasStakers::<Runtime>::iter_key_prefix(current_era).collect();
+    let n_all_validators = all_validators.len() as u32;
+    let current_session = Session::current_index();
+    let n_members = MembersPerSession::get();
+
+    let first_validator = current_session % n_all_validators * n_members;
+    Some(
+        (first_validator..first_validator + n_members)
+            .map(|i| all_validators[(i % n_all_validators) as usize].clone())
+            .collect(),
+    )
+}
+
+use primitives::SessionIndex;
+type SM = pallet_session::historical::NoteHistoricalRoot<Runtime, Staking>;
+pub struct SubsampleSessionManager;
+
+impl pallet_session::SessionManager<AccountId> for SubsampleSessionManager {
+    fn new_session(new_index: SessionIndex) -> Option<Vec<AccountId>> {
+        SM::new_session(new_index);
+        subsample()
+    }
+
+    fn end_session(end_index: SessionIndex) {
+        SM::end_session(end_index)
+    }
+
+    fn start_session(start_index: SessionIndex) {
+        SM::start_session(start_index)
+    }
+
+    fn new_session_genesis(new_index: SessionIndex) -> Option<Vec<AccountId>> {
+        SM::new_session_genesis(new_index)
+    }
 }
 
 impl pallet_session::Config for Runtime {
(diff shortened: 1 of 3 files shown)


bin/runtime/src/lib.rs
```diff
@@ -336,13 +338,67 @@ parameter_types! {
     pub const Offset: u32 = 0;
 }
 
+// Choose a subset of all the validators for current era that contains all the
+// reserved nodes. Non reserved ones are chosen in consecutive batches for every session
+fn sample() -> Option<Vec<AccountId>> {
```
kostekIV, line 343: nit: rename the function as it is not sampling but rather choosing.

pallets/elections/src/lib.rs
```diff
@@ -37,6 +37,8 @@ pub mod pallet {
         type DataProvider: ElectionDataProvider<Self::AccountId, Self::BlockNumber>;
         #[pallet::constant]
         type SessionPeriod: Get<u32>;
+        #[pallet::constant]
+        type MembersPerSession: Get<u32>;
```
kostekIV, line 41: I see 2 problems with keeping this as a constant:
* We won't be able to change nr of members without changing runtime
* If this should reflect the number of validators it might be harder to maintain as we now have more than one source of this number

bin/runtime/src/lib.rs
mike1729: The name was chosen to be forward compatible ;) But I may change that to `rotate` for example?

kostekIV: rotate sounds good

pallets/elections/src/lib.rs
kostekIV: discussed offline

mike1729: I agree with the first point, I did it this was because it was faster. As to the second, we discussed that tet a tet.

bin/runtime/src/lib.rs
```diff
@@ -336,13 +337,67 @@ parameter_types! {
...
+    // this is tricky: we might change the number of reserved nodes for (current_era + 1) while still
+    // in current_era. This is not a problem as long as we enlarge the list.
+    // A solution could be to store reserved nodes per era, but this is somewhat cumbersome.
+    // There is a ticket to improve that in later iteration
+    let mut validators = pallet_staking::Invulnerables::<Runtime>::get();
+    all_validators.retain(|v| !validators.contains(v));
+    let n_all_validators = all_validators.len();
+
+    let n_validators = MembersPerSession::get() as usize;
+    let free_sits = n_validators.checked_sub(validators.len()).unwrap();
```
kostekIV, line 358: unwrap -> expect.
Do we want to panic here if n_validators is less than `validators.len()`? If not `saturated_sub` will set the free_sits to 0 in such case

mike1729: Generally, this code is experimental and PoC, so I'm fine with unwrap for the time being. It will have to be refactored at later point, so if it works in current version then I'd leave it as it is.

kostekIV: ok

kostekIV (approved):

fixxxedpoint (approved):
LGTM. Only nitpicks would be to add some description to declared constants and avoid to use `as usize` (not refactor friendly, bla bla).

bin/runtime/src/lib.rs
```diff
@@ -336,13 +337,82 @@ parameter_types! {
...
+    };
+    let mut all_validators: Vec<AccountId> =
+        pallet_staking::ErasStakers::<Runtime>::iter_key_prefix(current_era).collect();
+
+    let mut validators = pallet_elections::ErasReserved::<Runtime>::get(current_era);
+    all_validators.retain(|v| !validators.contains(v));
+    let n_all_validators = all_validators.len();
+
+    let n_validators = MembersPerSession::get() as usize;
+    let free_sits = n_validators.checked_sub(validators.len()).unwrap();
```
piotrMocz, line 355: Typo (I think): `s/free_sits/free_seats/g` :)

bin/node/src/chain_spec.rs
```diff
@@ -334,7 +335,7 @@ fn generate_genesis_config(
             validator_count: authorities.len() as u32,
             // to satisfy some e2e tests as this cannot be changed during runtime
             minimum_validator_count: 4,
-            invulnerables: to_account_ids(&authorities).collect(),
+            invulnerables: to_account_ids(&authorities[..2]).collect(),
```
piotrMocz, line 338: This seems like it could _theoretically_ fail at runtime

piotrMocz (approved):
LGTM, I left some minor remarks

bin/node/src/chain_spec.rs
kostekIV: this is called at the point of bootsraping the chain so this is not executed in the runtime. Also if we have less than 2 authorities it doesn't make sense to run the chain where the minimum validator count is equal to 4 among other stuff

bin/runtime/src/lib.rs
```diff
@@ -336,13 +337,82 @@ parameter_types! {
...
+
+    let current_session = Session::current_index() as usize;
+    let first_validator = current_session * free_sits;
```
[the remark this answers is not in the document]
kostekIV: added

bin/runtime/src/lib.rs

bin/node/src/chain_spec.rs
piotrMocz: Fair enough 🙂

fixxxedpoint (approved):
lgtm

PR #370, opened 2022-04-06
Rewards based on stake and uptime

mike1729:
# Description

## Type of change

# Checklist:


bin/runtime/src/elections/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+                &validator,
+                scaled_total_exposure,
+            ),
+        )
+    });
+
+    pallet_staking::Pallet::<Runtime>::reward_by_ids(session_rewards);
+}
+
+fn calculate_adjusted_session_points<F: Fn(EraIndex, &T) -> u128, T: Clone + PartialEq>(
```
fixxxedpoint, line 59: This `F: Fn...` looks unecessary here.

bin/runtime/src/elections/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+    Staking, Vec,
+};
+use pallet_elections::MembersPerSession;
+use primitives::{SessionIndex, TOKEN};
+
+fn blocks_to_produce_per_session() -> u32 {
+    SessionPeriod::get() / MembersPerSession::<Runtime>::get()
+}
+
+fn scale_total_exposure(total: u128) -> u128 {
```
fixxxedpoint, line 12: Perhaps this should be returning `u32` so there would less type juggling later.

bin/runtime/src/elections/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+
+    scale_total_exposure(total)
+}
+
+fn validator_points_per_block(
+    total_exposure_in_tokens: u128,
+    sessions_per_era: u32,
+    blocks_to_produce_per_session: u32,
+) -> u32 {
+    (Perquintill::from_rational(1, (blocks_to_produce_per_session * sessions_per_era) as u64)
```
fixxxedpoint, line 28: What exactly is happening here? Mulitplied `u32` value promoted to `u64` based `rational`, then to `u128` then truncated to `u32`?

bin/runtime/src/elections/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+        * total_exposure_in_tokens) as u32
+}
+
+fn reward_for_session_non_committee(committee: &[AccountId]) {
```
fixxxedpoint, line 32: Suggestion: it would be clearer if you would just use here a list of members not in committee instead or not use any parameter (similarly to other functions here).

bin/runtime/src/elections/mod.rs
Marcin-Radecki: Ok, will do.

bin/runtime/src/elections/mod.rs
Marcin-Radecki: Let me try to change it to plain integer division `/` and if tests pass

bin/runtime/src/elections/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+use pallet_elections::MembersPerSession;
+use primitives::{SessionIndex, TOKEN};
+
+fn blocks_to_produce_per_session() -> u32 {
+    SessionPeriod::get() / MembersPerSession::<Runtime>::get()
+}
+
+fn scale_total_exposure(total: u128) -> u128 {
+    // to cover minimal possible theoretical stake (25k) and avoid loss of accuracy we need to scale
+    100 * total / TOKEN
```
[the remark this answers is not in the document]
Marcin-Radecki: That's a good point but then we won't be able to show reward points progress in UI. Let me ask around whether getting rid of points column from UI is viable idea.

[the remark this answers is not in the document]
Marcin-Radecki: For first iteration, let's stick to what we have now.

bin/runtime/src/elections/mod.rs
```diff
@@ -0,0 +1,316 @@
+use crate::{
```
Marcin-Radecki, line 1: We'll refactor this to pallet elections.

bin/runtime/src/elections/mod.rs
Marcin-Radecki: Add comment that we round here to nearest integer

pallets/elections/src/impls.rs
```diff
@@ -0,0 +1,316 @@
...
+    fn note_author(validator: T::AccountId) {
+        let active_era = pallet_staking::ActiveEra::<T>::get().unwrap().index;
+
+        let total = Self::scaled_total_exposure(active_era, &validator);
+        let sessions_per_era = T::SessionsPerEra::get();
+
+        let validator_points_per_block =
+            validator_points_per_block(total, sessions_per_era, Self::blocks_to_produce_per_session());
+
+        pallet_staking::Pallet::<T>::reward_by_ids(
```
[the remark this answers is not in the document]
mike1729: I believe that 1. is not much of a problem really but point 3 is enough to justify new storage and I agree that performance hit should be marginal. The only downside I see is that we need to change UI so that at least it shows an info that we update points every 15mins. This however has low prio in my opinion.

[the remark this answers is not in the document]
DamianStraszak: In the new version the points will be actually hard to interpret (as they drip proportionally to stake), so it's better if people just ignore them. Ideally we should remove points from the UI, but that's probably too much work. Anyway, one column with random noise that changes every 15 minutes have never hurt anyone :P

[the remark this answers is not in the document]
Marcin-Radecki: In my opinion, we can hide the `points` column from users entirely. There's another column just right to it `last #` which shows the progress that validators are producing blocks. We can treat (and announce) `points` as an internal metric, which will directly translate into rewards, but those are the internals. One can still query them from storage if needed. With the lenient uptime feature as noted above some users might notice some incontinence at the end of an era when points are adjusted. 

image

bin/runtime/src/lib.rs
```diff
@@ -343,6 +340,7 @@ impl pallet_elections::Config for Runtime {
     type Event = Event;
     type DataProvider = Staking;
     type SessionPeriod = SessionPeriod;
+    type SessionManager = pallet_session::historical::NoteHistoricalRoot<Runtime, Staking>;
```
Marcin-Radecki, line 343: We miss that line in the original change? What's the reason for adding it?

kostekIV: this is a way of providing the `SessionManager` into elections pallet

bin/runtime/src/elections/mod.rs
kostekIV: done

bin/runtime/src/elections/mod.rs
kostekIV: done

pallets/elections/src/impls.rs
```diff
@@ -0,0 +1,316 @@
...
+        .filter(|v| !reserved.contains(v))
+        .collect();
+    let n_all_validators_without_reserved = validators_without_reserved.len();
+
+    // The validators for the committee at the session `n` are chosen as follow:
+    // 1. Reserved validators are always chosen.
+    // 2. Given non-reserved list of validators the chosen ones are from the range:
+    // `n * free_seats` to `(n + 1) * free_seats` where free_seats is equal to free number of free
+    // seats in the committee after reserved nodes are added.
+    let free_seats = n_validators.checked_sub(reserved.len()).unwrap();
```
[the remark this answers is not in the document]
kostekIV: done

pallets/elections/src/impls.rs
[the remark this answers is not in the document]
kostekIV: not related to ui: the code change is in place :)

pallets/elections/src/impls.rs
```diff
@@ -0,0 +1,343 @@
+use crate::{Config, ErasReserved, MembersPerSession, Pallet, SessionValidatorBlockCount};
+use frame_election_provider_support::sp_arithmetic::Perquintill;
+use frame_support::{pallet_prelude::Get, traits::Currency};
+use primitives::TOKEN;
+use sp_staking::{EraIndex, SessionIndex};
+use sp_std::vec::Vec;
+
+fn scale_total_exposure(total: u128) -> u128 {
+    // to cover minimal possible theoretical stake (25k) and avoid loss of accuracy we need to scale
+    100 * total / TOKEN
```
Marcin-Radecki, line 10: This is not needed now? As we don't count points per block, but per session

kostekIV: It is as you said, removed

pallets/elections/src/impls.rs
```diff
@@ -0,0 +1,316 @@
+use crate::{Config, ErasReserved, MembersPerSession, Pallet, SessionValidatorBlockCount};
+use frame_election_provider_support::sp_arithmetic::Perquintill;
+use frame_support::{pallet_prelude::Get, traits::Currency};
+use primitives::TOKEN;
+use sp_staking::{EraIndex, SessionIndex};
+use sp_std::vec::Vec;
+
+fn scale_total_exposure(total: u128) -> u128 {
+    // to cover minimal possible theoretical stake (25k) and avoid loss of accuracy we need to scale
+    total / TOKEN
```
DamianStraszak, line 10: This scaling can be improved. If for some reason one would run it with some strange token distribution on a dev chain, this has the potential to panic (at the place where we cast to u32). And in general this is not quite robust.

I think the correct way to do that would be to make the total (sum over validators) maximum possible reward per era equal to u32::MAX. And then scale according to that. This will give you the optimal precision, and will never lead to unexpected behavior...

DamianStraszak: u32::MAX can be replaced by some large round number, if we prefer that, like 10**9.

kostekIV: Done

DamianStraszak: This still looks similar as before, only some code was moved. 
What you want to avoid is this kind of random scaling `(exposure.total.into() / TOKEN) as u32` because, why would you scale by `TOKEN`? That's a pretty random constant in this context. 

Let `v1, ..., vN` be the validators for this era. Let `e1, e2, ..., eN` be their exposure (as u128) and let `e=e1+e2+...+eN`. You want the scaling `(e1', e2', ..., eN')` (as `u32`) to be such that `e' = e1'+e2'+...+eN'` \approx `MAX_REWARD` (you will probably not get equality because of rounding errors, but make sure we always round down). Now, for validator `vi` the target reward for a single session should be `ei'/sessions_per_era`, i.e.:
-  if it is not in the committee, or it created >=90% fraction of blocks it was supposed to create, then it would get `ei'/sessions_per_era` reward
- otherwise `<ei'/sessions_per_era` proportionally to how many blocks it created.

kostekIV: Ok, now it should be as you described. Also added lenient uptime in this pr

DamianStraszak: Great, this looks good now. But you don't need to consider the case `sum_totals <= MAX_REWARD` separately. Just use this formula for scaling:
`(t*MAX_REWARD)/sum_totals` 
If you are scared that `t*MAX_REWARD` might overflow `u128` then you can either:
1. ignore it, because it will not,
2. do the multiplication in U256,
3. use `saturating_mul`

I think 3. is the cleanest of them.

pallets/elections/src/impls.rs
```diff
@@ -0,0 +1,398 @@
...
+    // when produced between 90% to 100% expected blocks get 100% possible reward for session
+    if performance >= LENIENT_THRESHOLD && blocks_to_produce_per_session >= blocks_created {
+        return (Perquintill::from_rational(1, sessions_per_era as u64)
+            * total_exposure_in_tokens as u64) as u32;
+    }
+
+    (Perquintill::from_rational(
+        blocks_created as u64,
+        (blocks_to_produce_per_session * sessions_per_era) as u64,
+    ) * total_exposure_in_tokens as u64) as u32
```
DamianStraszak, line 28: There is this question of whether this should be, like that, or it should be further divided by `LENIENT_THRESHOLD` so that the reward function is continuous. I don't have a strong preference here.

pallets/elections/src/impls.rs
```diff
@@ -0,0 +1,398 @@
+use crate::{Config, ErasReserved, MembersPerSession, Pallet, SessionValidatorBlockCount};
+use frame_election_provider_support::sp_arithmetic::Perquintill;
+use frame_support::{pallet_prelude::Get, traits::Currency};
+use sp_staking::{EraIndex, SessionIndex};
+use sp_std::{collections::btree_map::BTreeMap, vec::Vec};
+
+const MAX_REWARD: u32 = u32::MAX;
```
DamianStraszak, line 7: I'm getting scared that because of possible issues with rounding (and the fact that the staking pallet uses `+` instead of `saturating_add` when handling rewards), we will at some point hit an overflow here, which, no suprise, would be quite bad. For this reason I would go for one of the safe options: 10^9, 2*10^9, or 4*10^9.

kostekIV: changed to 10^9

pallets/elections/src/impls.rs
kostekIV: done, I applied the option 3.

pallets/elections/src/impls.rs
```diff
@@ -0,0 +1,398 @@
...
+    fn adjust_rewards_for_session() {
+        let active_era = match pallet_staking::ActiveEra::<T>::get() {
+            Some(ae) if ae.index > 0 => ae.index,
+            _ => return,
+        };
+
+        let (committee, non_committee) = Self::get_committee_and_non_committee(active_era);
+        let nr_of_sessions = T::SessionsPerEra::get();
+        let blocks_per_session = Self::blocks_to_produce_per_session();
+        let validator_totals = Self::compute_validator_scaled_totals(active_era);
```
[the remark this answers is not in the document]
kostekIV: added to jira (A-952)

pallets/elections/src/impls.rs
kostekIV: me neither, can we leave it like it is for now?

mike1729:
Approve.

DamianStraszak (approved):

pallets/elections/src/impls.rs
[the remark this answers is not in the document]
kostekIV: jira issue for it is here: [A0-955](https://cardinal-cryptography.atlassian.net/browse/A0-955)

pallets/elections/src/impls.rs
```diff
@@ -0,0 +1,389 @@
...
+
+impl<T> Pallet<T>
+    where
+        T: Config + pallet_session::Config + pallet_staking::Config,
+        <<T as pallet_staking::Config>::Currency as Currency<<T as frame_system::Config>::AccountId>>::Balance: Into<u128>,
+        <T as pallet_session::Config>::ValidatorId: From<T::AccountId>,
+        <T as pallet_session::Config>::ValidatorId: Into<T::AccountId>,
+{
+    fn compute_validator_scaled_totals(era: EraIndex) -> BTreeMap<T::AccountId, u32> {
+        let scaled_totals_by_token = pallet_staking::ErasStakers::<T>::iter_prefix(era)
```
fixxxedpoint, line 91: variable name left from previous version

kostekIV: thx, renamed

fixxxedpoint (approved):

PR #371, opened 2022-04-06
E2e everybody wins

kostekIV:
# Description
Adds test for elections

## Type of change

# Checklist:


maciejzelaszczyk (approved):
Solid work. Probably have to wait with merging this until _Everybody wins_ passes review.

.github/scripts/run_consensus.sh
```diff
@@ -10,8 +10,8 @@ source docker/env
 # Generate chainspec and populate comittee keystores
 # hidden assumption: --chain-id must be present and different from a0dnet1, otherwise default chainspec
 # would generate some non-empty validators, and we want to start from the scratch
-docker run -v $(pwd)/docker/data:/data --entrypoint "/bin/sh" -e DAMIAN -e TOMASZ -e ZBYSZKO -e HANSU -e JULIA -e RUST_LOG=info \
- aleph-node:latest -c "aleph-node bootstrap-chain --base-path /data --chain-id a0dnet1-e2e --account-ids $DAMIAN,$TOMASZ,$ZBYSZKO,$HANSU,$JULIA --sudo-account-id $DAMIAN > /data/chainspec.json"
+docker run -v $(pwd)/docker/data:/data --entrypoint "/bin/sh" -e DAMIAN -e TOMASZ -e ZBYSZKO -e HANSU -e JULIA -e MARCIN -e RUST_LOG=info \
```
Marcin-Radecki, line 13: <3

e2e-tests/src/test/validators_rotate.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    let block_number = connection
+        .get_header::<Header>(None)
+        .expect("Could not fetch header")
+        .expect("Block exists; qed")
+        .number;
+    wait_for_finalized_block(&connection, block_number)?;
+
+    Ok(())
+}
```
fixxxedpoint, line 114: Consider adding a more "black box" test as well, something like: count how many times each node was elected during an era and check whether these values differ by <= 1 (excluding reserved nodes).

kostekIV: added

maciejnems (approved):

e2e-tests/src/test/validators_rotate.rs
fixxxedpoint: I think you should test it against the list returned by `get_authorities_for_session` (exclude reserved nodes), not the one generated by test itself.

kostekIV: since the other 3 asserts passed in the loop it is equivalent to the list returned by `get_authorities_for_session` minus the nodes that are reserved.

fixxxedpoint: You're right. I was thinking about this as if it was a separate test.

PR #373, opened 2022-04-08
Not so slow staking.

mike1729:
# Description

## Type of change

# Checklist:


bin/runtime/src/weights/pallet_bags_list.rs
```diff
@@ -0,0 +1,25 @@
+use frame_support::{traits::Get, weights::Weight};
+use sp_std::marker::PhantomData;
+
+/// Weight functions for `pallet_bags_list`.
+pub struct WeightInfo<T>(PhantomData<T>);
```
DamianStraszak, line 5: This is the same as Polkadot?

bin/runtime/src/custom_migrations.rs
```diff
@@ -0,0 +1,27 @@
+use sp_std::marker::PhantomData;
+
+use super::*;
+use frame_election_provider_support::SortedListProvider;
+use frame_support::{log, traits::OnRuntimeUpgrade, weights::Weight};
+
+pub struct UpgradeToBags<T>(PhantomData<T>);
```
DamianStraszak, line 7: I need to test it, but this migration will be triggered at every runtime update, right? Unless we remove this code from the runtime after this update :)

Anyway, this requires a bit of thought whether we can somehow make it better.

DamianStraszak (commented):
Thanks for a quick a high-quality PR!

I like it in general, but before merging I would like to run a few small tests to check how the migration works.

bin/runtime/src/custom_migrations.rs
mike1729: To run it conditionally we would need some methods as for storage version eg current/next storage version but I couldn't find anything like that. So I would just delete the mod and remove the upgrade from Executive leaving a code artifact documenting the upgrade.

bin/runtime/src/lib.rs
```diff
@@ -60,6 +60,10 @@ use sp_runtime::traits::One;
 pub use sp_runtime::BuildStorage;
 pub use sp_runtime::{FixedPointNumber, Perbill, Permill};
 
+mod bag_thresholds;
+mod custom_migrations;
+mod weights;
+
```
DamianStraszak, line 66: You forgot to bump `spec_version`.

DamianStraszak: And it's better to bump `transaction_version` as well, as the docs say to do that when a new pallet enters.

mike1729: Done.

bin/runtime/src/bag_thresholds.rs
```diff
@@ -0,0 +1,202 @@
+pub const THRESHOLDS: [u64; 200] = [
```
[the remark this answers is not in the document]
mike1729: Done.

bin/runtime/src/custom_migrations.rs
```diff
@@ -0,0 +1,27 @@
...
+            "Nominator count: {}",
+            pre_migrate_nominators_count
+        );
+        let moved = T::SortedListProvider::regenerate(nominators, weight_of_fn);
+        log::info!(target: "CustomMigration::UpgradeToBags", "Moved {} nominators", moved);
+
+        let after_migrate_nominators_count = T::SortedListProvider::count() as usize;
+        assert_eq!(pre_migrate_nominators_count, after_migrate_nominators_count);
+
+        BlockWeights::get().max_block
```
kostekIV, line 25: why max here?

kostekIV (approved):

bin/runtime/src/custom_migrations.rs
mike1729: Hard to determine the exact number, but it will be slow. There is no point in allowing other txs in the block enacting the migration.

PR #380, opened 2022-04-12
make sure validator keys in chainspec match S3 keystore

fbielejec:
# Description

Fixes finalization stalled on devnet. Bootstrap command generates new keys (aura/aleph) if no keystore is present. These keys did not match the ones stored on S3, but were written to the chainspec and subsequently present in the fork genesis block.

In this fix we sync the keystore from S3, rename validator paths to the ones that `bootstrap` command expects and only then generate chainspec skeleton, preserving the validator keys.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/.github/workflows/deploy-to-devnet.yml
+++ b/.github/workflows/deploy-to-devnet.yml
@@ -4,6 +4,7 @@ on:
   workflow_run:
     workflows: ["e2e-tests-main-devnet"]
     branches:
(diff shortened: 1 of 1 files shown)


bartoszjedrzejewski (approved):

mike1729 (approved):
Nice work!

PR #382, opened 2022-04-14
cached nix build

fixxxedpoint:
# Description

Using nix-community/naersk for better caching of cargo dependencies. Moreover, our previous attempt at using nix was not valid from nix's point of view (download of dependencies should be managed by nix, not cargo directly) - this could lead to some confusion among users. This PR attempts at integrating nix-build with some of our pipelines - build node and runtime. This is a result of several attempts of incorporating nix to our build procedure - previous one that was using a tool called `crate2nix` end up being too complicated (it was managing each cargo dependency separately and was using rustc directly instead of cargo). This one tries to simplify the whole process and manages all crate dependencies on per Cargo.lock basis (it builds all dependencies as a single cached package - changing Cargo.lock forces it to rebuild it).

## Type of change

Build procedure.

- [x] This change requires a documentation update

# Checklist:


default.nix
```diff
@@ -1,35 +1,41 @@
...
-  });
+  # declares a build environment where C and C++ compilers are delivered by the llvm/clang project
+  # in this version build process should rely only on clang, without access to gcc
+  llvm = versions.llvm;
+  env = versions.stdenv;
+  llvmVersionString = "${nixpkgs.lib.getVersion env.cc.cc}";
 
   # we use a newer version of rocksdb than the one provided by nixpkgs
-  # we disable all compression algorithms and force it to use SSE 4.2 cpu instruction set
+  # we disable all compression algorithms, force it to use SSE 4.2 cpu instruction set and disable its `verify_checksum` mechanism
```
Marcin-Radecki, line 28: I know it's a bit late comment, but what about we separate bumping RocksDB version into a separate change? This is to decouple the risk of significant rework of our build system, and changing one of crucial dependencies in a significant way (ie we provide a patch to a third party lib, this is imho quite invasive change). I'd be more comfortable to review this PR when RocksDB bump would be a separate PR, what do you think?

default.nix
```diff
@@ -1,35 +1,41 @@
-{ rocksDBVersion ? "6.29.3" }:
+{ versions ? import ./nix/versions.nix
+, release ? true
+, name ? "aleph-node"
+, crates ? { "aleph-node" = []; }
+, runTests ? false
+, singleStep ? false
+, rustflags ? "-C target-cpu=native"
```
Marcin-Radecki, line 7: Does this come from any particular place? What `RUSTFLAGS` we use now?

default.nix
```diff
@@ -39,92 +45,125 @@ let
...
-    export BINDGEN_EXTRA_CLANG_ARGS=" \
-        ${"-isystem ${llvm.libclang.lib}/lib/clang/${llvmVersionString}/include"} \
-        $BINDGEN_EXTRA_CLANG_ARGS
-    "
-    export ROCKSDB_LIB_DIR="${customRocksdb}/lib"
-    export ROCKSDB_STATIC=1
+    ${rocksDbShellHook}
+    export SUBSTRATE_CLI_GIT_COMMIT_HASH=${SUBSTRATE_CLI_GIT_COMMIT_HASH}
+    export RUSTFLAGS="${rustflags}"
+    export LIBCLANG_PATH=${LIBCLANG_PATH};
```
Marcin-Radecki, line 130: Similar question to all this envs, who sets them in our current build process, and what are their values? Cargo does it?

default.nix
fixxxedpoint: github pipelines use `-C target-cpu=x86-64-v2` - I saw once `ILLEGAL INSTRUCTION` and got pretty scared.

default.nix
fixxxedpoint: Mostly, they are not set by cargo, just some magic happens in some of the build.rs scripts which are trying to find them in some `standard` paths. Unfortunately, those paths are meaningless under nix-build.

default.nix
```diff
@@ -39,92 +45,125 @@ let
...
+  RUSTFLAGS="${rustflags}";
+  ROCKSDB_STATIC=1;
+  LIBCLANG_PATH="${llvm.libclang.lib}/lib";
+  PROTOC="${protobuf}/bin/protoc";
+  # From: https://github.com/NixOS/nixpkgs/blob/1fab95f5190d087e66a3502481e34e15d62090aa/pkgs/applications/networking/browsers/firefox/common.nix#L247-L253
+  # Set C flags for Rust's bindgen program. Unlike ordinary C
+  # compilation, bindgen does not invoke $CC directly. Instead it
+  # uses LLVM's libclang. To make sure all necessary flags are
+  # included we need to look in a few places.
+  BINDGEN_EXTRA_CLANG_ARGS=" \
```
Marcin-Radecki, line 160: This seems a bit of hacky. Where do we use `bindgen` exactly?

fixxxedpoint: e.g. https://github.com/rust-rocksdb/rust-rocksdb/blob/3a46074c36a3f1f1f066ca0c9be1a2efc91563e2/librocksdb-sys/build.rs#L36

default.nix
Marcin-Radecki: As discussed, let's keep this value as `generic`

.github/workflows/build-node-and-runtime.yml
```diff
@@ -19,59 +19,55 @@ jobs:
...
-          path: target/release/wbuild/aleph-runtime/aleph_runtime.compact.wasm
+          path: result/lib/aleph_runtime.compact.wasm
           if-no-files-found: error
           retention-days: 7
 
       - name: Build test binary
-        run: cargo build --release -p aleph-node --features "short_session"
+        run: |
+          chmod +x docker/nix-build.sh
+          RUSTFLAGS='"-C target-cpu=x86-64-v2"' CRATES='{ "aleph-node" = ["short_session"]; "aleph-runtime" = ["short_session"]; }' NAME='"short_session"' docker/nix-build.sh
```
Marcin-Radecki, line 57: Could you please change to `generic`

Marcin-Radecki, line 57: `NAME` argument seems not relevant anymore

.github/actions/cache_install_nix/action.yml
```diff
@@ -0,0 +1,42 @@
+name: 'Cache install nix'
```
Marcin-Radecki, line 1: Please revert all workflow changes. We'll merge them in the later PR.

default.nix
```diff
@@ -39,92 +45,125 @@ let
...
-    export CARGO_HOME="$out/cargo"
-    export CARGO_BUILD_TARGET="x86_64-unknown-linux-gnu"
-
-    cargo build --locked --release -p aleph-node
   '';
-
-  installPhase = ''
-    mkdir -p $out/bin
-    mv target/x86_64-unknown-linux-gnu/release/aleph-node $out/bin/
+  postInstall = ''
```
Marcin-Radecki, line 139: Could you please add comment here why we need a custom post-install step

default.nix
```diff
@@ -39,92 +45,125 @@ let
...
+  gitCommitDrv = nixpkgs.runCommand "gitCommit" { nativeBuildInputs = [nixpkgs.git]; } ''
+    GIT_DIR=${gitFolder} git rev-parse --short HEAD > $out
+  '';
+  gitCommit = builtins.readFile gitCommitDrv;
+
+  modePath = if release then "release" else "debug";
+  pathToWasm = "target/" + modePath + "/wbuild/aleph-runtime/target/wasm32-unknown-unknown/" + modePath + "/aleph_runtime.wasm";
+  pathToCompactWasm = "target/" + modePath + "/wbuild/aleph-runtime/aleph_runtime.compact.wasm";
+
+  features =
```
Marcin-Radecki, line 80: Please improve readability here a little bit by using meaningful names

default.nix
```diff
@@ -39,92 +45,125 @@ let
...
-  buildInputs = [
-    rustToolchain
-    llvm.clang
-    openssl.dev
-    protobuf
-    customRocksdb
-    pkg-config
-    cacert
+with nixpkgs; naersk.buildPackage rec {
+  inherit name release src singleStep;
```
Marcin-Radecki, line 109: Shall we comment `singleStep` what's this all about

default.nix
```diff
@@ -39,92 +45,125 @@ let
...
+  doCheck = runTests;
```
Marcin-Radecki, line 110: Can we please have syntax with `@` here to name this script arguments via `args` aliast

default.nix
```diff
@@ -39,92 +45,125 @@ let
...
-    findutils
-    patchelf
+    cacert
+    pkg-config
+    llvm.libclang
+    protobuf
   ];
-
+  buildInputs = [
+    openssl.dev
```
Marcin-Radecki, line 119: Maybe this is not needed as well?

default.nix
```diff
@@ -39,92 +45,125 @@ let
...
+  ] ++ nixpkgs.lib.optional useCustomRocksDb customRocksdb;
+  cargoBuildOptions = opts:
+    packageFlags
+    ++ [featuresFlag]
+    ++ ["--locked" "--offline"]
```
Marcin-Radecki, line 124: This actually changes the way we use `cargo`. Please comment here or document the change.

default.nix
```diff
@@ -39,92 +45,125 @@ let
...
+    if [ -f ${pathToCompactWasm} ]; then
+      mkdir -p $out/lib
+      cp ${pathToCompactWasm} $out/lib/
+    fi
   '';
 
-  fixupPhase = ''
-    find $out -type f -exec patchelf --shrink-rpath '{}' \; -exec strip '{}' \; 2>/dev/null
-  '';
+  SUBSTRATE_CLI_GIT_COMMIT_HASH="${gitCommit}";
```
Marcin-Radecki, line 150: Please elaborate each value from where it originated from

docker/nix-build.sh
```diff
@@ -4,6 +4,13 @@ set -euo pipefail
 SPAWN_SHELL=${SPAWN_SHELL:-false}
```
Marcin-Radecki, line 4: Please move this file and `Dockerfile_build` to `nix` folder. Please update `build.md` accordingly.

Marcin-Radecki, line 4: Can we please increase `max-jobs` param since now it takes approx 10 mins to download all the derivations from `nixpkgs`?

shell.nix
```diff
@@ -0,0 +1,18 @@
+{ versions ? import ./nix/versions.nix
```
Marcin-Radecki, line 1: Can we please whether rust analyzer and clippy is installed in this shell?

docker/nix-build.sh
Marcin-Radecki, line 4: Can we please change the way docker displays logs from `nix-build.sh` ? So it does not display logs in an interactive way. There's a limit in docker of 1 MB now which should be solved by that as well.

default.nix
```diff
@@ -1,35 +1,60 @@
-{ rocksDBVersion ? "6.29.3" }:
```
Marcin-Radecki, line 1: Can we please add two tasks to JIRA:
1. Check forward and backward compatibility between nix `aleph-node` and current `aleph-node` 
2. Make sure WASM produced by nix `aleph-node` is exactly the same as WASM produced by current node.

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -64,11 +64,10 @@ jobs:
...
 
-      - name: Install Rust Toolchain
-        uses: actions-rs/toolchain@v1
-
-      - name: Install WASM target
-        run: rustup target add wasm32-unknown-unknown
+      - name: Install and restore nix
+        uses: ./.github/actions/cache_install_nix
+        with:
+          cache-key: ${{ runner.os }}-nix_store-v4-${{ hashFiles('**/Cargo.lock') }} # bump cache version if you need to clear it, see https://github.com/actions/cache/issues/485 and https://stackoverflow.com/questions/63521430/clear-cache-in-github-actions
```
fbielejec, line 70: not that it matters much, but is there a reason to start with the cache key versioned as 4 already?

default.nix
```diff
@@ -39,92 +64,117 @@ let
...
+  cargoBuildOptions = opts:
+    packageFlags
+    ++ [featuresFlag]
+    ++ ["--locked" "--offline"]
+    ++ opts;
   shellHook = ''
-    export RUST_SRC_PATH="${rustToolchain}/lib/rustlib/src/rust/src"
+    ${rocksDbShellHook}
+
+    # this is the way we can to pass additional arguments to rustc that is called by cargo, e.g. list of available cpu features
```
fbielejec, line 142:
```suggestion
    # this is the way we can pass additional arguments to rustc that is called by cargo, e.g. list of available cpu features
```

shell.nix
fixxxedpoint: It is installed by default.

docker/nix-build.sh
fixxxedpoint: Moved.

fixxxedpoint: `--max-jobs` is now `auto` (number of cpus)

shell.nix
fixxxedpoint: Moreover, `nix-shell` should provide you most common tools for developing our project -  cargo, rust, rust-analyzer, clippy, rust-src, clang, clang++.

docker/nix-build.sh
fixxxedpoint: Procedure described by`BUILD.md` should force docker disable log clipping (it's ugly...).

.github/workflows/e2e-tests-main-devnet.yml
fixxxedpoint: This integration is removed from this PR.

default.nix
fixxxedpoint: Added.

default.nix
fixxxedpoint: Fixed.

Marcin-Radecki:
Could we please add `!nix/` to `.dockerignore` as without it we run into following error
>Step 2/8 : COPY nix/nix-build.sh /node/
COPY failed: file not found in build context or excluded by .dockerignore: stat nix/nix-build.sh: file does not exist

BUILD.md
```diff
@@ -14,7 +14,8 @@ In order to build a binary for `aleph-node` using docker we first need to instal
 distribution, by executing `sudo apt install docker.io` (please consult your distribution's manual describing docker
 installation procedure). Build procedure can be invoked by running:
 ```
-sudo DOCKER_BUILDKIT=1 docker build -t aleph-node/build -f docker/Dockerfile_build .
+sudo DOCKER_BUILDKIT=1 BUILDKIT_PROGRESS=plain BUILDKIT_STEP_LOG_MAX_SIZE=-1
```
Marcin-Radecki, line 17: Nit: we can change this to `export` syntax to support both `bash` and `fish`

.github/workflows/build-node-and-runtime.yml
```diff
@@ -19,59 +19,55 @@ jobs:
...
+      - name: Install and restore nix
+        uses: ./.github/actions/cache_install_nix
+        with:
+          cache-key: ${{ runner.os }}-nix_store-v4-${{ hashFiles('**/Cargo.lock') }} # bump cache version if you need to clear it, see https://github.com/actions/cache/issues/485 and https://stackoverflow.com/questions/63521430/clear-cache-in-github-actions
 
       - name: Build binary and runtime
-        run: cargo build --release -p aleph-node
+        run: |
+          chmod +x docker/nix-build.sh
+          RUSTFLAGS='"-C target-cpu=x86-64-v2"' docker/nix-build.sh
```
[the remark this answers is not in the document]
fixxxedpoint: This was removed

fbielejec (approved):

Marcin-Radecki (approved):
Great work!

PR #392, opened 2022-04-25
Remove slack notifications.

bartoszjedrzejewski:


fbielejec (approved):
I thought we will replace it with some other workflow for notifications, but removing works too

PR #403, opened 2022-04-27
Fix node's identity crisis

h4nsu:
The goal is to change the node's name displayed in the wallet:
image

to Aleph Node

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [x] Infrastructure updated accordingly
- [x] Bump `aleph-client` version if relevant

### DIFF
--- a/bin/node/src/cli.rs
+++ b/bin/node/src/cli.rs
@@ -20,7 +20,7 @@ pub struct Cli {
 
 impl SubstrateCli for Cli {
     fn impl_name() -> String {
-        "Substrate Node".into()
+        "Aleph Node".into()
     }
 
     fn impl_version() -> String {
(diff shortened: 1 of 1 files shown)


pmikolajczyk41 (approved):

timorl (approved):

pmikolajczyk41 (commented):
actually, `support_url` could also be more appropriate :P

timorl:
Good point, but I'm not sure we have one. :/

deuszx (approved):

h4nsu:
I thought about adjusting `support_url` too, but I haven't found if and where it is displayed in the wallet... So not sure what we should put there...

kostekIV (approved):

pmikolajczyk41:
it is surely displayed in logs, so external validators will see it in case of failure

I would paste maybe Discord channel link or something like this

timorl:
I quickly asked Antek, and he suggested pointing to docs.alephzero.org .

PR #406, opened 2022-04-29
A0 774 split e2e tests

maciejzelaszczyk:
# Description

This provides a split to e2e tests which is aimed at making it possible to run them in parallel using Github workflows. A separate Github Action is defined to handle the running of tests.

## Type of change

- [X] New feature (non-breaking change which adds functionality)

# Checklist:

- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation


aleph-client/src/session.rs
```diff
@@ -58,15 +58,22 @@ pub fn set_keys(connection: &Connection, new_keys: Keys, status: XtStatus) {
     send_xt(connection, xt, Some("set_keys"), status);
 }
 
+/// Get the number of the current session.
+/// Panics for the 0th session.
 pub fn get_current(connection: &Connection) -> u32 {
+        try_get_current(connection).unwrap()
```
kostekIV, line 64: actually using here unwrap_or(0) and removing the function `try_get_current` makes more sense.

e2e-tests/src/main.rs
```diff
@@ -33,6 +34,53 @@ fn init_env() {
...
+                run_test_case(test_case, test_name, config)?;
+                Ok(())
+            }
+        ).collect::<Vec<_>>();
+    Ok(())
+}
+
+/// Runs specified test cases in sequence.
+/// Checks whether each provided test case is valid.
+fn run_specified_test_cases(test_cases: Vec<String>, possible_test_cases: HashMap<&str, TestCase>, config: &Config) -> anyhow::Result<()> {
```
kostekIV, line 52: nit: a really long line. You should run `cargo fmt` in the e2e-tests directory (and we need to add this to pipelines&checks but not in this pr).

e2e-tests/src/cases.rs
```diff
@@ -0,0 +1,32 @@
...
+    finalization as test_finalization, staking_era_payouts as test_staking_era_payouts,
+    staking_new_validator as test_staking_new_validator, token_transfer as test_token_transfer,
+    treasury_access as test_treasury_access,
+};
+use crate::config::Config;
+
+/// Wrapper around the type of the test case.
+#[derive(Copy, Clone)]
+pub enum TestCase {
+    BlockNumberResult(fn(&Config) -> anyhow::Result<u32>),
```
kostekIV, line 15: we dont really use the return value from only one tests that returns something. I would rather rewrite the function `test_finalization` so it returns unit and remove the TestCase enum.

e2e-tests/src/main.rs
```diff
@@ -33,6 +34,53 @@ fn init_env() {
...
+    let _ = test_cases
+        .iter()
```
kostekIV, line 54: instead of `iter -> map -> collect` use for_each or simple for loop.

kostekIV (commented):
Remove the `.idea` files :P

maciejzelaszczyk:
> Remove the `.idea` files :P

Aww, this is what opening separate parts of the project in multiple windows does :( Removed.

aleph-client/src/session.rs
maciejzelaszczyk: Agreed.

e2e-tests/src/main.rs
maciejzelaszczyk: Right, forgot to run a local pipeline which does this.

e2e-tests/src/cases.rs
maciejzelaszczyk: In hindsight, this indeed seems like a more sensible choice, changed.

e2e-tests/src/main.rs
maciejzelaszczyk: OK, reworked. I had some trouble here with _Result_ in a closure, hence the _map_, but now _for_each_ seems to work fine.

kostekIV: Sorry for the confusion, actually I would go even further and just use simple for loop syntax

e2e-tests/src/test/finalization.rs
```diff
@@ -2,7 +2,8 @@ use aleph_client::{create_connection, wait_for_finalized_block};
 
 use crate::config::Config;
 
-pub fn finalization(config: &Config) -> anyhow::Result<u32> {
+pub fn finalization(config: &Config) -> anyhow::Result<()> {
     let connection = create_connection(&config.node);
-    wait_for_finalized_block(&connection, 1)
+    let _ = wait_for_finalized_block(&connection, 1)?;
```
kostekIV, line 7: nit:
```suggestion
    wait_for_finalized_block(&connection, 1)?;
```

e2e-tests/src/main.rs
```diff
@@ -33,6 +33,38 @@ fn init_env() {
     env_logger::init();
 }
 
+/// Runs default test cases in sequence.
+fn run_default_test_cases(
```
kostekIV, line 37: rename to `run_all_test_cases`

e2e-tests/src/main.rs
maciejzelaszczyk: No problem, in the end I was wrong and _for_each_ didn't do the trick, so I went with a simple _for_ loop. Will ping you once all the pipelines pass.

e2e-tests/src/test/finalization.rs
maciejzelaszczyk: Makes sense :)

e2e-tests/src/cases.rs
```diff
@@ -0,0 +1,25 @@
...
+       ("token_transfer", test_token_transfer),
+       ("channeling_fee", test_channeling_fee),
+       ("treasury_access", test_treasury_access),
+       ("batch_transactions", test_batch_transactions),
+       ("staking_era_payouts", test_staking_era_payouts),
+       ("staking_new_validator", test_staking_new_validator),
+       ("change_validators", test_change_validators),
+       ("fee_calculation", test_fee_calculation),
+    ])
+}
```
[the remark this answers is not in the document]
maciejzelaszczyk: Should be OK now.

e2e-tests/src/main.rs
maciejzelaszczyk: @kostekIV The e2e pipeline went green, you can re-review at your convenience.

.github/scripts/run_e2e_test.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
kostekIV, line 1: does it works with the script `run_local_e2e_pipeline.sh`?

maciejzelaszczyk: It does, a simplified log:

```
     Running unittests (target/debug/deps/aleph_node-9c6f605cf1eff32e)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests (target/debug/deps/aleph_runtime-54abc09c46a98ea6)

running 1 test
test __construct_runtime_integrity_test::runtime_integrity_tests ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests (target/debug/deps/finality_aleph-f4bd448316b6f4fe)

running 127 tests

...

test result: ok. 127 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 53.54s

     Running unittests (target/debug/deps/pallet_aleph-621c615a61bba40d)

running 8 tests

...

test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s

     Running unittests (target/debug/deps/pallet_elections-6760ee537da352c7)

running 2 tests
test mock::__construct_runtime_integrity_test::runtime_integrity_tests ... ok
test tests::test_elect ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s

     Running unittests (target/debug/deps/primitives-27c8394dd7942b6e)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

Yithis (approved):
Looks good to me, when I read it all, but I'm really not that familiar with this part of code. Does github checks already tried this code in this PR or is it somehow not that obvious since it is tests code?

maciejzelaszczyk:
> Looks good to me, when I read it all, but I'm really not that familiar with this part of code. Does github checks already tried this code in this PR or is it somehow not that obvious since it is tests code?

Github worklows passed, for whatever reason the repo has a non-existent check listed as required, will verify this before merging.

maciejzelaszczyk:
@kostekIV There was a problem with local test runs. Now, it is fixed by introducing deterministic order when more then one case is run on one machine. Ready for re-review :)

.github/scripts/run_e2e_test.sh
maciejzelaszczyk: Now it does :P

e2e-tests/src/lib.rs
```diff
@@ -1,13 +1,8 @@
+pub use cases::{possible_test_cases, PossibleTestCases, TestCase};
```
kostekIV, line 1: TestCase is not used

e2e-tests/src/main.rs
```diff
@@ -1,28 +1,28 @@
 use std::{env, time::Instant};
 
+use aleph_e2e_client::{possible_test_cases, Config, PossibleTestCases, TestCase};
```
kostekIV, line 3: TestCase is not used

kostekIV (approved):
Great work

PR #413, opened 2022-05-10
Everybody wins

kostekIV:
# Description

## Type of change

# Checklist:


bin/runtime/src/lib.rs
```diff
@@ -344,13 +345,89 @@ parameter_types! {
...
+    // The validators for the committee at the session `n` are chosen as follow:
+    // 1. Reserved validators are always chosen.
+    // 2. Given non-reserved list of validators the chosen ones are from the range:
+    // `n * free_seats` to `(n + 1) * free_seats` where free_seats is equal to free number of free
+    // seats in the committee after reserved nodes are added.
+    let current_session = Session::current_index() as usize;
+    let first_validator = current_session * free_seats;
+
+    validators.extend(
+        (first_validator..first_validator + free_seats)
```
Marcin-Radecki, line 374: Are we planning to add later some unit tests to `rotate`? Mainly for sake of the readability of any future reader. I know it's very simple algo, yet maybe readability can benefit from a simple UT which imitates runtime of `run_nodes.sh -v 6`? Or maybe e2e test would be a better choice here

bin/runtime/src/lib.rs
```diff
@@ -344,13 +345,89 @@ parameter_types! {
...
+use primitives::SessionIndex;
+type SM = pallet_session::historical::NoteHistoricalRoot<Runtime, Staking>;
+pub struct SampleSessionManager;
+
+impl pallet_session::SessionManager<AccountId> for SampleSessionManager {
+    fn new_session(new_index: SessionIndex) -> Option<Vec<AccountId>> {
+        SM::new_session(new_index);
+        // new session is always called before the end_session of the previous session
+        // so we need to populate reserved set here not on start_session nor end_session
+        let committee = rotate();
```
Marcin-Radecki, line 405: We need to call `rotate` before `populate_reserved_on_next_era_start` right? Since `populate_reserved_on_next_era_start` might modify `ErasReserved` which is read in `rotate`. And we're doing it so in order to follow how pallet staking works, ie we're planning for the future here

pallets/elections/src/lib.rs
```diff
@@ -50,6 +46,16 @@ pub mod pallet {
     #[pallet::without_storage_info]
     pub struct Pallet<T>(_);
 
+    #[pallet::storage]
+    #[pallet::getter(fn members)]
+    pub type Members<T: Config> = StorageValue<_, Vec<T::AccountId>, ValueQuery>;
+
+    #[pallet::storage]
+    pub type MembersPerSession<T> = StorageValue<_, u32, OptionQuery>;
```
Marcin-Radecki, line 54: Does it need to be `Option` here? Is there any logic that depends or will depend in the future on `Option` state?

Marcin-Radecki (changes requested):
Just some nits and comment requests, also some questions.

bin/runtime/src/lib.rs
kostekIV: yep

bin/runtime/src/lib.rs
```diff
@@ -330,6 +330,7 @@ impl_opaque_keys! {
 
 parameter_types! {
     pub const SessionPeriod: u32 = DEFAULT_SESSION_PERIOD;
+    pub const MembersPerSession: u32 = 4;
```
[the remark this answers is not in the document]
kostekIV: Added a constant to the primitives but also removed this type. Now we are using MembersPerSession from pallet_elections (this will allow to change nr of members in the future)

bin/runtime/src/lib.rs
```diff
@@ -344,13 +345,89 @@ parameter_types! {
...
+fn populate_reserved_on_next_era_start(start_index: SessionIndex) {
+    let current_era = match Staking::active_era() {
+        Some(ae) => ae.index,
+        _ => return,
+    };
+    // this will be populated once for the session `n+1` on the start of the session `n` where session
+    // `n+1` starts a new era.
+    if let Some(era_index) = Staking::eras_start_session_index(current_era + 1) {
+        if era_index == start_index {
+            let reserved = pallet_staking::Invulnerables::<Runtime>::get();
```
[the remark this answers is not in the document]

pallets/elections/src/lib.rs
kostekIV: changed to value

bin/node/src/chain_spec.rs
```diff
@@ -481,7 +482,7 @@ fn generate_genesis_config(
             validator_count,
             // to satisfy some e2e tests as this cannot be changed during runtime
             minimum_validator_count: 4,
-            invulnerables: accounts_config.members,
+            invulnerables: accounts_config.members[0..2].to_vec(),
```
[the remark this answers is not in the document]
kostekIV: added a comment that this is for tests. We needed some number less than 4.

bin/runtime/src/lib.rs
```diff
@@ -344,13 +345,89 @@ parameter_types! {
...
+            pallet_elections::ErasReserved::<Runtime>::put(reserved);
+        }
+    }
+}
+
+use primitives::SessionIndex;
+type SM = pallet_session::historical::NoteHistoricalRoot<Runtime, Staking>;
+pub struct SampleSessionManager;
```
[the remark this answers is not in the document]

Marcin-Radecki (approved):
Splendid rework.

bin/runtime/src/lib.rs
```diff
@@ -344,13 +344,109 @@ parameter_types! {
...
+    current_session: SessionIndex,
+    n_validators: usize,
+    mut all_validators: Vec<T>,
+    mut reserved: Vec<T>,
+) -> Option<Vec<T>> {
+    if current_era == 0 {
+        return None;
+    }
+
+    all_validators.retain(|v| !reserved.contains(v));
```
piotrMocz, line 358: What I dislike about this function is the overloading of names because of mutability.
Here `all_validators` won't actually contain all of the validators and at the end of the function the `reserved` vector won't contain only the reserved nodes. Seems to me that since `T: Clone` anyway, we could do this without making `all_validators` and `reserved` mutable (i.e. use `filter` instead of `retain` and so on).

That said, this is still quite a nice readable piece of code, so I'm just leaving this here for future consideration 🙂

piotrMocz (approved):
LGTM, the comment is just something to consider in free time 😄

bin/runtime/src/lib.rs

fixxxedpoint (approved):
Good rework.

PR #414, opened 2022-05-10
A0-873: Vesting module for `aleph-client`

pmikolajczyk41:
# Description

This PR introduces the vesting module to the aleph-client lib crate. It enables vesting (for self or on behalf), making vested transfers and managing vesting schedules. This is needed for implementing corresponding scenarios.

## Type of change

# Checklist:

- [x] Bump `aleph-client` version if relevant

### DIFF
--- a/aleph-client/src/lib.rs
+++ b/aleph-client/src/lib.rs
@@ -180,7 +180,9 @@ fn storage_key(module: &str, version: &str) -> [u8; 32] {
 /// * `call` name of the pallet's call
 ///
 /// # Example
-/// ```
+/// ```no_run
+/// use aleph_client::get_storage_key;
+///
 /// let staking_nominate_storage_key = get_storage_key("Staking", "Nominators");
 /// assert_eq!(staking_nominate_storage_key, String::from("5f3e4907f716ac89b6347d15ececedca9c6a637f62ae2af1c7e31eed7e96be04"));
 /// ```
--- a/aleph-client/src/lib.rs
+++ b/aleph-client/src/lib.rs
@@ -180,7 +180,7 @@ fn storage_key(module: &str, version: &str) -> [u8; 32] {
 /// * `call` name of the pallet's call
 ///
 /// # Example
-/// ```no_run
+/// ```
 /// use aleph_client::get_storage_key;
 ///
 /// let staking_nominate_storage_key = get_storage_key("Staking", "Nominators");
--- a/aleph-client/src/lib.rs
+++ b/aleph-client/src/lib.rs
@@ -2,8 +2,7 @@ use std::{thread::sleep, time::Duration};
 
 use codec::Encode;
 use log::{info, warn};
-use sp_core::{sr25519, Pair, H256};
-use sp_core::storage::StorageKey;
+use sp_core::{sr25519, storage::StorageKey, Pair, H256};
 use sp_runtime::{generic::Header as GenericHeader, traits::BlakeTwo256};
 use substrate_api_client::{
     rpc::ws_client::WsRpcClient, std::error::Error, AccountId, Api, ApiResult, RpcClient,
@@ -34,6 +33,10 @@ pub use system::set_code;
 pub use transfer::{
     batch_transfer as balances_batch_transfer, transfer as balances_transfer, TransferTransaction,
 };
+pub use vesting::{
+    get_schedules, merge_schedules, vest, vest_other, vested_transfer, VestingError,
+    VestingSchedule,
+};
 pub use waiting::{wait_for_event, wait_for_finalized_block};
 
 mod account;
@@ -45,6 +48,7 @@ mod session;
 mod staking;
 mod system;
 mod transfer;
+mod vesting;
 mod waiting;
 
 pub trait FromStr: Sized {
--- a/aleph-client/src/vesting.rs
+++ b/aleph-client/src/vesting.rs
@@ -15,9 +15,17 @@ pub enum VestingError {
     #[error("🦺❌ Account has no active vesting schedules.")]
     NotVesting,
 }
-const PALLET: &str = "Vesting";
+
 pub type VestingSchedule = VestingInfo<Balance, BlockNumber>;
 
+const PALLET: &str = "Vesting";
+
+/// Calls `pallet_vesting::vest` for `who`, i.e. makes all unlocked balances transferable.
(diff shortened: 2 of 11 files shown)


aleph-client/src/vesting.rs
```diff
@@ -0,0 +1,123 @@
...
+/// Calls `pallet_vesting::vest` for `who`, i.e. makes all unlocked balances transferable.
+///
+/// Does not expect `connection` to be signed. Fails if transaction could not have been sent.
+///
+/// *Note*: This function returns `Ok(_)` even if the account has no active vesting schedules
+/// and thus the extrinsic was not successful. However, semantically it is still correct.
+pub fn vest(connection: &Connection, who: KeyPair) -> AnyResult<()> {
+    // Ensure that we make a call as `account`.
+    let connection = connection.clone().set_signer(who.clone());
+    let xt = compose_extrinsic!(connection, "Vesting", "vest");
```
deuszx, line 32: You could be using `PALLET` here instead of `"Vesting"` (and below).

pmikolajczyk41: sure, I must have missed that - fixed

aleph-client/src/lib.rs
```diff
@@ -176,7 +180,9 @@ fn storage_key(module: &str, version: &str) -> [u8; 32] {
 /// * `call` name of the pallet's call
 ///
 /// # Example
-/// ```
+/// ```no_run
+/// use aleph_client::get_storage_key;
```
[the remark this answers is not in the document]
pmikolajczyk41: sure, completely unnecessary and unwanted - removed

aleph-client/src/vesting.rs
```diff
@@ -0,0 +1,123 @@
...
+///
+/// *Note*: This function returns `Ok(_)` even if the account has no active vesting schedules
+/// and thus the extrinsic was not successful. However, semantically it is still correct.
+pub fn vest_other(
+    connection: Connection,
+    caller: KeyPair,
+    vest_account: AccountId,
+) -> AnyResult<()> {
+    // Ensure that we make a call as `caller`.
+    let connection = connection.set_signer(caller);
```
deuszx, line 55: In the same vain as @obrok 's comment - why pass in the `connection` and `caller` only to set it as a signer and not set it on the call site? That would allow you to pass in the `Connection` by reference (I assume).

pmikolajczyk41:
Oh, I failed to mention - for testing `merge_schedules` or `get_schedules` (for which I found adding special commands to `cliain` quite artificial and unwanted): I prepared you some short snippet in one of the early commits (which was then reverted)

aleph-client/src/vesting.rs
pmikolajczyk41: that's actually more subtle - suppose that, as a user, you want to e.g. call `pallet_vesting::vest()` for some account: you should obviously pass a connection (just like to any other function which should lead to some interaction with the chain) and additionally all the arguments that are required by `pallet_vesting::vest()`, in this case, just the account

the actual requirement that the account to be vested should sign the connection and thus implicitly passed seems more like the implementation detail, which does not have to be imposed on the 'end user'; in the way it is now, the API is less concise but more legible

Marcin-Radecki: I think the signer of the transactions is the account is enough degrees of freedom we need here:
* for `vest`, in Wallet we don't have another argument, so in example below `Alice` is the both signer and effect of the `vest` call. 
image
* for `vestOther` , indeed we have another argument `target`, but it is already passed here as `AccountId` called `vest_account`.

There are two non-written rules of `cliain` and `aleph-wallet`:
1. For all **target** account args, we pass `AccountId` type (like correctly we did here with `vest_other`. We can pass many such args depends on the call we're wrapping.
2. We wrap signer account in `Connection` object, so in `aleph-client` we assume this is the only one account that requires both keys to be created. In `cliain` , we control this by using `--seed` parameter. 

This comment applies to all wrapped extrinsic call in this file.

bin/cliain/src/vesting.rs
```diff
@@ -0,0 +1,61 @@
+use aleph_client::{
+    account_from_keypair, keypair_from_string, BlockNumber, Connection, KeyPair, VestingSchedule,
+};
+use log::{error, info};
+use primitives::{Balance, TOKEN};
+
+/// Retrieves signer from `connection`.
+///
+/// Panics if `connection` is not signed.
+fn get_caller(connection: &Connection) -> KeyPair {
```
Marcin-Radecki, line 10: See other comment in `aleph-client` - imho we don't need a separate signer as `Connection` is signed already by `--seed`.

Marcin-Radecki (changes requested):
Solid work

aleph-client/src/vesting.rs
pmikolajczyk41: okay, changed

aleph-client/src/vesting.rs
```diff
@@ -0,0 +1,123 @@
+use anyhow::Result as AnyResult;
```
h4nsu: Everywhere else we use it as `Result`, please don't introduce inconsistency

aleph-client/src/vesting.rs
h4nsu: I also don't like the pattern of putting `set_signer` inside these calls. In my opinion this is better handled outside. General rule of thumb I usually follow: in a library the most fine grained building blocks of the api should be as small and "single-responsibility" as possible. Your "user" will be a programmer, they'll know better what they want.

bin/cliain/src/vesting.rs
h4nsu: +1
Having to do that kind of shaneningans simply shows that `set_signer` does not belong inside `aleph-client` functions

h4nsu (changes requested):

h4nsu (approved):

bin/cliain/src/vesting.rs
deuszx: And that there should be a separate type - `SignedConnection` that differentiates at the type level.

pmikolajczyk41: let me handle this here:
https://cardinal-cryptography.atlassian.net/browse/A0-569

Marcin-Radecki (approved):

PR #415, opened 2022-05-10
Add pallet contracts

h4nsu:


bin/runtime/src/lib.rs
```diff
@@ -707,6 +711,41 @@ impl pallet_utility::Config for Runtime {
     type PalletsOrigin = OriginCaller;
 }
 
+parameter_types! {
+    // The following weights are pulled out of thin air. A separate task to adjust them is already in JIRA.
+    pub const DepositPerItem: Balance = TOKEN / 100;  //  10 milli-AZERO per item
+    pub const DepositPerByte: Balance = TOKEN / 100_000;      // ~10 milli-AZERO per kB storage
+    // The lazy deletion runs inside on_initialize.
```
pmikolajczyk41, line 718: since you have already provided the real values for `DepositPer_`, I would encourage you to add similar estimations for the two parameters below

bin/runtime/src/lib.rs
```diff
@@ -707,6 +711,41 @@ impl pallet_utility::Config for Runtime {
     type PalletsOrigin = OriginCaller;
 }
 
+parameter_types! {
+    // The following weights are pulled out of thin air. A separate task to adjust them is already in JIRA.
```
pmikolajczyk41, line 715: maybe `///` instead of `//`?

bin/runtime/src/lib.rs
```diff
@@ -889,6 +929,51 @@ impl_runtime_apis! {
...
+
+        fn call(
+            origin: AccountId,
+            dest: AccountId,
+            value: Balance,
+            gas_limit: u64,
+            storage_deposit_limit: Option<Balance>,
+            input_data: Vec<u8>,
+        ) -> ContractExecResult<Balance> {
+            Contracts::bare_call(origin, dest, value, gas_limit, storage_deposit_limit, input_data, true)
```
pmikolajczyk41, line 943: I would extract this enigmatic `true`s into some constant, like `DEBUG_CONTRACTS`

bin/runtime/src/lib.rs
h4nsu: What's the difference? ^^'

pmikolajczyk41: in the built docs: either some comment or none :P

bin/node/src/rpc.rs
```diff
@@ -52,8 +54,10 @@ where
     )));
 
     io.extend_with(TransactionPaymentApi::to_delegate(TransactionPayment::new(
-        client,
+        client.clone(),
     )));
 
+    io.extend_with(ContractsApi::to_delegate(Contracts::new(client)));
```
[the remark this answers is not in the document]
h4nsu: This difference was introduced to their repo literally 3 hours ago and I think they just changed the underlying JSONRPC library. Their previous version looked exactly like the one in this PR.

bin/runtime/src/lib.rs
h4nsu: This comment is kinda internal and meant to be removed soon. Let's leave `//`

bin/runtime/src/lib.rs
h4nsu: Done

bin/runtime/Cargo.toml
```diff
@@ -45,14 +50,14 @@ sp-core = { default-features = false, git = "https://github.com/paritytech/subst
 sp-inherents = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.19"}
 sp-offchain = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.19"}
 sp-runtime = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.19"}
+sp-sandbox = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.19"}
```
fixxxedpoint, line 53: Do we need this here for some reason? Definitely the .wasm output differs without it.

h4nsu: I have no idea how this ended up here ^^ Removed

pmikolajczyk41 (approved):
looks okay for me - I guess we will have to test it now extensively and review the configuration before deploying to testnet

fixxxedpoint (approved):
lgtm

PR #423, opened 2022-05-16
A0-932 Make e2e/devnet/local run use the same chain spec

Marcin-Radecki:
# Description

## Type of change

# Checklist:

### DIFF
--- a/.github/actions/run-e2e-test/action.yml
+++ b/.github/actions/run-e2e-test/action.yml
@@ -40,7 +40,7 @@ runs:
       run: ./.github/scripts/run_e2e_test.sh -t ${{ inputs.test-case }}
 
     - name: Run finalization e2e test
-      if: ${{ inputs.finalization-check == true }}
+      if: ${{ inputs.follow-up-finalization-check == true }}
       shell: bash
       run: ./.github/scripts/run_e2e_test.sh -t finalization
 
--- a/.github/actions/run-e2e-test/action.yml
+++ b/.github/actions/run-e2e-test/action.yml
@@ -39,6 +39,11 @@ runs:
       shell: bash
       run: ./.github/scripts/run_e2e_test.sh -t ${{ inputs.test-case }}
 
+    - name: Run finalization e2e test
+      if: ${{ inputs.finalization-check == true }}
+      shell: bash
+      run: ./.github/scripts/run_e2e_test.sh -t finalization
+
     - name: Print debug if failed
       if: ${{ failure() }}
       shell: bash
--- a/.github/actions/run-e2e-test/action.yml
+++ b/.github/actions/run-e2e-test/action.yml
@@ -39,11 +39,6 @@ runs:
       shell: bash
       run: ./.github/scripts/run_e2e_test.sh -t ${{ inputs.test-case }}
 
-    - name: Run finalization e2e test
-      if: ${{ inputs.follow-up-finalization-check == true }}
-      shell: bash
-      run: ./.github/scripts/run_e2e_test.sh -t finalization
-
     - name: Print debug if failed
       if: ${{ failure() }}
       shell: bash
--- a/.github/scripts/run_consensus.sh
+++ b/.github/scripts/run_consensus.sh
@@ -7,11 +7,8 @@ mkdir -p docker/data/
 # source account ids
 source docker/env
 
-# Generate chainspec and populate comittee keystores
-# hidden assumption: --chain-id must be present and different from a0dnet1, otherwise default chainspec
-# would generate some non-empty validators, and we want to start from the scratch
 docker run -v $(pwd)/docker/data:/data --entrypoint "/bin/sh" -e DAMIAN -e TOMASZ -e ZBYSZKO -e HANSU -e JULIA -e RUST_LOG=info \
- aleph-node:latest -c "aleph-node bootstrap-chain --base-path /data --chain-id a0dnet1-e2e --account-ids $DAMIAN,$TOMASZ,$ZBYSZKO,$HANSU,$JULIA --sudo-account-id $DAMIAN > /data/chainspec.json"
+ aleph-node:latest -c "aleph-node bootstrap-chain --base-path /data --account-ids $DAMIAN,$TOMASZ,$ZBYSZKO,$HANSU,$JULIA --sudo-account-id $DAMIAN > /data/chainspec.json"
 
 # get bootnote peer id
 export BOOTNODE_PEER_ID=$(docker run -v $(pwd)/docker/data:/data --entrypoint "/bin/sh" -e DAMIAN -e RUST_LOG=info aleph-node:latest -c "aleph-node key inspect-node-key --file /data/$DAMIAN/p2p_secret")
--- a/.github/scripts/run_consensus.sh
+++ b/.github/scripts/run_consensus.sh
@@ -7,6 +7,7 @@ mkdir -p docker/data/
 # source account ids
 source docker/env
 
+# generate chain spec and keystores based on authorites passed via --account-ids
 docker run -v $(pwd)/docker/data:/data --entrypoint "/bin/sh" -e DAMIAN -e TOMASZ -e ZBYSZKO -e HANSU -e JULIA -e RUST_LOG=info \
  aleph-node:latest -c "aleph-node bootstrap-chain --base-path /data --account-ids $DAMIAN,$TOMASZ,$ZBYSZKO,$HANSU,$JULIA --sudo-account-id $DAMIAN > /data/chainspec.json"
 
--- a/bin/node/src/chain_spec.rs
+++ b/bin/node/src/chain_spec.rs
@@ -207,22 +207,14 @@ pub fn local_config(
     chain_params: ChainParams,
     authorities: Vec<AuthorityKeys>,
 ) -> Result<ChainSpec, String> {
-    let authority_accounts: Vec<AccountId> = to_account_ids(&authorities).collect();
-    let controller_accounts: Vec<AccountId> =
-        authority_accounts
-            .into_iter()
-            .enumerate()
-            .map(|(index, _account)| {
-                get_account_id_from_seed::<sr25519::Public>(
-                    format!("//{}//Controller", index).as_str(),
-                )
-            })
-            .collect();
-    generate_chain_spec_config(
-        chain_params,
-        authorities,
-        controller_accounts,
-    )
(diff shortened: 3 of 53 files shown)


.github/scripts/run_consensus.sh
```diff
@@ -7,11 +7,8 @@ mkdir -p docker/data/
 # source account ids
 source docker/env
 
-# Generate chainspec and populate comittee keystores
```
pmikolajczyk41, line 10: I would leave a word or two telling what these two long lines do. 

Sorry for my bias, but I am very lame at bash and I strongly prefer scripts accompanied with human explanation :man_shrugging:

.github/scripts/run_consensus.sh
```diff
@@ -7,11 +7,8 @@ mkdir -p docker/data/
...
-# hidden assumption: --chain-id must be present and different from a0dnet1, otherwise default chainspec
```
pmikolajczyk41, line 11: for now, I would say that after removing the flag, we start expecting `chain-id` set in other way -- what do I miss? how is it set now?

bin/node/src/chain_spec.rs
```diff
@@ -201,38 +201,26 @@ fn system_properties(token_symbol: String) -> serde_json::map::Map<String, Value
...
-    chain_params: ChainParams,
-    authorities: Vec<AuthorityKeys>,
-) -> Result<ChainSpec, String> {
-    generate_chain_spec_config(chain_params, authorities, None, None)
-}
-
 /// Generate chain spec for local runs.
-/// Stash and controller accounts are generated for the specified authorities.
+/// Controller accounts are generated for the specified authorities.
 pub fn local_config(
```
pmikolajczyk41, line 206: now, when `config` is gone, I would rename `local_config` to just `config`, since as I understand, this is used in _all_ environments, not only local setting, right?

bin/node/src/chain_spec.rs
```diff
@@ -201,38 +201,26 @@ fn system_properties(token_symbol: String) -> serde_json::map::Map<String, Value
...
-        chain_params,
-        authorities,
-        Some(controller_accounts),
-        Some(stash_accounts),
-    )
+    let controller_accounts: Vec<AccountId> = to_account_ids(&authorities)
+        .into_iter()
+        .enumerate()
+        .map(|(index, _account)| {
+            get_account_id_from_seed::<sr25519::Public>(format!("//{}//Controller", index).as_str())
```
pmikolajczyk41, line 214: out of curiosity (I have never worked with this code before):
 - why do we use different derivation method than in `aleph-client`? I mean `get_account_id_from_seed`. From what I can see, here we hide under some traits, apparently trying to be generic, but actually this is the same code as in `aleph-client`. Maybe we should stick to one method?
 - why don't we make a use of `_account` for controller (previously also stash) accounts? is this for convenience? seems suspicious

bin/node/src/chain_spec.rs
```diff
@@ -322,12 +283,11 @@ struct AccountsConfig {
 }
 
 /// Provides accounts for GenesisConfig setup based on distinct staking accounts.
-/// Assumes validator, controller and stash are all separate accounts.
-fn configure_distinct_staking_accounts(
+/// Assumes validator == stash, but controller is a distinct account
```
pmikolajczyk41, line 286: oh, that's a twist for me - I would definitely put this assumption to the chainspec-generating command documentation in order to make sure that user is aware of this

bin/node/src/chain_spec.rs
```diff
@@ -337,21 +297,14 @@ fn configure_distinct_staking_accounts(
                 .into_iter()
                 .map(|account| (account, TOKEN)),
         )
-        .chain(
-            stashes
```
pmikolajczyk41, line 341: shouldn't we endow `authorities` with this `MIN_VALIDATOR_BOND` amount then?

bin/node/src/chain_spec.rs
```diff
@@ -337,21 +297,14 @@ fn configure_distinct_staking_accounts(
...
     let keys = authorities
         .iter()
-        .zip(stashes.clone())
-        .map(|(auth, stash)| {
+        .map(|auth| {
             (
-                stash.clone(),
-                stash,
+                auth.account_id.clone(),
+                auth.account_id.clone(),
```
pmikolajczyk41, line 307: do we have to clone this two times?

pmikolajczyk41 (changes requested):
I've left some comments, but I have also some more general reflexions:
 - This PR contains two very different things: `follow-up-finalization-check` for GH e2e tests and the chainspec-related; I would split this into two independent PRs
 - I'm bad at chainspec topics so I am not so sure: now, no matter in what environment, the chainspec generated from the node commands will be the same (we do not pass some flag for `chain-id` etc), so actually this PR is: "Make _all runs_ use the same chain spec", isn't it? For production environments we would have to create it manually, yes?
 - I do not get it why do we give up stash accounts :(

I guess you can answer some of these questions in PR description, which I would love to see less template :P

bin/node/src/chain_spec.rs
maciejzelaszczyk: Actually, this seems to be a pretty important point. @Marcin-Radecki reworked this from my previous commit which used this derivation, which in turn came from convenience functions for chain spec generation. Right now, with more knowledge about e2e tests, it seems that using `get_account_id_from_seed` will probably bite us. You may correct me if I am wrong, but this will result in different keys than using `keypair_from_string` from `aleph-client`. So, I would propose either to go with something based on `kaypair_from_string` or something which gives us the same keys we produce in the bench performance test (e.g. https://github.com/Cardinal-Cryptography/aleph-node/blob/3f72b3fe25ea433cdaba954f8d1c0f0c2f5236a8/benches/payout-stakers/src/main.rs#L73).

maciejzelaszczyk: Might go slightly further than that - we probably want to make sure that all the key pairs and accounts based on them are consistent with `alpeh-client` and `payout-stakers`, not only the lines @pmikolajczyk41 initially mentioned.

.github/actions/run-e2e-test/action.yml
```diff
@@ -39,6 +39,11 @@ runs:
       shell: bash
       run: ./.github/scripts/run_e2e_test.sh -t ${{ inputs.test-case }}
 
+    - name: Run finalization e2e test
+      if: ${{ inputs.follow-up-finalization-check == true }}
```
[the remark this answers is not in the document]
Marcin-Radecki: Will add - I used PR as a playground for testing that feature.

btw I'll split to other PR as suggested

.github/scripts/run_consensus.sh
Marcin-Radecki: `--chain-id` has a default value `a0dnet1`
https://github.com/Cardinal-Cryptography/aleph-node/blob/main/bin/node/src/chain_spec.rs#L112

we use that assumption here now and in `run_nodes.sh`, ie we don't pass `--chain-id` explictely. This was a technical debt introduced in the very first work on staking, which this PR tries to address finally.

bin/node/src/chain_spec.rs
Marcin-Radecki: Good point

bin/node/src/chain_spec.rs
Marcin-Radecki: Good points, let's use `keypair_from_string` everywhere we use now `get_account_id_from_seed`

Marcin-Radecki: @pmikolajczyk41  On the other hand, this implies adding `aleph-client` dependency to `aleph-node`. Not sure if it's worth it, to decrease one-liner duplication we'd need to build 395 packages more etc. There's another argument, this time about architecture - `aleph-client` was intended to use with non-production code crates. One can argue this does not make any sense since every code is a production code, test code or not, however, in practice, only test-code crates use `aleph-client` for now. So I'll stick to `get_account_id_from_seed`  for now.

Marcin-Radecki: @maciejzelaszczyk how's exactly get_account_id_from_seed going to give us different keys than keypair_from_string ?

bin/node/src/chain_spec.rs
Marcin-Radecki: Let me add some docs about it

bin/node/src/chain_spec.rs
pmikolajczyk41: nonono, no way of depending on aleph-client; however, iirc `keypair_from_string` is a very short code which does almost the same as the snippet here - I would just copy it to remove some unnecessary abstractions from here

bin/node/src/chain_spec.rs
Marcin-Radecki: We don't need to, since in local runs validator == stash, and validators are already endowed

bin/node/src/chain_spec.rs
Marcin-Radecki: We're iterating over shared reference (`iter()`) and later on we're using it in line 316 so we can't pass ownership here

bin/node/src/chain_spec.rs
Marcin-Radecki: Okay that's better idea, let me do it

bin/node/src/chain_spec.rs
pmikolajczyk41: okay, you have also put it in the ADR so it's okay with me now

pmikolajczyk41 (approved):

bin/node/src/chain_spec.rs
```diff
@@ -360,66 +304,37 @@ fn configure_distinct_staking_accounts(
...
         members,
         keys,
-        stakers: vec![],
+        stakers,
     }
 }
 
 /// Configure initial storage state for FRAME modules.
-/// Possible to provide distinct controller and stash accounts.
+/// Possible to provide distinct controller accounts.
```
maciejzelaszczyk, line 331: Controller accounts are now required.

maciejzelaszczyk (changes requested):
One nit regarding the doc comment. Apart from that looks good. Speedy job.

bin/node/src/chain_spec.rs
Marcin-Radecki: thanks, I've removed comment

maciejzelaszczyk (approved):

PR #429, opened 2022-05-17
Add new testnet chainspec

piotrMocz:
# Description

Add the new testnet chainspec to our repo so that it doesn't get lost in the abyss that is our Slack.


h4nsu (approved):

maciejnems (approved):
Archivist with this chainspec works with testnet, so looks good

PR #435, opened 2022-05-20
A0-888: Request justification for blocks that must have been finalized

DamianStraszak:
# Description

In case that after restart of many validator nodes, some of them did not have the most recent justification they would get stuck on proposals because of their parents being not finalized. This is fixed by requesting justifications in such a case. I also reduced some delays to make this kick in faster. 

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# Checklist:

### DIFF
--- a/finality-aleph/src/data_io/data_store.rs
+++ b/finality-aleph/src/data_io/data_store.rs
@@ -107,6 +107,8 @@ pub struct DataStoreConfig {
     pub max_messages_pending: usize,
     pub available_proposals_cache_capacity: usize,
     pub periodic_maintenance_interval: Duration,
+    // Specifies how much time must pass from receiving a given proposal for the first time, till we
+    // perform a request for either a block or a justification required to let this proposal through.
     pub request_block_after: Duration,
 }
 
@@ -316,16 +318,22 @@ where
                                 self.chain_info_provider.get_parent_hash(&bottom_block)
                             {
                                 let parent_num = bottom_block.num - NumberFor::<B>::one();
-                                if self
-                                    .chain_info_provider
-                                    .get_finalized_at(parent_num)
-                                    .is_err()
+                                if let Ok(finalized_block) =
+                                    self.chain_info_provider.get_finalized_at(parent_num)
                                 {
-                                    debug!(target: "aleph-data-store", "Requesting a justification for block {:?} {:?} after it has been missing for {:?} secs.", parent_num, parent_hash, time_waiting.as_secs());
+                                    if parent_hash != finalized_block.hash {
+                                        warn!(target: "aleph-data-store", "The proposal {:?} is pending because the parent: \
+                                        {:?}, does not agree with the block finalized at this height: {:?}.", proposal, parent_hash, finalized_block);
+                                    } else {
+                                        warn!(target: "aleph-data-store", "The proposal {:?} is pending even though blocks \
+                                            have been imported and parent was finalized.");
+                                    }
+
+                                } else {
+                                    debug!(target: "aleph-data-store", "Requesting a justification for block {:?} {:?} \
+                                        after it has been missing for {:?} secs.", parent_num, parent_hash, time_waiting.as_secs());
                                     self.block_requester
                                         .request_justification(&parent_hash, parent_num);
-                                } else {
-                                    warn!(target: "aleph-data-store", "The proposal {:?} is pending even though blocks were imported and parent was finalized", proposal);
                                 }
                             } else {
                                 warn!(target: "aleph-data-store", "Expected the block below the proposal {:?} to be imported", proposal);
@@ -629,7 +637,8 @@ where
                     {
                         proposals.push(proposal);
                     } else {
-                        warn!(target: "aleph-data-store", "Message {:?} dropped as it contains proposal {:?} not within bounds.", message, unvalidated_proposal);
+                        warn!(target: "aleph-data-store", "Message {:?} dropped as it contains \
+                            proposal {:?} not within bounds.", message, unvalidated_proposal);
                         return;
                     }
                 }
--- a/finality-aleph/src/data_io/data_store.rs
+++ b/finality-aleph/src/data_io/data_store.rs
@@ -117,8 +117,8 @@ impl Default for DataStoreConfig {
             max_proposals_pending: 80_000,
             max_messages_pending: 40_000,
             available_proposals_cache_capacity: 8000,
-            periodic_maintenance_interval: Duration::from_secs(60),
-            request_block_after: Duration::from_secs(100),
+            periodic_maintenance_interval: Duration::from_secs(25),
+            request_block_after: Duration::from_secs(20),
         }
     }
 }
@@ -309,6 +309,27 @@ where
                             debug!(target: "aleph-data-store", "Requesting a stale block {:?} after it has been missing for {:?} secs.", block, time_waiting.as_secs());
                             self.block_requester
                                 .request_stale_block(block.hash, block.num);
+                        } else {
+                            // The whole branch has been imported, what's holding us must be that the parent of the base is not finalized
+                            let bottom_block = proposal.bottom_block();
+                            if let Ok(parent_hash) =
+                                self.chain_info_provider.get_parent_hash(&bottom_block)
+                            {
+                                let parent_num = bottom_block.num - NumberFor::<B>::one();
+                                if self
+                                    .chain_info_provider
+                                    .get_finalized_at(parent_num)
+                                    .is_err()
+                                {
+                                    debug!(target: "aleph-data-store", "Requesting a justification for block {:?} {:?} after it has been missing for {:?} secs.", parent_num, parent_hash, time_waiting.as_secs());
+                                    self.block_requester
+                                        .request_justification(&parent_hash, parent_num);
+                                } else {
+                                    warn!(target: "aleph-data-store", "The proposal {:?} is pending even though blocks were imported and parent was finalized", proposal);
+                                }
+                            } else {
+                                warn!(target: "aleph-data-store", "Expected the block below the proposal {:?} to be imported", proposal);
+                            }
                         }
                     }
                 }
--- a/finality-aleph/src/data_io/data_store.rs
+++ b/finality-aleph/src/data_io/data_store.rs
@@ -302,45 +302,46 @@ where
 
         let now = time::SystemTime::now();
         for (proposal, first_occurrence) in proposals_with_timestamps {
-            if !self.bump_proposal(&proposal) {
-                // `bump_proposal` returns false if the bump didn't make the proposal available, meaning that it is still pending
-                if let Ok(time_waiting) = now.duration_since(first_occurrence) {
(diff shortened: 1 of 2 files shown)


finality-aleph/src/data_io/data_store.rs
```diff
@@ -309,6 +309,27 @@ where
...
+                                {
+                                    debug!(target: "aleph-data-store", "Requesting a justification for block {:?} {:?} after it has been missing for {:?} secs.", parent_num, parent_hash, time_waiting.as_secs());
+                                    self.block_requester
+                                        .request_justification(&parent_hash, parent_num);
+                                } else {
+                                    warn!(target: "aleph-data-store", "The proposal {:?} is pending even though blocks were imported and parent was finalized", proposal);
+                                }
+                            } else {
+                                warn!(target: "aleph-data-store", "Expected the block below the proposal {:?} to be imported", proposal);
+                            }
```
kostekIV, line 332: this is a comment to the whole for loop section but github doesn't allow commenting on the unchanged lines :/

nit: to reduce nesting:
1.
```rust
            if self.bump_proposal(&proposal) {
                continue;
            }
```

2.
``` rust
            let time_waiting = match now.duration_since(first_occurrence) {
                Ok(tw) if tw >= self.config.request_block_after => tw,
                _ => continue,
            };
```

3.
```rust
            if !self.chain_info_provider.is_block_imported(&block) {
                debug!(target: "aleph-data-store", "Requesting a stale block {:?} after it has been missing for {:?} secs.", block, time_waiting.as_secs());
                self.block_requester
                    .request_stale_block(block.hash, block.num);
                continue;
            }
```

4.
```rust
            let parent_hash = match self.chain_info_provider.get_parent_hash(&bottom_block)
            {
                Ok(ph) => ph,
                _ =>  {
                    warn!(target: "aleph-data-store", "Expected the block below the proposal {:?} to be imported", proposal);
                    continue;
                }
            };
```

1 -> line 303
2 -> lin3 305
3 -> line 308
4 -> line 315

finality-aleph/src/data_io/data_store.rs
```diff
@@ -309,6 +309,27 @@ where
...
+                            // The whole branch has been imported, what's holding us must be that the parent of the base is not finalized
+                            let bottom_block = proposal.bottom_block();
+                            if let Ok(parent_hash) =
+                                self.chain_info_provider.get_parent_hash(&bottom_block)
+                            {
+                                let parent_num = bottom_block.num - NumberFor::<B>::one();
+                                if self
+                                    .chain_info_provider
+                                    .get_finalized_at(parent_num)
+                                    .is_err()
```
kostekIV, line 322: this could be more concise if we would add a method `is_finalized(hash) -> bool` to the chain_info_provider.

finality-aleph/src/data_io/data_store.rs
```diff
@@ -309,6 +309,27 @@ where
...
+                                {
+                                    debug!(target: "aleph-data-store", "Requesting a justification for block {:?} {:?} after it has been missing for {:?} secs.", parent_num, parent_hash, time_waiting.as_secs());
+                                    self.block_requester
+                                        .request_justification(&parent_hash, parent_num);
+                                } else {
+                                    warn!(target: "aleph-data-store", "The proposal {:?} is pending even though blocks were imported and parent was finalized", proposal);
```
kostekIV, line 328: could it be possible that even though some block at `parent_num` is finalized it is not the parent of the `bottom_block`?

finality-aleph/src/testing/data_store.rs
```diff
@@ -442,6 +448,38 @@ async fn sends_block_request_on_missing_block() {
...
 
+#[tokio::test]
+async fn sends_justification_request_when_not_finalized() {
+    run_test(|mut test_handler| async move {
+        let blocks = test_handler
+            .initialize_single_branch(MAX_DATA_BRANCH_LEN * 10)
+            .await;
+        test_handler.import_branch(blocks.clone()).await;
+
+        let blocks_branch = blocks[2..3].to_vec();
```
pmikolajczyk41, line 459: my feeling is that `vec![blocks[2]]` is less confusing, feel free to ignore

finality-aleph/src/testing/data_store.rs
```diff
@@ -442,6 +448,38 @@ async fn sends_block_request_on_missing_block() {
...
+        let test_data: TestData = vec![aleph_data_from_blocks(blocks_branch)];
+        test_handler.send_data(test_data.clone());
```
pmikolajczyk41, line 461: you don't use `test_data` anywhere else, so there is no need to clone it, yes? also type ascription shouldn't be needed as you are passing it to `send_data` which expects `TestData`

finality-aleph/src/data_io/data_store.rs
```diff
@@ -309,6 +309,27 @@ where
                             debug!(target: "aleph-data-store", "Requesting a stale block {:?} after it has been missing for {:?} secs.", block, time_waiting.as_secs());
```
pmikolajczyk41, line 309: nit: wrap these long lines in `run_maintenance`

finality-aleph/src/data_io/data_store.rs
pmikolajczyk41, line 328: why this is not an error?

pmikolajczyk41: we were requesting parent of `BlockHashNum` so for both height and hash, so I guess we are safe

finality-aleph/src/testing/data_store.rs
```diff
@@ -442,6 +448,38 @@ async fn sends_block_request_on_missing_block() {
     .await;
 }
 
+#[tokio::test]
+async fn sends_justification_request_when_not_finalized() {
```
pmikolajczyk41, line 452: I'm just thinking out loud: we have here multiple scenarios testing different ordering of events; had we checked this particular configuration before, we would be aware of the problem before outage

and hence my random thought: is it possible/feasible to automatically and deftly test all event orderings? or at least increase coverage somehow?

finality-aleph/src/data_io/data_store.rs
kostekIV: But we only checked if some block at height of `parent_num` is finalized

DamianStraszak: This is possible but only in the presence of malicious nodes (follows from the fact that honest nodes always "root" their proposal in a finalized block). In such a case this proposal will not go through and will stay pending, which is OK. There is a pruning mechanism which will get rid of it if we run low on memory. We could attempt removing it right away, but that would add complexity to the code.

finality-aleph/src/data_io/data_store.rs
```diff
@@ -117,8 +117,8 @@ impl Default for DataStoreConfig {
             max_proposals_pending: 80_000,
             max_messages_pending: 40_000,
             available_proposals_cache_capacity: 8000,
-            periodic_maintenance_interval: Duration::from_secs(60),
-            request_block_after: Duration::from_secs(100),
+            periodic_maintenance_interval: Duration::from_secs(25),
+            request_block_after: Duration::from_secs(20),
```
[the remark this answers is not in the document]
DamianStraszak: In every maintenance run we always touch every proposal, but the `request_block_after` determines how much time must pass from the point in time when we first received a given proposal till when we first attempt requesting a block or justification. This means that after we receive a proposal and it stays pending, then either in the first or the second maintenance, something will be requested. I added a comment explaining this parameter.

In general these delays are probably quite large, but we are really stepping on thin ice here, because there is uncertainty on how well the network code handles multiple requests in short periods of time (i.e. whether it will lead to mass banning of nodes because of some random thing).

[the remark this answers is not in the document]
DamianStraszak: And this stuff is virtually impossible to test properly :/

finality-aleph/src/data_io/data_store.rs
DamianStraszak: Ok, done -- you mean something like
```
"bla bla bla \
bla bla bla"
```
right?

finality-aleph/src/data_io/data_store.rs
DamianStraszak: That's correct, but I actually changed this a little to log more accurately, and now `bool` would not be enough. So not adding it :)

finality-aleph/src/data_io/data_store.rs
DamianStraszak: I've added more precise logging to make it possible to see in case this happens.

DamianStraszak: There are two cases here now, one can occur due to malicious behavior of some node (when the finalized block is different than expected) -- so it's just `warn`. The second case I believe might happen due to some super unlucky execution ordering of threads -- so technically not an `error` either.

finality-aleph/src/data_io/data_store.rs
pmikolajczyk41: yeah, or you can turn on (for a while) an unstable option of `rustfmt` for formatting inside macro

finality-aleph/src/testing/data_store.rs
DamianStraszak: Fine, although it must be `vec![blocks[2].clone()]` for it to work.

finality-aleph/src/data_io/data_store.rs
DamianStraszak: Addressed -- everything super-flat now. Like Earth.

finality-aleph/src/data_io/data_store.rs
DamianStraszak: Ok, went for the manual slicing.

finality-aleph/src/data_io/data_store.rs
```diff
@@ -300,18 +302,46 @@ where
...
+
+            let block = proposal.top_block();
+            if !self.chain_info_provider.is_block_imported(&block) {
+                debug!(target: "aleph-data-store", "Requesting a stale block {:?} after it has been missing for {:?} secs.", block, time_waiting.as_secs());
+                self.block_requester
+                    .request_stale_block(block.hash, block.num);
+                continue;
+            }
+            // The top block (thus the whole branch, in the honest case) has been imported. What's holding us
+            // must be that the parent of the base is not finalized.
```
pmikolajczyk41, line 322: is this possible for an honest scenario?

pmikolajczyk41: oh, it must have been a hopeless fork, right? if so I would encourage you to mention it, otherwise please correct me

pmikolajczyk41 (approved):

finality-aleph/src/data_io/data_store.rs
DamianStraszak: This is either because of a malicious proposal, or because we are not up-to-date with finalization (which is the main thing to deal with here). I've adjusted the comment to reflect that piece of information.

PR #437, opened 2022-05-23
A0 864: Set balance when forking off

pmikolajczyk41:
# Description

This PR introduces two new configuration options to `fork-off` that enable specifying initial account balances (in System.Account map).

## Type of change

# Checklist:


fork-off/README.md
```diff
@@ -31,13 +33,15 @@ The tool will perform the following actions, in this order:
...
-5. Replace the genesis state in the chainspec by the one from the snapshot WITH THE EXCEPTION of states of paths provided via a comma separated list using `--storage_keep_state`. The default setting is `--storage_keep_state=Aura,Aleph,Balances,Sudo,Staking,Session,Elections,System.Account` and it's likely you don't want to change it.
-6. The final, new chainspec is saved to the path provided via `--combined-spec-path`.
+5. Replace the genesis state in the chainspec by the one from the snapshot WITH THE EXCEPTION of states of paths provided via a comma separated list using `--storage_keep_state`. The default setting is `--storage_keep_state=Aura,Aleph,Sudo,Staking,Session,Elections` and it's likely you don't want to change it.
+6. If you have passed `--accounts_path` pointing to a file with a configuration for some accounts, then it will be written into chainspec (`System.Account` map). For an example see `AccountInfo-Template.json` file.
+7. Alternatively to `--accounts_path` you can just pass `--balances` flag with which you can specify initial free balances for some accounts. Similarly, it will be saved to `System.Account` map.
```
Marcin-Radecki, line 38: Can we elaborate on some details on both `--accounts_path` and `--balances` ie what type of account we pass (ie SS58 or hex), and what format exactly have `--balances`. This is to enrich the user experience. :)

fork-off/src/config.rs
```diff
@@ -47,7 +42,54 @@ pub struct Config {
         multiple_occurrences = true,
         takes_value = true,
         value_delimiter = ',',
-        default_value = "Aura,Aleph,Balances,Sudo,Staking,Session,Elections,System.Account"
+        default_value = "Aura,Aleph,Sudo,Staking,Session,Elections"
```
Marcin-Radecki, line 45: I am worried that 
* copying `Balances` from Testnet (`Balances.locks` in particular) might cause local chainspec config to be overwritten from Testnet, and Testnet has different validators than Devnet, which will cause Devnet malfunction. On local run we're creating some `locks` just in the genesis block.
* copying `System.Account` from testnet would cause validators on Devnet to have 0 balance, preventing above mentioned feature to work correctly

fork-off/src/config.rs
```diff
@@ -47,7 +42,54 @@ pub struct Config {
...
+        multiple_occurrences(true))
+    ]
+    pub balances: Option<Vec<(AccountId, Balance)>>,
+}
+
+fn parse_balances(s: &str) -> Result<(AccountId, Balance), Box<dyn Error + Send + Sync + 'static>> {
+    let sep_pos = s.find('=').ok_or("Invalid ACCOUNT=BALANCE: no `=` found")?;
+
+    let account_raw: String = s[..sep_pos].parse()?;
+    let account = AccountId::new(&account_raw);
```
Marcin-Radecki, line 65: This accepts hex-based `AccountId`, is it not the case for end-users who mostly use the `SS58` format? Just wondering what would be most convenient for them to use.

fork-off/src/config.rs
```diff
@@ -47,7 +42,54 @@ pub struct Config {
...
+    let balance = s[sep_pos + 1..].parse()?;
+    Ok((account, balance))
+}
+
+impl Display for Config {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
```
Marcin-Radecki, line 71: Does not `clap` provide some default debug formatter? Can we display `Config` with `{:?}`?

fork-off/src/main.rs
```diff
@@ -1,29 +1,30 @@
...
-type StorageKey = String;
-type StorageValue = String;
-type Storage = HashMap<StorageKey, StorageValue>;
+mod types;
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
+    env_logger::Builder::from_env(Env::default().default_filter_or("info")).init();
+    let config = Config::parse();
+    info!(target: "fork", "{}", config);
```
Marcin-Radecki, line 26: Maybe `target` should be `fork-off`?

fork-off/src/account_setting.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+pub fn apply_account_setting(mut state: Storage, setting: AccountSetting) -> Storage {
+    let account_map: StorageKey = get_account_map().into();
+    for (account, info) in setting {
+        let account_hash = account.clone().into();
+        let key = &account_map.join(&account_hash);
+
+        state.insert(key.clone(), info.clone().into());
+        info!("Account info of `{:?}` set to `{:?}`", account, info);
```
Marcin-Radecki, line 63: Please add `target`

fork-off/Cargo.toml
```diff
@@ -1,18 +1,16 @@
 [package]
 name = "fork-off"
-version = "1.0.1"
+version = "1.2.0"
 edition = "2021"
 
 [dependencies]
-# client
-substrate-api-client = { git = "https://github.com/Cardinal-Cryptography/substrate-api-client.git", branch = "polkadot-v0.9.19" }
```
Marcin-Radecki, line 8: While I agree that dependency on `substrate-api-client` is not ideal, but I believe we can w/a this by depending on some core frames from Substrate to access common storage-releated code, see other comments in this PR

Marcin-Radecki (changes requested):
First-round to get on the same page with overall code architecture.

fork-off/src/types.rs
```diff
@@ -0,0 +1,173 @@
+//! This module defines common types, in particular those that are needed for storage navigation.
```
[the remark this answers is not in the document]
pmikolajczyk41: no, actually `substrate-api-client` wasn't used at all and was useless dependency

these internal types are only for easier navigation through chainspec and they serve as type-safety mechanism and type-backed semantics

fork-off/src/account_setting.rs
```diff
@@ -0,0 +1,66 @@
+use std::{collections::HashMap, str::FromStr};
+
+use crate::Storage;
+use codec::Encode;
+use log::info;
+use serde::Deserialize;
+
+use crate::types::{AccountId, Balance, StorageKey, StoragePath, StorageValue};
+
+/// This struct is copied and type-specialized from `pallet_balances::AccountData`.
```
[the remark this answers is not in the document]
pmikolajczyk41: it cost wrapping and implementing `serde::Deserialize`, but I changed it according to the suggestion

fork-off/src/types.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// Check whether `other` is prefix of `self`.
+    pub fn is_prefix_of(&self, other: &StorageKey) -> bool {
+        // We have to ensure that both items are in the same form.
+        let shorter = as_hex(&self.0);
+        let longer = as_hex(&other.0);
+        longer.starts_with(&shorter)
+    }
+}
+
+/// Copied from `frame_support`.
```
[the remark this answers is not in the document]
pmikolajczyk41: yeap, changed

fork-off/src/config.rs
pmikolajczyk41: uhm, I was struggling with encoding account to the correct hash for a long time and for now it is the only way I can do this;

you can always use `subkey` to see this hex-based public key

fork-off/Cargo.toml
pmikolajczyk41: not relevant I believe

fork-off/src/account_setting.rs
Marcin-Radecki: We can add  `target` here since we're adding/modifying this line anyway,  but we don't need to modify all `log!` lines in this PR.

pmikolajczyk41: okay, done

fork-off/src/config.rs
Marcin-Radecki: I think leaving `System.Account` and `Balances` won't hurt - we're copying a lot from Testnet anyway. On the other hand we might get rid of `fork-off` so maybe we should not invest much more time into it. We can pass palelts via GH arguments so we're good.

Marcin-Radecki (approved):

fork-off/src/config.rs
fbielejec: You could try:

- trim 0x
- let bytes = hex::decode_to_slice(trimmed)
- AccountId::new(bytes)

There is also : [AccountId::from_string](https://github.com/paritytech/substrate/blob/polkadot-v0.9.19/primitives/core/src/crypto.rs#L420)

fbielejec (commented):
can you give passing SS58 pubkeys one more try pls (see comment :P)

fbielejec (approved):

PR #438, opened 2022-05-23
A0-927 Make pallet_staking::Invulnerables empty. Store reserved node list in pallet elections

kostekIV:
# Description
Add storage for storing reserved members.

## Type of change

# Checklist:


bin/node/src/chain_spec.rs
```diff
@@ -361,6 +361,8 @@ fn generate_genesis_config(
     let accounts_config =
         configure_chain_spec_fields(unique_accounts_balances, authorities, controller_accounts);
 
+    let reserved = accounts_config.members[0..2].to_vec();
```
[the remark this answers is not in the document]
pmikolajczyk41: okay, I see below that it was already this way, but still, I think it won't hurt

pmikolajczyk41 (changes requested):
generally seems useful and okay, but I do not know these areas nor the context :(

however, I would insist on adding a corresponding testcase to the pallet elections

bin/node/src/chain_spec.rs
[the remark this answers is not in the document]
Marcin-Radecki: The thing is for local runs, we need some positive value in order to get that feature working. Other solution would be to set it to 0, but then anyone to get committee rotated would need to issue `change_reserved_members`. When I'm thinking it now I think this is better @kostekIV . This is because we want to be as close as possible to Testnet and Mainnet deployments, and those deployments would require such manual steps anyway.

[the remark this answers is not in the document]
kostekIV: That makes sense, but would also make e2e tests a little bit harder, with a few more setup steps:
* issue change_reserved_members
* wait an era
* actual test

pmikolajczyk41:
aaand change PR title pls

bin/node/src/chain_spec.rs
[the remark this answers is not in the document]
Marcin-Radecki: I think it's the price we can pay for avoiding mistakes during releases.

pallets/elections/src/lib.rs
```diff
@@ -59,6 +59,9 @@ pub mod pallet {
     #[pallet::storage]
     pub type MembersPerSession<T> = StorageValue<_, u32, ValueQuery>;
```
Marcin-Radecki, line 60: Technically now we should set it to 0 as default, and set to any given value in e2e test to be aligned with approach for reserved.

Marcin-Radecki: But the problem is local runs with `run_nodes.sh` would end up in empty committee produced each new session, so this is not that easy I guess. So one would need to run `run_nodes.sh` and set `MembersPerSession` manually via extrinsics, which is not good

bin/node/src/chain_spec.rs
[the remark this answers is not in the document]
Marcin-Radecki: @pmikolajczyk41  feel free to resolve, from my end it looks okay

pallets/elections/src/lib.rs
Marcin-Radecki: So we can discuss offline or I am happy for you to resolve this :)

kostekIV: I would leave it as it is, since it would make local runs pain for all of us :P

Marcin-Radecki (approved):

pallets/elections/src/lib.rs
```diff
@@ -59,6 +59,9 @@ pub mod pallet {
...
 
+    #[pallet::storage]
+    pub type ReservedMembers<T: Config> = StorageValue<_, Vec<T::AccountId>, ValueQuery>;
```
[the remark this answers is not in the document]
kostekIV: Added a small doc with terminology and storages. `Members` are confusing because actually are used in multiple contexts. With the upcoming cleanup you proposed this will go away (for now I left the todo mark)

aleph-client/src/session.rs
```diff
@@ -58,6 +58,27 @@ pub fn change_members(
...
+        new_members
+    );
+    let xt = compose_extrinsic!(
+        sudo_connection.as_connection(),
+        "Sudo",
+        "sudo_unchecked_weight",
+        call,
+        0_u64
+    );
+    send_xt(sudo_connection, xt, Some("sudo_unchecked_weight"), status);
```
pmikolajczyk41, line 80: Uh, I would rename this xt to something related to changing reserved members

pallets/elections/src/lib.rs
```diff
@@ -1,6 +1,22 @@
...
+//! ### Terminology
+//! For definition of session, era, staking see pallet_session and pallet_staking.
+//! - Committee: Set of nodes that produce and finalize blocks in the era.
+//! - Validator: Node that can become a member of committee (or already is) via rotation.
+//! - (TODO: remove this to remove confusion) Member: Usually same as validator, sometimes means member of the committee
+//! - ReservedMembers: Validators that are chosen to be in committee every single session.
+//!
+//! ### Storage
+//! - `Members` - List of possible validators.
+//! - `MembersPerSession` - Number of committee size.
```
pmikolajczyk41, line 14: just committee size

pallets/elections/src/lib.rs
```diff
@@ -1,6 +1,22 @@
...
+//! - `ReservedMembers` - List of reserved nodes.
+//! - `ErasReserved` - List of reserved nodes for the current era.
```
pmikolajczyk41, line 16: does it mean that reserved members are (potentially) changed each era?

pallets/elections/src/lib.rs
```diff
@@ -87,11 +106,23 @@ pub mod pallet {
 
             Ok(())
         }
+
+        #[pallet::weight((T::BlockWeights::get().max_block, DispatchClass::Operational))]
+        pub fn change_reserved_members(
```
pmikolajczyk41, line 111: tbh I was thinking about pallet tests likewise `pallet_aleph` has

pallets/elections/src/lib.rs
kostekIV: yep

pallets/elections/src/lib.rs
kostekIV: Tests will be added but first, we need to decouple some dependencies from this pallet to make it easier. Definitely not in this pr

pallets/elections/src/lib.rs
kostekIV: done

aleph-client/src/session.rs
kostekIV: isn't the whole function called `change_reserved_members`?

pmikolajczyk41: yeah, but in logs you will see only `"Transaction sudo_unchecked_weight was included in block with hash {}."`, which I think is less useful that  `"Transaction change_reserved_members was included in block with hash {}."`

pallets/elections/src/lib.rs
pmikolajczyk41: fair enough, is there any Jira for this?

pallets/elections/src/lib.rs
pmikolajczyk41: I would add a word about this then

pmikolajczyk41 (approved):

pallets/elections/src/lib.rs
kostekIV: the line 17 is not enough?

pallets/elections/src/lib.rs
kostekIV: added as 981

PR #442, opened 2022-05-25
A0-984 Refactor e2e test to used unified validators accounts with local

Marcin-Radecki:
runs.

# Description

## Type of change

# Checklist:

### DIFF
--- a/.github/scripts/run_consensus.sh
+++ b/.github/scripts/run_consensus.sh
@@ -2,21 +2,30 @@
 
 set -e
 
+# change it when increasing number of node containers
+export NODE_COUNT=5
+
 mkdir -p docker/data/
 
-# source account ids
-source docker/env
+echo "Generating accounts ids..."
+declare -a account_ids
+for node_index in $(seq 0 $((NODE_COUNT - 1))); do
+  account_ids+=($(docker run -v $(pwd)/docker/data:/data --entrypoint "/bin/sh" -e node_index aleph-node:latest -c "aleph-node key inspect //$node_index | grep \"SS58 Address:\" | awk \"{print \\\$3;}\""))
+done
+# space separated ids
+validator_ids="${account_ids[*]}"
+# comma separated ids
+validator_ids="${validator_ids//${IFS:0:1}/,}"
 
-# generate chain spec and keystores based on authorites passed via --account-ids
-docker run -v $(pwd)/docker/data:/data --entrypoint "/bin/sh" -e DAMIAN -e TOMASZ -e ZBYSZKO -e HANSU -e JULIA -e RUST_LOG=info \
- aleph-node:latest -c "aleph-node bootstrap-chain --base-path /data --account-ids $DAMIAN,$TOMASZ,$ZBYSZKO,$HANSU,$JULIA --sudo-account-id $DAMIAN > /data/chainspec.json"
+echo "Generate chainspec and keystores..."
+docker run -v $(pwd)/docker/data:/data --entrypoint "/bin/sh" -e RUST_LOG=info -e validator_ids aleph-node:latest -c \
+"aleph-node bootstrap-chain --base-path /data --account-ids $validator_ids  > /data/chainspec.json"
 
-# get bootnote peer id
-export BOOTNODE_PEER_ID=$(docker run -v $(pwd)/docker/data:/data --entrypoint "/bin/sh" -e DAMIAN -e RUST_LOG=info aleph-node:latest -c "aleph-node key inspect-node-key --file /data/$DAMIAN/p2p_secret")
+echo "Genereting bootnode peer id..."
+bootnote_account=${account_ids[0]}
+export BOOTNODE_PEER_ID=$(docker run -v $(pwd)/docker/data:/data --entrypoint "/bin/sh" -e bootnote_account -e RUST_LOG=info aleph-node:latest -c "aleph-node key inspect-node-key --file /data/$bootnote_account/p2p_secret")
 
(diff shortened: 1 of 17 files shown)


.github/scripts/run_consensus.sh
```diff
@@ -2,21 +2,30 @@
...
-export BOOTNODE_PEER_ID=$(docker run -v $(pwd)/docker/data:/data --entrypoint "/bin/sh" -e DAMIAN -e RUST_LOG=info aleph-node:latest -c "aleph-node key inspect-node-key --file /data/$DAMIAN/p2p_secret")
-
-echo "BOOTNODE_PEER_ID : $BOOTNODE_PEER_ID"
+mkdir -p docker/data/
 
-# Run consensus party
+echo "Generating accounts ids..."
+declare -a account_ids
+for node_index in $(seq 0 $((NODE_COUNT - 1))); do
+  account_ids+=($(docker run -v $(pwd)/docker/data:/data --entrypoint "/bin/sh" -e node_index aleph-node:latest -c "aleph-node key inspect //$node_index | grep \"SS58 Address:\" | awk \"{print \\\$3;}\""))
```
fixxxedpoint, line 13: Add `--rm` to all calls of `docker run ...`

fixxxedpoint, line 13: Are these `-e ...` for variables needed? Bash should substitute these values (works on my machine :))

.github/scripts/run_consensus.sh
```diff
@@ -2,21 +2,30 @@
 
 set -e
```
fixxxedpoint, line 4: It looks good for me. Would be also a good idea to make this script little bit more verbose, e.g. describe in a comment (or output to terminal?) that ids are generated from `//id` now and also what is the `sudo` account (use it somewhere here like in the previous version?) instead of making someone to read the whole script and looking for a default config value for `sudo`. In general, it would nice to see these components little bit more coupled together, i.e. instead of hard-coding and changing from time to time here and there a method by which those ids are derived, just use some configuration method (env variable, input arguments, whatever else).

Marcin-Radecki: Good point, done

.github/scripts/run_consensus.sh
Marcin-Radecki: Done

Marcin-Radecki: Done

docker/common.yml
```diff
@@ -0,0 +1,26 @@
...
+    environment:
+      - CUSTOM_ARGS=-lpallet_aleph=debug,runtime=debug
+      # key derived from "//0"
+      - BASE_PATH=/data/5D34dL5prEUaGNQtPPZ3yN5Y6BnkfXunKXXz6fo7ZJbLwRRH
+      - NODE_KEY_PATH=/data/5D34dL5prEUaGNQtPPZ3yN5Y6BnkfXunKXXz6fo7ZJbLwRRH/p2p_secret
+
+  AlephNonBootNode:
+    extends: AlephNodeService
+    environment:
+      - BOOT_NODES=/ip4/127.0.0.1/tcp/30333/p2p/$BOOTNODE_PEER_ID
```
fbielejec, line 26: how hard would it be to fix the peer id of the bootnode too? it would simplify the whole setup considerably.
 
Seems like it's just about generating the same libp2p key for that node (the seed could be anything, it's node-name account-id, or something else)

.github/scripts/run_consensus.sh
```diff
@@ -2,21 +2,51 @@
...
-# source account ids
-source docker/env
+function generate_authorities {
+  local authorities_count="$1"
+
+  echo "Generating ${authorities_count} authorities accounts ids..." >&2
+  declare -a account_ids
+  for node_index in $(seq 0 $((authorities_count - 1))); do
+    echo "Generating authority ${node_index} from key //${node_index}" >&2
+    account_ids+=($(docker run --rm -v $(pwd)/docker/data:/data --entrypoint "/bin/sh" aleph-node:latest -c "aleph-node key inspect //$node_index | grep \"SS58 Address:\" | awk \"{print \\\$3;}\""))
```
fbielejec, line 17: could we specify the image (export NODE_IMAGE=aleph-node:latest) as a top level var as well for all commands that call to the dockerized node?

This would make it easier to change it in one place if testing against some other version (say testnet or smartnet version)

docker/docker-compose.yml
```diff
@@ -1,108 +1,65 @@
...
-services:
+# When increasing number of nodes in this file, change default value of --validators-count param in e2e-tests/src/config.rs
 
-  # bootnode
-  damian:
-    image: aleph-node:latest
-    container_name: damian
-    network_mode: host
+services:
+  Node0:
```
fbielejec, line 4: maybe explicitly call it BootNode?

.github/scripts/run_consensus.sh
fixxxedpoint: another solution `docker tag <image_id> aleph-node:latest`

fbielejec: I would forget the next day I have a image tagged differently and spend hours debugging :rofl:

fixxxedpoint (approved):

.github/scripts/run_consensus.sh
Marcin-Radecki: Will do

docker/common.yml
Marcin-Radecki: As discussed offline, we'll keep it as it is for now, to limit PR scope.

fbielejec (approved):

PR #445, opened 2022-05-27
Write migration for pallet elections

kostekIV:
# Description

## Type of change

# Checklist:


pallets/elections/src/tests.rs
```diff
@@ -18,3 +31,44 @@ fn test_elect() {
         assert_eq!(elected.unwrap(), &[(1, supp.clone()), (2, supp)]);
     });
 }
+
+#[test]
+fn migration_from_v0_to_v1_works() {
+    new_test_ext(vec![1, 2, 3], vec![]).execute_with(|| {
```
Marcin-Radecki, line 37: Can we add another test to make sure we discard previous values?

Marcin-Radecki (changes requested):

pallets/elections/src/migrations/v0_to_v1.rs
```diff
@@ -0,0 +1,69 @@
...
+pub fn migrate<T: Config, P: PalletInfoAccess>() -> Weight {
+    log::info!(target: "pallet_elections", "Running migration from STORAGE_VERSION 0 to 1");
+    let members = Members::<T>::get().expect("");
+
+    let members_per_session = members.len();
+    let mut write_count = 0;
+
+    match MembersPerSession::translate(|old| match old {
+        Some(Some(x)) => Some(x),
+        _ => Some(members_per_session as u32),
```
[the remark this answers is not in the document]
DamianStraszak: I'm having trouble understanding this point: are you saying that it should not be `members.len()`? Why is that?

pallets/elections/src/tests.rs
DamianStraszak: I'm having this fundamental problem of not understanding how the test works. So for the test to make sense it would need to migrate from the old version of the pallet to the new one. How does it have access to the old one? The mock seems to have the new one?

pallets/elections/src/lib.rs
```diff
@@ -88,6 +92,28 @@ pub mod pallet {
     #[pallet::without_storage_info]
     pub struct Pallet<T>(_);
 
+    #[pallet::hooks]
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        fn on_runtime_upgrade() -> frame_support::weights::Weight {
+            let on_chain = <Pallet<T> as GetStorageVersion>::on_chain_storage_version();
+            T::DbWeight::get().reads(1)
+                + match on_chain {
```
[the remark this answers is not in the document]
kostekIV: no longer here

pallets/elections/src/impls.rs
```diff
@@ -52,34 +61,34 @@ fn rotate<T: Clone + PartialEq>(
     current_era: EraIndex,
     current_session: SessionIndex,
     n_validators: usize,
-    all_validators: Vec<T>,
     reserved: Vec<T>,
+    non_reserved: Vec<T>,
 ) -> Option<Vec<T>> {
     if current_era == 0 {
```
mike1729, line 67: Maybe return when `non_reserved.is_empty()`, then you don't need `current_era` in this method and in `rotate_members`.

pallets/elections/src/impls.rs
```diff
@@ -208,12 +215,12 @@ where
     }
 
     fn adjust_rewards_for_session() {
-        let active_era = match T::EraInfoProvider::active_era() {
-            Some(ae) if ae > 0 => ae,
+        match T::EraInfoProvider::active_era() {
```
mike1729, line 218: `if T::EraInfoProvider::active_era().unwrap_or(0) == 0 { return }`
or 
`if T::EraInfoProvider::active_era().filter(|e| e>0).is_some() == 0 { return }`

pallets/elections/src/lib.rs
```diff
@@ -112,10 +144,29 @@ pub mod pallet {
     #[pallet::call]
     impl<T: Config> Pallet<T> {
         #[pallet::weight((T::BlockWeights::get().max_block, DispatchClass::Operational))]
-        pub fn change_members(origin: OriginFor<T>, members: Vec<T::AccountId>) -> DispatchResult {
+        pub fn change_members(
```
mike1729, line 147: I would remove all the calls except this one. There are potential storage writes that could be optimized but it's not worth it in my opinion as we would get one call instead of four and we call it very infrequently so the storage write overhead is negligible.

mike1729 (changes requested):

pallets/elections/src/lib.rs
kostekIV: ok, done

pallets/elections/src/impls.rs
kostekIV: first one is the winner

pallets/elections/src/impls.rs
kostekIV: done

e2e-tests/src/test/validators_change.rs
```diff
@@ -11,44 +11,81 @@ use sp_core::Pair;
...
     }
     wait_for_event(
         &connection,
         ("Elections", "ChangeMembers"),
         |e: NewMembersEvent| {
-            info!("[+] NewMembersEvent: members{:?}", e.members);
+            info!("[+] NewMembersEvent: reserved: {:#?}, non_reserved: {:#?}, members_per_session: {:#?}", e.reserved, e.non_reserved, e.non_reserved);
 
-            e.members == new_members
+            e.reserved == new_members[0..2].to_vec()
```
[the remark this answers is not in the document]
kostekIV: fixed

pallets/elections/src/impls.rs
```diff
@@ -49,37 +58,32 @@ fn compute_validator_scaled_total_rewards<V>(validator_totals: Vec<(V, u128)>) -
...
-
     // The validators for the committee at the session `n` are chosen as follow:
     // 1. Reserved validators are always chosen.
     // 2. Given non-reserved list of validators the chosen ones are from the range:
     // `n * free_seats` to `(n + 1) * free_seats` where free_seats is equal to free number of free
     // seats in the committee after reserved nodes are added.
     let free_seats = n_validators.saturating_sub(reserved.len());
+
+    let non_reserved_len = non_reserved.len();
     let first_validator = current_session as usize * free_seats;
```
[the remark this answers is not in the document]
DamianStraszak: With current parameters (900 blocks per session) the overflow will probably never happen, but still, I think the rule of thumb when doing arithmetic in the runtime should be to use saturating operations whenever it makes sense.

pallets/elections/src/migrations/v0_to_v1.rs
```diff
@@ -0,0 +1,55 @@
...
+///
+/// There is one storage in the pallet elections `Members` containing current set of validators.
+/// After migration the state should be as follows:
+/// - `Members` are no longer present.
+/// - `MembersPerSession` is u32 storage set to size of the `Members` set.
+/// - `ReservedMembers` contains the content of the `Members`
+/// - `NonReservedMembers` are empty
+/// - `ErasMembers` contain tuple of (content of `Members`, empty vector).
+pub fn migrate<T: Config, P: PalletInfoAccess>() -> Weight {
+    log::info!(target: "pallet_elections", "Running migration from STORAGE_VERSION 0 to 1");
```
DamianStraszak, line 28: You can also add to the message that this comes from the Election pallet, because the targets are not necessarily shown in logs.

DamianStraszak (changes requested):
Very nice :) I like it a lot.

pallets/elections/src/impls.rs
```diff
@@ -32,9 +32,18 @@ fn calculate_adjusted_session_points(
...
-fn compute_validator_scaled_total_rewards<V>(validator_totals: Vec<(V, u128)>) -> Vec<(V, u32)> {
+pub fn compute_validator_scaled_total_rewards<V>(
+    validator_totals: Vec<(V, u128)>,
+) -> Vec<(V, u32)> {
     let sum_totals: u128 = validator_totals.iter().map(|(_, t)| t).sum();
 
+    if sum_totals == 0 {
+        return validator_totals
+            .into_iter()
+            .map(|(v, t)| (v, t as u32))
```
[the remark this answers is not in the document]
kostekIV: done

pallets/elections/src/impls.rs
```diff
@@ -95,11 +99,10 @@ where
         ValidatorEraTotalReward::<T>::put(ValidatorTotalRewards(scaled_totals.collect()));
     }
 
-    fn get_committee_and_non_committee(
-        current_era: EraIndex,
-    ) -> (Vec<T::AccountId>, Vec<T::AccountId>) {
+    fn get_committee_and_non_committee() -> (Vec<T::AccountId>, Vec<T::AccountId>) {
         let committee: Vec<T::AccountId> = T::SessionInfoProvider::current_committee();
```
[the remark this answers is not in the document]
kostekIV: done (with BTreeSet)

pallets/elections/src/migrations/v0_to_v1.rs
kostekIV: done

pallets/elections/src/impls.rs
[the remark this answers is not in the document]
kostekIV: done

DamianStraszak (approved):

mike1729 (approved):

PR #446, opened 2022-05-27
A0-917: Add tool for testnet validator candidate verification

maciejnems:
# Description

Peer signer tool can be used for verification of candidates for testnet validators. It checks if candidate node is up to date with block creation and that candidate actually has access to a specific peer id

## Type of change
New feature (non-breaking change which adds functionality)

# Checklist:

 have commented my code, particularly in hard-to-understand areas
New documentation created


peer-signer/README.md
```diff
@@ -0,0 +1,26 @@
+# Peer Signer
```
pmikolajczyk41, line 1: I have some problem with telling which part is more important here: signing the secret message (and then verifying is just an additional feature for ensuring that everything is ok) or the verifying setup, where signing is the chosen way. 

Inferring from the package name, former seems to be the crucial thing... but anyway this is a tad confusing to have this kind of structure that the whole package is named after one of its crates, while the other apparently seem to be of equal importance.

I would either rename the package or be more explanatory in README (my preference) / Cargo.toml description.

peer-signer/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Step one: signature generation
+To generate signature candidate needs a generated p2p address. For whole process to work they also need to have a node set up with up to date block creation and public p2p port, so it can be added as a peer. To generate signature:
+```console
+$ signer --message "secret message" --p2p-secret "some_path_to_p2p_secret/p2p_secret"
+Public key: 08011220c94d0593359ea30451f3aaf61040756c1bfdb7bb779489d4dafda00cce038c0a
+Signed message: 7ef2490075425bcd9fc77a994069183a21746f0be3561d464ae6a89e4b9bdaf26d9cb75c0b11325ae54e66c8ce8a9156a8b8e1d0120309fafe6971db53d6c104
+```
+
+## Step two: communication
```
pmikolajczyk41, line 12: suggestion: this section seems to be completely useless and could be incorporated into the following one

peer-signer/README.md
```diff
@@ -0,0 +1,26 @@
...
+User needs to send the generated public key and signature to verifier. So in this case needs to send:
+```
+08011220c94d0593359ea30451f3aaf61040756c1bfdb7bb779489d4dafda00cce038c0a
+
+7ef2490075425bcd9fc77a994069183a21746f0be3561d464ae6a89e4b9bdaf26d9cb75c0b11325ae54e66c8ce8a9156a8b8e1d0120309fafe6971db53d6c104
+```
+
+## Step three: verification
+To verify you need a node setup, that you can send unsafe rpc calls to. Let's say we have a running node with rpc socket on `http://127.0.0.1:9933`. This node needs to be up to date with block creation. Also for this process to work we need to have candidate in our peers, so to make sure we connected, we can add it to bootnodes of our node. Now we call `verifier` to verify peer and the signature:
```
pmikolajczyk41, line 21: fat nit: in markdown I usually recommend putting each sentence on a separate line: it does not influence the rendered view, but makes it easier for commenting on a particular sentence (instead of a whole paragraph).

peer-signer/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+```console
+$ verifier --node "http://127.0.0.1:9933" --peer-id "12D3KooWPNAEZ9Xru6SgaKoXj4XGeDeVbUjYQHoc3DXpHJ9eeQeM" --message "secret message" --public-key 08011220c94d0593359ea30451f3aaf61040756c1bfdb7bb779489d4dafda00cce038c0a --signature 7ef2490075425bcd9fc77a994069183a21746f0be3561d464ae6a89e4b9bdaf26d9cb75c0b11325ae54e66c8ce8a9156a8b8e1d0120309fafe6971db53d6c104
+Signature for peer 12D3KooWPNAEZ9Xru6SgaKoXj4XGeDeVbUjYQHoc3DXpHJ9eeQeM is correct and peer is up to date with block creation at 1386065
```
pmikolajczyk41, line 25: from the first glance it seems that the code does not provide very friendly communication about errors: since this tool is targeting external users, I would increase legibility in case of failure and probably add a word or two in this README, e.g. what user will see when the node is not accessible (wrong port or blocked unsafe rpc) or the signature is invalid

pmikolajczyk41 (changes requested):

peer-signer/src/verifier.rs
```diff
@@ -0,0 +1,129 @@
...
+    })
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+struct Peer {
+    best_hash: String,
+    best_number: u64,
+    peer_id: String,
+    roles: String,
```
pmikolajczyk41, line 69: since you are already ignoring `protocol_version`, can we skip `roles` as well?

pmikolajczyk41 (changes requested):

peer-signer/README.md
piotrMocz: Yes, the split is even deeper: the signer is intended to be shipped as a part of the "setup package" in the `aleph-node-runner` repo, meaning it will eventually be exposed in a binary form to the end-users. The verifier, on the other hand, is meant for internal use by the Aleph team. I think it could be worth mentioning.

peer-signer/src/signer.rs
```diff
@@ -0,0 +1,38 @@
...
+use std::fs;
+
+#[derive(Debug, Parser)]
+#[clap(version = "1.0")]
+pub struct Config {
+    /// Message we want to sign.
+    #[clap(long)]
+    pub message: String,
+
+    /// Path to p2p secret.
```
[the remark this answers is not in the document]
piotrMocz: This is a nice idea (and eventually we should do that) but in the most common use case the `p2p_secret` is generated during the first run of a node.

peer-signer/src/signer.rs
```diff
@@ -0,0 +1,38 @@
...
+
+fn main() {
+    let Config {
+        message,
+        p2p_secret: p2p_secret_path,
+    } = Config::parse();
+
+    let mut file_content = fs::read(&p2p_secret_path).expect("Can not read from p2p secret path");
+    let secret_key = libp2p_ed25519::SecretKey::from_bytes(&mut file_content)
+        .expect("Incorrect secret. Can't create a secret key.");
```
piotrMocz, line 25: Nit: how about `"Incorrect secret format. Failed to create a secret key."`?

peer-signer/src/verifier.rs
```diff
@@ -0,0 +1,129 @@
...
+) -> T {
+    let mut response: Value = client
+        .post(endpoint)
+        .json(&body)
+        .send()
+        .await
+        .expect("Storage request has failed")
+        .json()
+        .await
+        .expect("Could not deserialize response as JSON");
```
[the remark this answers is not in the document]
piotrMocz: As a sidenote, I think this should be discussed company-wide at some point.

peer-signer/src/signer.rs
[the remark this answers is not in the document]
maciejnems: This secret is generated during the run of node, so probably noone will remove it, as it is necessary for restart. So this option could be nice, but I don't think it is necessary right now.

peer-signer/src/signer.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    let keypair = libp2p_ed25519::Keypair::from(secret_key);
+    let public = PublicKey::Ed25519(keypair.public());
+
+    println!(
+        "Public key: {}",
+        hex::encode(&public.to_protobuf_encoding())
```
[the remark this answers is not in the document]
maciejnems: I think it is a little bit more complicated then that. In wallet we se `peer_ids` and this is public key, that allows us, to generate `peer_id`. Libp2p suggests to use protobuf encoding, to encoding and decoding Public Key. After that we can hex encode that so that it is user friendly, or base64 encode that. Right now I'm actually not sure, which encoding should we use here. Base64 or Hex.

peer-signer/src/signer.rs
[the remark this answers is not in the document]
maciejnems: renamed it to `p2p_secret_path`

peer-signer/src/verifier.rs
```diff
@@ -0,0 +1,129 @@
...
+    /// Signature we want to check.
+    #[clap(long)]
+    pub signature: String,
+
+    /// Max block difference with head.
+    #[clap(long, default_value = "10")]
+    pub block_difference: u64,
+}
+
+async fn make_request_and_parse_result<T: serde::de::DeserializeOwned>(
```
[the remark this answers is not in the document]
maciejnems: Me too. This is a simple one case usage, so I think a copy of part of `forf-off` is fine here. My view on this is "it does the job".

peer-signer/src/verifier.rs
```diff
@@ -0,0 +1,129 @@
...
+        public_key,
+        signature,
+        block_difference,
+    } = Config::parse();
+
+    let req = construct_json_body("system_peers", Value::Null);
+    let block_req = construct_json_body("chain_getBlock", Value::Null);
+
+    let client = Client::new();
+    let res: Vec<Peer> = make_request_and_parse_result(&client, &node, req).await;
```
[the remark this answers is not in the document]
maciejnems: Renamed variable and moved requesting peers to a separate function.

peer-signer/src/verifier.rs
maciejnems: Well, `protocol_version` is not skipped, as I take the result but indeed you can skip roles. I did that, and also skipped best_hash.

peer-signer/src/verifier.rs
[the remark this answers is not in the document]
maciejnems: I think this should be discussed company-wide, but for now I would stick with first option here.

peer-signer/README.md
maciejnems: Maybe it could be renamed to `peer-verification`? I know this places a tad more importance on verifier but this is a tool intended for peer verification. Signer part will be used by a validator candidate, and Verifier part will be used by us. I can add more explanation in README regarding a split of this package.

maciejnems: For now I updated the README. Right now I'm leaning more towards name `peer-verifier`/`peer-verification`. Please let me know what do you think about those.

peer-signer/src/verifier.rs
```diff
@@ -0,0 +1,129 @@
...
+    if let Some(peer) = res.iter().find(|r| r.peer_id == peer_id) {
+        if peer.best_number + block_difference < best_number {
+            panic!(
+                "Peer is not up to date. Should have {:?} block number. Has {:?} block number",
+                best_number, peer.best_number
+            )
+        }
+        if best_number + block_difference < peer.best_number {
+            panic!("Peer is too far in the future. Should have {:?} block number. Has {:?} block number", best_number, peer.best_number)
+        }
```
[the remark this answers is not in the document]
maciejnems: I see what is the value in that but for now we haven't thought about how to detect errors of this tool when using in pipeline. We probably don't want to run it by hand for every peer XD So for now I would stick with asserts and panics, as they are simplest to implement and are already done. In the future when we incorporate this tool so that a lot of peers can be verified automatically we can think about it again.

peer-signer/README.md
pmikolajczyk41: in that case I would support you in changing the name; `peer-verification` sounds subtly better for me, since `peer-verifier` suggests a single tool, whereas `peer-verification` should be a fair umbrella term

peer-signer/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+Verifier part will be used by Aleph Zero team.
+With this program we are going to check whether supplied public key derives to supplied peer id, signature of secret message is correct and this peer is connected and up to date with our network.
+
+## Step one: signature generation by Validator candidate
+To generate signature, candidate needs a generated p2p address.
+For whole process to work they also need to have a node set up with up to date block creation and public p2p port, so it can be added as a peer.
+To generate signature:
+```console
+$ signer --message "secret message" --p2p-secret "some_path_to_p2p_secret/p2p_secret"
```
pmikolajczyk41, line 18: I do not know how you are going to distribute this binary, but if the user has to build it from source then I would put here (and for verifier) information about running `cargo run --bin signer`

peer-signer/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+Signer part will be used by the candidate.
+With this program candidate is going to sign a secret message provided by us to prove they own the p2p secret they claim to own.
+
+
+Verifier part will be used by Aleph Zero team.
+With this program we are going to check whether supplied public key derives to supplied peer id, signature of secret message is correct and this peer is connected and up to date with our network.
+
+## Step one: signature generation by Validator candidate
+To generate signature, candidate needs a generated p2p address.
```
pmikolajczyk41, line 14: how do I generate such address? when I pass `p2p_secret` file that was generated from `./scripts/run_nodes.sh` I get:
```
thread 'main' panicked at 'Incorrect secret format. Failed to create a secret key.: DecodingError { msg: "Ed25519 secret key", source: Some(signature::Error { source: Some(SecretKey must be 32 bytes in length) }) }', src/signer.rs:30:10
```

maciejnems: This is caused by our `./scripts/run_nodes.sh` generating p2p secret in hex encoded format. Currently script for running external validator that we are developing saves it in raw format. I changed signer, to be generic in sense that it first tries to generate Secret Key  from raw bytes and if it fails, signer tries to first decode from hex and after that generate Secret Key.

pmikolajczyk41 (approved):
well done!

peer-verification/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+[[bin]]
+name = "verifier"
+path = "src/verifier.rs"
+
+[dependencies]
+clap = {version = "3.0", features = ["derive"]}
+hex = "0.4"
+libp2p = "0.44"
+reqwest = { version = "0.11", features = ["json"] }
+serde = "1"
```
piotrMocz, line 19: 1.0?

peer-verification/README.md
```diff
@@ -0,0 +1,58 @@
...
+Signer part will be used by the candidate.
+With this program candidate is going to sign a secret message provided by us to prove they own the p2p secret they claim to own.
+
+
+Verifier part will be used by Aleph Zero team.
+With this program we are going to check whether supplied public key derives to supplied peer id, signature of secret message is correct and this peer is connected and up to date with our network.
+
+## Step one: signature generation by Validator candidate
+To generate signature, candidate needs a generated p2p secret.
+For whole process to work they also need to have a node set up with up to date block creation and public p2p port, so it can be added as a peer.
```
piotrMocz, line 15:
```suggestion
For whole process to work they also need to have a node set up with up to date block creation and public p2p port, so it can be discovered by other nodes as a peer.
```

peer-verification/README.md
```diff
@@ -0,0 +1,58 @@
...
+To generate signature:
+```console
+$ signer --message "secret message" --p2p-secret-path "some_path_to_p2p_secret/p2p_secret"
+Public key: 08011220c94d0593359ea30451f3aaf61040756c1bfdb7bb779489d4dafda00cce038c0a
+Signed message: 7ef2490075425bcd9fc77a994069183a21746f0be3561d464ae6a89e4b9bdaf26d9cb75c0b11325ae54e66c8ce8a9156a8b8e1d0120309fafe6971db53d6c104
+```
+Or alternatively with building:
```
piotrMocz, line 22:
```suggestion
Or, alternatively, using cargo, if for some reason the pre-built binary is unavailable:
```

peer-verification/README.md
```diff
@@ -0,0 +1,58 @@
...
+```console
+$ cargo run --bin signer -- --message "secret message" --p2p-secret-path "some_path_to_p2p_secret/p2p_secret"
+Public key: 08011220c94d0593359ea30451f3aaf61040756c1bfdb7bb779489d4dafda00cce038c0a
+Signed message: 7ef2490075425bcd9fc77a994069183a21746f0be3561d464ae6a89e4b9bdaf26d9cb75c0b11325ae54e66c8ce8a9156a8b8e1d0120309fafe6971db53d6c104
+```
+After that candidate needs to send the generated public key and signature to Aleph Zero team.
```
piotrMocz, line 28:
```suggestion
After that, the candidate needs to send the generated public key and signature to Aleph Zero team using the form found on alephzero.org website.
```

peer-verification/README.md
```diff
@@ -0,0 +1,58 @@
...
+So in this case needs to send:
```
piotrMocz, line 29:
```suggestion
So in this case one would send the following:
```

peer-verification/README.md
```diff
@@ -0,0 +1,58 @@
...
+$ verifier --node "http://127.0.0.1:9933" --block-difference=10 --peer-id "12D3KooWPNAEZ9Xru6SgaKoXj4XGeDeVbUjYQHoc3DXpHJ9eeQeM" --message "secret message" --public-key 08011220c94d0593359ea30451f3aaf61040756c1bfdb7bb779489d4dafda00cce038c0a --signature 7ef2490075425bcd9fc77a994069183a21746f0be3561d464ae6a89e4b9bdaf26d9cb75c0b11325ae54e66c8ce8a9156a8b8e1d0120309fafe6971db53d6c104
+Signature for peer 12D3KooWPNAEZ9Xru6SgaKoXj4XGeDeVbUjYQHoc3DXpHJ9eeQeM is correct and peer is up to date with block creation at 1386065
+```
+Or alternatively with building:
+```console
+$ cargo run --bin verifier -- --node "http://127.0.0.1:9933" --block-difference=10 --peer-id "12D3KooWPNAEZ9Xru6SgaKoXj4XGeDeVbUjYQHoc3DXpHJ9eeQeM" --message "secret message" --public-key 08011220c94d0593359ea30451f3aaf61040756c1bfdb7bb779489d4dafda00cce038c0a --signature 7ef2490075425bcd9fc77a994069183a21746f0be3561d464ae6a89e4b9bdaf26d9cb75c0b11325ae54e66c8ce8a9156a8b8e1d0120309fafe6971db53d6c104
+Signature for peer 12D3KooWPNAEZ9Xru6SgaKoXj4XGeDeVbUjYQHoc3DXpHJ9eeQeM is correct and peer is up to date with block creation at 1386065
+```
+
+Arguments are:
```
piotrMocz, line 52:
```suggestion
Available arguments:
```

peer-verification/src/signer.rs
```diff
@@ -0,0 +1,52 @@
...
+
+fn main() {
+    let Config {
+        message,
+        p2p_secret_path,
+    } = Config::parse();
+
+    let path = Path::new(&p2p_secret_path);
+    if !path.exists() {
+        panic!("Can not find p2p secret file: {:?}", p2p_secret_path);
```
piotrMocz, line 34: I see you're using past tense for errors throughout the file, so perhaps it would be best to stick to this convention (i.e. `Could not find [...]`)

piotrMocz (approved):
Looks really nice, some minor suggestions about wording 🙂

PR #447, opened 2022-05-30
Port GHA caching from traffic-maker

obrok:
# Description

Splits GHA caching into 3 parts:

  1. The ~/.cargo cache
  2. The ./target cache
  3. The sccache cache along with starting the sccache server

All caches are keyed with cache_version along with a hash of all Cargo.lock files in the project. As a fallback they
take the newest cache with any hash of Cargo.lock files to rebuild from a partial cache in case of a Cargo.lock change.

Thanks to this, we should now correctly cache `target`. Each workflow might generate different contents in `target`, because the workflows compile different things and use or don't `--release`. However, the cache action only allows a single version of the cache under a given key. When caching `target` along with `~/.cargo`, this leads to a race condition where one workflow manages to upload the cache and then all the other workflows need to use this version. However, that version will not contain any of the files that would be useful for those other builds and they will need to rebuild those. Note that if a cache hit occurs, the cache action __wil not__ upload an updated version of that cache at the end.

Additionally, this PR triggers the `unit-test` and `check-excluded-packages` actions on pushes to `main`. This is to build the cache in the context of `main`. Caches are normally isolated per-branch with the exception of the `main` cache being available in other branches. This should lead to the first build of a PR to be fast as well.

## Type of change

- [x] CI config, no code changes

### DIFF
--- a/.github/workflows/build-node-and-runtime.yml
+++ b/.github/workflows/build-node-and-runtime.yml
@@ -19,7 +19,6 @@ jobs:
     runs-on: ubuntu-latest
     env:
       RUST_BACKTRACE: full
-      # SCCACHE_RECACHE: 1 # to clear cache uncomment this, let the workflow run once, then comment it out again
     steps:
       - name: Checkout source code
         uses: actions/checkout@v2
@@ -34,6 +33,9 @@ jobs:
 
       - name: Restore cache
         uses: ./.github/actions/restore-cache
+        with:
+          target_key: release
+          cache_version: v1
 
       - name: Build binary and runtime
         run: cargo build --release -p aleph-node
@@ -73,5 +75,5 @@ jobs:
           if-no-files-found: error
           retention-days: 7
 
-      - name: Stop cache
+      - name: Cleanup cache
         uses: ./.github/actions/post-cache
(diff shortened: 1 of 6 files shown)


maciejzelaszczyk (approved):

.github/workflows/build-node-and-runtime.yml
```diff
@@ -34,6 +33,9 @@ jobs:
 
       - name: Restore cache
         uses: ./.github/actions/restore-cache
+        with:
+          target-key: release
+          cache-version: v1
```
fbielejec, line 38: maybe a comment that clearing the cache is now done via changing the version?

.github/workflows/check-excluded-packages.yml
```diff
@@ -1,7 +1,14 @@
 name: Check excluded packages
 
 on:
-  pull_request
+  pull_request:
```
fbielejec, line 4: I was wondering if we could start triggering this (and other) auxiliary with `path` argument - i.e. on changes in the listed directories? As it stands it will always be triggered

fbielejec (approved):
Good work!

.github/workflows/check-excluded-packages.yml
obrok: Hm, I think I wouldn't go for that. It seems much worse to not build by accident (because the `path` is incorrect) than to build when not needed. I suspect most changes are code-related after all. If we notice that there is a lot of builds that only touch some non-code files we could exclude those files specifically in the future.

PR #464, opened 2022-06-03
A0-956: Abstract away PeerId and Multiaddr in network

timorl:
# Description

This is a refactor to remove (most of) the dependency of our `network` module on the substrate network, replacing it by abstract traits. For now makes testing marginally easier, in the future will make switching to a different network easier.

Mostly wrote it, because I was fed up with reading code/writing docs and needed some coding to soothe my soul.

Sorry for the size, hard to do it in smaller chunks, it's almost exclusively moving stuff around/renaming though.

## Type of change

Cool, there is no "refactor" in our change types. :D

# Checklist:

- I have added tests


finality-aleph/src/network/manager/discovery.rs
```diff
@@ -31,38 +32,51 @@ impl DiscoveryMessage {
...
 }
 
-impl Discovery {
+impl<M: Multiaddress> Discovery<M> {
     /// Create a new discovery handler with the given response/broadcast cooldown.
     pub fn new(cooldown: Duration) -> Self {
         Discovery {
             cooldown,
             last_broadcast: HashMap::new(),
+            _phantom: PhantomData,
```
fbielejec, line 71: why do you need PhantomData here if it doesn't hold a generic?

finality-aleph/src/network/manager/service.rs
```diff
@@ -242,7 +248,7 @@ impl<NI: NetworkIdentity, D: Data> Service<NI, D> {
             None => {
                 let (data, data_from_network) =
                     self.start_validator_session(pre_session, addresses).await?;
-                return Ok((None, data, data_from_network));
+                return Ok(((None, data), data_from_network));
```
fbielejec, line 251: naming is hard :sweat_smile: but at the same time it's kinda hard to follow what is going on here with the tuples  usd for these complex types - could we use structs (with named fields) to untangle it a bit?

finality-aleph/src/network/manager/session.rs
```diff
@@ -48,19 +50,38 @@ pub enum HandlerError {
     MultiplePeerIds,
 }
 
-fn retrieve_peer_id(addresses: &[Multiaddr]) -> Result<PeerId, HandlerError> {
+fn get_common_peer_id<M: Multiaddress>(addresses: &[M]) -> Option<M::PeerId> {
+    addresses
+        .iter()
+        .try_fold(None, |maybe_peer_id, address| match address.get_peer_id() {
```
fbielejec, line 56: why not just `fold` doesn't look like anything can fail here? You could then get rid of `flatten` too

finality-aleph/src/network/service.rs
```diff
@@ -555,13 +502,13 @@ mod tests {
...
-            identities
+            peer_ids
                 .iter()
                 .take(opened_authorities_n)
-                .map(|identity| {
+                .map(|peer_id| {
                     messages
                         .iter()
-                        .map(move |m| (m.clone(), identity.1, Cow::Borrowed(ALEPH_PROTOCOL_NAME)))
+                        .map(move |m| (m.clone(), *peer_id, Protocol::Generic))
```
fbielejec, line 511: you might be able to get rid of deref and clone by using `into_iter`, But I'm Not a Compiler :slightly_smiling_face:

finality-aleph/src/network/service.rs
```diff
@@ -689,32 +626,25 @@ mod tests {
...
         }
 
-        let identities: Vec<_> = (0..4)
-            .map(|_| MockNetworkIdentity::new().identity())
-            .collect();
+        let peer_ids: Vec<_> = (0..4).map(|_| MockPeerId::random()).collect();
         let message: Vec<u8> = vec![1, 2, 3];
 
-        identities.iter().for_each(|identity| {
+        peer_ids.iter().for_each(|peer_id| {
```
fbielejec, line 632: `into_iter` ?

finality-aleph/src/network/service.rs
```diff
@@ -689,32 +626,25 @@ mod tests {
...
-                    role: ObservedRole::Authority,
-                })
+                .emit_event(MockEvent::StreamOpened(*peer_id, Protocol::Validator));
         });
 
         // We do this only to make sure that NotificationStreamOpened events are handled
         test_data.wait_for_events_handled().await;
 
-        identities.iter().for_each(|identity| {
+        peer_ids.iter().for_each(|peer_id| {
```
fbielejec, line 641: `into_iter`

finality-aleph/src/network/service.rs
```diff
@@ -819,32 +744,25 @@ mod tests {
...
         }
 
-        let identities: Vec<_> = (0..4)
-            .map(|_| MockNetworkIdentity::new().identity())
-            .collect();
+        let peer_ids: Vec<_> = (0..4).map(|_| MockPeerId::random()).collect();
         let message: Vec<u8> = vec![1, 2, 3];
 
-        identities.iter().for_each(|identity| {
+        peer_ids.iter().for_each(|peer_id| {
```
fbielejec, line 750: `into_iter`

finality-aleph/src/network/service.rs
```diff
@@ -819,32 +744,25 @@ mod tests {
...
-                    role: ObservedRole::Authority,
-                })
+                .emit_event(MockEvent::StreamOpened(*peer_id, Protocol::Validator));
         });
 
         // We do this only to make sure that NotificationStreamOpened events are handled
         test_data.wait_for_events_handled().await;
 
-        identities.iter().for_each(|identity| {
+        peer_ids.iter().for_each(|peer_id| {
```
fbielejec, line 759: `into_iter`

finality-aleph/src/network/service.rs
```diff
@@ -864,10 +782,10 @@ mod tests {
         );
 
         let expected_messages = HashSet::from_iter(
-            identities
+            peer_ids
                 .iter()
```
fbielejec, line 786: the same :slightly_smiling_face:

finality-aleph/src/network/service.rs
```diff
@@ -993,32 +901,25 @@ mod tests {
...
         }
 
-        let identities: Vec<_> = (0..4)
-            .map(|_| MockNetworkIdentity::new().identity())
-            .collect();
+        let peer_ids: Vec<_> = (0..4).map(|_| MockPeerId::random()).collect();
         let message: Vec<u8> = vec![1, 2, 3];
 
-        identities.iter().for_each(|identity| {
+        peer_ids.iter().for_each(|peer_id| {
```
fbielejec, line 907: `into_iter`

finality-aleph/src/network/service.rs
```diff
@@ -993,32 +901,25 @@ mod tests {
...
-                    role: ObservedRole::Authority,
-                })
+                .emit_event(MockEvent::StreamOpened(*peer_id, Protocol::Generic));
         });
 
         // We do this only to make sure that NotificationStreamOpened events are handled
         test_data.wait_for_events_handled().await;
 
-        identities.iter().for_each(|identity| {
+        peer_ids.iter().for_each(|peer_id| {
```
fbielejec, line 916: and `into_iter`

finality-aleph/src/substrate_network.rs
```diff
@@ -0,0 +1,413 @@
...
+    }
+}
+
+impl MultiaddressT for Multiaddress {
+    type PeerId = PeerId;
+
+    fn get_peer_id(&self) -> Option<Self::PeerId> {
+        self.0
+            .iter()
+            .try_fold(None, |maybe_peer_id, protocol| match peer_id(&protocol) {
```
fbielejec, line 116: use `fold` if possible ?

finality-aleph/src/network/manager/discovery.rs
timorl: It holds `M: Multiaddress`, as written in the struct definition. I think I could rewrite this whole thing to only have the methods parametrized by `M`, but this would be less clear imo.

finality-aleph/src/network/service.rs
timorl: Used below. :<

finality-aleph/src/network/service.rs
timorl: Used below, but changed there.

finality-aleph/src/network/service.rs
timorl: Changed.

finality-aleph/src/network/service.rs
timorl: Used below.

finality-aleph/src/network/service.rs
```diff
@@ -1038,10 +939,10 @@ mod tests {
         );
 
         let expected_messages = HashSet::from_iter(
-            identities
+            peer_ids
```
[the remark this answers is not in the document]
timorl: Changed.

finality-aleph/src/network/manager/discovery.rs
```diff
@@ -31,38 +32,51 @@ impl DiscoveryMessage {
...
 /// Handles creating and responding to discovery messages.
-pub struct Discovery {
+pub struct Discovery<M: Multiaddress> {
     cooldown: Duration,
     last_broadcast: HashMap<NodeIndex, Instant>,
+    _phantom: PhantomData<M>,
 }
 
-type DiscoveryCommand = (DiscoveryMessage, DataCommand);
+type DiscoveryCommand<M> = (
```
piotrMocz, line 41: @fbielejec mentions it below but I'm generally against using tuples for anything other than trivial unpacking of values or enumeration in lists.
```suggestion
struct DiscoveryCommand<M> {
  message: DiscoverMessage<M>,
  command: DataCommand<<M as Multiaddress>::PeerId>,
}
```

finality-aleph/src/network/manager/discovery.rs
```diff
@@ -129,9 +143,9 @@ impl Discovery {
     /// that we should send as a result of it.
     pub fn handle_message(
         &mut self,
-        message: DiscoveryMessage,
-        handler: &mut SessionHandler,
-    ) -> (Vec<Multiaddr>, Vec<DiscoveryCommand>) {
+        message: DiscoveryMessage<M>,
+        handler: &mut SessionHandler<M>,
+    ) -> (Vec<M>, Vec<DiscoveryCommand<M>>) {
```
piotrMocz, line 148: Same thing about tuples: here it's a public API so I'm even more convinced :smile:

finality-aleph/src/network/manager/service.rs
```diff
@@ -96,6 +96,22 @@ impl Config {
...
 
+type MessageForNetwork<D, M> = (NetworkData<D, M>, DataCommand<<M as Multiaddress>::PeerId>);
+
+pub struct ServiceActions<D: Data, M: Multiaddress> {
+    maybe_command: Option<ConnectionCommand<M>>,
+    data: Vec<MessageForNetwork<D, M>>,
+}
+
+impl<D: Data, M: Multiaddress> ServiceActions<D, M> {
+    fn noop() -> Self {
```
piotrMocz, line 107: I could imagine rewriting it so that we derive `Default` for `ServiceActions` and here simply do:
```rust
fn noop() -> Self { Default::default() }
```

Although one could argue that doing this explicitly is better (and I would probably agree) ;)

finality-aleph/src/network/manager/service.rs
```diff
@@ -134,30 +150,35 @@ impl<NI: NetworkIdentity, D: Data> Service<NI, D> {
...
         self.sessions.remove(&session_id);
         self.to_retry
             .retain(|(pre_session, _)| pre_session.session_id() != session_id);
         Self::delete_reserved(self.connections.remove_session(session_id))
     }
 
     fn network_message(
-        (message, command): (DiscoveryMessage, DataCommand),
-    ) -> (NetworkData<D>, DataCommand) {
+        (message, command): (DiscoveryMessage<NI::Multiaddress>, DataCommand<NI::PeerId>),
```
piotrMocz, line 173: At the risk of sounding like a broken record: structs here would make it more future-proof as you could change the field order :sunglasses:

finality-aleph/src/network/manager/session.rs
piotrMocz: Maybe you could try extracting this closure into its own function? It's certainly complex enough :smile:

finality-aleph/src/substrate_network.rs
```diff
@@ -0,0 +1,413 @@
...
+                Some(peer_id) => match maybe_peer_id {
+                    Some(other_peer_id) => match peer_id == other_peer_id {
+                        true => Some(maybe_peer_id),
```
piotrMocz, line 119: Is there any advantage in using `match` on boolean expressions instead of `if`?

finality-aleph/src/substrate_network.rs
```diff
@@ -0,0 +1,413 @@
...
+        let addresses = peers.into_iter().map(|peer_id| peer_id.0).collect();
+        self.remove_peers_from_reserved_set(protocol_name(&protocol), addresses);
+    }
+}
+
+impl<B: Block, H: ExHashT> NetworkIdentity for Arc<NetworkService<B, H>> {
+    type PeerId = PeerId;
+    type Multiaddress = Multiaddress;
+
+    fn identity(&self) -> (Vec<Self::Multiaddress>, Self::PeerId) {
```
piotrMocz, line 328: This function looks like LISP code :smile: Not saying it's necessarily a bad thing but... structs >> tuples

finality-aleph/src/substrate_network.rs
timorl: I think it's clearer when you want to return something from the statement, especially if the things are single line, e.g. I would pretty clearly prefer:
```
if condition {
 do_some_work();
 do_some_more_work();
} else { //although preferably avoid `else` altogether and just return in the `if` perhaps splitting into functions; perhaps this is a big part of the reason for preferring `match`, I just dislike `else`?
 do_other_work();
}
```
but I'm ambivalent about
```
let result = if condition {
 let result = create_result();
 result.modify_result();
 result
} else {
 create_result()
}
```
vs
```
let result = match condition {
 true => {
  let result = create_result();
  result.modify_result();
  result
 },
 false => create_result()
}
```
And in the case when both cases are single-line and returning something I clearly prefer `match`.

Definitely mostly a personal preference though.

finality-aleph/src/network/manager/service.rs
timorl: Then consider that argued. :P

To make this more explicit – I think `Default` implies there is a default value of something, and it would be pretty hard to convince me that a `ServiceAction` should by default be `noop`.

finality-aleph/src/network/manager/discovery.rs
timorl: As discussed elsewhere we will leave this for now and try agreeing on a policy on tuples vs structs later. I sent you the task.

finality-aleph/src/network/manager/session.rs
timorl: Changed into something more similar to what was there before the refactor, although unfortunately this resulted in some code duplication. :/

At this moment I'm a tad sad we don't write in a more functional language, there this kind of thing would be solved with completely generic functions. :<

finality-aleph/src/substrate_network.rs
timorl: Just as in the other case extracting the function removed the `try_`.

finality-aleph/src/substrate_network.rs
piotrMocz: Alright! It looks quite nice, I was just asking out of curiosity

finality-aleph/src/network/manager/service.rs
piotrMocz: Okay, that certainly convinces me, thanks :)

fbielejec (approved):

PR #469, opened 2022-06-07
Update substrate to 0.9.23

Yithis:
# Description

This PR updates substrate in our repository from branch `polkadot-v0.9.19` to `polkadot-v0.9.23`.

## Type of change

- Substrate update


bin/node/Cargo.toml
```diff
@@ -25,48 +26,48 @@ hex = "0.4"
...
+sp-timestamp = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.23" }
+pallet-staking = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.23" }
 
 aleph-runtime = { path = "../runtime"}
 finality-aleph = { path = "../../finality-aleph"}
 aleph-primitives = { package = "primitives", path = "../../primitives"}
 
 # These dependencies are used for the node's RPCs
+jsonrpsee = { version = "0.13.0", features = ["server"] }
 jsonrpc-core = "18.0"
```
h4nsu, line 59: Is this still needed after switching to `jsonrpsee`?

aleph-client/src/staking.rs
```diff
@@ -271,10 +272,21 @@ pub fn bonded<C: AnyConnection>(connection: &C, stash: &KeyPair) -> Option<Accou
...
 }
 
-pub fn ledger<C: AnyConnection>(
-    connection: &C,
-    controller: &KeyPair,
-) -> Option<pallet_staking::StakingLedger<AccountId, Balance>> {
+/// Since PR #10982 changed `pallet_staking::StakingLedger` to be generic over
+/// `T: pallet_staking::Config` (somehow breaking consistency with similar structures in other
+/// pallets) we have no easy way of retrieving ledgers from storage. Thus, we chose cloning
+/// (relevant part of) this struct instead of implementing `Config` trait.
```
timorl, line 278: I don't quite get the problem here. What is the problem with implementing the `Config` trait?

Also, I assume this will break as soon as there is a migration changing the shape of the data in any way, right? I don't think this is a major problem, just making sure I understand.

bin/node/src/service.rs
```diff
@@ -190,7 +190,7 @@ fn setup(
                 deny_unsafe,
             };
 
-            Ok(crate::rpc::create_full(deps))
+            crate::rpc::create_full(deps).map_err(Into::into)
```
timorl, line 193: Hmmm, doesn't `?` work here (it should be ok within a closure, shouldn't it?)?

bin/node/src/service.rs
```diff
@@ -200,7 +200,7 @@ fn setup(
         keystore: keystore_container.sync_keystore(),
         task_manager,
         transaction_pool,
-        rpc_extensions_builder,
+        rpc_builder: rpc_extensions_builder,
```
timorl, line 203: Nit: rename the variable to avoid the `: `.

rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2021-10-24
+nightly-2022-05-23
```
timorl, line 1: Hm, this has far-ranging consequences, you should probably announce this somewhere, and I believe we should also bump AlephBFT.

scripts/run_e2e.sh
```diff
@@ -4,6 +4,6 @@ set -e
 
 cd e2e-tests/
 
-RUST_LOG=aleph_e2e_client=info,aleph-client=info cargo run -- --node 127.0.0.1:9943 --test-cases members_rotate
+RUST_LOG=aleph_e2e_client=info,aleph-client=info cargo run -- --node 127.0.0.1:9943
```
timorl, line 7: Huh, what happened here?

timorl (approved):
Looks good, some minor comments.

mike1729 (approved):
Only some minor comments, great work :muscle:

bin/runtime/src/lib.rs
```diff
@@ -651,6 +687,9 @@ impl pallet_contracts::Config for Runtime {
     type Schedule = Schedule;
     type CallStack = [pallet_contracts::Frame<Self>; 31];
     type AddressGenerator = pallet_contracts::DefaultAddressGenerator;
+    type ContractAccessWeight = pallet_contracts::DefaultContractAccessWeight<BlockWeights>;
```
h4nsu, line 690: https://github.com/paritytech/substrate/blob/6cbe1772bf258793fa9845daa8f43ea0cadee596/frame/contracts/src/lib.rs#L347

aleph-client/src/staking.rs
pmikolajczyk41: The problem is that implementing `Config` trait is quite cumbersome - it has very broad API and this would seem very strange that `aleph-client` provides `pallet_x::Config` only for `x = staking`, which is used simply for reading a secondary struct from storage.

Yes, this will break as soon as there is a change influencing decoding of this structure.

Personally, I would rather make a PR to Substrate with reverting this parametrization instead of adjusting to it #rebel

scripts/run_e2e.sh
pmikolajczyk41: I'm afraid that there was some leftover

bin/node/Cargo.toml
Yithis: Done

bin/node/src/service.rs
Yithis: Can't do. It fails to deduce the correct type of variable then

bin/node/src/service.rs
Yithis: Done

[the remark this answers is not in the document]
Yithis: Done

bin/runtime/src/lib.rs
```diff
@@ -660,7 +699,7 @@ construct_runtime!(
         NodeBlock = opaque::Block,
         UncheckedExtrinsic = UncheckedExtrinsic
     {
-        System: frame_system::{Pallet, Call, Config, Storage, Event<T>} = 0,
+        System: frame_system,
```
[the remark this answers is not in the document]
Yithis: Done for all

bin/runtime/src/lib.rs

bin/node/src/service.rs
Yithis: My bad, forgot `Ok` and error wasn't helpful. Done

rust-toolchain
Yithis: I will announce it today, also already taking care of AlephBFT

PR #475, opened 2022-06-08
Adjust runtime constants

h4nsu:
# Description

Unifies and adjusts constants (mostly in runtime) that are responsible for defining various tokenomics related aspects of the chain.

## Type of change

- Refactor (non-breaking change which fixes an issue)

# Checklist:

### DIFF
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -113,23 +113,12 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("aleph-node"),
     impl_name: create_runtime_str!("aleph-node"),
     authoring_version: 1,
-    spec_version: 15,
+    spec_version: 16,
     impl_version: 1,
     apis: RUNTIME_API_VERSIONS,
     transaction_version: 6,
     state_version: 1,
 };
-
-pub const MILLISECS_PER_BLOCK: u64 = DEFAULT_MILLISECS_PER_BLOCK;
-
-pub const MILLISECS_PER_MINUTE: u64 = 60_000; // milliseconds
-pub const MILLISECS_PER_HOUR: u64 = MILLISECS_PER_MINUTE * 60;
-pub const MILLISECS_PER_DAY: u64 = MILLISECS_PER_HOUR * 24;
-
-/// Get the number of blocks produced in the period given by `hours`
-pub fn hours_as_block_num(hours: u64) -> BlockNumber {
-    (MILLISECS_PER_HOUR * hours / MILLISECS_PER_BLOCK) as BlockNumber
-}
 
 /// The version information used to identify this runtime when compiled natively.
 #[cfg(feature = "std")]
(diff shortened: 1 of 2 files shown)


bin/runtime/src/lib.rs
```diff
@@ -140,16 +129,24 @@ pub fn native_version() -> NativeVersion {
...
-const MAX_BLOCK_WEIGHT: Weight = 400 * WEIGHT_PER_MILLIS;
+pub const MAX_BLOCK_WEIGHT: Weight = 400 * WEIGHT_PER_MILLIS;
 // We agreed to 5MB as the block size limit.
 pub const MAX_BLOCK_SIZE: u32 = 5 * 1024 * 1024;
 
-pub const MILLICENTS: Balance = 100_000_000;
-pub const CENTS: Balance = 1_000 * MILLICENTS; // 10^12 is one token, which for now is worth $0.1
-pub const DOLLARS: Balance = 100 * CENTS; // 10_000_000_000
+// The storage deposit is roughly 1 TOKEN per 1kB
+pub const DEPOSIT_PER_BYTE: Balance = MILLI;
```
pmikolajczyk41, line 149: I'm not so sure, but e.g. in multisignatures it is just a cost (not refundable), so maybe just `BYTE_COST`/`STORAGE_BYTE_COST` is better?

pmikolajczyk41 (commented):
kudos for tidying things here! I left some suggestions and I also have some questions/requests:
 - is/will be there any Notion doc about constants? for now I feel that is not needed, _but_ I would love to have some human-friendly insight for them - one option is (mentioned in some comment) rust doc, wdyt?
 - please add Jira ID to PR title
 - I'm aware that there were some talks about it, but why have we finally given up a single point of adjusting all values? I mean a constant specifying (rough) ratio between token and USD? If all costs depended on it, then adapting prices when token goes up 10 times would have been much easier and straightforward

mike1729 (approved):
No other comments than Piotr's. Great step towards cleaning this file.

bin/runtime/src/lib.rs
h4nsu: If it's not refundable then why is it called "deposit"? The code seems to be using `T::Currency::reserve` and `T::Currency::unreserve` (https://github.com/paritytech/substrate/blob/bc7a1e6c19aec92bfa247d8ca68ec63e07061032/frame/multisig/src/lib.rs#L711) so I'd naively guess this is a refundable deposit?

bin/runtime/src/lib.rs
```diff
@@ -140,16 +129,24 @@ pub fn native_version() -> NativeVersion {
     }
 }
 
-const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(75);
+pub const MILLISECS_PER_BLOCK: u64 = DEFAULT_MILLISECS_PER_BLOCK;
+pub const BLOCKS_PER_HOUR: u32 = 60 * 60 * 1000 / (MILLISECS_PER_BLOCK as u32);
+
+pub const MILLI: Balance = TOKEN / 1000;
```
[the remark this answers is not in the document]
h4nsu: Done

bin/runtime/src/lib.rs
```diff
@@ -278,8 +268,12 @@ impl OnUnbalanced<NegativeImbalance> for EverythingToTheTreasury {
 }
 
 parameter_types! {
+    // This value increases the priority of `Operational` transactions by adding
```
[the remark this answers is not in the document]
h4nsu: Quite a lot of comments already start with `///`. The ones that do not seem to be more like internal notes ("We agreed to 5MB as the block size limit." "We allow at most 20 approvals in the queue at once.") or contain permalinks to other repos. Seems to me like not the best candidate for a docstring...

h4nsu:
@pmikolajczyk41
1. Doc with constants is already in Notion, yet not fully finished
2. Added Jira IDs, it seems like this PR closes 3 issues XD
3. I don't know, I somehow like the setup which expresses everything in native token much more than in usd. Doesn't look like we will be doing linear fee scaling in near future, I'd leave it for now the way it is.

bin/runtime/src/lib.rs
pmikolajczyk41: okay, then I wrongly remembered

pmikolajczyk41 (approved):

PR #480, opened 2022-06-09
add default timeout

fbielejec:
# Description

An attempt at fixing flaky action

## Type of change


fbielejec:
> So looks like `sccache` is taking a long time to start... Then I don't entirely understand this - the link you posted has the failing action take around 1:47, seems like `sccache --start-server` has some timeout of its own that's maybe around 1 minute. So seems to me like adding a 5 minute timeout on the action as a whole is not going to do anything?

I agree, there are various reports on GH but nothing super specific: https://github.com/mozilla/sccache/issues/739
We could increase logging and monitor, but closing for now.

PR #484, opened 2022-06-13
Identity pallet

kostekIV:
# Description
Adds an identity pallet

## Type of change

# Checklist:

### DIFF
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -112,7 +112,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("aleph-node"),
     impl_name: create_runtime_str!("aleph-node"),
     authoring_version: 1,
-    spec_version: 17,
+    spec_version: 18,
     impl_version: 1,
     apis: RUNTIME_API_VERSIONS,
     transaction_version: 6,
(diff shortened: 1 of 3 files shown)


bin/runtime/src/lib.rs
```diff
@@ -670,6 +670,30 @@ impl pallet_contracts::Config for Runtime {
     type RelaxedMaxCodeLen = ConstU32<{ 256 * 1024 }>;
 }
 
+parameter_types! {
+    pub const BasicDeposit: Balance = 258 * DEPOSIT_PER_BYTE;
+    pub const FieldDeposit: Balance = 66 * DEPOSIT_PER_BYTE;
+    pub const SubAccountDeposit: Balance = 85 * DEPOSIT_PER_BYTE;
```
pmikolajczyk41, line 676: it would be good to provide some explanation/calculations behind these values

pmikolajczyk41: for now it should be sufficient

pallets/elections/src/impls.rs
```diff
@@ -414,7 +414,7 @@ mod tests {
                     reserved.clone(),
                     non_reserved.clone(),
                 )
-                .expect("Expected non-empty rotated committee!")
+                .lletexpect("Expected non-empty rotated committee!")
```
pmikolajczyk41, line 417: O_O

pmikolajczyk41 (approved):
there is nothing to complain about it, so just two points:
 - what is the reason for adding it now? some Jira?
 - pallet's README is not very explanatory, so it would be good to create some documentation for us about it; it doesn't have to be long but at least some terminology (registrars etc) and some examples would be great (in particular, I do not see any UI support for this, so it's not so intuitive how to use it)

mike1729 (approved):

bin/runtime/src/lib.rs
```diff
@@ -112,7 +112,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("aleph-node"),
     impl_name: create_runtime_str!("aleph-node"),
     authoring_version: 1,
-    spec_version: 17,
+    spec_version: 18,
     impl_version: 1,
     apis: RUNTIME_API_VERSIONS,
     transaction_version: 6,
```
DamianStraszak, line 118: Checklist says you bumped `transaction_version` :)

DamianStraszak (changes requested):

DamianStraszak (approved):

PR #486, opened 2022-06-13
A0-991: e2e test for changing non-reserved member list

krzysztofziobro:
# Description

Adding new e2e test that checks whether changing non-reserved member list works properly.

## Type of change

- New e2e test


maciejnems (approved):
LGTM :slightly_smiling_face:

e2e-tests/src/test/non_reserved_change.rs
```diff
@@ -0,0 +1,106 @@
...
+use aleph_client::{
+    change_members, get_current_session, wait_for_finalized_block, wait_for_full_era_completion,
+    wait_for_next_era, wait_for_session, AnyConnection, Header, KeyPair, RootConnection,
+    SignedConnection,
+};
+use sp_core::Pair;
+use substrate_api_client::{AccountId, XtStatus};
+
+fn get_reserved_members(config: &Config) -> Vec<KeyPair> {
+    get_validators_keys(config)[0..1].to_vec()
```
kostekIV, line 14: nit: `..1`

e2e-tests/src/test/non_reserved_change.rs
```diff
@@ -0,0 +1,106 @@
...
+        XtStatus::InBlock,
+    );
+
+    let current_session = get_current_session(&connection);
+    wait_for_session(&connection, current_session + 1)?;
+
+    let (stored_non_reserved, eras_non_reserved) = get_pallets_non_reserved(&connection)?;
+
+    assert_eq!(stored_non_reserved, new_non_reserved_members);
+    assert_eq!(eras_non_reserved, initial_non_reserved_members);
```
kostekIV, line 88: add a message to asserts

e2e-tests/src/test/non_reserved_change.rs
```diff
@@ -0,0 +1,106 @@
...
+    get_validators_keys(config)[2..].to_vec()
+}
+
+fn get_pallets_non_reserved(
+    connection: &SignedConnection,
+) -> anyhow::Result<(Vec<AccountId>, Vec<AccountId>)> {
+    let stored_non_reserved: Vec<AccountId> = connection
+        .as_connection()
+        .get_storage_value("Elections", "NonReservedMembers", None)?
+        .unwrap();
```
kostekIV, line 31: nit: unwrap to expect

kostekIV (approved):
great job, just a couple of nits

kostekIV (changes requested):
Actually, there is one more thing to do here. We need to add this test to our pipelines.

1. In `.github/workflows/e2e-tests-main-devnet.yml` add a job named `run-e2e-name-of-your-test`
2. To do it, copy for example `run-e2e-era-payout` and change the following fields:
- `name`
- `test-case` to the name of your test (the one from the `cases.rs` file)
- set `timeout-minutes` to the appropriate number
3. Add your job name to the array in the `check-e2e-test-suite-completion`

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -293,6 +293,20 @@ jobs:
...
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@v2
+
+      - name: Run e2e test
+        uses: ./.github/actions/run-e2e-test
+        with:
+          test-case: change_non_reserved
+          follow-up-finalization-check: true
+        timeout-minutes: 15
```
kostekIV, line 309: 10 should be sufficient here

kostekIV (approved):
lgtm, one thing you can add is the same check but on the reserved list.

.github/workflows/e2e-tests-main-devnet.yml
krzysztofziobro: Add the check in the same test? Then I should probably also change name of the test, because it won't really reflect what it's doing?

PR #493, opened 2022-06-15
Compatibility for justifications

timorl:
# Description

This implements the changes to compatibility as proposed in the message compatibility ADR, but only for justifications. This is a bit out of order, because we want to introduce new types of justifications for A0-926, and introducing them before compatibility is implemented would have left us with more ugly decoding code (likely) forever.

## Type of change

The list of types of changes really isn't comprehensive, calling this a "bug fix" is quite a bit of a stretch. <_<

# Checklist:

- I have added tests


finality-aleph/src/justification/compatibility.rs
```diff
@@ -27,22 +32,242 @@ impl From<AlephJustificationV1> for AlephJustification {
...
+    fn decode<I: CodecInput>(input: &mut I) -> Result<Self, CodecError> {
+        use VersionedAlephJustification::*;
+        let version = u32::decode(input)?;
+        let num_bytes = u32::decode(input)?;
+        match version {
+            1 => Ok(V1(AlephJustificationV1::decode(input)?)),
+            2 => Ok(V2(AlephJustification::decode(input)?)),
+            _ => {
+                // TODO: figure out some limit for `num_bytes` and error out if this alloc would be
+                // too big
```
DamianStraszak, line 95: This does not seem like a TODO we would like to leave for later :P 

One signature is `64` bytes, so if we stay below `100` nodes, then something like `2^{13}` should be safe. 
Probably before we scale beyond `100` we will move to succinct multisigs, so the const will never need to grow, I guess?

timorl: Hey, cool, that would mean I can just switch this to a `u16` and that would be acceptable, wouldn't it?

DamianStraszak: Yes, and I believe the same can be done for version :)

finality-aleph/src/justification/compatibility.rs
```diff
@@ -27,22 +32,242 @@ impl From<AlephJustificationV1> for AlephJustification {
...
+                result
+            }
+            V2(justification) => {
+                let mut result = (2_u32).encode();
+                let mut justification = justification.encode();
+                let num_bytes = justification.len() as u32;
+                result.append(&mut num_bytes.encode());
+                result.append(&mut justification);
+                result
+            }
```
kostekIV, line 80: nit: all arms could be handled by one function `fn smt<E: Encode>(version, payload: E) -> Vec<u8>`

kostekIV: the version could be of the `type Version = u16` to enforce always one type for it

finality-aleph/src/justification/compatibility.rs
timorl: Done, also rebased on new main.

finality-aleph/src/justification/compatibility.rs
timorl: Slightly different than you suggested so that it works for `Other` as well, but done.

Also possibly fixed a bug on the way and added a test for it (in wrong order unfortunately, so I'm not quite sure the bug was there).

kostekIV (approved):

PR #497, opened 2022-06-21
Improvements to Python chainrunner and related tests

h4nsu:
# Description

- adds a method that waits for finalization to catch up
- adds a method that waits until all authorities are connected (prevents unlucky rolls in tests where one validator is killed - now we will never kill the finalization while killing that node) 
- adds a method to fork-off the state from an existing chain
- adds a `backup` switch for including or not `--backup-path` node argument
- removes Jupyter notebook - it's a true PITA to keep that thing in sync
- applies the above improvements to all the Python tests to (hopefully) make them less flaky and their output more friendly

## Type of change

- New feature
- Tests refactor

# Checklist:

- I have added tests


local-tests/chainrunner/chain.py
```diff
@@ -162,3 +157,57 @@ def purge(self, nodes=None):
...
+        nodes = [self.nodes[i] for i in nodes] if nodes else self.nodes
+        deadline = time.time() + timeout
+        while any((n.highest_block()[1] <= old_finalized + finalized_delta) for n in nodes):
+            time.sleep(5)
+            if time.time() > deadline:
+                raise TimeoutError(f'Block finalization stalled after {timeout} seconds')
+        if catchup:
+            def lags(node):
+                r, f = node.highest_block()
+                return r - f > catchup_delta
```
pmikolajczyk41, line 198: what about `lags = lambda node: operator.sub(node.highest_block())`?

local-tests/chainrunner/node.py
```diff
@@ -58,13 +80,19 @@ def greplog(self, regexp):
...
+        Return two ints: highest block and highest finalized block."""
         results = self.greplog(r'best: #(\d+) .+ finalized #(\d+)')
         if results:
             a, b = results[-1]
             return int(a), int(b)
         return -1, -1
 
+    def check_authorities(self):
+        """Find in the logs the number of authorities this node is connected to.
+        Return bool indicating if it's connected to all know authorities."""
```
pmikolajczyk41, line 92: typo: known

local-tests/chainrunner/node.py
```diff
@@ -46,7 +52,23 @@ def stop(self):
     def purge(self):
         """Purge chain (delete the database of the node)."""
         cmd = [self.binary, 'purge-chain', '-y'] + self._stdargs()
-        subprocess.run(cmd, stdout=subprocess.DEVNULL, check=True)
```
pmikolajczyk41, line 49: why not `check`?

local-tests/chainrunner/node.py
```diff
@@ -28,9 +28,15 @@ def __init__(self, binary, chainspec, path, logdir=None):
     def _stdargs(self):
         return ['--base-path', self.path, '--chain', self.chainspec]
 
-    def start(self, name):
-        """Start the node. `name` is used to name of the logfile and for --name flag."""
-        cmd = [self.binary, '--name', name] + self._stdargs() + flags_from_dict(self.flags)
+    def _nodeargs(self, backup=True):
+        res = ['--node-key-file', op.join(self.path, 'p2p_secret'), '--enable-log-reloading']
+        if backup:
```
pmikolajczyk41, line 33: I guess that we are missing: `else: res += ['--no_backup']` (one of these flags is mandatory, iirc)

local-tests/chainrunner/node.py
```diff
@@ -46,7 +52,23 @@ def stop(self):
...
+        Raises KeyError if not present."""
+        port = self.flags.get('rpc_port', self.flags.get('rpc-port'))
+        if port is None:
+            raise KeyError("RPC port unknown, please set rpc_port flag")
+        return port
+
+    def ws_port(self):
+        """Return WS port for this node. The value is taken from `flags` dictionary.
+        Raises KeyError if not present."""
+        port = self.flags.get('ws_port', self.flags.get('ws-port'))
```
pmikolajczyk41, line 68: meh, do we need such fallback? we have to pass `ws_port` to `set_flags()` anyway so `ws-port` is never gonna happen

pmikolajczyk41: oh, okay, in Python you can pass kwargs like `**{"ws-port": 9933}`, but I think that such people are asking for troubles

local-tests/chainrunner/utils.py
```diff
@@ -43,11 +44,34 @@ def flags_from_dict(d):
...
     print('Blocks seen by nodes:')
     print('  Highest:   ', *highest)
     print('  Finalized: ', *finalized)
 
     return finalized
+
+
+def check_version(nodes, verbose=False):
+    """Query given nodes for aleph-node (host) version and runtime version.
+    Print the summary to the standard output and return the runtime version.
```
pmikolajczyk41, line 59: maybe we should clarify that we will return the max runtime version

local-tests/test_catch_up.py
```diff
@@ -33,45 +36,45 @@
...
-finalized_before_kill_per_node = check_finalized(chain)
-print('Killing one validator and one nonvalidator')
+printt('Waiting for finalization')
+chain.wait_for_finalization(0)
+printt('Waiting for authorities')
+chain.wait_for_authorities()
 
-chain[3].stop()
-chain[4].stop()
+printt('Killing one validator and one nonvalidator')
```
pmikolajczyk41, line 47: how does this statement work with `nonvalidators=all_accounts[4:]` ?

local-tests/test_catch_up.py
```diff
@@ -33,45 +36,45 @@
...
+chain.stop(nodes=[3, 4])
+finalized_before_kill = check_finalized(chain)
```
pmikolajczyk41, line 49: actually, it is finalized _after_ kill :stuck_out_tongue_closed_eyes:

local-tests/test_catch_up.py
```diff
@@ -33,45 +36,45 @@
...
 sleep(30 * 2)
 
-print('restarting nodes')
-finalized_before_start_per_node = check_finalized(chain)
+finalized_before_start = check_finalized(chain)
 
 # Check if the finalization didn't stop after a kill.
-if finalized_before_start_per_node[0] - finalized_before_kill_per_node[0] < 10:
-    print('Finalization stalled')
+if finalized_before_start[0] - finalized_before_kill[0] < 10:
```
pmikolajczyk41, line 57: nit: I know that it is sufficient, but I think that `max(finalized_before_start) - max(finalized_before_kill)` is a tad better

local-tests/test_catch_up.py
```diff
@@ -33,45 +36,45 @@
...
+validator_diff = finalized_after[3] - finalized_before_start[3]
 
-ALLOWED_DELTA = 5
+delta = 5
 
-# Checks if the murdered nodes started catching up with reasonable nr of blocks.
-if nonvalidator_diff <= ALLOWED_DELTA:
-    print(f"too small catch up for nonvalidators: {nonvalidator_diff}")
+# Check if the murdered nodes started catching up with reasonable nr of blocks.
+if nonvalidator_diff <= delta:
```
pmikolajczyk41, line 74: probably I've some blackout, but why it is _small_ difference that bothers us?

local-tests/test_multiple_restarts.py
```diff
@@ -2,27 +2,28 @@
 import os
 import sys
 from os.path import abspath, join
-from time import sleep
+from time import sleep, ctime
 
 from chainrunner import Chain, Seq, generate_keys, check_finalized
 
+def printt(s): print(ctime() + ' | ' + s)
```
pmikolajczyk41, line 9: what about `print = lambda s: print(ctime() + ' | ' + s)` :stuck_out_tongue:

local-tests/test_multiple_restarts.py
```diff
@@ -33,49 +34,40 @@
...
 
-    print('Restarting nodes')
-    finalized_before_start_per_node = check_finalized(chain)
+    finalized_before_start = check_finalized(chain)
 
     # Check if the finalization didn't stop after a kill.
-    # Use a reduced rate since 1/4 nodes are offline, and fudge by 0.9 to give a margin of error.
-    if finalized_before_start_per_node[0] - finalized_before_kill_per_node[0] < stop_duration * (3/4) * 0.9:
-        print('Finalization stalled')
+    if finalized_before_start[0] - finalized_before_kill[0] < 10:
```
pmikolajczyk41, line 56: let's keep this `10` as `delta`

local-tests/test_multiple_restarts.py
```diff
@@ -33,49 +34,40 @@
...
+        printt('Finalization stalled')
         sys.exit(1)
 
-    chain.start('aleph', nodes=[3, 4])
-
-    print(f'Waiting {catch_up_duration}s for catch up')
-    sleep(catch_up_duration)
-    finalized_after_catch_up_per_node = check_finalized(chain)
+    printt('Restarting nodes')
+    chain[3].start('aleph')
```
pmikolajczyk41, line 61: let's extract all `aleph`s to `chain-name`, just to be explicit about a single chain :P

local-tests/test_multiple_restarts.py
```diff
@@ -33,49 +34,40 @@
...
-        print(f"too small catch up for nonvalidators: {nonvalidator_diff}")
-        sys.exit(1)
+    finalized_after = check_finalized(chain)
+    diff = finalized_after[3] - finalized_before_start[3]
+    delta = 5
 
-    if validator_diff <= ALLOWED_DELTA:
-        print(f"too small catch up for validators: {validator_diff}")
+    # Check if the murdered node started catching up with reasonable nr of blocks.
+    if diff <= delta:
```
pmikolajczyk41, line 71: the same question as in the other test

local-tests/test_multiple_restarts.py
```diff
@@ -33,49 +34,40 @@
...
-    sleep(run_duration)
-
-    finalized_before_kill_per_node = check_finalized(chain)
-    print('Killing one validator and one nonvalidator')
+printt('Waiting for finalization')
+chain.wait_for_finalization(0)
+printt('Waiting for authorities')
+chain.wait_for_authorities()
 
+for sleep_duration in [21, 37, 15]:
```
pmikolajczyk41, line 45: okay, I get papal numbers, but I'm afraid that the experiment will not fit into a single session as `README.md` claims (at least with `short-session` feature)

pmikolajczyk41 (approved):
cool, nice improvement

I left a couple of minor comments and I think that you can handle all of them by yourself, so :white_check_mark: from me

local-tests/chainrunner/chain.py
h4nsu: Much less readable in my opinion. And requires new import ;)

local-tests/chainrunner/node.py
h4nsu: Someone was complaing that `purge` on an empty chain throws exception due to check and that makes it impossible to mimic the run_nodes workflow:
1. purge
2. bootstrap
3. run

local-tests/chainrunner/node.py
h4nsu: Nope. The idea behind this `backup` switch is to be able to run also older binaries, where the backup-path/no-backup was not yet a thing.

local-tests/chainrunner/node.py
h4nsu: I'm just trying to keep the `-`/`_` compatibility everywhere. No one will probably ever need that, but it's better to stay consistent

local-tests/test_multiple_restarts.py
h4nsu: "To understand recursion, you must first understand recursion" ;)

local-tests/test_multiple_restarts.py
h4nsu: 'aleph' is just the name of the logfile in this context. I can imagine someone working on this test failing who would like to use different names for different stages.

local-tests/test_multiple_restarts.py
h4nsu: This test is not meant to be run with `short-session` (see line 13)

local-tests/test_catch_up.py
h4nsu: I don't understand :/ Nodes 0,1,2,3 are validators, while 4 and 5 are non-validators.

local-tests/test_catch_up.py
h4nsu: With `max` we can compare two numbers from two different nodes. How is this better?

local-tests/chainrunner/node.py
pmikolajczyk41: so for newer binaries we always use backup, right?

local-tests/test_catch_up.py
pmikolajczyk41: I was always bad at indices and digits in general :disappointed:

local-tests/test_catch_up.py
pmikolajczyk41: `max` stands for consensus state in general; possible difference between `max` and single node state should be only due to communication layer

local-tests/test_multiple_restarts.py
pmikolajczyk41: uhm, I thought that new `print` will be visible after declaration :|

local-tests/test_catch_up.py
h4nsu: Because that means finalization did not restart

local-tests/test_multiple_restarts.py
h4nsu: the same answer ;)

local-tests/chainrunner/node.py
h4nsu: Yes, for now `--no-backup` is not supported

local-tests/test_catch_up.py
h4nsu: Please mind the fact that this information is taken from the logfile, where the corresponding line is put once every 5 seconds. So it's not only the communication layer, differences mostly come from the fact that the general "heartbeat" of the whole thing being these long 5 seconds and nodes obviously don't synchronize when to put that line into logs.

maciejnems (approved):

PR #499, opened 2022-06-23
[Not for merging] A0-912: disabling mechanics

pmikolajczyk41:
For presentation use only.

### DIFF
--- a/pallets/elections/src/lib.rs
+++ b/pallets/elections/src/lib.rs
@@ -64,7 +64,7 @@ pub mod pallet {
         pallet_prelude::{BlockNumberFor, OriginFor},
     };
     use pallet_session::SessionManager;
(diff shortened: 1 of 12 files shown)


pallets/elections/src/lib.rs
```diff
@@ -271,6 +271,13 @@ pub mod pallet {
         NonUniqueListOfValidators,
     }
 
+    impl<T: Config> OnValidatorDisabled for Pallet<T> {
+        fn disable_validator(_validator_index: u32) {
```
deuszx, line 275: Is this unimplemented on purpose?

pmikolajczyk41: yes yes, this is only for presentation and I wasn't sure how proper handling should look like here :man_shrugging:

PR #500, opened 2022-06-28
A0-1092 logging number of discovered authorities at info

DamianStraszak:
# Description

Switched logging number of discovered authorities in network to info. Crucial for diagnostics in production.

### DIFF
--- a/finality-aleph/src/network/manager/discovery.rs
+++ b/finality-aleph/src/network/manager/discovery.rs
@@ -6,7 +6,7 @@ use crate::{
     NodeIndex, SessionId,
 };
 use codec::{Decode, Encode};
-use log::{debug, trace, warn};
+use log::{debug, info, trace, warn};
 use std::{
     collections::HashMap,
     marker::PhantomData,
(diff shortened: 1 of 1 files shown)


timorl (approved):
We should probably implement `Display` for `NodeCount` and our other types at some point. Just a thought triggered by this PR, no action here needed ofc.

finality-aleph/src/network/manager/discovery.rs
```diff
@@ -84,7 +84,7 @@ impl<M: Multiaddress> Discovery<M> {
 
         let missing_authorities = handler.missing_nodes();
         let node_count = handler.node_count();
-        debug!(target: "aleph-network", "{:?}/{:?} authorities known for session {:?}.", node_count.0-missing_authorities.len(), node_count, handler.session_id());
+        info!(target: "aleph-network", "{}/{} authorities known for session {}.", node_count.0-missing_authorities.len(), node_count.0, handler.session_id().0);
```
deuszx, line 87: Sometimes it's useful to know the keys of the authorities as well.

DamianStraszak: Yes -- more info about current session authorities is logged at a different place actually. This line prints only some minimal info on the status of network connectivity between authorities :)

PR #501, opened 2022-06-28
A0-1083: Fix flaky finalization start

maciejnems:


local-tests/test_multiple_restarts.py
```diff
@@ -11,7 +11,7 @@
 # Path to the aleph-node binary (important DON'T use short-session feature):
 binary = abspath(os.getenv('ALEPH_NODE_BINARY', join(workdir, 'aleph-node')))
 
-phrases = [f'//{i}' for i in range(6)]
+phrases = [f'//{i}' for i in range(5)]
```
timorl, line 14: Huh, why 6→5?

timorl (approved):
Looks good, one small question only.

local-tests/test_multiple_restarts.py
maciejnems: It goes with my hypothesis, that 6 nodes is a little bit much for our pipeline. I see the point of 1 non validator which is killed, but I don't think that 1 non validator that is not killed is needed. I believe in the past we had some pipelines failing because they were running too many nodes.

timorl: Perhaps this should end up in a comment here then, sounds like relevant information and we don't really have it anywhere outside our brains. <_<"

maciejnems: This seems like a strange thing for adding here :// there are other tests where this also might be a case (like test_catch_up). Comment in a specific place of code makes sense in context of this change, but seems to me like we should either fully diagnose it and document in a more global place... or maybe this is not the case and we have some strange bug that makes this test more flaky on github. Also maybe this comment should be added in all places where we define the number of validators for a test.

timorl: Hm, very good point. I'm not sure where the info should be either... Perhaps some constant (`MAX_PIPELINE_NODES`?) in a common file for the tests? Do we even have such a thing?

h4nsu (approved):
The fix stops working when `chain_type` is set to `live` and no `public_addr` flag is given to the nodes. Is this the intended behavior?

local-tests/test_multiple_restarts.py
maciejnems: Actually I don't think we will decrease the number of nodes. Test seem flaky anyway :( This would be more fixed after we implement wait_for_finalization in tests

PR #503, opened 2022-06-29
A0-1041: Treasury code cleanup

pmikolajczyk41:
# Description

Moved some functionalities from e2e to aleph-client. Local refactors.

Next steps (another PRs):
 - [unify reading storage/constants in aleph-client](https://cardinal-cryptography.atlassian.net/browse/A0-1096)
 - [extend cliain with treasury commands](https://cardinal-cryptography.atlassian.net/browse/A0-1098)

## Type of change

# Checklist:

### DIFF
--- a/aleph-client/src/balances.rs
+++ b/aleph-client/src/balances.rs
@@ -1,0 +1,13 @@
+use crate::AnyConnection;
+use primitives::Balance;
+
+/// Reads from the storage how much balance is currently on chain.
+///
+/// Performs a single storage read.
+pub fn total_issuance<C: AnyConnection>(connection: &C) -> Balance {
+    connection
+        .as_connection()
+        .get_storage_value("Balances", "TotalIssuance", None)
+        .expect("Key `Balances::TotalIssuance` should be present in storage")
+        .unwrap_or(0)
+}
--- a/aleph-client/src/balances.rs
+++ b/aleph-client/src/balances.rs
@@ -9,5 +9,5 @@ pub fn total_issuance<C: AnyConnection>(connection: &C) -> Balance {
         .as_connection()
         .get_storage_value("Balances", "TotalIssuance", None)
         .expect("Key `Balances::TotalIssuance` should be present in storage")
-        .unwrap_or(0)
+        .unwrap()
 }
--- a/aleph-client/src/treasury.rs
+++ b/aleph-client/src/treasury.rs
@@ -2,7 +2,6 @@ use crate::{try_send_xt, wait_for_event, AnyConnection, RootConnection, SignedCo
 use ac_primitives::ExtrinsicParams;
 use codec::Decode;
 use frame_support::PalletId;
-use log::info;
 use primitives::Balance;
 use sp_core::{Pair, H256};
 use sp_runtime::{traits::AccountIdConversion, AccountId32};
@@ -45,12 +44,7 @@ pub fn staking_treasury_payout<C: AnyConnection>(connection: &C) -> Balance {
         .expect("Constant `Timestamp::MinimumPeriod` should be present");
 
     let millisecs_per_era = millisecs_per_block * session_period as u64 * sessions_per_era as u64;
-    let treasury_era_payout_from_staking = primitives::staking::era_payout(millisecs_per_era).1;
-    info!(
-        "[+] Possible treasury gain from staking is {}",
-        treasury_era_payout_from_staking
-    );
-    treasury_era_payout_from_staking
+    primitives::staking::era_payout(millisecs_per_era).1
 }
 
 /// Creates a proposal of spending treasury's funds.
@@ -82,10 +76,7 @@ fn wait_for_rejection<C: AnyConnection>(connection: &C, proposal_id: u32) -> Any
     wait_for_event(
         connection,
         ("Treasury", "Rejected"),
-        |e: ProposalRejectedEvent| {
-            info!("[+] Rejected proposal {:?}", e.proposal_id);
-            proposal_id.eq(&e.proposal_id)
-        },
+        |e: ProposalRejectedEvent| proposal_id.eq(&e.proposal_id),
     )
     .map(|_| ())
 }
@@ -107,7 +98,7 @@ fn send_rejection(connection: &RootConnection, proposal_id: u32) -> ApiResult<Op
 
 /// Rejects proposal with id `proposal_id` and waits for the corresponding event.
 ///
-/// Fails if either sending extrinsic failed or the expected event has not been observed.
+/// Be careful, since execution might never end (we may stuck on waiting for the event forever).
 pub fn reject(connection: &RootConnection, proposal_id: u32) -> AnyResult<()> {
     let listener = {
         let (c, p) = (connection.clone(), proposal_id);
@@ -142,22 +133,19 @@ fn wait_for_approval<C: AnyConnection>(connection: &C, proposal_id: u32) -> AnyR
             .expect("Key `Treasury::Approvals` should be present in storage")
             .unwrap_or_default();
         if approvals.contains(&proposal_id) {
-            info!("[+] Proposal {:?} approved successfully", proposal_id);
             return Ok(());
         } else {
-            info!(
-                "[+] Still waiting for approval for proposal {:?}",
-                proposal_id
-            );
             sleep(Duration::from_millis(500))
         }
     }
 }
 
-/// Approves proposal with id `proposal_id` and waits (in the loop) until pallet storage is updated.
+/// Approves proposal with id `proposal_id` and waits (in a loop) until pallet storage is updated.
 ///
 /// Unfortunately, pallet treasury does not emit any event (like while rejecting), so we have to
 /// keep reading storage to be sure. Hence, it may be an expensive call.
+///
+/// Be careful, since execution might never end.
 pub fn approve(connection: &RootConnection, proposal_id: u32) -> AnyResult<()> {
     send_approval(connection, proposal_id)?;
     wait_for_approval(connection, proposal_id)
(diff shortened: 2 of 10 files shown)


aleph-client/src/treasury.rs
```diff
@@ -0,0 +1,164 @@
...
+        .as_connection()
+        .get_constant::<u64>("Timestamp", "MinimumPeriod")
+        .expect("Constant `Timestamp::MinimumPeriod` should be present");
+
+    let millisecs_per_era = millisecs_per_block * session_period as u64 * sessions_per_era as u64;
+    let treasury_era_payout_from_staking = primitives::staking::era_payout(millisecs_per_era).1;
+    info!(
+        "[+] Possible treasury gain from staking is {}",
+        treasury_era_payout_from_staking
+    );
```
obrok, line 52: Question: why does this function log? It feels a little weird to me that a getter like this would log... Makes more sense to me that whatever is using this would potentially choose to log the value in its own context. This doesn't seem to be a part of some grand pattern of logging every response either - e.g. `proposals_counter` does not log its output.

aleph-client/src/treasury.rs
```diff
@@ -0,0 +1,164 @@
...
+
+fn wait_for_approval<C: AnyConnection>(connection: &C, proposal_id: u32) -> AnyResult<()> {
+    loop {
+        let approvals: Vec<u32> = connection
+            .as_connection()
+            .get_storage_value("Treasury", "Approvals", None)
+            .expect("Key `Treasury::Approvals` should be present in storage")
+            .unwrap_or_default();
+        if approvals.contains(&proposal_id) {
+            info!("[+] Proposal {:?} approved successfully", proposal_id);
```
obrok, line 145: Comment: same here, these logs seem out of place to me in a client lib.

aleph-client/src/treasury.rs
```diff
@@ -0,0 +1,164 @@
...
+            info!(
+                "[+] Still waiting for approval for proposal {:?}",
+                proposal_id
+            );
+            sleep(Duration::from_millis(500))
+        }
+    }
+}
+
+/// Approves proposal with id `proposal_id` and waits (in the loop) until pallet storage is updated.
```
obrok, line 157:
```suggestion
/// Approves proposal with id `proposal_id` and waits (in a loop) until pallet storage is updated.
```

obrok (approved):
LGTM overall, but concerned about publishing a function that can block forever, as indicated.

aleph-client/src/treasury.rs
pmikolajczyk41: I hope I will learn articles before retiring

aleph-client/src/treasury.rs
pmikolajczyk41: I removed all logs in this module

aleph-client/src/treasury.rs
pmikolajczyk41: yeah, ugly leftover - I removed it

aleph-client/src/treasury.rs
```diff
@@ -0,0 +1,164 @@
...
+///
+/// Unfortunately, pallet treasury does not emit any event (like while rejecting), so we have to
+/// keep reading storage to be sure. Hence, it may be an expensive call.
+pub fn approve(connection: &RootConnection, proposal_id: u32) -> AnyResult<()> {
```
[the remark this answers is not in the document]
pmikolajczyk41: yes, this is confusing, but I didn't intend to change things too much here; the problem lies deeper, in `wait_for_event` method - we should come up with a better listening mechanics in general;

in this case I just added warning in the docs

aleph-client/src/treasury.rs
obrok: I don't think there is any hope for us - as Slavs we just don't have articles in our blood.

aleph-client/src/treasury.rs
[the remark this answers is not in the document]
obrok: :+1:

aleph-client/src/balances.rs
```diff
@@ -0,0 +1,13 @@
...
+
+/// Reads from the storage how much balance is currently on chain.
+///
+/// Performs a single storage read.
+pub fn total_issuance<C: AnyConnection>(connection: &C) -> Balance {
+    connection
+        .as_connection()
+        .get_storage_value("Balances", "TotalIssuance", None)
+        .expect("Key `Balances::TotalIssuance` should be present in storage")
+        .unwrap_or(0)
```
Marcin-Radecki, line 12: Here we change the semantics of this method, previously we `panic`ed when `TotalIssuance` cannot be decoded, now we return 0. Former seems to be more defensive but in light of this is in library, imho we want to be more strict just in case we slip a false positive e2e test run only if storage was borked (unlikely, I know) by somehow e2e passing with 0 value (depends on e2e test logic in general case, agree with currents tests we would catch error anyway ). One can even say this is a major semver change,

This can be fixed by passing some lambda what user of this function want to do in case entry cannot be decoded, ie wither `unwrap` anyway or return 0 etc

Another possibility is that we would go into direction that upon such error scenario we return the default value for given storage value

aleph-client/src/treasury.rs
```diff
@@ -0,0 +1,152 @@
...
+pub fn staking_treasury_payout<C: AnyConnection>(connection: &C) -> Balance {
+    let sessions_per_era = connection
+        .as_connection()
+        .get_constant::<u32>("Staking", "SessionsPerEra")
+        .expect("Constant `Staking::SessionsPerEra` should be present");
+    let session_period = connection
+        .as_connection()
+        .get_constant::<u32>("Elections", "SessionPeriod")
+        .expect("Constant `Elections::SessionPeriod` should be present");
+    let millisecs_per_block = 2 * connection
```
Marcin-Radecki, line 41: I know this was in the original code already, but this `2*` is quite sneaky meaning it makes code more coupled - ie technically this line now depends on line https://github.com/Cardinal-Cryptography/aleph-node/blob/main/bin/runtime/src/lib.rs#L525 . On the other hand I don't have an idea of how to make this read from storage so I think this was the only way then

aleph-client/src/treasury.rs
```diff
@@ -0,0 +1,152 @@
...
+    connection
+        .as_connection()
+        .get_storage_value("Treasury", "ProposalCount", None)
+        .expect("Key `Treasury::ProposalCount` should be present in storage")
+        .unwrap_or(0)
+}
+
+/// Calculates how much balance will be paid out to the treasury after each era.
+///
+/// Every call to this method is potentially expensive - it requires three storage reads.
```
Marcin-Radecki, line 31: Depends how we look at it - those get key-value pair RPC calls are not that expensive as e.g. sending  extrinsic that requires more encoding usually. Making this comment only because other methods in this file does not have such comment. If we take metric of how many TCP calls we need to do under the hood then I agree 3 requests are more than 1 :)

aleph-client/src/treasury.rs
```diff
@@ -0,0 +1,152 @@
+use crate::{try_send_xt, wait_for_event, AnyConnection, RootConnection, SignedConnection};
```
Marcin-Radecki, line 1: In general to this file: what do you think to make `pub fn`s organized as first in this file, followed by private `fn`? Motivation is for a reader which usually wants to read public API of this module first

Marcin-Radecki (changes requested):
Nice rework, code is cleaner.

aleph-client/src/treasury.rs
obrok: This is also my preferred order (public functions first, and then the functions used by them, toposorted), but I was under the impression that most files in our codebase toposort the other way and was keeping to that. Is there any agreed practice in place?

aleph-client/src/balances.rs
pmikolajczyk41: I'm about to start implementing such task: https://cardinal-cryptography.atlassian.net/browse/A0-1096, where such concerns have even broader consequences. I propose leaving this particular method as it is (or I can revert the behavior - whatever you prefer) and then, in the next PR, I will provide two methods for reading storage (and analogous pair for constants):
 - `read_storage(connection: AnyConnection, module: String, key: String)` - panicable, force reading
 - `read_storage_or_else(connection: AnyConnection, module: String, key: String, handler: Fn: () -> T)` - defensive reading with possibility of providing default or some handling

wdyt?

aleph-client/src/treasury.rs
pmikolajczyk41: yeah, you are right, I was trying to be kinda informative, but it was silly; removed

aleph-client/src/treasury.rs
pmikolajczyk41: I changed it to read directly from `primitives`

aleph-client/src/treasury.rs
pmikolajczyk41: I reordered it; this is also my taste, but some time ago both reviewers were opting for absolute toposort, hence this previous outcome; but we should fight for our feelings!

obrok: Honestly, I feel like `rustfmt` or, at the most, `clippy` should just enforce the chosen toposort :)

aleph-client/src/balances.rs
Marcin-Radecki: Great, let's revert this particular behavior if it's not a big problem (let's try to be consistent with this semver semantics). 

Those two methods sound like a decent design.

aleph-client/src/treasury.rs
Marcin-Radecki: Yes, we should at some point enforce this by the tooling.

Marcin-Radecki (approved):

PR #506, opened 2022-06-30
Uniform import order

pmikolajczyk41:
# Description

This is a proposition of having one strict imports order, so that we avoid dumb changes in PRs that comes from different local settings.

## Type of change

# Checklist:

### DIFF
--- a/aleph-client/Cargo.lock
+++ b/aleph-client/Cargo.lock
@@ -93,7 +93,7 @@ dependencies = [
 
 [[package]]
 name = "aleph_client"
-version = "1.2.0"
+version = "1.2.1"
 dependencies = [
  "ac-primitives",
  "anyhow",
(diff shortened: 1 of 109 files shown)


DamianStraszak (approved):

artemis-cc (approved):
:+1: Does what it says on the tin; looks good to me.

That said, I haven't seen many others weighing in on whether we _should_ do this, so I'm not sure we should actually merge it until we get more feedback (or at least a longer period without objections :stuck_out_tongue:).

pmikolajczyk41:
@artemis-cc I talked with hans and he was good with it, so I do not expect any strong objections :) however, in case any appear, we will rediscuss the issue; for now I think it is safe to merge it; at least we gain a consistent import order in the current code

PR #508, opened 2022-06-30
Match changed system_version rpc method in pipelines for testnet.

bartoszjedrzejewski:
New substrate version changed syntax of system_version rpc call, so pipelines needed matching.


pmikolajczyk41 (approved):
since this command:
`(curl -s -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "system_version"}' https://rpc.test.azero.dev | jq -r '.result' | cut -d "-" -f 2 | head -c 7)` returns `650e6cc`, I believe we are good now :)

PR #511, opened 2022-07-04
Add e2e test checking points calculation after stake change

krzysztofziobro:
# Description

Adding a new e2e test that checks if points are calculated correctly after validators have bonded additional funds.

## Type of change

Added new e2e test.


e2e-tests/src/test/rewards.rs
```diff
@@ -40,6 +41,127 @@ fn get_non_reserved_members_for_session(config: &Config, session: SessionIndex)
...
+
+    stash_accounts_key_pairs
+        .iter()
+        .zip(additional_stakes.iter())
+        .for_each(|(key, additional_stake)| {
+            let id = account_from_keypair(key);
+            balances_transfer(
+                &root_connection.as_signed(),
+                &id,
+                *additional_stake + TOKEN,
```
pmikolajczyk41, line 72: Question: why `+TOKEN`? if this is for being over existential deposit, then I would either add comment or create some constant like `MIN_ACCOUNT_DEPOSIT = TOKEN` and use it here

e2e-tests/src/test/rewards.rs
```diff
@@ -40,6 +41,127 @@ fn get_non_reserved_members_for_session(config: &Config, session: SessionIndex)
...
+                XtStatus::InBlock,
```
pmikolajczyk41, line 73: Question: don't we want `Finalized`?

e2e-tests/src/test/rewards.rs
```diff
@@ -40,6 +41,127 @@ fn get_non_reserved_members_for_session(config: &Config, session: SessionIndex)
...
+        .map(account_from_keypair)
+        .collect();
+
+    balances_batch_transfer(&root_connection.as_signed(), controller_accounts, TOKEN);
+
+    stash_accounts_key_pairs
+        .iter()
+        .zip(additional_stakes.iter())
+        .for_each(|(key, additional_stake)| {
+            let id = account_from_keypair(key);
```
pmikolajczyk41, line 68: Suggestion: please rename this variable to smth more explicit

e2e-tests/src/test/rewards.rs
```diff
@@ -40,6 +41,127 @@ fn get_non_reserved_members_for_session(config: &Config, session: SessionIndex)
     non_reserved.iter().map(account_from_keypair).collect()
 }
 
+fn validators_bond_extra_stakes(config: &Config, additional_stakes: Vec<u128>) {
```
pmikolajczyk41, line 44: Question: should we put such function in `aleph-client`? Or is it too specific?

e2e-tests/src/test/rewards.rs
```diff
@@ -40,6 +41,127 @@ fn get_non_reserved_members_for_session(config: &Config, session: SessionIndex)
...
+    let root_connection = config.create_root_connection();
+
+    let reserved_members: Vec<_> = get_reserved_members(config)
+        .iter()
+        .map(account_from_keypair)
+        .collect();
+    let non_reserved_members: Vec<_> = get_non_reserved_members(config)
+        .iter()
+        .map(account_from_keypair)
+        .collect();
```
pmikolajczyk41, line 99: Suggestion: what about changing these (and twin lines in the other test) to some function like `members_accounts` which would return two vecs?

pmikolajczyk41 (approved):
basically lgtm

e2e-tests/src/test/rewards.rs
krzysztofziobro: Probably a good idea

e2e-tests/src/test/rewards.rs
krzysztofziobro: If I understand correctly, we need some additional funds to make sure that fees can be covered. I will add a comment.

e2e-tests/src/test/rewards.rs
pmikolajczyk41: one left afais

e2e-tests/src/test/rewards.rs
```diff
@@ -40,23 +41,151 @@ fn get_non_reserved_members_for_session(config: &Config, session: SessionIndex)
...
-    let reserved_members: Vec<_> = get_reserved_members(config)
+    let validator_seeds = get_validators_seeds(config);
+    let controller_seeds: Vec<_> = validator_seeds
         .iter()
-        .map(account_from_keypair)
+        .map(|v| format!("{}//Controller", v))
         .collect();
-    let non_reserved_members: Vec<_> = get_non_reserved_members(config)
+
+    let controller_accounts_key_pairs = accounts_seeds_to_keys(&controller_seeds);
```
fixxxedpoint, line 67: Maybe we can keep this `knowledge` of which key is what in `config.rs/accounts.rs` using that `NodeKeys` struct or something similar, i.e. add method `get_node_keys -> Vec<NodeKeys>`. No more questions about keys...:)

e2e-tests/src/test/rewards.rs
```diff
@@ -40,23 +41,151 @@ fn get_non_reserved_members_for_session(config: &Config, session: SessionIndex)
...
+        ]
+        .to_vec(),
+    );
+
+    let session_period = get_session_period(&connection);
+    let sessions_per_era = get_sessions_per_era(&connection);
+
+    let era_block = session_period * sessions_per_era * wait_for_next_era(&root_connection)?;
+    let era_block_hash = get_block_hash(&connection, era_block);
+    let start_era_session = get_session(&connection, Some(era_block_hash));
```
fixxxedpoint, line 140: Can be simplified to `sessions_per_era * wait_for_next_era(&root_connection)`?

e2e-tests/src/test/rewards.rs
```diff
@@ -40,23 +41,151 @@ fn get_non_reserved_members_for_session(config: &Config, session: SessionIndex)
...
+
+    let next_era_block = session_period * sessions_per_era * wait_for_next_era(&root_connection)?;
+    let next_era_block_hash = get_block_hash(&connection, next_era_block);
+    let end_era_session = get_session(&connection, Some(next_era_block_hash));
```
fixxxedpoint, line 144: Can be simplified to `sessions_per_era * wait_for_next_era(&root_connection)`?

fixxxedpoint (changes requested):
LGTM. Left few small comments.

fixxxedpoint (approved):
well done

PR #512, opened 2022-07-04
A0-989: Force new era e2e test for rewards per stake and uptime

maciejzelaszczyk:
# Description

This test is part of a slew of tests for rewards based on stake and uptime. The individual test case here simulates a situation when a new era is forced by the `sudoer`. We check reward points immediately following the `ForceNewEra` extrinsic is included and in the two subsequent sessions. Overall, we want to test that the points are calculated correctly (within a margin of error) in sessions where the new era is not yet in effect (sessions `k` and `k + 1` where `k` is the session in which we make the call) and the first session of the forced era (`k + 2`).

In a slightly tangential modification, we change the stakes set in `chain_spec.rs` to vary - previously they were uniform.

## Type of change

# Checklist:

- I have added tests


fixxxedpoint (approved):
LGTM. A follow up for this and @krzysztofziobro work should be a test that changes `staking` (changes significantly staking proportions) and then forces a new era.

e2e-tests/src/test/rewards.rs
```diff
@@ -109,3 +120,74 @@ pub fn disable_node(config: &Config) -> anyhow::Result<()> {
...
+        let non_reserved_for_session =
+            get_non_reserved_members_for_session(config, session_to_check);
+        let non_reserved_bench =
+            get_bench_members(non_reserved_members.clone(), &non_reserved_for_session);
+
+        check_points(
+            &connection,
+            session_to_check,
+            era_to_check,
+            reserved_members
```
piotrMocz, line 183: Nit: I think I would extract it into a separate `let` binding just for the sake of naming it (feel free to ignore this :))

piotrMocz (approved):
Looks good to me 👍

e2e-tests/src/test/rewards.rs
maciejzelaszczyk: Good call. Made me think some more about the naming convention here and I included this along with minuscule changes in other variable names.

piotrMocz (approved):

PR #518, opened 2022-07-07
A0-895: aleph-party: Delay start until initial sync is done or finalization stops progressing

artemis-cc:
# Description

This may also fix the excessive volume of errors observed during initial sync (though I was unable to reproduce that).

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/party/mod.rs
+++ b/finality-aleph/src/party/mod.rs
@@ -16,7 +16,7 @@ use sp_runtime::traits::{Block, Header};
 use tokio::{task::spawn_blocking, time::sleep};
 
 use crate::{
-    crypto::{AuthorityPen, AuthorityVerifier, KeyBox},
+    crypto::{AuthorityPen, AuthorityVerifier, Keychain},
     data_io::{ChainTracker, DataStore, OrderedDataInterpreter},
     default_aleph_config,
     justification::{AlephJustification, JustificationNotification, Verifier},
@@ -153,7 +153,7 @@ where
     async fn spawn_authority_subtasks(
         &self,
         node_id: NodeIndex,
-        multikeychain: KeyBox,
+        multikeychain: Keychain,
         data_network: SessionNetwork<SplitData<B>>,
         session_id: SessionId,
         authorities: Vec<AuthorityId>,
@@ -241,7 +241,7 @@ where
                 .await
                 .expect("The keys should sign successfully");
 
-        let keybox = KeyBox::new(node_id, authority_verifier.clone(), authority_pen.clone());
+        let keychain = Keychain::new(node_id, authority_verifier.clone(), authority_pen.clone());
 
         let data_network = self
             .session_manager
@@ -253,7 +253,7 @@ where
         let authority_subtasks = self
             .spawn_authority_subtasks(
                 node_id,
-                keybox,
+                keychain,
                 data_network,
                 session_id,
                 authorities,
@@ -282,7 +282,7 @@ where
 
         // Early skip attempt -- this will trigger during catching up (initial sync).
         if self.client.info().best_number >= last_block {
-            // We need to give the JustificationHandler some time to pick up the keybox for the new session,
+            // We need to give the JustificationHandler some time to pick up the keychain for the new session,
             // validate justifications and finalize blocks. We wait 2000ms in total, checking every 200ms
             // if the last block has been finalized.
             for attempt in 0..10 {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -84,9 +84,9 @@ dependencies = [
 
 [[package]]
 name = "aleph-bft"
-version = "0.14.0"
+version = "0.15.0"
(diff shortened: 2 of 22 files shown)


finality-aleph/src/party/mod.rs
```diff
@@ -432,9 +432,20 @@ where
     }
 
     pub async fn run(mut self) {
-        let last_finalized_number = self.client.info().finalized_number;
+        let mut finalized_number = self.client.info().finalized_number;
+        let mut previous_finalized_number;
+        loop {
```
timorl, line 437: The above two lines are pretty unusual style for our code. I feel like it might be more readable if you:
1. explicitly initialize `previous_finalized_number`
2. replace the `loop` with a `while` – I think you could do 1. in a way that ensures the numbers differ (e.g. `finalized_number + 1`) and that would just work. also we save 1 second in the optimistic case :D
The second point would probably require a comment for why the initialization is weird, but this would still be more readable than this? Disputable I guess.

timorl (approved):
Looks good, only one readability complaint.

finality-aleph/src/party/mod.rs
```diff
@@ -432,9 +432,20 @@ where
...
+            sleep(Duration::from_secs(1)).await;
```
fixxxedpoint, line 438: Perhaps just name the magic number for better readability, e.g. EXPECTED_TIME_TO_NEW_FINALIZED_BLOCK ( could be longer ;) )

fixxxedpoint (approved):
LGTM. Just left one suggestion.

finality-aleph/src/party/mod.rs
```diff
@@ -432,9 +432,17 @@ where
     }
 
     pub async fn run(mut self) {
-        let last_finalized_number = self.client.info().finalized_number;
+        let mut finalized_number = self.client.info().finalized_number;
```
fixxxedpoint, line 435: Suggestion: move this into a seperate method -  it would keep that `run` method minimalistic and tidy + its name would be part of `documentation`.

finality-aleph/src/network/mod.rs
```diff
@@ -126,6 +126,13 @@ pub trait RequestBlocks<B: Block>: Clone + Send + Sync + 'static {
     /// Clear all pending justification requests. We need this function in case
     /// we requested a justification for a block, which will never get it.
     fn clear_justification_requests(&self);
+
+    /// Are we in the process of downloading the chain?
+    ///
+    /// Like [`NetworkService::is_major_syncing`][1].
+    ///
+    /// [1]: sc_network::NetworkService::is_major_syncing
+    fn is_major_syncing(&self) -> bool;
```
fixxxedpoint, line 135: Question/suggestion: is this method necessary here? It doesn't look for me like it belong here, so perhaps we shouldn't extend this trait with some ad-hoc method like this. Or perhaps some other consumers of this trait are going to care about its syncing state?

finality-aleph/src/party/mod.rs
fixxxedpoint: Yeah, it sounds like an overkill. :) Since it's arbitrary and less than 1hr (or anything that can confuse a user) I leave it up to you.

timorl: > (it's how long we sleep before checking if finalized_number has changed)

Oh, I only now noticed this, in this case it definitely should be below 1s – in normal operation we finalize 1 block/s, so this would slow down many nodes joining (they would eventually join because of random delays, but bleh). I would say something in the range 0.5-0.8s would be good.

finality-aleph/src/network/mod.rs
timorl: I suggested adding this here, so perhaps I'll explain my reasoning. <_<" I thought it would fit here relatively well as far as responsibilities of components go – one component is responsible for syncing blocks, and information about whether we are in a major sync is part of these responsibilities. On the other hand I agree the API users are only interested in parts of the API, but I'm also not sure it will stay that way.

fixxxedpoint: How about creating a new trait that extends `RequestBlock` (or perhaps composition fits here better) - only party.rs would see that new method. It also looks like an implementation detail of how networking is implemented currently, users of this - I would even consider having a separate trait for something `low-level` like this.

timorl: I'm fine with either approach (as well as leaving it as-is).

I don't think it's strictly (at most partially) an implementation detail – whatever component takes care of block syncing is extremely likely to be aware of "major syncing" or a similar state regardless of implementation. The fact that it's part of network is an implementation detail though.

timorl (approved):
LGTM

finality-aleph/src/party/mod.rs
fixxxedpoint: lgtm. Is there any reason for making it public?

fixxxedpoint (approved):
lgtm

PR #519, opened 2022-07-08
Rotating reserved nodes

kostekIV:
# Description
Added support for rotating reserved nodes.

Migrations tested on local node with latest main and on testnet with try-runtime.

## Type of change

# Checklist:

### DIFF
--- a/pallets/elections/src/lib.rs
+++ b/pallets/elections/src/lib.rs
@@ -41,12 +41,34 @@ pub struct EraValidators<AccountId> {
 
 impl<AccountId> Default for EraValidators<AccountId> {
     fn default() -> Self {
-        EraValidators {
+        Self {
             reserved: vec![],
             non_reserved: vec![],
         }
     }
 }
+
+#[derive(Decode, Encode, TypeInfo, Debug, Clone, Copy, PartialEq)]
+pub struct CommitteeSeats {
+    pub reserved_seats: u32,
+    pub non_reserved_seats: u32,
+}
+
+impl CommitteeSeats {
+    fn size(&self) -> u32 {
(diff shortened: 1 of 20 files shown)


pallets/elections/src/lib.rs
```diff
@@ -133,14 +151,17 @@ pub mod pallet {
...
+                }
                 _ => Err("Bad storage version"),
             }
         }
         #[cfg(feature = "try-runtime")]
         fn post_upgrade() -> Result<(), &'static str> {
             let on_chain = <Pallet<T> as GetStorageVersion>::on_chain_storage_version();
             match on_chain {
-                _ if on_chain == STORAGE_VERSION => migrations::v1_to_v2::post_upgrade::<T, Self>(),
+                _ if on_chain == STORAGE_VERSION => migrations::v2_to_v3::post_upgrade::<T, Self>(),
```
piotrMocz, line 164: Shouldn't we somehow match it so that we can migrate from any version to the latest? I've seen it somewhere in Polkadot I think

piotrMocz (approved):
Looks great!

pallets/elections/src/lib.rs
kostekIV: We should and it is done in the fn `on_runtime_upgrade` line 116. This is fn for performing checks after the upgrade already happened.

kostekIV: We should and it is done in the fn `on_runtime_upgrade` line 116. This is fn for performing checks after the upgrade already happened.

kostekIV: We should and it is done in the fn `on_runtime_upgrade` line 116. This is fn for performing checks after the upgrade already happened.

PR #521, opened 2022-07-11
Shared contract traits

fbielejec:
# Description

Implementation of the second games logic.

In this PR I factor-out shared code as common traits, as much as it's possible in ink!

- games implement `button::IButtonGame` ink trait, which defined game contracts API shared between all the games.
- they get default implementation logic by implementing `button::ButtonGame` (can be overridden if needed)
  - still responsible for defining and calling events (I could not find an easy way to share event definitions, without access to Environment (ink(topic), ink(event)) which is deep behind ink! macros, but if someone has an idea that'd be awesome)
  - self.env() is only defined for ink::contract, therefore all the calls to the `Environment` are also the implementer responsibility, again could not find an easy way to have that in the shared traits.  
- underlying storage is defined by the `ButtonData` struct, that works just fine
- Because of the multiple calls to `Self::env()` constructors are not part of the common traits and are duplicated (maybe some bits could be refactored out, but not sure how much lines of code would it actually save, if any) 

## Type of change

# Checklist:

- I have added tests


contracts/button/button.rs
```diff
@@ -0,0 +1,399 @@
...
+    }
+
+    fn deadline(&self) -> u32 {
+        let ButtonData {
+            last_press,
+            button_lifetime,
+            ..
+        } = self.get();
+        last_press + button_lifetime
+    }
```
pmikolajczyk41, line 134: I'm just wondering where is the moment we would like to keep such information in storage instead of computing it every time

do you have some rule of thumb?

contracts/button/button.rs
```diff
@@ -0,0 +1,399 @@
...
+
+    fn can_play(&self, user: AccountId) -> bool {
+        self.get().can_play.get(&user).unwrap_or(false)
+    }
+
+    fn access_control(&self) -> AccountId {
+        self.get().access_control
+    }
+
+    fn set_access_control(
```
pmikolajczyk41, line 148: when do we need such functionality?

contracts/button/button.rs
```diff
@@ -0,0 +1,399 @@
...
+        self.check_role(caller, required_role)?;
+        self.get_mut().access_control = new_access_control;
+        Ok(())
+    }
+
+    fn last_presser(&self) -> Option<AccountId> {
+        self.get().last_presser
+    }
+
+    fn button_token(&self) -> Result<AccountId> {
```
pmikolajczyk41, line 167: why do we wrap it in `Result`?

contracts/button/button.rs
```diff
@@ -0,0 +1,399 @@
...
+        Ok(self.get().button_token)
+    }
+
+    fn balance(&self, balance_of_selector: [u8; 4], this: AccountId) -> Result<Balance> {
+        let button_token = self.get().button_token;
+        let balance = build_call::<DefaultEnvironment>()
```
pmikolajczyk41, line 173: I'm not familiar with environments here, but maybe we should be generic over it?

contracts/button/button.rs
```diff
@@ -0,0 +1,399 @@
...
+            .returns::<Balance>()
+            .fire()?;
+        Ok(balance)
+    }
+
+    fn transfer_tx(
+        &self,
+        transfer_to_selector: [u8; 4],
+        to: AccountId,
+        value: u128,
```
pmikolajczyk41, line 185: u/128/Balance?

contracts/button/button.rs
```diff
@@ -0,0 +1,399 @@
...
+            return Err(Error::NotWhitelisted);
+        }
+
+        let score = self.score(now);
+
+        self.get_mut().presses.insert(&caller, &score);
+        self.get_mut().press_accounts.push(caller);
+        self.get_mut().last_presser = Some(caller);
+        self.get_mut().last_press = now;
+        self.get_mut().total_scores += score;
```
pmikolajczyk41, line 275: is there any way of making a single read and a single write?

contracts/button/button.rs
```diff
@@ -0,0 +1,399 @@
...
+    ) -> Result<()> {
+        let ButtonData {
+            total_scores,
+            last_presser,
+            press_accounts,
+            presses,
+            ..
+        } = self.get();
+
+        let total_balance = self.balance(balance_of_selector, this)?;
```
pmikolajczyk41, line 294: maybe we should check firstly whether there is any pressiah to avoid unnecessary call in case when nobody has pressed the button

contracts/button/button.rs
```diff
@@ -0,0 +1,399 @@
...
+        this: AccountId,
+    ) -> Result<()> {
+        let ButtonData {
+            total_scores,
+            last_presser,
+            press_accounts,
+            presses,
+            ..
+        } = self.get();
+
```
pmikolajczyk41, line 293: shouldn't we assert that `self.is_dead`?

pmikolajczyk41: unless I'm missing something, anyone can call this though `IButtonGame::death`

pmikolajczyk41 (commented):
I didn't find any blockers, just a few doubts

in general it is an awesome piece of rework and definitely a great improvement; nevertheless, I would split this PR into two, but I do not insist

contracts/button/button.rs
fbielejec: that would be sad, but we can do that optimization :)

contracts/button/button.rs
fbielejec: that *should*^TM do it

contracts/button/button.rs
fbielejec: changed

contracts/button/button.rs
fbielejec: sure, why not

contracts/button/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+    "ink_env/std",
+    "ink_lang/std",
+    "ink_metadata/std",
+    "ink_prelude/std",    
+    "ink_primitives/std",
+    "ink_storage/std",
+    "scale-info/std",
+    "scale/std",
+]
+ink-as-dependency = []
```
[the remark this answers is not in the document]
fbielejec: done

contracts/README.md
```diff
@@ -13,34 +13,36 @@ TheButton is based on the famous [game](https://en.wikipedia.org/wiki/The_Button
...
+In this scenario the rewards are reversed - players get rewarded for extending the button's life further into the future, i.e.:
 
 ```
 score = now - start
 ```
 
-## Blue button
+The Pressiah gets 50% of the total reward pool.
+
+## Yellow Button
```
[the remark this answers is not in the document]
fbielejec: done

.github/scripts/assert.sh
```diff
@@ -1,248 +0,0 @@
-#!/usr/bin/env bash
-
```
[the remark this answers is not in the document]
fbielejec: was used to have assertions in bash. Not needed now since:

1. ButtonDeath event carries no data anyway (it would be cumbersome to return it as Button::death and IButton::death would have different signatures, plus there is no need since this data can be queried) 
2. parity merged a PR that factors out transcode bits form cargo-contract into a standalone library, so next PR or next-next PR can introduce proper (Rust) e2e test

contracts/button/lib.rs
```diff
@@ -0,0 +1,3 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+pub mod button;
```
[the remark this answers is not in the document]
fbielejec: there is only one concept here : a game logic, so hard to find a seam to split things by

contracts/button/mod.rs
```diff
@@ -0,0 +1,3 @@
+mod button;
+
+pub use button::*;
```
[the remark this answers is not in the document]
fbielejec: no, could be all in lib.rs - call it convention or cargo cult. Did we ever specify what rules we go by with regards to grouping code?

contracts/back_to_the_future/lib.rs
```diff
@@ -0,0 +1,374 @@
...
+        /// Returns own code hash
+        #[ink(message)]
+        pub fn code_hash(&self) -> Result<Hash> {
+            self.env().own_code_hash().map_err(|why| {
+                Error::ContractCall(format!("Can't retrieve own code hash: {:?}", why))
+            })
+        }
+    }
+
+    // NOTE: can't test because off-chain environment does not support `own_code_hash`
```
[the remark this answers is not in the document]
fbielejec: done

contracts/button/button.rs
```diff
@@ -0,0 +1,399 @@
...
+)]
+pub struct ButtonData {
+    /// How long does TheButton live for?
+    pub button_lifetime: u32,
+    /// Stores a mapping between user accounts and the number of blocks they extended The Buttons life for
+    pub presses: Mapping<AccountId, u32>,
+    /// stores keys to `presses` because Mapping is not an Iterator. Heap-allocated so we might need Map<u32, AccountId> if it grows out of proportion
+    pub press_accounts: Vec<AccountId>,
+    /// stores total sum of user scores
+    pub total_scores: u32,
```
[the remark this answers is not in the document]
fbielejec: I aliased it as a BlockNumber from the `Environment` trait. 
(Scores are denominated in block numbers). The default env defines it as u32, but since we are generic over Environment - it can be easily changed if needed

contracts/button/button.rs
```diff
@@ -0,0 +1,399 @@
...
+    /// stores the last account that pressed The Button
+    pub last_presser: Option<AccountId>,
+    /// block number of the last press
+    pub last_press: u32,
+    /// AccountId of the ERC20 ButtonToken instance on-chain
+    pub button_token: AccountId,
+    /// accounts whitelisted to play the game
+    pub can_play: Mapping<AccountId, bool>,
```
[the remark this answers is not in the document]
fbielejec: good idea, saves some precious storage

contracts/button/button.rs
```diff
@@ -0,0 +1,399 @@
...
+    /// access control contract
+    pub access_control: AccountId,
+}
+
+/// Concrete implementations of the games API
```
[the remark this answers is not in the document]
fbielejec: it's meant to convey the fact you need to implement two traits, and only one has some default impls (cause ink!)

contracts/button/button.rs
fbielejec: when you update the on-chain access control contract

contracts/button/button.rs
fbielejec: I suppose storage >> computation, so when computing is cheap like here :shrug:

contracts/scripts/deploy.sh
```diff
@@ -69,99 +71,183 @@ BUTTON_TOKEN_CODE_HASH=$(echo "$CODE_HASH" | grep hash | tail -1 | cut -c 15-)
...
 
-EVENT=$(cargo contract call --url $NODE --contract $YELLOW_BUTTON --message press --suri $ALICE_SEED | grep ButtonDeath)
-EVENT=$(echo "$EVENT" | sed 's/^ *//g' | tr " " "\n")
+cargo contract call --url $NODE --contract $EARLY_BIRD_SPECIAL --message IButtonGame::death --suri $ALICE_SEED
+
+#
+# --- BACK_TO_THE_FUTURE GAME
+#
+
+# --- INITIALIZE TOKEN CONTRACT FOR THE BACK_TO_THE_FUTURE GAME
```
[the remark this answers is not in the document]
fbielejec: Sorry I don't understand this comment?

contracts/button/button.rs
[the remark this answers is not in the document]
pmikolajczyk41: I don't see why do we have to stick with `BlockNumber` semantics here. For custom `EarlyBirdSpecial` with a long deadline, say 1 million blocks (12 days), a couple of thousands presses can easily overflow `u32`

contracts/scripts/deploy.sh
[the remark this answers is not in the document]
pmikolajczyk41: if I'm not mistaken, we have two sets of deployment/run instructions for both contracts; these sets are very similar, so I was wondering whether we could do something like:
```
function do_all_the_stuff(params) {
   ...
}

do_all_the_stuff("blue_button");
do_all_the_stuff("yellow_button");
```

contracts/button/button.rs
pmikolajczyk41: do you mean that this is compiled into single-read-single-write?

contracts/button/button.rs
```diff
@@ -0,0 +1,448 @@
...
+        &self,
+        now: BlockNumber,
+        balance_of_selector: [u8; 4],
+        transfer_selector: [u8; 4],
+        this: AccountId,
+    ) -> Result<()>
+    where
+        E: Environment<AccountId = AccountId>,
+    {
+        if self.is_dead(now) {
```
pmikolajczyk41, line 323: maybe let's reduce indentation by just:
```rust
if !self.is_dead(now) {
    Err(Error::BeforeDeadline)
}
```

contracts/button/button.rs
fbielejec: I meant this: https://github.com/Cardinal-Cryptography/aleph-node/blob/blue-button/contracts/button/button.rs#L297
pulls the data from storage, mutates in-memory and writes again

contracts/button/button.rs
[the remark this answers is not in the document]
fbielejec: it is u64, env is custom for game contracts (the type for BlockNum is the only change)

contracts/button/button.rs
fbielejec: :ok_hand:

contracts/button/mod.rs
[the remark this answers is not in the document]
fbielejec: removed mod, however I'm not sure why would we want to split it - I don't ever see us using just a subset of them?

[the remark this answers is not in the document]
pmikolajczyk41: uh, I guess we won't use any proper subset - this is just my bias from old java school to have a very granular division :see_no_evil:

contracts/button/lib.rs
```diff
@@ -1,3 +1,451 @@
...
+
+pub enum ButtonGameEnvironment {}
+
+impl Environment for ButtonGameEnvironment {
+    const MAX_EVENT_TOPICS: usize = <DefaultEnvironment as Environment>::MAX_EVENT_TOPICS;
+
+    type AccountId = <DefaultEnvironment as Environment>::AccountId;
+    type Balance = <DefaultEnvironment as Environment>::Balance;
+    type Hash = <DefaultEnvironment as Environment>::Hash;
+    type BlockNumber = u64;
```
pmikolajczyk41, line 34: `type BlockNumber = BlockNumber`? I mean some rearrangement of definitions?

contracts/scripts/deploy.sh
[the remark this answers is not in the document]
fbielejec: I see, done

contracts/early_bird_special/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+                Event::ButtonPressed(ButtonPressed {
+                    by: caller,
+                    when: now,
+                }),
+            );
+            Ok(())
+        }
+
+        #[ink(message)]
+        fn death(&self) -> Result<()> {
```
DamianStraszak, line 119: I'm not sure when this is supposed to be called and by whom, but my general comment here is that `death` should not trigger the payouts, because:
-- who will call it, who will pay for the gas?
-- given enough players, death will not fit in one block, and we will be screwed.

Whenever a smart contract has a loop over users than it's already a red flag, because of the problems above. So this should be handled differently I believe:
1) death should be only marked with a boolean flag or so,
2) users need to `claim reward` with a transaction, post death.
And while I see that this is not "nice" UX-wise, there is just no alternative that I would see.

contracts/early_bird_special/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+    // becasue ink! does not allow generics or trait default implementations
+    impl IButtonGame for EarlyBirdSpecial {
+        #[ink(message)]
+        fn is_dead(&self) -> bool {
+            let now = Self::env().block_number();
+            ButtonGame::is_dead(self, now)
+        }
+
+        #[ink(message)]
+        fn press(&mut self) -> Result<()> {
```
DamianStraszak, line 104: I think that there is an opportunity to improve UX here by adding a parameter `press_number: u32` here that would be passed within the transaction. 
The meaning of this `press_number` argument within the `press` transaction is the following "I'm willing to press the button with the intention of my press being the number `press_number` counting from the start of the game." The contract should maintain in its state a variable `number_pressed` (or some such), and whenever a transaction comes saying press with some outdated number, then just ignore it (I mean, the press does not happen).

This helps a lot when lots of users are trying to land a button press at a particular moment in time (especially when competing with bots). If 5 of them send a tx at the same time with the current solution, then 1 of them will succeed and 4 of them will "waste" their press (and will be angry). While in what I propose, the one guy will win, but the remaining guys presses will be canceled, so they will keep their chance to score high.

DamianStraszak (commented):
Impressive work! 

I left 2 high-level comments about some specific parts of the design. I'm now taking a second pass to look at details.

contracts/early_bird_special/lib.rs
fbielejec: This is a great comment. Initially the idea was anyone can call it, thus paying the associated costs, but in practice for "fairness" we (aka the foundation) would. However as you rightly point out given a large enough pool of players this tx can never fit into one block. 

Initially I imagined a "batched" death, that always triggers rewards for next `n` accounts - but this does not go around the problem that the entire vector of accounts needs to be read from storage.

In the end the right design is a right design - this is now, per comment, refactored to `claim_reward`, can still be called by anyone (on behalf of the account), such that we can distribute the rewards post-mortem. I guess we should include this change in the wireframes for the UI? (after buttons death, `press` becomes claim_reward or something to that effect)?

contracts/early_bird_special/lib.rs
fbielejec: This is another great comment, that made me re-think how the game should work.

As it was, if multiple account recorded their press at the same time, they would all get recorded (so not wasted per se) with the same score. When it comes to the Pressiah than effectively the account that submitted the last tx processed for the last block of the Buttons life becomes the Pressiah - which could be seen as counter-intuitive.

What I propose (subject to discussion OC) is this:
- the block number (the `now`) becomes a type of a nonce for the `press` transactions. 
- in each block there can be only one such tx processed, so there is also a per-block competition to be that one included in it, in addition to the overall game competition
- if you were late, the tx returns an error, but you get to try again (your press is not recorded this time)

contracts/early_bird_special/lib.rs
DamianStraszak: Awesome, thanks for introducing `claim_reward`! I've looked at the UI and if it was a "one shot" game only then it would make most sense to have `claim` instead of `press` after pressing for the first time, but if we already plan that the game might have rounds, then probably I would place the `claim` button somewhere else. But I'm bad at UI :P

contracts/early_bird_special/lib.rs
DamianStraszak: I think the version you propose is good. Although I would still argue that for the `back_to_the_future` game this can lead to some unpleasant surprises for the users. For instance
-- The button is close to dead -- 5 seconds from the end
-- Two users click -- for one of them the transaction lands in block 896, and for the other 897
-- The second user is pissed, because it wanted ~900 points and not 1 xD

DamianStraszak: Also a general consideration -- relying on block numbers instead of timestamps (which ink exposes I would presume?) might sometimes have bad effects. If we end up relying on block numbers, we need to make sure the UI displays this accordingly. We should ideally not assume 1 block=1 sec then.

fbielejec: ad 1) Unless I'm missing or did not understand something, if we allow for passing a "desired" number in a press tx argument, then the game becomes very "gameable".
Taking for example the `back_to_the_future` game, where the reward is bigger the closer to the end of the Buttons life it is (even ignoring the whoel pressiah business).
- lets say the game starts in block number 0 and lives for 900 blocks
- just in the block 0 you submit press(900) and it gets into the block - your score becomes 900-0 and you already "won"

Regardless of the logic of the game it is now always desired to be the first one to play, because you get to submit the winning transaction right of the bat. I mean others can still play for the scrappy seconds, but for example the lucrative Pressiah status is taken right away :)

ad 2) timestamps are not exposed, just block number - maybe instead of a timer the UI should display block numbers counting down to the game end?

DamianStraszak: 1) That's not what I meant :) 
What I imagine is that you should initialize the state of the game with `number_pressed=0`
Then whoever sends `press(0)` first, presses the button for the first time and `number_pressed` gets incremented, then we expect `press(1)`, etc.
More generally: in the state `number_pressed=k` all `press(x)` transactions with `x!=k` are ignored, and only the first one with `x=k` gets accepted.

It's just "adding a nonce" being a counter of presses, so that people don't get screwed when pressing at roughly the same time.
2) OK, in that case I would be in favor of the UI measuring time in blocks and not in seconds...

PR #523, opened 2022-07-11
A0-1106 : Add e2e test for reward point calculation after changing stake and forcing new era

krzysztofziobro:


e2e-tests/src/test/rewards.rs
```diff
@@ -311,3 +304,82 @@ pub fn force_new_era(config: &Config) -> anyhow::Result<()> {
 
     Ok(())
 }
+
+pub fn change_stake_and_force_new_era(config: &Config) -> anyhow::Result<()> {
+    const MAX_DIFFERENCE: f64 = 0.07;
```
Marcin-Radecki, line 309: 1) I can see all tests in this file re-use the same value, does it make sense to extract it to a common value in this file (e.g. making `mod rewards` with that const)?
2) this is a relative difference right? and exactly from what two values? I can see it is used in `check_rewards` to compare some fractions. Maybe we can add some comment describing what we are actually comparing with that const?

e2e-tests/src/test/rewards.rs
```diff
@@ -311,3 +304,82 @@ pub fn force_new_era(config: &Config) -> anyhow::Result<()> {
...
+        "After ForceNewEra | era: {}, session: {}",
+        current_era, current_session
+    );
+
+    // Once a new era is forced in session k, the new era does not come into effect until session
+    // k + 2; we test points:
+    // 1) immediately following the call in session k,
+    // 2) in the interim session k + 1,
+    // 3) in session k + 2, the first session of the new era.
+    for idx in 0..3 {
```
Marcin-Radecki, line 354: Can this for loop be extracted to some common place, as the above test `force_new_era` seems to use it as well?

Marcin-Radecki (approved):
Very decent work.

e2e-tests/src/test/rewards.rs
```diff
@@ -311,3 +304,82 @@ pub fn force_new_era(config: &Config) -> anyhow::Result<()> {
...
+
+    let node = &config.node;
+    let accounts = get_validators_keys(config);
+    let sender = accounts.first().expect("Using default accounts").to_owned();
+    let connection = SignedConnection::new(node, sender);
+
+    let sudo = get_sudo_key(config);
+    let root_connection = RootConnection::new(node, sudo);
+
+    let (reserved_members, non_reserved_members) = get_member_accounts(config);
```
pmikolajczyk41, line 319: I suspect that we may need to `change_validators` here (like we recently forgot: https://github.com/Cardinal-Cryptography/aleph-node/pull/528)

e2e-tests/src/test/rewards.rs
```diff
@@ -311,3 +304,82 @@ pub fn force_new_era(config: &Config) -> anyhow::Result<()> {
 
     Ok(())
 }
+
+pub fn change_stake_and_force_new_era(config: &Config) -> anyhow::Result<()> {
```
pmikolajczyk41, line 308: maybe it is out of scope of this PR, but I feel that our E2E tests should be documented, i.e. every test function should have a docstring describing the scenario and the expected outcome

if you agree, you can add such description here and then we will create Jira with your test as the example

pmikolajczyk41 (commented):
cool, well done

Marcin-Radecki (approved):

PR #525, opened 2022-07-11
nix-build for github actions

fixxxedpoint:
This PR integrates nix based build procedure into our github pipelines. Since direct installation of nix on our github images creates additional problems on its own (e.g. caching of the /nix folder, problems with permissions), it uses the docker-approach described in BUILD.md. It caches docker containers (per **/Cargo.lock **/rust-toolchain) between builds to speed it up. Size of such images is around 4GB. They contain pre-installed rust, all other native dependencies (e.g. protobuf) and pre-compiled rust dependencies. This nix-based approach uses directly cargo in order to simplify troubleshooting (some other nix-based tools are way more complicated and for example uses rustc directly).


.github/workflows/build-node-and-runtime.yml
```diff
@@ -25,55 +25,79 @@ jobs:
...
-          path: target/release/wbuild/aleph-runtime/aleph_runtime.compact.wasm
+          path: result/lib/aleph_runtime.compact.wasm
           if-no-files-found: error
           retention-days: 7
 
       - name: Build test binary
-        run: cargo build --release -p aleph-node --features "short_session enable_treasury_proposals"
+        shell: bash
+        run: |
+          # we need to clean after previous build
```
Marcin-Radecki, line 74: I'm not an expert in GH actions, but I wonder whether we do have two steps that are related (ie outcome of the first one impacts the run of the second one), then does it make sense to have two separate steps? Not sure what's the best practice

.github/workflows/build-node-and-runtime.yml
```diff
@@ -25,55 +25,79 @@ jobs:
...
+          mv result result.main
```
Marcin-Radecki, line 75: Do we need to keep `result.main` directory at all? We don't use them later in this workflow

.github/workflows/build-node-and-runtime.yml
```diff
@@ -25,55 +25,79 @@ jobs:
...
 
-      - name: Cleanup cache
-        uses: ./.github/actions/post-cache
+      - name: Save docker image
+        shell: bash
+        run: |
+          if [[ ! -f .image.loaded ]]; then
+            docker commit build.env aleph-node/build
+            docker rm build.env
+            docker save -o image.tar aleph-node/build
```
Marcin-Radecki, line 101: I do not quite understand what happens next with `image.tar` when it comes to a lifecycle of a workflow.

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -64,29 +64,53 @@ jobs:
...
-          cargo build --release -p aleph-runtime
-          sha256sum -c checksum.sha256
+          rm -rf "$TARGET_DIR"
+
+      - name: Checkout docker image
+        uses: actions/cache@v2
+        with:
+          path: |
+            image.tar
+          key: ${{ runner.os }}-build-docker-image-${{ hashFiles('**/Cargo.lock', '**/rust-toolchain', 'nix/nix-build.sh') }}-v0
```
Marcin-Radecki, line 87: I wonder whether it makes sense to extract additional workflow (with trigger `on workflow_call` etc) , and call it from this one and `build-node-and-runtime.yml`, in order to decrease code duplication. Things we call twice now is e.g. `hashFiles('**/Cargo.lock', '**/rust-toolchain', 'nix/nix-build.sh')` which is a concrete logic that might change in the future, so better to have it in one place. Also, there's some boilerplate code around calling nix with docker, so I imagine this new workflow would have arguments like
* `crates` 
* `single_step`
And would define those steps
* "Build or load our `build` docker image"
* "Build package"
* ... etc

nix/nix-build.sh
```diff
@@ -32,14 +32,30 @@ then
     nix-shell --pure $SHELL_NIX_FILE
 else
     ARGS=(--arg crates "${CRATES}" --arg singleStep "${SINGLE_STEP}" --arg rustflags "${RUSTFLAGS}")
-    nix-build --max-jobs auto $SHELL_NIX_FILE "${ARGS[@]}"
+
+    # first we download all dependencies
+    echo fetching depedencies...
+    CARGO_HOME="$(realpath ~/cargo-home)"
+    nix-shell --pure --run "CARGO_HOME=$CARGO_HOME cargo fetch --locked --offline || CARGO_HOME=$CARGO_HOME cargo fetch --locked"
```
Marcin-Radecki, line 39: I think we might want to split this into two commands with more verbose output to a user, just in case one of them fail, so the user is not confused with an output

nix/nix-build.sh
```diff
@@ -32,14 +32,30 @@ then
...
+
+    echo building...
+    nix-build --max-jobs auto --option sandbox true --arg cargoHomePath "$CARGO_HOME" --show-trace $SHELL_NIX_FILE "${ARGS[@]}"
+    echo build finished
+
+    echo copying results...
```
Marcin-Radecki, line 45: I think we should have sth in the `result` so it resembles it us actually _a result of some build_ , so I propose to name it `build_results`, or just `build` would be more straighforward

Marcin-Radecki, line 45: I do not quite understand what we're trying to achieve in this copy section, could you please elaborate?

Marcin-Radecki (changes requested):

Marcin-Radecki:
Some high-level questions: 
1. How to build `aleph-runtime` with nix - do we want to specify some sub-paragraph in this file, mostly for sake of devs 
2. When I try to use the command from `BUILD.md`, it starts from the below error (though it seems to proceed further)
```
marcin@marcin-Latitude-3520 ~/g/aleph-node-fixedpoint (nix-integration-workflows.PR) [1]> docker run -ti --volume=$(pwd):/node/build aleph-node/build                                               (base) 
fetching depedencies...
error: failed to get `pallet-contracts-rpc` as a dependency of package `aleph-node v0.6.0 (/node/build/bin/node)`

Caused by:
  failed to load source for dependency `pallet-contracts-rpc`

Caused by:
  Unable to update https://github.com/Cardinal-Cryptography/substrate.git?branch=aleph-v0.9.23#2d851685

Caused by:
  can't checkout from 'https://github.com/Cardinal-Cryptography/substrate.git': you are in the offline mode (--offline)
    Updating crates.io index
    Updating git repository `https://github.com/Cardinal-Cryptography/substrate.git`
```

3. We need nix to work with `cargo-remote` to support compilation servers
4.  We might as well edit https://github.com/Cardinal-Cryptography/aleph-node/blob/main/BUILD.md?plain=1#L20 to as it not quite right atm - we have `result` dir in repo root, then `bin` and `lib` inside it.  

```
marcin@marcin-Latitude-3520 ~/g/aleph-node-fixedpoint (nix-integration-workflows.PR)> cd result/                                                                                                    (base) 
marcin@marcin-Latitude-3520 ~/g/a/result (nix-integration-workflows.PR)> ll                                                                                                                         (base) 
total 8,0K
dr-xr-xr-x 2 root root 4,0K júl   18 09:05 bin/
dr-xr-xr-x 2 root root 4,0K júl   18 09:05 lib/
marcin@marcin-Latitude-3520 ~/g/a/result (nix-integration-workflows.PR)> cd bin/                                                                                                                    (base) 
marcin@marcin-Latitude-3520 ~/g/a/r/bin (nix-integration-workflows.PR)> ls -ltr                                                                                                                     (base) 
total 196772
-rwxr-xr-x 1 root root 201490000 júl   18 09:05 aleph-node*
```
5. Why binary size is 2 times bigger than the current one?
```
marcin@marcin-Latitude-3520 ~/g/a/r/bin (nix-integration-workflows.PR)> du -sh aleph-node                                                                                                           (base) 
193M	aleph-node
```
6. This page says cache size is 10 GB (https://github.com/actions/cache#cache-limits), and since our docker has 4 GB, does it mean we have a space for 2 `image.tar` only?
7. Do we enable in this PR by default to build RocksDB with patched diff?
8.  This command `docker run -ti --volume=$(pwd):/node/build aleph-node/build -s` fails on my Ubuntu 20.04 woth below error:
```
building '/nix/store/nnfcvwkp738kl2brg7lb08bj6c2pnpz3-aleph-node-deps-unknown.drv'...
unpacking sources
unpacking source archive /nix/store/3319f7s3qaan3w7y5k2y707khz2fvljj-dummy-src
source root is dummy-src
patching sources
configuring
[naersk] cargo_version (read): 1.63.0-nightly (a4c1cd0eb 2022-05-18)
[naersk] cargo_message_format (set): json-diagnostic-rendered-ansi
[naersk] cargo_release: --release
[naersk] cargo_options: 
[naersk] cargo_build_options: --package aleph-node  --locked --offline $cargo_release -j "$NIX_BUILD_CORES" --message-format=$cargo_message_format
[naersk] cargo_test_options: $cargo_release -j "$NIX_BUILD_CORES"
[naersk] RUST_TEST_THREADS: 8
[naersk] cargo_bins_jq_filter: .
[naersk] cargo_build_output_json (created): /tmp/nix-build-aleph-node-deps-unknown.drv-0/tmp.EfqkHGA2ka
[naersk] crate_sources: /nix/store/62r3gvnsjcax7dhin7i2gq9j7nyk3k8j-crates-io
[naersk] RUSTFLAGS: -C target-cpu=generic
[naersk] CARGO_BUILD_RUSTFLAGS: 
[naersk] RUSTFLAGS (updated): -C target-cpu=generic --remap-path-prefix /nix/store/62r3gvnsjcax7dhin7i2gq9j7nyk3k8j-crates-io=/sources
building
cargo build --package aleph-node --locked --offline $cargo_release -j "$NIX_BUILD_CORES" --message-format=$cargo_message_format
error: source `https://github.com/Cardinal-Cryptography/substrate?rev=2d85168534f2981a92dfbae78cfa9e48f32c6bfc` defines source https://github.com/Cardinal-Cryptography/substrate.git?branch=aleph-v0.9.23, but that source is already defined by `https://github.com/Cardinal-Cryptography/substrate.git?rev=2d85168534f2981a92dfbae78cfa9e48f32c6bfc`
note: Sources are not allowed to be defined multiple times.
[naersk] cargo returned with exit code 101, exiting
error: builder for '/nix/store/nnfcvwkp738kl2brg7lb08bj6c2pnpz3-aleph-node-deps-unknown.drv' failed with exit code 101
```
9. Are we going to maintain two build systems? Nix and the current one? Or only Nix? Asking since we do not touch files in https://github.com/Cardinal-Cryptography/aleph-node/tree/main/docker in this PR.

.github/workflows/build-node-and-runtime.yml
```diff
@@ -25,55 +25,79 @@ jobs:
...
+          key: ${{ runner.os }}-build-docker-image-${{ hashFiles('**/Cargo.lock', '**/rust-toolchain', 'nix/nix-build.sh') }}-v0
+
+      - name: Build or load our `build` docker image
+        shell: bash
+        run: |
+          if [[ ! -f image.tar ]]; then
+            docker build -t aleph-node/build -f nix/Dockerfile.build .
+          else
+            docker load -i image.tar
+            touch .image.loaded
```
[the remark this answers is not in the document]
fixxxedpoint: I think we should be able to use github's `failure()` (executing another step only if some other failed...) mechanism here instead of lock-files.

.github/workflows/build-node-and-runtime.yml
```diff
@@ -25,55 +25,79 @@ jobs:
...
-      - name: Install Rust toolchain
-        uses: actions-rs/toolchain@v1
-
-      - name: Install WASM target
-        run: rustup target add wasm32-unknown-unknown
-
-      - name: Restore cache
-        uses: ./.github/actions/restore-cache
+      - name: Checkout docker image
+        uses: actions/cache@v2
```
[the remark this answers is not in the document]
fixxxedpoint: Changed.

.github/workflows/build-node-and-runtime.yml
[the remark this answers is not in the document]
fixxxedpoint: Changed.

.github/workflows/build-node-and-runtime.yml
```diff
@@ -25,55 +25,79 @@ jobs:
...
+            touch .image.loaded
+            rm image.tar
+          fi
 
       - name: Build binary and runtime
-        run: cargo build --release -p aleph-node
+        shell: bash
+        run: |
+          export CRATES='{ "aleph-node" = []; }'
+          docker run --name build.env -e CRATES --volume=$(pwd):/node/build aleph-node/build
```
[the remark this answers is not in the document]
fixxxedpoint: Changed.

.github/workflows/build-node-and-runtime.yml
fixxxedpoint: Changed.

.github/workflows/build-node-and-runtime.yml
fixxxedpoint: Changed. In general, `cache` defines a `post` action that stores it in cache.

.github/workflows/e2e-tests-main-devnet.yml
fixxxedpoint: Moved to a separate action.

nix/nix-build.sh
fixxxedpoint: Changed.

nix/nix-build.sh
fixxxedpoint: I don't have any opinion here. I don't know what is the reason for nix using `result` as its name.

fixxxedpoint:
1. Using `CRATES` env variable. I've added it to BUILD.md.                                                                                                                                                                                                                                                                                                      
2. Fixed.                                                                                                                                                                                                                                                                                                                                                       
3. `cargo-remote` makes no sense in context of nix (build should have no internet access). nix has its own mechanisms for that purpose (e.g. nix-cache, remote builders).                                                                                                                                                                                       
4. updated                                                                                                                                                                                                                                                                                                                                                      
5. I was unable to reproduce it so far.                                                                                                                                                                                                                                                                                                                         
6. I don't know.                                                                                                                                                                                                                                                                                                                                                
7. No.                                                                                                                                                                                                                                                                                                                                                          
8. Removed. It was unnecessary. Modified description in BUILD.md accordingly. Additionally, this error was related with some fancy mechanism of naersk (nix lib for building rust code), namely dependency vendoring (manually downloading rust dependencies). This version intentionally disables this mechanism, since it uses explicitly provided CARGO_HOME.
9. I don't know. I would answer: nix or nix+docker or something similar/equal - current build system is pretty flaky.

BUILD.md
```diff
@@ -14,10 +14,14 @@ In order to build a binary for `aleph-node` using docker we first need to instal
 distribution, by executing `sudo apt install docker.io` (please consult your distribution's manual describing docker
 installation procedure). Build procedure can be invoked by running:
 ```
-sudo docker build -t aleph-node/build -f nix/Dockerfile.build .
-sudo docker run -ti --volume=$(pwd):/node/build aleph-node/build
+sudo docker build -t aleph-build -f nix/Dockerfile.build .
+sudo docker run -ti --volume=$(pwd):/node/build aleph-build
```
Marcin-Radecki, line 18: I've tried to run it for the second time and got 
```
...
   Compiling finality-aleph v0.5.0 (/tmp/nix-build-aleph-node-unknown.drv-0/aleph-source/finality-aleph)
    Finished release [optimized] target(s) in 2m 52s
installing
[naersk] Using file /tmp/nix-build-aleph-node-unknown.drv-0/tmp.YqVpaGUDB1 to retrieve build (executable) products
[naersk] found executable aleph-node -> /nix/store/5kl74gf982qrg9nwxwnm01p4nsvzc8ii-aleph-node-unknown/bin/aleph-node
post-installation fixup
shrinking RPATHs of ELF executables and libraries in /nix/store/5kl74gf982qrg9nwxwnm01p4nsvzc8ii-aleph-node-unknown
shrinking /nix/store/5kl74gf982qrg9nwxwnm01p4nsvzc8ii-aleph-node-unknown/bin/aleph-node
strip is /nix/store/p792j5f44l3f0xi7ai5jllwnxqwnka88-binutils-2.31.1/bin/strip
stripping (with command strip and flags -S) in /nix/store/5kl74gf982qrg9nwxwnm01p4nsvzc8ii-aleph-node-unknown/lib  /nix/store/5kl74gf982qrg9nwxwnm01p4nsvzc8ii-aleph-node-unknown/bin 
patching script interpreter paths in /nix/store/5kl74gf982qrg9nwxwnm01p4nsvzc8ii-aleph-node-unknown
checking for references to /tmp/nix-build-aleph-node-unknown.drv-0/ in /nix/store/5kl74gf982qrg9nwxwnm01p4nsvzc8ii-aleph-node-unknown...
error: cannot create symlink '/node/build/result'; already exists
```

Marcin-Radecki (changes requested):

.github/actions/build_with_docker/action.yml
```diff
@@ -0,0 +1,69 @@
...
+    default: false
+
+runs:
+  using: "composite"
+  steps:
+    - name: Check if docker image is already loaded
+      id: check_loaded
+      shell: bash
+      run: |
+        rm image.tar 2>/dev/null || true
```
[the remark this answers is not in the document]
fixxxedpoint: Changed to `rm -f`,

.github/actions/build_with_docker/action.yml
```diff
@@ -0,0 +1,69 @@
...
+      shell: bash
+      run: |
+        if [ "${{ steps.check_loaded.outputs.loaded }}" == "false" ]; then
+          if [ "${{ steps.check_cache.outputs.cache-hit }}" == "true" ]; then
+            docker load -i image.tar
+          else
+            docker build -t aleph-build:${{ steps.check_loaded.outputs.cache_key }} -f nix/Dockerfile.build .
+          fi
+        fi
+        rm image.tar 2>/dev/null || true
```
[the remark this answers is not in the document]
fixxxedpoint: Changed to `rm -f`.

BUILD.md
fixxxedpoint: Added verbose `backup/mv` of the `result` folder.

BUILD.md
```diff
@@ -14,10 +14,14 @@ In order to build a binary for `aleph-node` using docker we first need to instal
...
+```
+Binary will be stored at `$(pwd)/result/bin/aleph-node`.
```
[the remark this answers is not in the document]
fixxxedpoint: I don't think it's necessary. Devs locally will use `nix-shell` which is independent of this mechanism. Another thing is that I would be sad if `nix-build` would override my `cargo target` dir.

Marcin-Radecki (approved):
Very nice!

gasiordev:
Just a quick info. I've tested it on macOS Monterey with docker 10.20.8 and it works as well. Although the docker VM needed at least 4GB of RAM and a bit more disk space.

gasiordev:
This is a great work you've done here, @fixxxedpoint!

PR #527, opened 2022-07-12
A0-1125: Elect only staking validators

pmikolajczyk41:
# Description

We ensure that `pallet_elections::ElectionProvider::elect()` returns only staking validators. Before it was possible for validators to chill and unbond their funds and still be elected to the committee.

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/aleph-client/Cargo.lock
+++ b/aleph-client/Cargo.lock
@@ -128,9 +128,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.57"
+version = "1.0.58"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "08f9b8508dccb7687a1d6c4ce66b2b0ecef467c94667de27d8d7fe1f8d2a9cdc"
+checksum = "bb07d2053ccdbe10e2af2995a2f116c1330396493dc1269f6a91d0ae82e19704"
 
 [[package]]
 name = "approx"
@@ -187,9 +187,9 @@ checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "backtrace"
-version = "0.3.65"
+version = "0.3.66"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "11a17d453482a265fd5f8479f2a3f405566e6ca627837aaddb85af8b1ab8ef61"
+checksum = "cab84319d616cfb654d03394f38ab7e6f0919e181b1b57e1fd15e7fb4077d9a7"
 dependencies = [
  "addr2line",
  "cc",
@@ -226,9 +226,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitvec"
-version = "1.0.0"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1489fcb93a5bb47da0462ca93ad252ad6af2145cce58d10d46a83931ba9f016b"
+checksum = "1bc2832c24239b0141d5674bb9174f9d68a8b5b3f2753311927c172ca46f7e9c"
 dependencies = [
  "funty",
  "radium",
@@ -381,9 +381,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-channel"
-version = "0.5.4"
+version = "0.5.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5aaa7bd5fb665c6864b5f963dd9097905c54125909c7aa94c9e18507cdbe6c53"
+checksum = "4c02a4d71819009c192cf4872265391563fd6a84c81ff2c0f2a7026ca4c1d85c"
 dependencies = [
  "cfg-if 1.0.0",
  "crossbeam-utils",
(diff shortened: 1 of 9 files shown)


pallets/elections/src/lib.rs
```diff
@@ -311,15 +311,27 @@ pub mod pallet {
...
+                .collect::<BTreeSet<_>>();
+            let reserved_validators: BTreeSet<_> = NextEraReservedValidators::<T>::get()
+                .into_iter()
+                .collect::<BTreeSet<_>>();
+            let non_reserved_validators: BTreeSet<_> = NextEraNonReservedValidators::<T>::get()
+                .into_iter()
+                .collect::<BTreeSet<_>>();
+
+            let eligible_validators =
+                &(&staking_validators & &non_reserved_validators) | &reserved_validators;
```
mike1729, line 332: Hilarious, this looks almost like python!

mike1729 (approved):
Great catch!

pallets/elections/src/lib.rs
pmikolajczyk41: and it is a way more handy than `intersection()` / `union()` methods that return intermediate lazy structs, which you have to `iter().collect()` to perform another operation

Marcin-Radecki: Those `&` before values obfuscate the whole expression a bit but agree `&` and `|` are handy.

pallets/elections/src/impls.rs
```diff
@@ -199,13 +202,22 @@ where
         // this will be populated once for the session `n+1` on the start of the session `n` where session
         // `n+1` starts a new era.
         Self::if_era_starts_do(active_era + 1, session, || {
+            let elected_committee =
+                BTreeSet::from_iter(T::EraInfoProvider::elected_validators(active_era + 1));
+
+            let retain_elected = |vals: Vec<T::AccountId>| -> Vec<T::AccountId> {
+                vals.iter()
+                    .filter_map(|v| elected_committee.contains(&v).then_some(v.clone()))
```
kostekIV, line 210:
```suggestion
                vals.into_iter()
                    .filter(|v| elected_committee.contains(&v))
                    .collect()
```

pmikolajczyk41: simplified

pallets/elections/src/tests.rs
```diff
@@ -1,20 +1,67 @@
...
+            (7, 10, bounded_vec![7]),
+        ]);
+
+        let elected = <Elections as ElectionProvider>::elect().expect("`elect()` should succeed");
+
+        assert_eq!(
+            elected,
+            &[
+                (1, support(10, vec![(1, 10)])),
+                (2, no_support()),
```
DamianStraszak, line 39: I have this doubt of why we have this entry with `no_support()` instead of omitting it? What difference would it make?

pallets/elections/src/impls.rs
```diff
@@ -81,7 +84,7 @@ fn choose_for_session<T: Clone>(
     let first_index = session.saturating_mul(count) % validators_len;
     let mut chosen = Vec::new();
 
-    for i in 0..count {
+    for i in 0..count.min(validators.len()) {
```
DamianStraszak, line 87: use `validators_len` for consistency

DamianStraszak (commented):
Awesome, I just left one question. I will now take a second pass to read the e2e tests.

pallets/elections/src/impls.rs
pmikolajczyk41: sure, I must have missed it; done

e2e-tests/src/test/electing_validators.rs
```diff
@@ -0,0 +1,266 @@
...
+        set_keys(&connection, keys, XtStatus::Finalized);
+        staking_validate(&connection, 10, XtStatus::Finalized);
+    }
+}
+
+/// Produce storage key to `ErasStakers::era`.
+///
+/// Since in `substrate-api-client` it seems impossible to get prefix for the first key in double
+/// map, we have to do it by hand.
+fn get_eras_stakers_storage_key(era: EraIndex) -> StorageKey {
```
kostekIV, line 86: what about method `get_storage_map_key_prefix`

e2e-tests/src/test/electing_validators.rs
```diff
@@ -0,0 +1,266 @@
...
+    let xt = compose_extrinsic!(connection.as_connection(), "Staking", "chill");
+    send_xt(
+        connection,
+        xt,
+        Some("chilling validator"),
+        XtStatus::InBlock,
+    );
+}
+
+/// Chill all validators in `malicious`.
```
kostekIV, line 188: I would not use the word `malicious` as it is kinda misleading. They are doing what is allowed, nothing is evil about them :P

kostekIV (approved):
lgtm

e2e-tests/src/test/electing_validators.rs
pmikolajczyk41: this works with Map, not DoubleMap, so it won't work since corresponding structure will not be found in metadata

e2e-tests/src/test/electing_validators.rs
pmikolajczyk41: yeah, right; renamed

e2e-tests/src/test/electing_validators.rs
```diff
@@ -0,0 +1,264 @@
...
+        staking_bond(
+            &connection,
+            MIN_VALIDATOR_BOND,
+            controller,
+            XtStatus::Finalized,
+        );
+    }
+
+    for controller in accounts.controller_keys.iter() {
+        let keys = rotate_keys(connection).expect("Failed to retrieve keys from chain");
```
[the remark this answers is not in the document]
pmikolajczyk41: right, corrected

DamianStraszak (approved):

pmikolajczyk41:
regarding finalization problems: https://cardinal-cryptography.atlassian.net/browse/A0-1152

pmikolajczyk41:
Test runtime update check won't pass due to `chain.start('old', backup=False)`, which results in:
```
error: The following required arguments were not provided:
    <--no-backup|--backup-path <PATH>>
```

I will fix the test in another PR. Apart from this, that workflow check is to be deleted soon. Since it is not a blocker, I'm merging this PR without successful pass.

PR #529, opened 2022-07-12
A0-926: Basic CCR support

timorl:
# Description

Implements the validation and pallet parts of the chain crash recovery method based on an emergency finalizer key. There will be two follow up PRs, one introducing an RPC call to actually use the finalization method and another adding related capabilities to `cliain`. Both have already been mostly prepared and the whole setup has been tested locally.

## Type of change

# Checklist:

- I have added tests


Yithis (approved):
Looks good to me, but I'm not pallet calls specialist. Also could be divided in 2 PR's. One with new justification type and second with pallet changes.

pallets/aleph/src/lib.rs
```diff
@@ -84,6 +106,39 @@ pub mod pallet {
         pub(crate) fn update_authorities(authorities: &[T::AuthorityId]) {
             <Authorities<T>>::put(authorities);
         }
+
+        pub(crate) fn update_emergency_finalizer() {
```
mike1729, line 110: I believe that names`QueuedEmergencyFinalizer` and  `NextEmergencyFinalizer` instead of `FutureEmergencyFinalizer` and `NextFutureEmergencyFinalizer` respectively would be more in line with the names in pallet session.

mike1729 (changes requested):
Just on request for changing names. Despite that great work :)

pallets/aleph/src/lib.rs
timorl: Done.

PR #531, opened 2022-07-14
A0-1143 Updated Testnet chainspec with code_substitute to use correct runtime from block 5268502  till block 5434491

Marcin-Radecki:
# Description

This fixes issue introduced during https://github.com/Cardinal-Cryptography/aleph-node/releases/tag/r-6.1 release. It used the Substrate `code_substitute` feature by telling the node to use the given runtime from block 5268502. It's the same runtime as `correct-aleph_runtime.compact.wasm` from that release Artifacts. It works since that runtime has the same `spec_version`, 21, as the runtime from block, hence substitution is used. From block 5434491, `spec_version` is 22 so substitute is no longer used.

Testing: an archivist connecting to Testnet that successfully past block 5434491 imports and finalizes blocks.

## Type of change

# Checklist:


DamianStraszak (approved):
Tested, works.

PR #537, opened 2022-07-15
Make cliain support CCR

timorl:
# Description

Add functions in `cliain` to support setting the emergency finalizer and using it to finalize blocks.

## Type of change

# Checklist:


pmikolajczyk41 (changes requested):
generally it looks cool, I have left some comments/questions, but none is a blocker; I also tested it locally and works well :muscle: 

apart from this, I have two questions regarding CCR in general (this is my first interaction with it):
 - why do we need this distinction with `NextEmergencyFinalizer` and `QueuedEmergencyFinalizer`? further, why don't we erase queued change as soon as it is applied? I see that it was easier to update it every session, but maybe keeping `None` would be cleaner (see pic)?
 - how can we trigger RPC call through UI? I can see that it is available, but not clickable (see pic) :| maybe it is not registered in all places?

image

image

pmikolajczyk41 (approved):
:ok_hand:

Yithis (approved):

PR #541, opened 2022-07-19
A0-1125: Give up testing updateability

pmikolajczyk41:
# Description

No longer do we require that every commit on main is ‘updateable’ from the previous one. Therefore we do not have to run test update when the `spec_version` has been bumped.


h4nsu (approved):
HELL YEAH! This needs to go so badly :)

kostekIV (commented):
What do we do with the python files that are no longer in use?

PR #545, opened 2022-07-20
[A0-1141] Enforce disabled pruning

h4nsu:
# Description

Quick and a bit dirty fix for the fact that we don't support pruning. The value of the corresponding flag is forcefully set at the beginning of `main()`.

## Type of change

### DIFF
--- a/bin/node/src/main.rs
+++ b/bin/node/src/main.rs
@@ -16,11 +16,11 @@ fn main() -> sc_cli::Result<()> {
         .import_params
         .pruning_params
         .pruning
+        .replace(String::from("archive"))
         .map_or(false, |x| x != "archive")
     {
         println!("Pruning not supported. Switching to 'archive' mode.");
     }
-    cli.run.import_params.pruning_params.pruning = Some(String::from("archive"));
 
     match &cli.subcommand {
         Some(Subcommand::BootstrapChain(cmd)) => cmd.run(),
--- a/bin/node/src/main.rs
+++ b/bin/node/src/main.rs
@@ -9,7 +9,20 @@ use sc_network::config::Role;
 use sc_service::PartialComponents;
 
 fn main() -> sc_cli::Result<()> {
-    let cli = Cli::parse();
+    let mut cli = Cli::parse();
+
+    match cli.run.import_params.pruning_params.pruning.as_ref() {
+        Some(mode) => {
+            if mode != "archive" {
+                println!("Pruning not supported. Switching to 'archive' mode.");
+                cli.run.import_params.pruning_params.pruning = Some(String::from("archive"));
+            }
+        }
+        None => {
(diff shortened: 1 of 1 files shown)


bin/node/src/main.rs
```diff
@@ -9,7 +9,20 @@ use sc_network::config::Role;
 use sc_service::PartialComponents;
 
 fn main() -> sc_cli::Result<()> {
-    let cli = Cli::parse();
+    let mut cli = Cli::parse();
+
+    match cli.run.import_params.pruning_params.pruning.as_ref() {
+        Some(mode) => {
+            if mode != "archive" {
+                println!("Pruning not supported. Switching to 'archive' mode.");
```
DamianStraszak, line 17: It's better to log it I think -- `info!` probably.

DamianStraszak (approved):

bin/node/src/main.rs
maciejnems: Logger is not initialized here yet :( Actually, where is it initialized @h4nsu? Haven't found that. Does it happen inside Substrate?

h4nsu: It's not possible to use the logger here, it gets initialized later, in differerent `match` branches separately.

maciejnems (approved):

bin/node/src/main.rs
h4nsu: @maciejnems I think somewhere in `cli.create_runner`

h4nsu: [client/cli/src/config.rs](https://github.com/Cardinal-Cryptography/substrate/blob/5f0aa1feb7250ac7b8c1b9928f87b2420b530e22/client/cli/src/config.rs#L651)

bin/node/src/main.rs
```diff
@@ -9,7 +9,20 @@ use sc_network::config::Role;
...
+                cli.run.import_params.pruning_params.pruning = Some(String::from("archive"));
+            }
+        }
+        None => {
+            cli.run.import_params.pruning_params.pruning = Some(String::from("archive"));
+        }
+    }
```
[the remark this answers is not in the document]
h4nsu: I don't know :/ I find it kinda cryptic and difficult to read, but maybe I'm just not familiar with idiomatic Rust ^^' I changed it according to your suggestion.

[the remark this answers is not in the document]
maciejnems: Could be even shorter :3
```
if cli
    .run
    .import_params
    .pruning_params
    .pruning
    .replace(String::from("archive"))
    .map_or(false, |x| x != "archive")
{
    println!("Pruning not supported. Switching to 'archive' mode.");
}
```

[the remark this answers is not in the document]
h4nsu: Even more cryptic ^^

PR #546, opened 2022-07-21
Bump Cargo.locks to match updated Substrate version

pmikolajczyk41:
# Description

We bump version to include: https://github.com/Cardinal-Cryptography/substrate/pull/23.

I bumped spec_version as well, since all pallets depend on the changed dependency (`tracing`-related) so as their logging part.

## Type of change

# Checklist:

### DIFF
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -107,7 +107,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("aleph-node"),
     impl_name: create_runtime_str!("aleph-node"),
     authoring_version: 1,
-    spec_version: 26,
+    spec_version: 27,
     impl_version: 1,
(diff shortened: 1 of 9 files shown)


Marcin-Radecki (approved):

bin/runtime/src/lib.rs
```diff
@@ -107,7 +107,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("aleph-node"),
     impl_name: create_runtime_str!("aleph-node"),
     authoring_version: 1,
-    spec_version: 26,
+    spec_version: 27,
```
fbielejec, line 110: needed? if so why?

pmikolajczyk41: as in the PR description:

> I bumped spec_version as well, since all pallets depend on the changed dependency (tracing-related) so as their logging part.

this potentially can change wasm bytecode

PR #550, opened 2022-07-22
A0-1100 Justification Decoding Error for pre-versioning justifications

Marcin-Radecki:
# Description

This is a bug fix to a minor issue in which pre-compatibility justifications are recognized as justifications from the future. 

## Type of change

# Checklist:

- I have added tests


kostekIV (approved):

lesniak43 (approved):

fixxxedpoint:
Do we have any task for further invastigating this issue beyond this bug-fix? After quick glance, it looks like the old version might `pretend` that is any other version (current and possibly from the future) and it will be found at the verification step (too late). Perhaps it can't, but it seems to be then a some very specific detail of how decode works and how we were represented justifications (list of Options). We might never see it, cuz it depends on network's configuration (size, etc.). Since new version should be future-proof (no more duality in parsing), returning from this method something like `(Maybe<Old>, Maybe<New>)` would solve it once for all without the need of proving (?).

Marcin-Radecki:
> Do we have any task for further investigate this issue beyond this bug-fix? After quick glance, it looks like the old version might `pretend` that is any other version (current and possibly from the future) and it would be found at the verification step (too late). Perhaps it can't, but it seems to be then a some very specific detail of how decode works and how we represented justifications (vec of Options). We might never see it, cuz it depends on network's configuration (size, etc.). Since new version should be future-proof (no more new dualities in parsing), returning from this method something like `(Maybe<Old>, Maybe<New>)` would solve it once for all without the need of proving.

Good point, added https://cardinal-cryptography.atlassian.net/browse/A0-1167

PR #554, opened 2022-07-26
Bump aleph-node and spec_version

kostekIV:


Marcin-Radecki (approved):

maciejnems (approved):

bin/node/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aleph-node"
-version = "0.6.0"
```
maciejnems, line 3: Actually, don't we also want ot update runtime and other crates?

maciejnems (commented):
Actually, don't we also want ot update runtime and other crates?

bin/node/Cargo.toml
kostekIV: runtime updated, other crates not related imo

PR #555, opened 2022-07-26
A0-1097: Include try-runtime check in pipelines

pmikolajczyk41:
# Description

We introduce a new runtime update check to our pipelines.

## Type of change

# Notes

I have temporarily bumped `spec_version` to trigger workflow


.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -532,6 +532,148 @@ jobs:
...
+      - name: Install WASM target
+        run: rustup target add wasm32-unknown-unknown
+
+      - name: Restore cache
+        uses: ./.github/actions/restore-cache
+        with:
+          target-key: try-runtime
+          cargo-key: old
+
+      - name: Build old binary
```
Marcin-Radecki, line 587: There're two approaches to this feature:
1) build the binary from `main` as a baseline for a runtime upgrade
2) use Testnet as such
Now they obviously test different things, but I think what we want to test here is the latter. This is because we want to be prepared for upcoming Testnet updates, so essentialy test runtime upgrade from version `X` to `Y`, where `X` is top of `main`, and `Y` is Testnet one

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -532,6 +532,148 @@ jobs:
...
+
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.8'
+
+      - name: Test update
+        timeout-minutes: 20
+        env:
+          OLD_BINARY: old-aleph-node/aleph-node
```
Marcin-Radecki, line 665: If we go with the second approach mentioned above, we just need to build one binary more (with `--try-runtime` feature), instead of building two.

local-tests/test_runtime_update.py
```diff
@@ -0,0 +1,113 @@
+#!/bin/env python
```
Marcin-Radecki, line 1: I'll review this file after first review round :)

Marcin-Radecki (changes requested):
Solid work , but let's agree on general approach to this test

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -532,6 +532,148 @@ jobs:
...
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+
+      - name: Install WASM target
+        run: rustup target add wasm32-unknown-unknown
+
+      - name: Restore cache
+        uses: ./.github/actions/restore-cache
+        with:
+          target-key: try-runtime
```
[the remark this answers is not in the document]
pmikolajczyk41: there's another `feature` set

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -532,6 +532,148 @@ jobs:
...
+        with:
+          name: old-aleph-node
+          path: target/release/aleph-node
+          if-no-files-found: error
+          retention-days: 7
+
+      - name: Cleanup cache
+        uses: ./.github/actions/post-cache
+
+  build-new-runtime-and-try_runtime:
```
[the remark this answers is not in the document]
pmikolajczyk41: actually... the same answer :(

local-tests/test_runtime_update.py
```diff
@@ -0,0 +1,113 @@
...
+           '--uri', 'ws://localhost:9944']
+    logging.info('Running `try-runtime` check...')
+    subprocess.run(cmd, check=True)
+    logging.info('✅ Update has been successful!')
+
+
+def run_test(old_binary: Path, new_runtime: Path, try_runtime: Path):
+    chain = start_chain(old_binary)
+    chainspec = file(os.path.join(WORKDIR, 'chainspec.json'))
+    save_runtime_to_chainspec(chainspec, new_runtime)
```
DamianStraszak, line 105: This mutates the original chainspec. I have an impression that it would be cleaner to keep the old file and save the runtime to a new copy...

DamianStraszak (approved):

.github/scripts/check_runtime_change.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+if [ $# -ne 3 ]; then
+  echo "Expected 3 arguments (paths to the repositories and the output name)"
+  exit 2
+fi
+
+grep "spec_version:" "${1%/}/bin/runtime/src/lib.rs" | grep -o '[0-9]*' > old.version
```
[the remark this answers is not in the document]
pmikolajczyk41: I removed the script anyway

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -532,6 +532,148 @@ jobs:
...
+
+      - name: Checkout current source code
+        uses: actions/checkout@v3
+        with:
+          path: ${{ env.NEW_REPO_PATH }}
+
+      - name: Inspect versions
+        id: inspect
+        run: |
+          ./${{ env.NEW_REPO_PATH }}/.github/scripts/check_runtime_change.sh \
```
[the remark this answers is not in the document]
pmikolajczyk41: I used this command but still as another step - it was much easier for me to preserve this as separate step and trigger further depending on the output (I couldn't make it clean otherwise)

.github/workflows/e2e-tests-main-devnet.yml
pmikolajczyk41: done, but I'm a little worried about storage scraping time... we'll see

.github/workflows/e2e-tests-main-devnet.yml
pmikolajczyk41: done

local-tests/test_runtime_update.py
pmikolajczyk41: now you have much less to review :)

local-tests/test_runtime_update.py
pmikolajczyk41: changed

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -532,6 +532,103 @@ jobs:
...
+        env:
+          NEW_RUNTIME: aleph-runtime-try_runtime/aleph_runtime.compact.wasm
+          TRY_RUNTIME: try_runtime/aleph-node
+        working-directory: local-tests/
+        run: |
+          chmod +x ${{ env.TRY_RUNTIME }}
+          pip install -r requirements.txt
+          ./test_runtime_update.py \
+              wss://ws.test.azero.dev:443 \
+              ../bin/node/src/resources/testnet_chainspec.json \
```
Marcin-Radecki, line 628: We need to use _new_ runtime, ie generated locally from what's on PR. Hence we need to generate chain spec from new node version.

Marcin-Radecki (changes requested):
One more change plx :)

DamianStraszak (approved):

.github/workflows/e2e-tests-main-devnet.yml
Marcin-Radecki: My bad, I did not read review carefully, all good.

Marcin-Radecki (approved):

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -532,6 +532,103 @@ jobs:
...
+    steps:
+      - name: Checkout current and old source code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Inspect versions
+        id: inspect
+        run: |
+          if ! git diff HEAD origin/main -- bin/runtime/src/lib.rs | grep -q ' spec_version: '
```
Marcin-Radecki, line 549: I think we need here at least 
```
if test ! $(git diff HEAD origin/main -- bin/runtime/src/lib.rs | grep -q ' spec_version: ')
```

or
```
if [ ! $(git diff HEAD origin/main -- bin/runtime/src/lib.rs | grep -q ' spec_version: ') ]
```

Marcin-Radecki (changes requested):
One more thing?

.github/workflows/e2e-tests-main-devnet.yml
pmikolajczyk41: done

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -532,6 +532,103 @@ jobs:
...
+
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.8'
+
+      - name: Test update
+        timeout-minutes: 20
+        env:
+          NEW_RUNTIME: aleph-runtime-try_runtime/aleph_runtime.compact.wasm
```
Marcin-Radecki, line 620: Runtime needs to be build with `try-runtime` as well

Marcin-Radecki (approved):

.github/workflows/e2e-tests-main-devnet.yml
pmikolajczyk41: done

PR #556, opened 2022-07-26
Add default AlephBFT backup folder

maciejnems:


bin/node/src/aleph_cli.rs
```diff
@@ -3,9 +3,12 @@ use std::path::PathBuf;
 use aleph_primitives::DEFAULT_UNIT_CREATION_DELAY;
 use clap::{ArgGroup, Parser};
 use finality_aleph::UnitCreationDelay;
+use sc_service::Configuration;
+
+const DEFAULT_BACKUP_FOLDER: &str = "backup-stash";
```
h4nsu, line 8: If you make this a constant, maybe `bootstrap-chain` should use it too? (struct `ChainParams` in `chain_spec.rs`)

bin/node/src/aleph_cli.rs
```diff
@@ -37,4 +40,18 @@ impl AlephCli {
...
+        if !self.no_backup {
+            let path = self.backup_path.get_or_insert(
+                config
+                    .base_path
+                    .as_ref()
+                    .expect("Please specify base path")
+                    .path()
+                    .join(DEFAULT_BACKUP_FOLDER),
+            );
+            eprintln!("No backup path provided, using default path: {:?} for AlephBFT backups. Please do not remove this folder", path);
```
h4nsu, line 54: This will be printed also when `backup_path` was `Some` at the beginning (that means the user DID actually provide backup path)

bin/node/src/aleph_cli.rs
```diff
@@ -37,4 +40,18 @@ impl AlephCli {
     pub fn backup_path(&self) -> Option<PathBuf> {
         self.backup_path.clone()
     }
+
+    pub fn insert_default_backup_path(&mut self, config: &Configuration) {
```
h4nsu, line 44: Let's use the path itself as a parameter instead of the whole config

h4nsu (changes requested):

bin/node/src/aleph_cli.rs
maciejnems: done

bin/node/src/aleph_cli.rs
maciejnems: Yep, thanks for noticing. I tried to do something smart

bin/node/src/aleph_cli.rs
maciejnems: Done

bin/node/src/aleph_cli.rs
```diff
@@ -3,9 +3,8 @@ use std::path::PathBuf;
 use aleph_primitives::DEFAULT_UNIT_CREATION_DELAY;
 use clap::{ArgGroup, Parser};
 use finality_aleph::UnitCreationDelay;
-use sc_service::Configuration;
 
-const DEFAULT_BACKUP_FOLDER: &str = "backup-stash";
+pub const DEFAULT_BACKUP_FOLDER: &str = "backup-stash";
```
h4nsu, line 7: Now defining this here doesn't make much sense. `chain_spec.rs` seems like a good place, there are many other constants defined there

h4nsu (approved):
Please move the constant and we're good to go

bin/node/src/service.rs
```diff
@@ -25,12 +25,33 @@ use sp_runtime::{
...
+        return None;
+    }
+    if let Some(path) = aleph_config.backup_path() {
+        Some(path)
+    } else {
+        let path = config
+            .base_path
+            .as_ref()
+            .expect("Please specify base path")
+            .path()
```
pmikolajczyk41, line 48: this unwrapping seems a bit unrelated to the `get_backup_path` semantics - I would pass to this method ready `base_path` instead of whole `Configuration` object

pmikolajczyk41 (approved):

bin/node/src/service.rs
maciejnems: done

bin/node/src/aleph_cli.rs
maciejnems: done

PR #558, opened 2022-07-26
A0-1175: Full genesis init for pallet elections

pmikolajczyk41:
# Description

This PR introduces two changes:
 - we enable setting non-reserved seats for genesis build as well
 - we initialize all the (reasonable) storage items in genesis build 

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/pallets/elections/src/lib.rs
+++ b/pallets/elections/src/lib.rs
@@ -280,9 +280,14 @@ pub mod pallet {
     #[pallet::genesis_build]
     impl<T: Config> GenesisBuild<T> for GenesisConfig<T> {
         fn build(&self) {
-            <NextEraNonReservedValidators<T>>::put(&self.non_reserved_validators);
             <CommitteeSize<T>>::put(&self.committee_seats);
+            <NextEraCommitteeSize<T>>::put(&self.committee_seats);
+            <NextEraNonReservedValidators<T>>::put(&self.non_reserved_validators);
             <NextEraReservedValidators<T>>::put(&self.reserved_validators);
+            <CurrentEraValidators<T>>::put(&EraValidators {
+                reserved: self.reserved_validators.clone(),
+                non_reserved: self.non_reserved_validators.clone(),
+            });
         }
     }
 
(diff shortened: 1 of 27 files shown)


pallets/elections/src/tests.rs
```diff
@@ -14,54 +21,86 @@ fn support(total: Balance, voters: Vec<(AccountId, Balance)>) -> Support<Account
     Support { total, voters }
 }
 
+#[test]
+fn storage_is_initialized_already_in_genesis() {
+    static RESERVED: [AccountId; 3] = [1, 2, 3];
+    static NON_RESERVED: [AccountId; 2] = [4, 5];
+    static COMMITTEE_SEATS: CommitteeSeats = CommitteeSeats {
```
kostekIV, line 28: use const instead of static

pallets/elections/src/lib.rs
```diff
@@ -48,7 +51,8 @@ impl<AccountId> Default for EraValidators<AccountId> {
     }
 }
 
-#[derive(Decode, Encode, TypeInfo, Debug, Clone, Copy, PartialEq, Eq, Default)]
+#[derive(Decode, Encode, TypeInfo, Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
```
kostekIV, line 55: we can move this to primitives to avoid adding `serde` dependency to the pallet

kostekIV (approved):

pallets/elections/src/lib.rs
pmikolajczyk41: the other alternative is to take just `reserved_seats: u32` and `non_reserved_seats: u32` in `GenesisConfig`; this has such an advantage that `CommitteeSeats` would stay in pallet where it belongs; wdyt?

obrok (approved):
LGTM

pallets/elections/src/lib.rs
kostekIV: I would move it to primitives anyway, an additional reason is that this struct is used in other places as well (e2e tests, chain_spec.rs etc)

pmikolajczyk41: done

PR #559, opened 2022-07-27
Bump alephbft version to 0.18.1

mike1729:
# Description

Change implementation of `DataProvider` to return `Option`.

## Type of change

- Refactor

# Checklist:

- I have updated tests

### DIFF
--- a/finality-aleph/src/data_io/data_provider.rs
+++ b/finality-aleph/src/data_io/data_provider.rs
@@ -115,7 +115,7 @@ struct ChainInfo<B: BlockT> {
 
 /// ChainTracker keeps track of the best_block in a given session and allows to generate `AlephData`.
 /// Internally it frequently updates a `data_to_propose` field that is shared with a `DataProvider`, which
-/// in turn is a tiny wrapper around this single shared resource that outputs `data_to_propose` whenever
+/// in turn is a tiny wrapper around this single shared resource that takes out `data_to_propose` whenever
 /// `get_data` is called.
 pub struct ChainTracker<B, SC, C>
 where
--- a/finality-aleph/src/data_io/data_provider.rs
+++ b/finality-aleph/src/data_io/data_provider.rs
@@ -82,9 +82,9 @@ where
         let num_last = finalized_block.num + <NumberFor<B>>::saturated_from(branch.len());
         // The hashes in `branch` are ordered from top to bottom -- need to reverse.
         branch.reverse();
-        Ok(AlephData::HeadProposal(UnvalidatedAlephProposal::<B>::new(
-            branch, num_last,
-        )))
+        Ok(AlephData {
+            head_proposal: UnvalidatedAlephProposal::new(branch, num_last),
+        })
     } else {
         // By backtracking from the best block we reached a block conflicting with best finalized.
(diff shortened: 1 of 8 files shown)


finality-aleph/src/data_io/data_provider.rs
```diff
@@ -307,20 +307,21 @@ struct DataProvider<B: BlockT> {
 //    at most MAX_DATA_BRANCH_LEN.
 #[async_trait]
 impl<B: BlockT> aleph_bft::DataProvider<AlephData<B>> for DataProvider<B> {
-    async fn get_data(&mut self) -> AlephData<B> {
-        let data = (*self.data_to_propose.lock().await).clone();
+    async fn get_data(&mut self) -> Option<AlephData<B>> {
+        let data_to_propose = (*self.data_to_propose.lock().await).take();
```
kostekIV, line 311: In case the data_to_propose is not updated, previously multiple calls to get_data would return the same proposals (because of the clone), now only the first call will return data, and the next ones will return None.

kostekIV: Just making sure it will not result in any unexpected behaviour changes in the aleph-bft

mike1729: That was intended :) There is no point in ordering the same hash multiple times.

kostekIV (approved):

finality-aleph/src/data_io/mod.rs
```diff
@@ -24,34 +24,22 @@ pub const MAX_DATA_BRANCH_LEN: usize = 7;
...
-        match self {
-            AlephData::Empty => {
-                (0u8).hash(state);
-            }
-            AlephData::HeadProposal(proposal) => {
-                (1u8).hash(state);
-                proposal.hash(state);
-            }
-        }
+        (1u8).hash(state);
```
DamianStraszak, line 34: this line is not needed anymore

finality-aleph/src/data_io/data_provider.rs
DamianStraszak, line 311: This change I like, although I'm sure it would be controversial if timorl was around :P

Anyway, could you perhaps add some little comment to `ChainTracker` that the other thread is "consuming" (and leaving `None`) from `data_to_propose`. I imagine one could sometime in the future make a wrong assumption that `get_data` does not modify `data_to_propose`.

finality-aleph/src/data_io/mod.rs
DamianStraszak: You can also try to derive `Hash` -- previously it was not possible, but we are a couple of versions of rustc ahead.

DamianStraszak (commented):
Nice, the PR "feels right" :)

finality-aleph/src/data_io/data_provider.rs
mike1729: Done.

finality-aleph/src/data_io/mod.rs
mike1729: Removed the line but unfortunately, BlockT contains sp_std::hash::Hash and not std::hash::Hash and derive still don't work.

DamianStraszak (approved):

PR #560, opened 2022-07-27
Perform word splitting on custom arguments

artemis-cc:
# Description

Custom arguments set via the Docker environment file are passed to the entrypoint script as a single string (rather than anything more structured like an array). Use xargs to split this string into shell words, enabling the passing of custom arguments which individually contain whitespace (such as `--telemetry-url 'wss://telemetry.polkadot.io/submit/ 1'`).

Fixes Cardinal-Cryptography/aleph-node-runner#14.

## Type of change

# Checklist:


piotrMocz (approved):
Awesome! :100:

fixxxedpoint (approved):
LGTM. Seems like old version + `eval aleph-node "${ARGS[@]}"` achieves same (not sure).

PR #564, opened 2022-07-29
A0-1179: Stateful checks for pallet elections migrations

pmikolajczyk41:
# Description

Goal: we verify whether the storage content has been transferred correctly (using set_temp_storage / get_temp_storage). Additionally, we include these checks in unit tests.

## Type of change

# Checklist:

- I have added tests


krzysztofziobro (approved):
Looks fine

pallets/elections/src/migrations/mod.rs
```diff
@@ -1,14 +1,26 @@
...
+    pallet_prelude::StorageVersion, storage::storage_prefix, traits::PalletInfoAccess,
+};
 use frame_support::{traits::OnRuntimeUpgrade, weights::Weight};
+use sp_std::vec::Vec;
 
 pub mod v0_to_v1;
 pub mod v1_to_v2;
 pub mod v2_to_v3;
 
+type ValidatorsVec<T> = Vec<<T as frame_system::Config>::AccountId>;
```
fbielejec, line 14: could be called just `Validators` imho

fbielejec (approved):
lgtm

pallets/elections/src/migrations/mod.rs
pmikolajczyk41: renamed

PR #565, opened 2022-08-01
debugging featurenet(s)

fbielejec:
# Description

fixes for fork-off tool: with addition of contracts pallet there are now child state keys in storage. These keys throw when bootstrapping node from a forked chainspec (https://github.com/Cardinal-Cryptography/substrate/blob/98c2eeea74413044ae8ccfca1b6d56d01b57a76b/client/db/src/lib.rs#L778). We omit those keys

## Type of change

# Checklist:


fork-off/src/chainspec_combining.rs
```diff
@@ -48,7 +49,7 @@ pub fn combine_states(
                 removed_per_path_count.bump(path);
                 false
             }
-            None => true,
+            None => !well_known_keys::is_child_storage_key(&Into::<Vec<u8>>::into(k)), // remove child keys, they are not accepted as part of storage, see https://github.com/Cardinal-Cryptography/substrate/blob/98c2eeea74413044ae8ccfca1b6d56d01b57a76b/client/db/src/lib.rs#L778
```
obrok, line 52: Rust newb question: why do we need `Into::<Vec<u8>>::into(k)`? I would have expected `k.into()` or maybe `k.into::<Vec<_>>()` to work.

obrok: Substrate newb question: why is it OK to drop these keys during fork-off? Are they somehow unimportant?

obrok (approved):
Looks reasonable, questions are to satisfy my own curiosity.

fork-off/src/chainspec_combining.rs
pmikolajczyk41: `Into::<Vec<u8>>::into(k)` should be equivalent to `k.into::<Vec<_>>()`; `k.into()` might be ambiguous though - depending on the context

fbielejec: ad1) into() doesn't take generic arguments:
```
    fn into(self) -> U {
        U::from(self)
    }
```
ad2) see code-comment and the PR description

obrok: > ad2) see code-comment and the PR description

I did read those, and I can see that they are not accepted. What I'm wondering is why is it OK to just remove some keys? Seems like that will cause the forked-off chain to have a different state in some way, and what I don't get is in what way exactly.

bin/node/src/chain_spec.rs
```diff
@@ -131,7 +131,7 @@ pub struct ChainParams {
     token_symbol: String,
 
     /// AccountIds of authorities forming the committee at the genesis (comma delimited)
-    #[clap(long, require_value_delimiter = true, parse(from_str = parse_account_id))]
+    #[clap(long, require_delimiter = true, value_delimiter = ',', parse(from_str = parse_account_id))]
```
pmikolajczyk41, line 134: that's weird - we are using `clap = 3.2.12`, so `require_delimiter` should be deprecated in favor of `require_value_delimiter`... https://docs.rs/clap/3.2.12/clap/struct.Arg.html#method.require_delimiter

it seems that we should leave `require_value_delimiter`, but add:
 - [takes_value](https://docs.rs/clap/3.2.12/clap/builder/struct.Arg.html#method.takes_value)
 - [use_value_delimiter](https://docs.rs/clap/3.2.12/clap/builder/struct.Arg.html#method.use_value_delimiter)


(https://docs.rs/clap/3.2.12/clap/builder/struct.Arg.html#method.require_value_delimiter)

pmikolajczyk41 (approved):

bin/node/src/chain_spec.rs
fbielejec: done, although I kept the "," requirement - I don't see us using any other delimiter anyway

fork-off/src/chainspec_combining.rs
obrok: :+1:

pmikolajczyk41:
maybe change a bit the PR's title :P

PR #567, opened 2022-08-02
A0-1208: Stop asking for sudo in case of rpc

Yithis:
# Description

Stop asking for sudo in case of rpc "finalize" in cliain

## Type of change

# Checklist:

### DIFF
--- a/bin/cliain/src/main.rs
+++ b/bin/cliain/src/main.rs
@@ -22,7 +22,7 @@ struct Config {
     #[clap(long, default_value = "127.0.0.1:9944")]
     pub node: String,
 
-    /// The seed of the key to use for signing calls
+    /// The seed of the key to use for signing calls.
     /// If not given and the command is not rpc call,
     /// a user is prompted to provide seed
     #[clap(long)]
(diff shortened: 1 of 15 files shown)


bin/cliain/src/main.rs
```diff
@@ -54,7 +55,14 @@ fn main() {
         command,
     } = Config::parse();
 
-    let seed = read_secret(seed, "Provide seed for the signer account:");
+    let seed = match command {
+        Command::Finalize {
```
maciejnems, line 59: I guess RotateKeys and GetStorage also do not need the seed phrase, right? And also SeedToSS58 should take a seed as an argument to command not as a signer seed. Could we handle those in a similar manner?

Yithis: Done

maciejnems (approved):

bin/cliain/src/main.rs
DamianStraszak: Could you also move this piece of code to a separate function. `main` is already packed...

DamianStraszak (approved):

bin/cliain/src/main.rs
Yithis: Done

PR #572, opened 2022-08-03
A0-1221: Extract migration support to a new crate

pmikolajczyk41:
# Description

We move migration support code to a new crate, so that our other pallets (for now: `aleph`) could benefit from better storage migration checks (A0-1222).

## Type of change

# Checklist:


pallets/elections/Cargo.toml
```diff
@@ -37,8 +38,12 @@ std = [
     "pallet-session/std",
     "pallet-staking/std",
     "pallet-authorship/std",
+    "pallets-support/std",
```
Marcin-Radecki, line 41: Could you please bump `version`? Tbh not sure whether this is breaking change or not, but I guess `0.4.0` would to the job.

pallets/support/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "pallets-support"
```
Marcin-Radecki, line 2: Any particular reason to name it like this? Ie is it some substrate convention? I was thinking of anything `migration` in the name

Marcin-Radecki (approved):

pallets/support/Cargo.toml
pmikolajczyk41: I'm expecting that more things may appear here, so in general: support (like frame-support)

pallets/elections/Cargo.toml
pmikolajczyk41: sure, done

fbielejec (approved):

PR #573, opened 2022-08-04
A0-1232 Added full support for cliain to changeValidators

Marcin-Radecki:
# Description

As in the title

## Type of change


bin/cliain/src/commands.rs
```diff
@@ -108,9 +108,21 @@ pub enum Command {
...
-        validators: Vec<String>,
+        reserved_validators: Vec<String>,
+
+        /// The new non-reserved validators list
+        #[clap(long, value_delimiter = ',')]
+        non_reserved_validators: Vec<String>,
+
+        /// The new reserved committee size
+        #[clap(long)]
+        reserved_committee_size: Option<u32>,
```
kostekIV, line 121: now leaving it set it to 0 which can be error-prone. The user could expect not to change it at all. Either require this argument from user or let the `None` variant mean that this particular field will not be changed

kostekIV (changes requested):

bin/cliain/src/commands.rs
DamianStraszak: Yes, agreed, I'm of the opinion that we should be able to pass arguments to this in exactly the same format as `change_validators` expects -- and since it is impossible to achieve with a flat list of cli arguments, I would suggest passing a file with json which gets parsed to a struct like this:
```
#[derive(serde::Serialize, serde::Deserialize)]
struct ChangeValidatorArgs {
    reserved_validators: Option<Vec<AccountId>>,
    non_reserved_validators: Option<Vec<AccountId>>,
    committee_size: Option<CommitteeSeats>,
}
```

One create such a json very easily and parse it like that
`let change_validator_args: ChangeValidatorArgs = serde_json::from_reader(file).unwrap();`

This is the least error prone approach in my opinion.

DamianStraszak (changes requested):

bin/cliain/src/commands.rs
Marcin-Radecki: Agree with both comments, I think the just pushed version is significantly more cleaner. Rust does its magic and can even convert `AccountId` from `String` under the hood, since there exists `impl FromStr for AccountId` (https://paritytech.github.io/substrate/master/sp_core/crypto/struct.AccountId32.html#impl-FromStr-for-AccountId32) .

Here's now how a user can run `cliain`:
```bash
 ./target/release/cliain change-validators --change-validators-args '{"reserved_validators": ["5D34dL5prEUaGNQtPPZ3yN5Y6BnkfXunKXXz6fo7ZJbLwRRH","5GBNeWRhZc2jXu7D55rBimKYDk8PGk8itRYFTPfC8RJLKG5o"], "non_reserved_validators": ["5Dfis6XL8J2P6JHUnUtArnFWndn62SydeP8ee8sG2ky9nfm9","5F4H97f7nQovyrbiq4ZetaaviNwThSVcFobcA5aGab6167dK"], "committee_size": {"reserved_seats": 2, "non_reserved_seats": 2}}'
```
Our goal was achieved, now it is not possible to miss any of the fields `reserved_seats` or `non_reserved_seats` - both are required. 

Note that there are some quirks that can make life hard for users not experienced with json format and bash:
* one must put single quotes around the whole json, otherwise bash would escape `"` and json would be malformed,
* since `reserved_validators` and `non_reserved_validators` are `Option`s, it is possible to make a typo e.g. `reservde_validators` which will be hard to miss on the first look, but serde won't complain. This can be very annoying if user does not check whether extrinsic succeded on their own - we don't have now in `cliain` possibility to check whether tx was successfull, there's even a comment about it in `validators.rs`.

Marcin-Radecki: Let me know if you'd like me to add support for parsing JSON from file - it should be easy to extent current code to when json parse from str failed, logic will try to parse json from filename.

bin/cliain/src/commands.rs
```diff
@@ -108,9 +124,9 @@ pub enum Command {
 
     /// Change the validator set for the session after the next
     ChangeValidators {
-        /// The new validators
-        #[clap(long, value_delimiter = ',')]
-        validators: Vec<String>,
+        /// The new reserved validators list
+        #[clap(long)]
+        change_validators_args: ChangeValidatorArgs,
```
kostekIV, line 129: This means we have to paste json into cmd line right? Could you also add a support for passing path to file with a json?

bin/cliain/src/commands.rs
DamianStraszak: Awesome, I like this version a lot. If this is not much work you can try adding a version that reads from file, but it's up to you (don't bother if you are busy).

DamianStraszak (approved):

bin/cliain/src/commands.rs
Marcin-Radecki: Done:
```bash
marcin@marcin-Latitude-3520 ~/g/a/b/cliain (non-reserved-nodes-support-cliain) [1]> cat test.json                                                                                                   (base) 
{
  "reserved_validators": [
    "5D34dL5prEUaGNQtPPZ3yN5Y6BnkfXunKXXz6fo7ZJbLwRRH",
    "5GBNeWRhZc2jXu7D55rBimKYDk8PGk8itRYFTPfC8RJLKG5o"
  ],
  "non_reserved_validators": [
    "5Dfis6XL8J2P6JHUnUtArnFWndn62SydeP8ee8sG2ky9nfm9",
    "5F4H97f7nQovyrbiq4ZetaaviNwThSVcFobcA5aGab6167dK"
  ],
  "committee_size": {
    "reserved_seats": 2,
    "non_reserved_seats": 2
  }
}
```
And now it is possible just to do `./target/release/cliain change-validators --change-validators-args test.json` :)

kostekIV (approved):
Looks fantastics :)

PR #582, opened 2022-08-09
Improve experience of writing party tests

kostekIV:
## Type of change

- tests

# Checklist:

- I have added tests

### DIFF
--- a/finality-aleph/src/party/mod.rs
+++ b/finality-aleph/src/party/mod.rs
@@ -237,7 +237,6 @@ where
     pub async fn run(mut self) {
         let starting_session = self.catch_up().await;
         for curr_id in starting_session.0.. {
-            println!("Running session {:?}.", curr_id);
             info!(target: "aleph-party", "Running session {:?}.", curr_id);
             self.run_session(SessionId(curr_id)).await;
         }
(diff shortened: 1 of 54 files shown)


finality-aleph/src/party/mod.rs
```diff
@@ -324,275 +545,201 @@ mod tests {
...
+            validator_started: vec![SessionId(0), SessionId(1)],
+            early_started: vec![SessionId(1)],
+            stopped: vec![SessionId(0)],
+            non_validator_started: vec![],
+        };
+
+        test.set_authorities_for_session_at_block(0, authorities.clone(), SessionId(0))
+            .set_authorities_for_session_at_block(25, authorities, SessionId(1))
+            .set_node_id_for_session_at_block(0, Some(UintAuthorityId(0).to_public_key()))
+            .assert_session_states_at_block(28, state_1)
```
Marcin-Radecki, line 573: Name nitpick: I find the `expect` phrase better here since at the stage of calling `assert_session_states_at_block` we are not asserting anything. Also, `set_node_id_for_session_at_block` has already input args such as `expected_state` etc

Marcin-Radecki (approved):
I like this rework very much.

finality-aleph/src/party/mod.rs
kostekIV: done

piotrMocz (approved):
Great work :100:

PR #583, opened 2022-08-11
Create a separate docker image for OnFinality where entrypoint is not wrapped with a bash script

mikogs:
# Description

Adds `Dockerfile.onfinality` where `docker_entrypoint.sh` is not added and `/usr/local/bin/aleph-node` is used as an entrypoint. This is one of the requirements to run Aleph Node on OnFinality.

Also, a new step that builds the `aleph-node:latest-onfinality` image has been added. Along with steps to push it to Docker Hub (though these are commented out for now).


.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -448,6 +448,11 @@ jobs:
           chmod +x target/release/aleph-node
           docker build --tag aleph-node:latest -f ./docker/Dockerfile .
 
+      - name: Build docker image for onfinality
+        id: build-image-onfinality
+        run: |
+          docker build --tag aleph-node:latest-onfinality -f ./docker/Dockerfile.onfinality .
+
```
Kosiare, line 455: Maybe to speedup a build process I would add to this step `--cache-from` option from latest image. Example:
`docker build --cache-from aleph-node:latest-onfinality --tag aleph-node:latest-onfinality -f ./docker/Dockerfile.onfinality .`

Kosiare (commented):
Generally LGTM but I would add one thing that I described in other comment.

.github/workflows/e2e-tests-main-devnet.yml
mikogs: Added

Kosiare (approved):

PR #584, opened 2022-08-11
Change consensus engine ID

pmikolajczyk41:
# Description

Since https://github.com/paritytech/substrate/pull/8266 resolved the blocker, we can get rid of the default consensus engine ID, which previously was `FRNK`.

If you are curious what `FRNK` means, let these screenshots from https://github.com/paritytech/substrate/pull/2802 confuse you:
image
image

## Type of change

Changes in UI:
image

# Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -142,7 +142,7 @@ dependencies = [
 
 [[package]]
 name = "aleph-node"
-version = "0.7.0"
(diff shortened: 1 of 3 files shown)


pmikolajczyk41:
there is a need of compatibility support, which doesn't seem to be ready soon

PR #586, opened 2022-08-11
A0-1159: The Pressiah Cometh

h4nsu:
# Description

The third variant of the button game (tm).

## Type of change

# Checklist:

- I have added tests


contracts/scripts/deploy.sh
```diff
@@ -219,6 +222,22 @@ instrument_game_token BACK_TO_THE_FUTURE_TOKEN game_token 0x4261636B546F54686546
 
 deploy_and_instrument_game BACK_TO_THE_FUTURE back_to_the_future $BACK_TO_THE_FUTURE_TICKET $BACK_TO_THE_FUTURE_TOKEN
 
+#
+# --- THE_PRESSIAH_COMETH GAME
+#
+
+# --- CREATE AN INSTANCE OF THE TICKET CONTRACT FOR THE THE_PRESSIAH_COMETH GAME
+
+instrument_ticket_token THE_PRESSIAH_COMETH_TICKET ticket_token 0x4261636B546F546865467574752137 back_to_the_future BTF
```
fbielejec, line 231:
```suggestion
instrument_ticket_token THE_PRESSIAH_COMETH_TICKET ticket_token 0x7468655F70726573736961685F636F6D657468 the_pressiah_cometh TPC
```

contracts/scripts/deploy.sh
```diff
@@ -219,6 +222,22 @@ instrument_game_token BACK_TO_THE_FUTURE_TOKEN game_token 0x4261636B546F54686546
...
+
+# --- CREATE AN INSTANCE OF THE TOKEN CONTRACT FOR THE THE_PRESSIAH_COMETH GAME
+
+instrument_game_token THE_PRESSIAH_COMETH_TOKEN game_token 0x4261636B546F546865467574752137
```
fbielejec, line 235:
```suggestion
instrument_game_token THE_PRESSIAH_COMETH_TOKEN game_token 0x7468655F70726573736961685F636F6D657468
```

contracts/button/lib.rs
```diff
@@ -161,8 +148,7 @@ pub trait ButtonGame {
     /// where k is the number of players that participated until the button has died
     /// Can be overriden to some other custom calculation
     fn pressiah_score(&self) -> Balance {
-        let presses = self.get().presses;
-        (presses * num::integer::sqrt(presses)) as Balance
+        (self.get().total_rewards) as Balance
```
fbielejec, line 151: that is not what we established? We wanted to lower the Pressiah reward to subquadratic?

_as we decided to mint button tokens instead of distributing them from the predefined pool, we may actually use arbitrary function here, so how about using something subquadratic instead of the current proposal? I.e., sth like k*sqrt(k), which wouldn't be such a high percentage in case of a very long streak. (...) communicate it as a 'pressiah multiplier', i.e., the pressiah would get k due to clicking as k-th person in the row, and it will be multiplied by the multiplier, that could be a square root of k (rounded up)_

h4nsu: I missed that part, I thought the pressiah is always supposed to get the half of total distributed rewards

fbielejec (approved):

contracts/button/lib.rs
```diff
@@ -157,12 +144,10 @@ pub trait ButtonGame {
...
     /// Logic for calculating pressiah score
     ///
-    /// By defaul the pressiah score is defined as k * sqrt(k)
-    /// where k is the number of players that participated until the button has died
+    /// By default the pressiah gets 20% of the total rewards paid in the current game iteration
     /// Can be overriden to some other custom calculation
     fn pressiah_score(&self) -> Balance {
-        let presses = self.get().presses;
-        (presses * num::integer::sqrt(presses)) as Balance
+        (self.get().total_rewards / 4) as Balance
```
fbielejec, line 150: num dependency can be removed

h4nsu: done

contracts/back_to_the_future/lib.rs
```diff
@@ -5,17 +5,15 @@ use ink_lang as ink;
 /// This is the BackToTheFuture
 ///
 /// Larger rewards are distributed for postponing playing for as long as possible, but without letting TheButton die:
-/// user_score = now - start
-/// ThePressiah (the last player to click) still gets 50% of the tokens.
+/// user_score = now - previous_press
+/// ThePressiah (the last player to click) gets 20% of the tokens.
```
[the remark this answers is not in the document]
h4nsu: README updated

contracts/button/lib.rs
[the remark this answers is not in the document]
h4nsu: 1/4 of rewards paid so far results in pressiah reward being 1/5 of all paid rewards

PR #588, opened 2022-08-12
update nightly

fbielejec:
# Description

Bump nightly version to fix compilation errors

## Type of change

# Checklist:


mike1729 (approved):

rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2022-05-23
+nightly-2022-08-12
```
DamianStraszak, line 1: Maybe something like that would be more `pro`? It then automatically installs the wasm target.
```
[toolchain]
channel = "nightly-2022-08-12"
targets = [ "wasm32-unknown-unknown" ]
```

timorl (approved):
Looks good, although if @DamianStraszak's comment works it would be great to have.

rust-toolchain
fbielejec: I wrestled with it in a past, but the actions-rs/toolchain action does not understand toml and it's still in v1. Probably unmaintained too: https://github.com/actions-rs/toolchain/issues/216

DamianStraszak: Ok, fine, then you can ignore my comment.

PR #589, opened 2022-08-12
A0-1190 Prisoner's Dilemma SC

Marcin-Radecki:
# Description

Implementation of classic Prisoner's Dilemma (https://en.wikipedia.org/wiki/Prisoner%27s_dilemma) . 

There are currently two UI drawbacks, which rootcause lays most likely deep inside in a way how SC are executed:
* there are no concrete errors (defined in SC itself) shown, just `Contracts.Reverted` - it's very hard to figure out what's going on when things does not work,
* custom contract event are not shown in UI as well, they are just shown as `Contracts.EventsEmitted`, a generic event

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/contracts/two_prisoners_dilemma/.gitignore
+++ b/contracts/two_prisoners_dilemma/.gitignore
@@ -1,0 +1,9 @@
+# Ignore build artifacts from the local tests sub-crate.
+/target/
+
+# Ignore backup files creates by cargo fmt.
+**/*.rs.bk
+
+# Remove Cargo.lock when creating an executable, leave it for libraries
+# More information here http://doc.crates.io/guide.html#cargotoml-vs-cargolock
+Cargo.lock
--- a/contracts/two_prisoners_dilemma/lib.rs
+++ b/contracts/two_prisoners_dilemma/lib.rs
@@ -1,0 +1,527 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use ink_lang as ink;
+
+/// Smart Contract modelling classic Prisoner's Dilemma problem.
+/// To receive rewards, the contract needs to be pre-founded in the first place
+#[ink::contract]
+mod two_prisoners_dilemma {
+    use ink_lang::utils::initialize_contract;
+    use ink_prelude::string::{String, ToString};
+    use ink_storage::{
+        traits::{PackedLayout, SpreadAllocate, SpreadLayout},
+        Mapping,
+    };
+    use scale::{Decode, Encode};
+
+    // For now we harcode number of players to two, to model original dillema
+    const NUMBER_OF_PLAYERS: usize = 2;
+    pub const TOKEN_DECIMALS: u32 = 12;
+    pub const TOKEN: u128 = 10u128.pow(TOKEN_DECIMALS);
+
+    /// A choice of a player during the game
+    /// Initially all registered players are in `Waiting` state
+    #[derive(Debug, Encode, Decode, Clone, Copy, SpreadLayout, PackedLayout, PartialEq, Eq)]
+    #[cfg_attr(
+        feature = "std",
+        derive(scale_info::TypeInfo, ink_storage::traits::StorageLayout)
+    )]
+    enum Choice {
+        Waiting,
+        Cooperate,
+        Defect,
+    }
+
+    /// An event emitted when a reward was claimed and transfer succeeded
+    /// From acocunt is always the contract one
+    #[ink(event)]
+    pub struct Transfer {
+        #[ink(topic)]
+        to: AccountId,
+        value: Balance,
+    }
+
+    /// An event emitted when a player makes a legitimate choice
+    /// legitimate: not duplicate, and from registered players only
+    /// Obviously the exact choice is not revealed
+    #[ink(event)]
+    pub struct PlayerMadeChoice {
+        #[ink(topic)]
+        player: AccountId,
+    }
+
+    /// Written result of the game for a given player
+    #[ink(event)]
+    pub struct ResultOfTheGame {
+        #[ink(topic)]
+        player: AccountId,
+        result: String,
+    }
+
+    #[ink(storage)]
+    #[derive(SpreadAllocate)]
+    pub struct TwoPrisonersDilemma {
+        /// A reward a cooperating player receives when other one defect
+        bad_luck_payoff: u128,
+
+        /// A reward both players receive when they both defect
+        defect_payoff: u128,
+
+        /// A reward both players receive when they both cooperate
+        cooperation_payoff: u128,
+
+        /// A reward a defecting player gets when other one cooperate
+        temptation_payoff: u128,
+
+        /// Players choices, lookup of account to either choice of cooperate or defect
+        choices: Mapping<AccountId, Choice>,
+
+        /// How many legitimate votes has been made so far
+        /// legitimate: not duplicate, and from registered players only
+        vote_count: u32,
+
+        /// Registered player accounts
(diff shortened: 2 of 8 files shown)


contracts/two_prisoners_dilemma/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+    #[ink(event)]
+    pub struct Transfer {
+        #[ink(topic)]
+        to: AccountId,
+        value: Balance,
+    }
+
+    /// An event emitted when a player makes a legitimate choice
+    /// legitimate: not duplicate, and from registered players only
+    /// Obviously the exact choice is not revealed
```
h4nsu, line 46: The whole problem with this game is that the choice is actually revealed. It's stored in the contract storage and everybody can go there and see it. Not sure how that could be circumvented, but in the current form the game does not make much sense.

It's still fine if we're ok with the game being pointless and the sole purpose of this contract is to be a tutorial, but in that case some work needs to be done wrt storage optimisation (currently you're storing much more data than needed)

contracts/two_prisoners_dilemma/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+            initialize_contract(|contract: &mut Self| {
+                contract.players = players.clone();
+                for player in players {
+                    contract.choices.insert(&player, &Choice::Waiting);
+                }
+
+                contract.bad_luck_payoff = 0;
+                contract.defect_payoff = TOKEN;
+                contract.cooperation_payoff = 2 * TOKEN;
+                contract.temptation_payoff = 3 * TOKEN;
```
h4nsu, line 138: If all these are constants, why do you put them as storage items? It would be cheaper just to do:
```
const BAD_LUCK_PAYOFF = 0;
const DEFECT_PAYOFF = TOKEN;
etc
```

h4nsu (changes requested):
I'd like to have a quick chat about the assumptions behind this

contracts/two_prisoners_dilemma/lib.rs
fbielejec: I think you attacked quite a hard problem here, where the task was simpler imo - ([rewrite popular ethereum contract(s) in ink!](https://cardinal-cryptography.atlassian.net/browse/A0-1190)).

IMHO writing:
- an escrow
- a blind auction
- a multisig wallet
- some defi "primitives" like a stablecoin
- one of the popular standards not yet implementen by OpenBrush (like ERC-777 or ERC-1155)

would be a) simpler, with lots of existing materials to draw inspiration from b) more "educational" for peeps looking to switch from Sol -> ink!

contracts/two_prisoners_dilemma/.gitignore
```diff
@@ -0,0 +1,9 @@
+# Ignore build artifacts from the local tests sub-crate.
+/target/
```
fbielejec, line 2: there's no need to have another .gitignore, just specify them in the one in the projects root no? the whole .idea directory looks like something your IDE created, so likely does not belong under version control either?

contracts/two_prisoners_dilemma/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+
+        fn get_other_player_choice(&self, this_player: &AccountId) -> Choice {
+            let other_player_index: usize = match self.players[0] == *this_player {
+                true => 1,
+                false => 0,
+            };
+            return self.choices.get(self.players[other_player_index]).unwrap();
+        }
+
+        fn make_choice(&mut self, choice: &Choice) -> Result<()> {
```
fbielejec, line 160: you should conceal the choice, for example by submitting a keccak256(Choice, secret) and storing that in a AccountId => Hash mapping. Than you should have a reveal_choice with arguments Choice, secret that check whether it hashes to the same value.

There might be people that don't reveal in time: you could alter the make_choice to take a said hash and a plaintext FakeChoice, and using that when comes the time to reveal, something like that

contracts/two_prisoners_dilemma/lib.rs
Marcin-Radecki: To be honest it was not my intention to give out plain choices in the store. I realized this the same day afternoon after I published this PR. This is a serious flaw in the current versions when tokens are sent as rewards. Nevertheless, the code treated as an example is still a good use case of `ink!` features, such as storage, events, token transfer from contract, `ink!` unit tests, so I think it make sense to publish this example in either form:
1. Just change comments so that it is clearly said this is a non-production SC with transparent votes, only for tutorial purposes, or
2. Put some more effort and add hashing of the vote as per @fbielejec comment below, to get at least non-transparent votes. This version would be more or less similar to Blinded Auction (https://docs.soliditylang.org/en/v0.8.16/solidity-by-example.html#id2). 

My intention also was to come up with a new idea other than that already presented in https://github.com/paritytech/ink/tree/master/examples , which contains already name server(https://github.com/paritytech/ink/tree/master/examples/dns) , nevertheless but the list you provided looks very interesting. 

What do you think we still go with this PR with version 1. or 2., and I'll prepare another PR with a classic example like Escrow or MultiSig wallet?

contracts/two_prisoners_dilemma/lib.rs
Marcin-Radecki: My intention was, in the future this SC can have constructor which initilize these values to make the game more general, but I guess for the first version it is fine to hardcode it.

contracts/two_prisoners_dilemma/.gitignore
Marcin-Radecki: Not sure how this slipped through, I think `cargo contract new` made this .gitignore, but till I'm not sure how I added it to VCS. Anyway, let me remove it.

contracts/two_prisoners_dilemma/lib.rs
Marcin-Radecki: Right, as per blinded auction https://docs.soliditylang.org/en/v0.8.16/solidity-by-example.html#id2 right? What I don't understand is how one computes this a keccak256(Choice, secret) on the side. 

Btw do you know what's the equivalent of Solidity's `payable` in `ink!`'s world? E.g with below call user somehow transfers tokens to contract?
```
function bid(bytes32 blindedBid)
        external
        payable
        onlyBefore(biddingEnd)
    {
        bids[msg.sender].push(Bid({
            blindedBid: blindedBid,
            deposit: msg.value
        }));
    }
```

fbielejec: You have `#[ink(message, payable)]`.

deuszx: > What I don't understand is how one computes this a keccak256(Choice, secret) on the side.

If it was a game with a frontend then you could hash the input on user's side (in the browser) and never leak it oustide.

Marcin-Radecki: Turns out it is enough to paste sth llike this in https://test.azero.dev/#/js

>console.log(hashing.keccakAsHex("secret"));
console.log(hashing.keccakAsU8a("secret"));

Ouput

>0x65462b0520ef7d3df61b9992ed3bea0c56ead753be7c8b3614e0ce01e4cac41b
101,70,43,5,32,239,125,61,246,27,153,146,237,59,234,12,86,234,215,83,190,124,139,54,20,224,206,1,228,202,196,27

deuszx: Oh yes, I thought you were asking how to automate it and take the burden off of the user.

Marcin-Radecki:
Turned out this requires more work, closing for now

PR #593, opened 2022-08-18
A0-1236: minting and burning tokens

pmikolajczyk41:


contracts/game_token/lib.rs
```diff
@@ -35,39 +40,9 @@ pub mod game_token {
...
-                Event::Approval(Approval {
-                    owner: _owner,
-                    spender: _spender,
-                    value: _amount,
-                }),
-            );
-        }
-    }
+    impl PSP22Mintable for GameToken {}
+    impl PSP22Burnable for GameToken {}
```
fbielejec, line 45: you should override the default mint and burn to check the role of the caller

fbielejec:
Can we close? @pmikolajczyk41

PR #595, opened 2022-08-19
A0-1242 Aggregator Never Dies

fixxxedpoint:
Early exit of aggregator.rs was causing a cascade of errors related with its closed channels, i.e. split.rs was indicating it as an error. This PR tries to simplify several related APIs and properly handle early exits in authority.rs (not logging them as errors).

Main changes:
- refactored split.rs
- allowing aggregator.rs to early exit and not logging it as an error
- refactored ComponentNetwork (R, S channels) to simplify its lifetime management
- SpawnHandle: allowing to spawn Futures with Output = Result<(), ()>


finality-aleph/src/party/manager/task.rs
```diff
@@ -12,26 +12,43 @@ pub type Handle = Pin<Box<(dyn Future<Output = sc_service::Result<(), ()>> + Sen
 pub struct Task {
     handle: Handle,
     exit: oneshot::Sender<()>,
+    error_on_exit: Option<bool>,
```
obrok, line 15: Suggesion: I'm confused by this three-state bool here, so I would propose to change it to some enum with informational variant names. Maybe `status: Running/ExitedWithError/ExitedCleanly`? As I said I'm confused, so not sure what exact information this is supposed to convey, so my proposed names may be way off.

finality-aleph/src/party/manager/authority.rs
```diff
@@ -65,31 +67,39 @@ impl Subtasks {
...
         trace!(target: "aleph-party", "Refresher stopped");
-        self.data_store.stop().await;
+        if self.data_store.stop().await.is_err() {
+            warn!(target: "aleph-party", "DataStore stopped with en error");
+        }
         trace!(target: "aleph-party", "DataStore stopped");
     }
 
     /// Blocks until the task is done and returns true if it quit unexpectedly.
     pub async fn failed(mut self) -> bool {
```
obrok, line 89: Suggestion: You could change this to also return a `Result<(), ()>` - a `Result` is self-documenting to an extent, while a `bool` requires some extra knowledge of which value means what.

finality-aleph/src/network/split.rs
```diff
@@ -77,17 +131,22 @@ async fn forward_or_wait<
...
     left_sender: &mpsc::UnboundedSender<LeftData>,
     right_sender: &mpsc::UnboundedSender<RightData>,
+    name: &String,
 ) -> bool {
     match receiver.lock().await.next().await {
         Some(Split::Left(data)) => {
             if left_sender.unbounded_send(data).is_err() {
-                warn!(target: "aleph-network", "Failed send despite controlling receiver, this shouldn't've happened.");
+                // this is totally fine - the other half of the channel can be dropped by any reason
+                // but it's not our responsibility to track it here
```
obrok, line 140:
```suggestion
                // This is totally fine - the other half of the channel can be dropped for any reason
                // but it's not our responsibility to track it here.
```

finality-aleph/src/network/split.rs
```diff
@@ -77,17 +131,22 @@ async fn forward_or_wait<
...
+                debug!(target: "aleph-network", "Unable to send to LeftNetwork ({}) - already disabled", name);
             }
             true
         }
         Some(Split::Right(data)) => {
             if right_sender.unbounded_send(data).is_err() {
-                warn!(target: "aleph-network", "Failed send despite controlling receiver, this shouldn't've happened.");
+                // this is totally fine - the other half of the channel can be dropped by any reason
+                // but it's not our responsibility to track it here
```
obrok, line 148:
```suggestion
                // This is totally fine - the other half of the channel can be dropped for any reason
                // but it's not our responsibility to track it here.
```

finality-aleph/src/network/split.rs
obrok: Suggestion/nit: have this comment once before the `match` to reduce visual noise.

finality-aleph/src/party/manager/task.rs
```diff
@@ -12,26 +12,43 @@ pub type Handle = Pin<Box<(dyn Future<Output = sc_service::Result<(), ()>> + Sen
...
+            error_on_exit: None,
+        }
     }
 
     /// Cleanly stop the task.
-    pub async fn stop(self) {
+    pub async fn stop(self) -> Result<(), ()> {
+        if let Some(res) = self.error_on_exit {
+            return if res { Err(()) } else { Ok(()) };
+        }
```
obrok, line 32: Suggestion: Perhaps this would be clearer as:

```suggestion
        match self.error_on_exit {
            Some(true) => return Err(()),
            Some(false) => return Ok(()),
            None => ()
        }
```

This will also happen naturally if you go with my enum suggestion.

obrok (approved):
LGTM, left some stylistic suggestions.

finality-aleph/src/party/manager/task.rs
```diff
@@ -12,7 +12,7 @@ pub type Handle = Pin<Box<(dyn Future<Output = sc_service::Result<(), ()>> + Sen
 pub struct Task {
     handle: Handle,
     exit: oneshot::Sender<()>,
-    error_on_exit: Option<bool>,
+    cached_result: Option<Result<(), ()>>,
```
obrok, line 15: :+1: This naming makes it much clearer to me!

finality-aleph/src/party/manager/task.rs
fixxxedpoint: changed its type to `Option<Result<(), ()>>` and renamed to `cached_result`

finality-aleph/src/party/manager/authority.rs
fixxxedpoint: changed to `fn wait_completion(mut self) -> Result<(), ()>`

finality-aleph/src/network/split.rs
fixxxedpoint: moved

finality-aleph/src/aggregation/aggregator.rs
```diff
@@ -178,19 +178,19 @@ impl<
...
-                    return Some(res);
+                    return Ok(Some(res));
                 }
                 Err(AggregatorError::LastHashPlaced) => {
                     debug!(
                         target: "aleph-aggregator",
                         "Terminating next_multisigned_hash because the last hash has been signed."
                     );
-                    return None;
+                    return Ok(None);
```
timorl, line 193: I would say it's better to just remove this case – we already ensure in other places in the code that we will not attempt to create silly justifications here, they would not be accepted anyway, and this is the only (I believe?) component that checks its own special termination criterion.

If that whole refactor adding `Result`s to spawning was there only to support this, but you think it's actually a good idea in general you can leave it in place.

finality-aleph/src/network/component.rs
```diff
@@ -21,18 +20,8 @@ pub trait Receiver<D: Data>: Sync + Send {
...
 
-#[async_trait::async_trait]
-impl<D: Data, CN: Network<D>> DataNetwork<D> for CN {
-    fn send(&self, data: D, recipient: Recipient) -> Result<(), SendError> {
-        self.sender().send(data, recipient)
-    }
-    async fn next(&mut self) -> Option<D> {
-        self.receiver().lock_owned().await.next().await
-    }
+    fn into(self) -> (Self::S, Self::R);
```
timorl, line 24: Already discussed out of band, but this trait should be removed, and all its implementers should be made into various specializations of `SimpleNetwork` (perhaps now called `ComponentNetwork`). Then the latter struct should implement `Into<S, R>` and that should be enough for our purposes.

finality-aleph/src/network/split.rs
```diff
@@ -230,12 +247,14 @@ fn split_receiver<
             translated_receiver: left_receiver,
             left_sender: left_sender.clone(),
             right_sender: right_sender.clone(),
+            name: left_name.to_string(),
```
timorl, line 250: Hmm, couldn't you just pass the `&'static str` in here and not use `String` at all?

finality-aleph/src/network/split.rs
```diff
@@ -77,17 +131,21 @@ async fn forward_or_wait<
...
             if left_sender.unbounded_send(data).is_err() {
-                warn!(target: "aleph-network", "Failed send despite controlling receiver, this shouldn't've happened.");
+                debug!(target: "aleph-network", "Unable to send to LeftNetwork ({}) - already disabled", name);
             }
             true
         }
         Some(Split::Right(data)) => {
             if right_sender.unbounded_send(data).is_err() {
-                warn!(target: "aleph-network", "Failed send despite controlling receiver, this shouldn't've happened.");
+                debug!(target: "aleph-network", "Unable to send to LeftNetwork ({}) - already disabled", name);
```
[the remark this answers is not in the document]
timorl: On second thought with `name` maybe specifying `Left` and `Right` is unnecessary in the first place? I'm not sure, do as you will, just mentioning the possibility.

timorl (changes requested):
I absolutely _adore_ what you did with `Split`, it's brilliant! <3

Other than that some comments, a couple of which we already discussed in other channels, but not all of them.

finality-aleph/src/network/split.rs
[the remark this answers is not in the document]
fixxxedpoint: I would leave Left and Right so it will easier pinpoint this in our logs.

finality-aleph/src/aggregation/aggregator.rs
fixxxedpoint: Yep, this `Result<>` was added here to support a scenario where a task finishes early but without errors.

finality-aleph/src/network/split.rs
fixxxedpoint: changed

finality-aleph/src/aggregation/aggregator.rs
timorl: This might be good in general, but I would remove this specific case – this component really shouldn't be ending prematurely, just because it thinks that it's not responsible for any more blocks.

finality-aleph/src/network/session.rs
```diff
@@ -103,13 +87,13 @@ impl<D: Data> Manager<D> {
             .await
             .map_err(|_| ManagerError::NetworkReceiveFailed)?;
         let messages_for_network = self.messages_for_service.clone();
-        Ok(Network {
-            sender: Sender {
+        Ok(SimpleNetwork::new(
```
timorl, line 90: I think you can write `Network::new` here, and this should be more readable.

finality-aleph/src/network/split.rs
```diff
@@ -159,11 +179,9 @@ impl<
 {
     type S = LeftSender<LeftData, RightData, S>;
     type R = LeftReceiver<LeftData, RightData, R>;
-    fn sender(&self) -> &Self::S {
-        &self.sender
-    }
-    fn receiver(&self) -> Arc<Mutex<Self::R>> {
-        self.receiver.clone()
+
+    fn into(self) -> (Self::S, Self::R) {
```
timorl, line 183: Make `Left/RightNetwork` into instances of `SimpleNetwork` and return `impl ComponentNetwork` to avoid leaking types.

finality-aleph/src/testing/network.rs
```diff
@@ -59,6 +61,40 @@ impl NetworkIdentity for Authority {
     }
 }
 
+struct TestComponentNetwork<D: Data, R: ReceiverComponent<D>, S: SenderComponent<D>> {
```
timorl, line 64: This is exactly `SimpleNetwork`, cannot you use that?

timorl (approved):
Also there is some automated reference complaint, which seems correct.

finality-aleph/src/aggregation/aggregator.rs
fixxxedpoint: I am not sure what to do with this one. If I am supposed to remove this behavior, then probably the whole `AggregatorError::LastHashPlaced` should be removed, together with its `notify_last_hash` method - this component shouldn't know about any termination condition, caller should just stop querying it.

finality-aleph/src/network/split.rs
```diff
@@ -8,6 +8,7 @@ use futures::channel::mpsc;
 use log::{debug, trace};
 use tokio::sync::Mutex;
 
+use super::SimpleNetwork;
```
timorl, line 11: Small style thingy – we don't generally use `super` and I would keep it that way, especially since the line below is inconsistent with this.

finality-aleph/src/aggregation/aggregator.rs
timorl: Yup, I would say that's exactly what should happen.

timorl (approved):
<3

PR #596, opened 2022-08-19
Create justifications for all blocks

timorl:
# Description

Create justifications for all blocks. This not only fixes the linked bug (at least for validators), but also finally implements an ADR.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/data_io/status_provider.rs
+++ b/finality-aleph/src/data_io/status_provider.rs
@@ -413,5 +413,15 @@ mod tests {
             &fresh_proposal,
             Finalize(fresh_proposal.blocks_from_num(0).collect()),
         );
+
+        // Long proposals should finalize the appropriate suffix.
+        let long_proposal = proposal_from_blocks(blocks[0..6].to_vec());
+        verify_proposal_status(
+            &mut cached_cip,
+            &mut aux_cip,
+            &fresh_proposal,
+            // We are using fresh_proposal here on purpose, to only check the expected blocks.
+            Finalize(fresh_proposal.blocks_from_num(0).collect()),
+        );
     }
 }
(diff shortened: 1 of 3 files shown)


timorl:
I'm pretty sure the logic of status provider would now be even more natural if it iterated over the branch rather than properties, but this is urgent so I'm not touching that for now.

finality-aleph/src/data_io/status_provider.rs
```diff
@@ -238,7 +248,7 @@ mod tests {
                 cached_cip,
                 aux_cip,
                 &proposal,
-                ProposalStatus::Finalize(proposal.top_block()),
+                ProposalStatus::Finalize(proposal.blocks_from_num(0).collect()),
```
DamianStraszak, line 251: Would be great to have even a single test that checks the case when only some part of the branch is supposed to be finalized

DamianStraszak (changes requested):
Looks good, although I requested a small test to be added, which would nicely accompany the changes.

piotrMocz (approved):
Looks great!

finality-aleph/src/data_io/status_provider.rs
```diff
@@ -413,5 +413,15 @@ mod tests {
             &fresh_proposal,
             Finalize(fresh_proposal.blocks_from_num(0).collect()),
         );
+
+        // Long proposals should finalize the appropriate suffix.
+        let long_proposal = proposal_from_blocks(blocks[0..6].to_vec());
```
DamianStraszak, line 418: Hmm, this seems to be the only line which mentions `long_proposal` so I'm a little suspicious :P.

timorl: Whoopsie, should be better now.

DamianStraszak (approved):

PR #597, opened 2022-08-19
A0-1285: Implement common traits in aleph-client

obrok:
# Description

Applies [the trait ADR](https://www.notion.so/cardinalcryptography/Basic-traits-derivation-and-field-access-for-public-structs-0aa22c4887024547b613db5754e0664a) to public types in aleph-client.

## Type of change

# Checklist:

### DIFF
--- a/aleph-client/src/fee.rs
+++ b/aleph-client/src/fee.rs
@@ -3,7 +3,7 @@ use substrate_api_client::Balance;
 
 use crate::{AnyConnectionExt, Extrinsic};
 
-#[derive(Debug)]
+#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
 pub struct FeeInfo {
     pub fee_without_weight: Balance,
     pub unadjusted_weight: Balance,
--- a/aleph-client/src/fee.rs
+++ b/aleph-client/src/fee.rs
@@ -10,6 +10,12 @@ pub struct FeeInfo {
     pub adjusted_weight: Balance,
 }
 
+impl Default for FeeInfo {
+    fn default() -> Self {
+        FeeInfo { fee_without_weight: 0, unadjusted_weight: 0, adjusted_weight: 0 }
+    }
+}
+
 pub fn get_tx_fee_info<C: AnyConnectionExt, Call: Encode>(
     connection: &C,
     tx: &Extrinsic<Call>,
--- a/aleph-client/src/fee.rs
+++ b/aleph-client/src/fee.rs
@@ -3,21 +3,11 @@ use substrate_api_client::Balance;
 
 use crate::{AnyConnectionExt, Extrinsic};
 
-#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
+#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug, Default)]
 pub struct FeeInfo {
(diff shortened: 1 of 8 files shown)


aleph-client/src/fee.rs
```diff
@@ -3,7 +3,7 @@ use substrate_api_client::Balance;
 
 use crate::{AnyConnectionExt, Extrinsic};
 
-#[derive(Debug)]
+#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
```
pmikolajczyk41, line 6: [up to you] I would reconsider `Default` - zero fee is kinda natural default

aleph-client/src/multisig.rs
```diff
@@ -122,6 +125,7 @@ impl SignatureAggregation {
 
 /// `MultisigParty` is representing a multiparty entity constructed from
 /// a group of accounts (`members`) and a threshold (`threshold`).
+#[derive(Clone)]
```
pmikolajczyk41, line 128: [question] cannot we make it `(Partial)Eq`?

aleph-client/src/multisig.rs
```diff
@@ -91,7 +94,7 @@ struct Multisig {
 
 /// This represents the ongoing procedure of aggregating approvals among members
 /// of multisignature party.
-#[derive(Debug)]
+#[derive(Clone, Eq, PartialEq, Debug)]
```
pmikolajczyk41, line 97: [maybe stupid] I remember that I intentionally didn't make it `Clone` - such object was supposed to be created from `initiate_aggregation_with_hash`/`initiate_aggregation_with_call` and then, passed and destroyed with consecutive calls to `approve` (see the method signature - it takes `sig_agg: SignatureAggregation`); the intention was to never reuse `SignatureAggregation`, because it would lead to an unnecessary extrinsic submission

however, I won't argue that this care is not an overkill :)

aleph-client/src/staking.rs
```diff
@@ -301,7 +301,7 @@ pub fn bonded<C: AnyConnection>(connection: &C, stash: &KeyPair) -> Option<Accou
 /// `T: pallet_staking::Config` (somehow breaking consistency with similar structures in other
 /// pallets) we have no easy way of retrieving ledgers from storage. Thus, we chose cloning
 /// (relevant part of) this struct instead of implementing `Config` trait.
-#[derive(PartialEq, Eq, Clone, Debug, Encode, Decode)]
+#[derive(Clone, PartialEq, Eq, Debug, Encode, Decode)]
```
pmikolajczyk41, line 304: `Eq, PartialEq` :stuck_out_tongue:

pmikolajczyk41 (approved):
generally lgtm, thanks for doing that

aleph-client/src/multisig.rs
timorl: I would say that if this is semantically not supposed to be copied, then we definitely should not derive `Clone` for it.

timorl (approved):
LGTM, other than the already present comments.

aleph-client/src/multisig.rs
obrok: `KeyPair` is not `Eq`, so not just like that.

PR #598, opened 2022-08-22
A0-834: aleph-client: Always require std

artemis-cc:
# Description

In aleph-client, change the `std` feature requirements to be unconditional, and
remove the `std` feature (and the `[features]` section). This is accomplished
by opting back into default features, because a `std` feature is the only
default feature in all relevant crates.

Note that aleph-client already always required std (because it was not
`#[no_std]`).

## Type of change

# Checklist:


mike1729 (approved):

kostekIV (approved):
I cant comment on the unchanged code:
Bump version of the aleph_client

fbielejec:
> I cant comment on the unchanged code: Bump version of the aleph_client

done

PR #599, opened 2022-08-23
Change the way images without docker-entrypoint.sh are built

mikogs:
# Description
This change makes the workflows push docker images to the Docker Hub registry only when a version is pre-released or released. Images will no longer be added with every push to the `main` branch. 

## Type of change
Contains amends to GitHub workflow files and removes the `Dockerfile.onfinality` file which was originally used to build the images with an ENTRYPOINT using `/usr/local/bin/aleph-node` instead of `docker-entrypoint.sh`. Images will be now created just by adding an extra `ENTRYPOINT` entry on top of existing ECR image.


Kosiare (approved):
lgtm

PR #600, opened 2022-08-24
Update substrate to v0.9.28

Yithis:
# Description

## Type of change

# Checklist:

- I have added tests


e2e-tests/Cargo.toml
```diff
@@ -14,12 +14,15 @@ serde_json = "1.0"
 codec = { package = 'parity-scale-codec', version = "3.0", default-features = false, features = ['derive'] }
 rayon = "1.5"
 
-sp-core = { git = "https://github.com/Cardinal-Cryptography/substrate.git", branch = "aleph-v0.9.26", default-features = false, features = ["full_crypto"] }
-sp-runtime = { git = "https://github.com/Cardinal-Cryptography/substrate.git", branch = "aleph-v0.9.26", default-features = false }
-frame-support = { git = "https://github.com/Cardinal-Cryptography/substrate.git", branch = "aleph-v0.9.26", default-features = false }
-system = { git = "https://github.com/Cardinal-Cryptography/substrate.git", branch = "aleph-v0.9.26", package = "frame-system" }
-pallet-staking = { git = "https://github.com/Cardinal-Cryptography/substrate.git", branch = "aleph-v0.9.26", default-features = false }
-pallet-balances = { git = "https://github.com/Cardinal-Cryptography/substrate.git", branch = "aleph-v0.9.26", default-features = false }
+substrate-api-client = { git = "https://github.com/Cardinal-Cryptography/substrate-api-client.git", branch = "aleph-v0.9.28", features = ["staking-xt"] }
```
fixxxedpoint, line 17: Can you remove direct dependency on `substrate-api-client` in e2e-tests?

Yithis: Done

mike1729 (approved):
I approve under an assumption that this PR is not going to be merged :P

bin/node/src/main.rs
```diff
@@ -15,11 +15,11 @@ fn main() -> sc_cli::Result<()> {
         .run
         .import_params
         .pruning_params
-        .pruning
-        .replace(String::from("archive"))
-        .map_or(false, |x| x != "archive")
+        .blocks_pruning
```
DamianStraszak, line 18: Is there both `blocks_pruning` and `state_pruning` in the config now? 

If so, then this change is not quite equivalent to what was there before  -- I think you just want to disable both types of pruning.

Btw. I think we should have here a little "switch" in the form of a cli argument like `--experimental-pruning` which does not enforce pruning. This way we can test pruning without recompiling over and over again.

bin/runtime/src/lib.rs
```diff
@@ -107,7 +107,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("aleph-node"),
     impl_name: create_runtime_str!("aleph-node"),
     authoring_version: 1,
-    spec_version: 34,
+    spec_version: 35,
     impl_version: 1,
     apis: RUNTIME_API_VERSIONS,
     transaction_version: 9,
```
DamianStraszak, line 113: bump this as well.

bin/node/src/main.rs
DamianStraszak: The "switch" does not belong to this PR though (unless it's trivial).

bin/runtime/src/lib.rs

bin/node/src/main.rs
Yithis: Done as discussed yesterday

DamianStraszak (approved):

PR #605, opened 2022-08-30
nix - fix for package vendoring

fixxxedpoint:
This PR provides a subset of changes included in https://github.com/Cardinal-Cryptography/aleph-node/pull/525 . It fixes a mechanism by which nix attempts at downloading cargo dependencies and sometimes fails cuz it wrongly parses some of the Cargo.toml (it doesn't recognize configuration where you use same github dependency twice in your project, but one with .git suffix in its path, another without - its custom cargo config fails due to conflicting source mapping). Using it, one can provide an already propagated cargo-home directory (or it can be propagated by the nix-build.sh script).


Marcin-Radecki (changes requested):
Looks good, just one issue, when I run `docker run -ti --volume=$(pwd):/node/build aleph-build                     `, I get

```
checking for references to /tmp/nix-build-aleph-node-unknown.drv-0/ in /nix/store/6gddxl2ahivzbd3w7igmjg8r1zqx4mxz-aleph-node-unknown...
error: cannot create symlink '/node/build/result'; already exists
```

Marcin-Radecki (approved):
Very soilid work!

PR #607, opened 2022-08-31
A0-1152: Add possibility to set node count and min validator count via CLI

maciejzelaszczyk:
# Description

The `authorities_are_staking` test case stalls finalization due to changing the number of validators. This introduces two new params to our e2e test pipeline:

- `node-count` for setting the number of nodes to be run as consensus
- `min-validator-count` to set the minimum number of validators in chain spec at genesis.

By doing this, we are now able to run 6 nodes, chill 2 of them and do not hit the default `min-validator-count` limit. Also, if we need to specify different number of nodes and required validators for test cases, e.g. for `nightly`, we can now do this via GH pipelines.

## Type of change

# Checklist:


krzysztofziobro:
Is possibility to set min validator count in any way important in the context of finalization stall?

e2e-tests/src/test/electing_validators.rs
```diff
@@ -127,16 +124,16 @@ fn assert_enough_validators_left_after_chilling(
...
-/// generating new keys for new members (with `rotate_keys`), **FINALIZATION IS STALLED**. This is
-/// because a single node keeps in its keystore all Aleph keys, which is neither expected nor
-/// handled by our code. Fortunately, Aura handles this gently, so after changing committee block
-/// production keeps working. This is completetly enough for this test.
+///  - `pallet_staking` has `MinimumValidatorCount` (usually set to 4 in chain spec) and this cannot be
+/// changed on a running chain.
+///  - our e2e tests run with 5 validators by default.
+/// Thus, running on default settings and chilling 2 validators (1 reserved and 1 non reserved) is
+/// a no go: `pallet_staking` will protest and won't proceed with a new committee.
+/// To mitigate this, out e2e pipeline accepts a `node-count` parameter to specify the desired
```
krzysztofziobro, line 132:
```suggestion
/// To mitigate this, our e2e pipeline accepts a `node-count` parameter to specify the desired
```

maciejzelaszczyk:
> Is possibility to set min validator count in any way important in the context of finalization stall?

What `min-validator-count` does is it is passed to the chain spec:

https://github.com/Cardinal-Cryptography/aleph-node/blob/aa902b814814e31e0a33d5d2133bef2a6a77c3fc/bin/node/src/chain_spec.rs#L390

What this in turn does is persist a `StorageValue` available in `pallet_staking`:

https://github.com/paritytech/substrate/blob/324a18e3c5cbf333672c54f9367f530ea976928d/frame/staking/src/pallet/mod.rs#L229

This value is a threshold below which the chain enters an emergency state and finalization stalls. The important part is that it is not easily changed once the chain is up and running. In the `authorities_are_staking` test, we needed to chill (retire) `2` validators. Since our default chain starts with `5` nodes and the default `MinimumValidatorCount` is `4`, chilling `2` validators would hit the threshold and put the chain in emergency state.

By introducing `node-count`, we can set the initial number of nodes to, say, `6` and completely circumvent the finalization stall without changing the logic of the test. At the same time, it might be beneficial to be able to set different levels of `MinimumValidatorCount`, for instance for the purpose of randomizing test case parameters, and now we can do that.

krzysztofziobro (approved):

pmikolajczyk41 (approved):
I'm so glad that this has been solved in such a generic and robust way! kudos!

PR #608, opened 2022-08-31
A0-1323: Extend Runtime API to handle AlephBFT version

maciejzelaszczyk:
# Description

This extends the `AlephSessionApi` Runtime API, to handle the AlephBFT version. Storage items, getters, a setter and the declaration and implementation of appropriate methods are provided.

The solution is based on a `u32` version. The scheduled version change is persisted as a `StorageValue` -> `FinalityScheduledVersionChange`. This value stores the information about a scheduled AlephBFT version change, where `version_incoming` is the version to be set and `session` is the session on which the new version will be set. A `pallet_session::Session_Manager` is implemented to check whether a scheduled version change has moved into the past and, if so, to record it as the current version `StorageValue` -> `FinalityVersion`. When a version change is recorded as the current version, `FinalityScheduledVersionChange` is cleared.

Importantly, we can always reschedule a version change, and an already set version change may move into the past as time goes by. The value of the `current_session` is needed to check whether the version change needs to be recorded as the current version.

In order to cancel a scheduled version change, we need to use the version change setter (scheduler) and schedule a version change which will actually not change the current version.

## Type of change

# Checklist:

- I have added tests
- I have bumped the `spec_version` and `transaction_version`


bin/runtime/src/lib.rs
```diff
@@ -895,6 +896,10 @@ impl_runtime_apis! {
                 Aleph::queued_emergency_finalizer(),
             ))
         }
+
+        fn aleph_bft_version(session: SessionIndex) -> Result<AlephBFTVersion, AlephApiError> {
+           Aleph::aleph_bft_version(session).ok_or(AlephApiError::VersionNotSet)
```
kostekIV, line 901: I think what would be more suited is to get the latest set version. Lets say we set version to Legacy on the session nr `n-10` and did not change it since. Asking for the version for session `n` should return Version::Legacy. Also we need some default in case when session was never set - the legacy version seems like a perfect candidate

kostekIV (changes requested):

primitives/src/lib.rs
```diff
@@ -124,6 +125,13 @@ impl SessionAuthorityData {
     }
 }
 
+/// Current/legacy version identifier.
```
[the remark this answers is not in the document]
maciejzelaszczyk: Removed.

bin/runtime/src/lib.rs
maciejzelaszczyk: Changed logic to work this way.

pallets/aleph/src/lib.rs
```diff
@@ -121,6 +127,29 @@ pub mod pallet {
...
+
+            let session = version_change.session;
+            let version = version_change.version_incoming;
+
+            assert!(
+                session > previous_session,
+                "Tried to schedule an AlephBFT version change for a session ({}) prior to the session ({}) of the currently scheduled version change.",
+                session,
+                previous_session
+            );
```
kostekIV, line 144: this should be permitted imho

pallets/aleph/src/lib.rs
```diff
@@ -121,6 +127,29 @@ pub mod pallet {
...
+
+            assert_ne!(
+                previous_version, version,
+                "Tried to schedule an AlephBFT version change with the same version as the currently scheduled one: {:?}!",
+                previous_version
+            );
```
kostekIV, line 150: this should be permitted if previous_session is not in the past (think of it as adjustment of planned update)

kostekIV (approved):

pallets/aleph/src/lib.rs
maciejzelaszczyk: Correct, will change it to a `current_session` check.

pallets/aleph/src/lib.rs
maciejzelaszczyk: Agreed, will change this to check `current_session` as well.

maciejzelaszczyk:
Based on a fruitful conversation with @fixxxedpoint and @kostekIV, I am converting this back to draft, as we have converged on an alternative and potentially more robust scheme to handle version updates.

primitives/src/lib.rs
```diff
@@ -133,6 +141,7 @@ sp_api::decl_runtime_apis! {
         fn authority_data() -> SessionAuthorityData;
         fn session_period() -> u32;
         fn millisecs_per_block() -> u64;
+        fn aleph_bft_version() -> Result<Version, ApiError>;
```
timorl, line 144: You also need `next_session_aleph_bft_version`.

timorl (changes requested):
As far as I can tell brilliant!

Only one comment, but a pretty crucial one, thankfully should be quite simple to fix.

pallets/aleph/src/lib.rs
```diff
@@ -121,6 +141,43 @@ pub mod pallet {
...
+                );
+            }
+
+            // Can only schedule version changes with a version different than the current one.
+            // If a scheduled future version change is rescheduled to a different session,
+            // it should be possible to reschedule it with the same version as initially.
+            // If a scheduled version change has moved into the past, `SessionManager` records it
+            // as the current version.
+            if let Some(current_version) = <AlephBFTVersion<T>>::get() {
+                if current_version == version_to_schedule {
```
kostekIV, line 171: in this version we can relax this constraint. We could just allow rescheduling with the same version

timorl: Uhhh, right, this is tricky. If you want to undo a scheduling of a version change you have to schedule a "change" to the current version, right? So we definitely cannot be erroring out here, unless we add some other way of unscheduling a version change.

primitives/src/lib.rs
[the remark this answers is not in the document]
maciejzelaszczyk: In the current solution, it is possible that no version has ever been set and this is the case when the call can fail. This is only relevant at the very beginning when we start to use a version marker. Once the first version change is scheduled but has not yet come into effect, there is this possibility that no `AlephBFTVersion` has been manually set. After the first version change, this is not binding.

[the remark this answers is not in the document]
DamianStraszak: I would say it's much cleaner to output the default `0` here (the version we start from), instead of an error. Because otherwise what would the first version change look like? The version will be `Err` till some point and then jump to `0` (or rather `1`)? As mentioned somewhere else, you will need to deal with errors that the runtime API is not available at some blocks yet, having internal errors on top of that will only complicate things.

And in case you want to stick to this version then make the error name actually give some indication of what happened.

pallets/aleph/src/lib.rs
timorl: Wait, did you do _both_? That's somewhat silly, if you want an explicit unschedule then error out on attempts to schedule a change to the same version like before, since then it's an useful error to avoid confusion.

pallets/aleph/src/lib.rs
```diff
@@ -121,6 +142,57 @@ pub mod pallet {
...
+            <AlephBFTScheduledVersionChange<T>>::put(version_change);
+
+            Ok(())
+        }
+
+        pub(crate) fn remove_next_aleph_bft_version_change() -> Result<VersionChange, &'static str>
+        {
+            let version_change_to_remove = Self::aleph_bft_version_change()
+                .ok_or("No AlephBFT version change to unschedule!")?;
+            <AlephBFTScheduledVersionChange<T>>::kill();
```
timorl, line 179: You cannot do that when the session is too close, requires a check as in the other method.

But also, since you removed the identity check above maybe just remove this method to lower the bug surface? Sorry for suggesting opposite things in two different comments, I actually don't have an opinion on which you should do but please pick one. ^^"

timorl (changes requested):
Confusion!

primitives/src/lib.rs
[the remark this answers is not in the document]
maciejzelaszczyk: Since the error here pertains to a very slim possibility, let's go with the default-based version. I have introduced changes to reflect this.

pallets/aleph/src/lib.rs
maciejzelaszczyk: After consideration, I got rid of the remover.

pallets/aleph/src/lib.rs
maciejzelaszczyk: I was on the fence for this one, since the remover was more explicit but it resulted in a bit of bloat. The more implicit approach, however, results in a simpler implementation and I tend to agree with the remark about the bug surface. So, the remover is gone, and we can cancel a scheduled version change via a rescheduling with the same version as the one currently recorded.

Marcin-Radecki:
Is there a possibility to include some info from this PR description into pallet aleph's readme? In order to have the idea implemented in the PR reflected somewhere that we'll have dynamic versioning of dependencies in aleph-bft. After some thought, it's not intuitive to me we can have library A and can depend on the same library B in two different versions. Thanks to Cargo it is possible, as Rust encodes crate id in the mangled symbol (https://rust-lang.github.io/rfcs/2603-rust-symbol-name-mangling-v0.html).

maciejzelaszczyk:
> Is there a possibility to include some info from this PR description into pallet aleph's readme? In order to have the idea implemented in the PR reflected somewhere that we'll have dynamic versioning of dependencies in aleph-bft. After some thought, it's not intuitive to me we can have library A and can depend on the same library B in two different versions. Thanks to Cargo it is possible, as Rust encodes crate id in the mangled symbol (https://rust-lang.github.io/rfcs/2603-rust-symbol-name-mangling-v0.html).

Sure thing, I added an edited description.

timorl (approved):
LGTM.

pallets/aleph/src/lib.rs
```diff
@@ -121,6 +161,49 @@ pub mod pallet {
...
+        }
+
+        // If a scheduled future version change is rescheduled to a different session,
+        // it is possible to reschedule it with the same version as initially.
+        // To cancel a future version change, reschedule it with the current version.
+        // If a scheduled version change has moved into the past, `SessionManager` records it
+        // as the current version.
+        pub(crate) fn schedule_next_aleph_bft_version_change(
+            version_change: VersionChange,
+        ) -> Result<(), &'static str> {
```
DamianStraszak, line 176: Is there a specific reason we use strings here and not an enum?

pallets/aleph/src/lib.rs
```diff
@@ -121,6 +161,49 @@ pub mod pallet {
...
+
+            Ok(())
+        }
+
+        pub fn next_session_aleph_bft_version() -> Version {
+            let next_session = Self::current_session() + 1;
+            let scheduled_version_change = Self::aleph_bft_version_change();
+
+            if let Some(version_change) = scheduled_version_change {
+                if next_session == version_change.session {
```
DamianStraszak, line 200: An edge case comes to mind that on session boundary the session number changed and somehow the version change was not applied yet.

I'm 95% certain this is now correct, but just wanted to make sure you are also aware and certain :) 

Feel free to resolve after you familiarize yourself with the comment.

pallets/aleph/src/lib.rs
maciejzelaszczyk: This is due to `DispatchResult` being returned later on. If we do not want to include this information in the error variant, then it would be possible to get rid of the string, but the error itself would be less informative.

DamianStraszak: Ok. I see now. This is a helper functional that is called internally only. 

I think it has a slightly confusing name actually :) There is naming convention in substrate pallets (and much broader) for such functions: if the call is `something` then the internal function is called `do_something`. In this case this should be called `do_schedule_aleph_bft_version_change`. 

Feel free to incorporate if it makes sense and resolve all my comments.

DamianStraszak (commented):
My last doubt is regarding naming in this pallet. We want to use versioning not only for alephBFT but also for signature aggregation, and who knows what else in the future. I think its safe to use a more generic name here like `schedule_finality_version_change` or `schedule_consensus_version_change` instead of `schedule_aleph_bft_version_change`.

We ideally want to avoid changing the names ever, because it is in the runtime.

What do you think @timorl and @kostekIV ?

kostekIV:
> My last doubt is regarding naming in this pallet. We want to use versioning not only for alephBFT but also for signature aggregation, and who knows what else in the future. I think its safe to use a more generic name here like `schedule_finality_version_change` or `schedule_consensus_version_change` instead of `schedule_aleph_bft_version_change`.
> 
> We ideally want to avoid changing the names ever, because it is in the runtime.
> 
> What do you think @timorl and @kostekIV ?

Sounds good, either name is fine to me.

timorl:
The names you propose sound good, although it's a bit strange that these two things (abft & justifications) get conflated – to be clear I don't have a better solution and this one seems alright, just the versioning will correspond to "official" versions in a strange way.

At the same time I'm worried about the arguments you use – changing names of functions in the runtime (or other changes there for that matter) shouldn't be _that_ disastrous and we should probably spend some time figuring out how to do this better. I think I already mentioned this a long time ago, but it got forgotten... I'll add a task.

pallets/aleph/src/lib.rs
maciejzelaszczyk: Sure, `do` sounds much better then `next` in this context.

DamianStraszak:
Changing the names of runtime calls is fine (if these are sudo calls, because only we need to adapt), but changing the names in the runtime API is imho VERY non-trivial, because you need to add some compatibility layer in the host, otherwise a runtime update with new API would crash the node.

We should probably add this compatibility layer when writing the host part of compatibility.

That being said -- let's avoid changing names anyway, and use `consensus` or `finality` here (I would vote for the latter, as this is the nomenclature used by substrate and nicely reflects what apehbft+aggregator do).

maciejzelaszczyk:
> Changing the names of runtime calls (=extrinsics) is fine (if these are sudo calls, because only we need to adapt), but changing the names in the runtime API is imho VERY non-trivial, because you need to add some compatibility layer in the host, otherwise a runtime update with new API would crash the node.
> 
> We should probably add this compatibility layer when writing the host part of update logic.
> 
> That being said -- let's avoid changing names anyway, and use `consensus` or `finality` here (I would vote for the latter, as this is the nomenclature used by substrate and nicely reflects what apehbft+aggregator do).

`finality` it is :)

DamianStraszak (approved):
Awesome work!

DamianStraszak (changes requested):
Ehh. I approved but now I see that `spec_version` has not been bumped. Why is that?

maciejzelaszczyk:
> Ehh. I approved but now I see that `spec_version` has not been bumped. Why is that?

First time working on anything in the runtime. Have corrected this just now.

DamianStraszak:
You must also bump `transaction_version` -- these things are in the PR checklist exactly because we don't want to miss them :/

maciejzelaszczyk:
> You must also bump `transaction_version` -- these things are in the PR checklist exactly because we don't want to miss them :/

This should be OK now.

DamianStraszak (approved):

maciejzelaszczyk:
@DamianStraszak I have performed some final tests and there is one case that bothers me. Not a bug per se, but worth considering nonetheless.  Whenever you use `next_session_finality_version`, you have to make sure that your call comes after the `update_version_change_history` call from the `SessionManager`. Otherwise, the returned version for the next session will not be correct for the session when the change takes hold. I can change this behavior to always return the correct version number at the cost of introducing a storage item. Any thoughts?

PR #609, opened 2022-08-31
Message Compatibility vol 1

fixxxedpoint:
# Description

Hacky solution for message "duality" of alephbft sessions, i.e. we want to be able to send different messages in two following sessions.

## Type of change

Refactored component.rs + added a wrapper for the ComponentNetwork<D> that allows mappings between Vec<u8> on other types.


finality-aleph/src/network/component.rs
```diff
@@ -10,12 +11,73 @@ pub trait Sender<D: Data>: Sync + Send + Clone {
     fn send(&self, data: D, recipient: Recipient) -> Result<(), SendError>;
 }
 
+#[derive(Clone)]
+pub struct MapSender<D, S> {
+    sender: S,
+    _phantom: PhantomData<D>,
+}
+
+pub trait SenderMap<From: Data>: Sender<From> {
```
kostekIV, line 20: nit: rename From to something else as From is already a trait in rust and is a bit misleading

finality-aleph/src/network/component.rs
```diff
@@ -10,12 +11,73 @@ pub trait Sender<D: Data>: Sync + Send + Clone {
...
 pub trait Receiver<D: Data>: Sync + Send {
     async fn next(&mut self) -> Option<D>;
 }
 
+pub struct MapReceiver<D, R> {
+    receiver: R,
+    _phantom: PhantomData<D>,
+}
+
+pub trait ReceiverMap<From: Data>: Receiver<From> + Sized {
```
kostekIV, line 48: same nit here

finality-aleph/src/network/component.rs
```diff
@@ -10,12 +11,73 @@ pub trait Sender<D: Data>: Sync + Send + Clone {
...
+        loop {
+            let data = self.receiver.next().await;
+            let data = match data {
+                Some(data) => data,
+                None => return None,
+            };
+            match TryFrom::try_from(data) {
+                Ok(message) => return Some(message),
+                Err(e) => {
+                    warn!(target: "aleph-network", "Error decoding message in MapReceiver: {}", e)
```
kostekIV, line 74: should this be an error?

kostekIV (commented):
Is it possible to add tests here to see if we can indeed send two different data types and receive them?

finality-aleph/src/network/component.rs
fixxxedpoint: It is based on https://github.com/Cardinal-Cryptography/aleph-node/blob/1abe0ac671f4be8c5f0176f1b2dfcd0b0c73fe03/finality-aleph/src/network/service.rs#L221 . I am not sure. Some way of reasoning: error should for some serious problems, like some component can't operate - a user seeing it should be scared :). warn is more like - meh, failed, but lets move on.

kostekIV: sounds fair

finality-aleph/src/lib.rs
```diff
@@ -89,7 +89,82 @@ pub struct MillisecsPerBlock(pub u64);
...
+/// representations for provided type parameters with their version (they need to implement the `Versioned` trait).
+/// If one provides data types with equal versions, the first data parameter will have priority while decoding.
+/// Keep in mind that with this approach, `decode` might fail even if the second data type could parse (fast fail).
+#[derive(Clone)]
+pub enum VersionedEitherMessage<L, R> {
+    Left(L),
+    Right(R),
+}
+
+impl<L: Versioned + Decode, R: Versioned + Decode> Decode for VersionedEitherMessage<L, R> {
```
kostekIV, line 115: requiring Versioned here means we can only use this with types that we declared in finality-aleph (since we won't be able to implement this trait for foreign structs) or use some hacks like wrappers and so on. Could we just abandon this trait and use constants to denote versions for right/left variants of `VersionedEitherMessage`?

fixxxedpoint: You can implement local traits for foreign types without any additional wrappers. So should be able to implement it in finality-aleph for any Data type provided by aleph-bft. Or do you mean something different?

kostekIV (approved):

Yithis (approved):

PR #610, opened 2022-08-31
no direct dependency on substrate-api-client in e2e-tests

fixxxedpoint:
# Description

Created a collection of traits allowing to abstract away implementation details of how aleph-node is accessed by aleph-client. Using it we were able to remove direct dependency on substrate-api-client in e2e-tests. Moreover, this change should allow us to more easily switch to a different implementation of the aleph-client API (e.g. different libraries for some/all functions).

## Type of change

- Refactored API of aleph-client - collection of smaller, related traits. instead of a single api object
- removed direct dependency on substrate-api-client in e2e-tests by using this new api


aleph-client/src/system.rs
```diff
@@ -19,3 +19,19 @@ pub fn set_code(connection: &RootConnection, runtime: Vec<u8>, status: XtStatus)
...
+impl CallSystem for RootConnection {
+    type Error = substrate_api_client::error::Error;
+
+    fn fill_block(&self, target_ratio: u32, status: XtStatus) -> Result<(), Self::Error> {
+        let connection = self.as_connection();
+        let call = compose_call!(
+            connection.metadata,
+            "System",
+            "fill_block",
+            target_ratio * 10_000_000
```
obrok, line 32: Request: name this `10_000_000` constant - perhaps it's obvious to someone who's been in the project for some time, but I have no clue why it has the value it does.

obrok (changes requested):

aleph-client/src/lib.rs
```diff
@@ -105,18 +106,16 @@ pub trait AnyConnection: Clone + Send {
...
-pub trait AnyConnectionExt: AnyConnection {
+pub trait ReadStorage: AnyConnection {
     /// Reads value from storage. Panics if it couldn't be read.
     fn read_storage_value<T: Decode>(&self, pallet: &'static str, key: &'static str) -> T {
         self.read_storage_value_or_else(pallet, key, || {
             panic!("Value is `None` or couldn't have been decoded")
         })
     }
 
-    /// Reads value from storage at given block (empty means `best known`). Panics if it couldn't be read.
```
Marcin-Radecki, line 118: I found removed comment part `(empty means `best known`).` helpful.

e2e-tests/src/test/fee.rs
```diff
@@ -1,12 +1,8 @@
...
-    balances_transfer, get_next_fee_multiplier, get_tx_fee_info, send_xt, AccountId, AnyConnection,
-    AnyConnectionExt, Extrinsic, FeeInfo, RootConnection, SignedConnection, TransferTransaction,
-    XtStatus,
+    balances_transfer, get_next_fee_multiplier, AccountId, BalanceTransfer, CallSystem, FeeInfo,
+    GetTxInfo, ReadStorage, RootConnection, XtStatus,
 };
-use codec::Encode;
-use sp_core::Pair;
 use sp_runtime::{traits::One, FixedPointNumber, FixedU128};
-use substrate_api_client::{compose_call, compose_extrinsic, ExtrinsicParams, GenericAddress};
```
Marcin-Radecki, line 9: An alternative would be that we'll extract methods that use any of those (so we can still get rid of `substrate-api-client` dep on `e2e-client`), as we do now, and don't introduce additional traits interfaces.

e2e-tests/src/test/fee.rs
```diff
@@ -113,26 +109,17 @@ fn assert_no_scaling(
     );
 }
 
-fn prepare_transaction(connection: &SignedConnection) -> TransferTransaction {
+fn prepare_transaction<C: BalanceTransfer>(connection: &C) -> C::TransferTx {
     let bytes = [0u8; 32];
-    compose_extrinsic!(
```
Marcin-Radecki, line 118: Instead `compose_extrinsic` we can use https://github.com/Cardinal-Cryptography/aleph-node/blob/main/aleph-client/src/transfer.rs#L12

e2e-tests/src/test/fee.rs
```diff
@@ -113,26 +109,17 @@ fn assert_no_scaling(
...
-        Compact(0u128)
-    )
+    let account = AccountId::from(bytes);
+
+    connection.create_transfer_tx(account, 0u128.into())
 }
 
 fn fill_blocks(target_ratio: u32, blocks: u32, connection: &RootConnection) {
     for _ in 0..blocks {
-        let call = compose_call!(
```
Marcin-Radecki, line 129: This method we can extract to `aleph-client`  to `system` module https://github.com/Cardinal-Cryptography/aleph-node/blob/main/aleph-client/src/system.rs

e2e-tests/src/test/utility.rs
```diff
@@ -11,25 +8,15 @@ pub fn batch_transactions(config: &Config) -> anyhow::Result<()> {
 
     let (connection, to) = setup_for_transfer(config);
 
-    let call = compose_call!(
```
Marcin-Radecki, line 14: We can use `transfer` https://github.com/Cardinal-Cryptography/aleph-node/blob/main/aleph-client/src/transfer.rs#L12 here

e2e-tests/src/test/utility.rs
```diff
@@ -11,25 +8,15 @@ pub fn batch_transactions(config: &Config) -> anyhow::Result<()> {
...
-        GenericAddress::Id(to),
-        Compact(1000u128)
-    );
+    let call = connection.create_transfer_tx(to, 1000u128);
     let mut transactions = Vec::new();
     for _i in 0..NUMBER_OF_TRANSACTIONS {
         transactions.push(call.clone());
     }
 
-    let extrinsic =
```
Marcin-Radecki, line 26: This can we extract to new `utility` module in https://github.com/Cardinal-Cryptography/aleph-node/tree/main/aleph-client/src

Marcin-Radecki (changes requested):
Agree with the motivation to get rid of `substrate-api-client` dependency in `e2e-client` , however can this be done without introducing new traits? As current approach in `aleph-client` is just to wrap all `substrate-api-client` internals like `compose_call`, `compose_extrinsic` etc. , so we can just extract/create some more methods in `aleph-client` and use them in`e2e-client`?

e2e-tests/src/test/fee.rs
fixxxedpoint: I don't see how. Transfer also sends that transaction, but this code suppose to only create it.

fixxxedpoint:
> Agree with the motivation to get rid of `substrate-api-client` dependency in `e2e-client` , however can this be done without introducing new traits? As current approach in `aleph-client` is just to wrap all `substrate-api-client` internals like `compose_call`, `compose_extrinsic` etc. , so we can just extract/create some more methods in `aleph-client` and use them in`e2e-client`?

Sure it can be implemented without them. Main reason of introducing them was to encourage (make it easier to happen) to switch to some other library - currently aleph-client is just a huge collection of utility methods on substrate_api_client. Because of that, api of substrate_api_client spreads in other projects like a virus :). Next step can be to remove substrate-api-client from public interface of aleph-client. Preferably all external-access methods to aleph-node should be implemented by means of aleph-client (is there any other library for that?). This way we can also provide end-users a more tailored (and easier to use) rust-api, instead of giving them a generic substrate api library in form of substrate_api_client.
Other reasons for introducing mutliple traits (they can be also merged into one or I can provide a convenience trait extending all of them):
- We already use a trait for connections, but currently it is strictly tied to substrate_api_client [sac], which we are planning to at least partially switch for something else. Currently, whole api of aleph-client is forced to be just a wrapper for substrate_api_client - only interface used by it is that Connection provided by sac. Introducing any other implementation in aleph-client will require to manually switch it in all dependent code.
- They allow us to gradually switch to other implementation - you can just implement a subset of required methods using other library instead of being forced to implement all - interface segregation.

aleph-client/src/system.rs
fixxxedpoint: Actually, I have no clue as well. I will ask the author and introduce some descriptive name.

aleph-client/src/lib.rs
```diff
@@ -105,18 +106,16 @@ pub trait AnyConnection: Clone + Send {
...
-    fn read_storage_value_at_block<T: Decode>(
+    /// Reads value from storage. Panics if it couldn't be read.
+    fn read_storage_value_from_block<T: Decode>(
```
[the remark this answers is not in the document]
fixxxedpoint: changed

aleph-client/src/lib.rs
```diff
@@ -217,6 +216,48 @@ pub trait AnyConnectionExt: AnyConnection {
     }
 }
 
+pub trait BalanceTransfer {
```
Marcin-Radecki, line 219: I think we need to come up (I'm okay with not doing it in this PR, yet still I think it's an important thing to consider) with a holistic approach to such a new interface that abstracts `aleph-client` from any library that provides data to it. They are loosely coupled to me now, maybe we should make them resemble pallets and their calls? E.g.

```
trait PalletBalances {
    fn transfer(&self, tx: Self::TransferTx, status: XtStatus)
        -> Result<Option<H256>, Self::Error> { {// some default imp }
     fn force_transfer(&self, tx: Self::TransferTx, status: XtStatus)
        -> Result<Option<H256>, Self::Error> { {// some default imp  }
   ....
}

trait PalletSystem {
   fn fill_block(&self, target_ratio: u32, status: XtStatus) -> Result<(), Self::Error> {// some default imp }
};
```
 
We need to provide some defaults impl to actually make interface segregation possible. Also we can then extend those traits as we go with more pallet calls. Plus of that design is that pallets and their names in substrate should not change very often, as they are API themselves.

Marcin-Radecki:
> > Agree with the motivation to get rid of `substrate-api-client` dependency in `e2e-client` , however can this be done without introducing new traits? As current approach in `aleph-client` is just to wrap all `substrate-api-client` internals like `compose_call`, `compose_extrinsic` etc. , so we can just extract/create some more methods in `aleph-client` and use them in`e2e-client`?
> 
> Sure it can be implemented without them. Main reason of introducing them was to encourage (make it easier to happen) to switch to some other library - currently aleph-client is just a huge collection of utility methods on substrate_api_client. Because of that, api of substrate_api_client spreads in other projects like a virus :). Next step can be to remove substrate-api-client from public interface of aleph-client. Preferably all external-access methods to aleph-node should be implemented by means of aleph-client (is there any other library for that?). This way we can also provide end-users a more tailored (and easier to use) rust-api, instead of giving them a generic substrate api library in form of substrate_api_client. Other reasons for introducing mutliple traits (they can be also merged into one or I can provide a convenience trait extending all of them):
> 
>     * We already use a trait for connections, but currently it is strictly tied to substrate_api_client [sac], which we are planning to at least partially switch for something else. Currently, whole api of aleph-client is forced to be just a wrapper for substrate_api_client - only interface used by it is that Connection provided by sac. Introducing any other implementation in aleph-client will require to manually switch it in all dependent code.
> 
>     * They allow us to gradually switch to other implementation - you can just implement a subset of required methods using other library instead of being forced to implement all - interface segregation.

Thanks for explanation, see other comments

e2e-tests/src/test/fee.rs
Marcin-Radecki: I mean to refactor the code in this create so the creation of transaction object is not required here. We need such an object to send some transfers anyway, so in the end, we just need to be able to send some transfers from this created code.

Marcin-Radecki (commented):
Happy to approve once we agree that we don't need prior general design of the new interface. I'm making this point only because it was by design so that `aleph-client` was just a wrapper of `substrate-api-client`. I'm okay with changing that, yet we need to think whether to invest time of designing a new interface, or just migrate to some other tools like `subxt`.

aleph-client/src/lib.rs
```diff
@@ -464,3 +505,7 @@ pub fn get_current_block_number<C: AnyConnection>(connection: &C) -> BlockNumber
         .expect("Block exists; qed")
         .number
 }
+
+impl ReadStorage for Connection {}
```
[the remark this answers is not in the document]
Marcin-Radecki: Actually I was wrong, this is just to replace `AnyConnection`

aleph-client/src/lib.rs
Marcin-Radecki: Let's keep that in mind during `subxt` migration

Marcin-Radecki (approved):

aleph-client/src/system.rs
fixxxedpoint: Changed - now it should be more descriptive.

obrok: It still says `10_000_000` here, did you forget to push something?

fixxxedpoint: Check it now.

obrok: Can we use https://docs.rs/sp-runtime/6.0.0/sp_runtime/struct.Perbill.html#method.from_percent here? Or is this a different perbill?

fixxxedpoint: Changed it.

obrok (approved):

PR #612, opened 2022-09-08
Simple dex

fbielejec:
# Description

This contract is based on the Balancer design - automated market maker with an underlying multi-asset LP. 
A must-read before reviewing is at least glancing at the Balancer whitepaper: https://balancer.fi/whitepaper.pdf

There is one pool with 4 PSP22 game reward tokens + wrapped native AZERO token, with equal weights all. **Swaps** can be done between all pairs in the pool (see Out-Given-In).

Liquidity providing works for arbitrary deposits / withdrawals, without tracking per account pool shares, as the Foundation will be the sole liquidity provider for the pool.


contracts/access_control/lib.rs
```diff
@@ -34,6 +34,8 @@ mod access_control {
         Owner(AccountId),
         /// Indicates account can initialize a contract from a given code hash.
         Initializer(Hash),
+        /// Indicates account can add liquidity to a DEX contract (call certain functions)
+        LiquidityProvider(AccountId),
```
deuszx, line 38: Won't this force all users of this `Role` enum to handle this new variant as well? Even though they don't need it? Is there any versioning for this? If so, we should probably bump the version.

fbielejec: it's not versioned, but how'd you version an enum? A separate crate just for enum's sake?

pmikolajczyk41: I'm not sure whether such variant adding might break some compatibility, but maybe we will be more safe if we drop enum and use some plain string or byte slice as role identifier

deuszx: > How it's used so far does not force handling all variants - it's just a check for one of them, (for [example](https://github.com/Cardinal-Cryptography/aleph-node/blob/simple-dex/contracts/simple_dex/lib.rs?rgh-link-date=2022-09-13T08%3A00%3A45Z#L410))

That's how _you_ are using it _right now_. Maybe others would pattern match on the caller's `Role` and would have to handle all of the cases. Then, if they initially wrote a contract when `Role` had fewer entries, after you add a new one their code won't be compiling anymore.

fbielejec: if we drop enum now they also won't be compiling anymore :rofl:

deuszx: Not if you version it and they use its semver properly :P

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,660 @@
...
+    pub struct SimpleDex {
+        /// total supply of pool shares
+        pub total_liquidity: u128,
+        /// tracks pool shares per account
+        pub liquidity: Mapping<AccountId, u128>,
+        pub swap_fee: u128,
+        pub access_control: AccountId,
+        pub ubik: AccountId,
+        pub cyberiad: AccountId,
+        pub lono: AccountId,
```
pmikolajczyk41, line 125: these are cool themselves, but highly enigmatic as field names :P

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,660 @@
...
+            let caller = self.env().caller();
+            let this = self.env().account_id();
+
+            <Self as AccessControlled>::check_role(
+                self.access_control,
+                caller,
+                Role::Admin(this),
+                Self::cross_contract_call_error_handler,
+                Self::access_control_error_handler,
+            )?;
```
pmikolajczyk41, line 465: I would really love to work towards macro-ing / hiding access control; the solution I would appreciate would be like

```rust
#[ink(message)]
#[access_control(require = Admin)]
pub fn set_swap_fee(&mut self, swap_fee: Balance) -> Result<(), DexError> {
  // logic that works under assumption that the caller has all required roles
```

such macro in standard Rust env would be really easy to implement (I can volutneer :P) but I'm not so sure how they work with ink!

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,660 @@
...
+        }
+
+        /// Transfers a given amount of a PSP22 token to a specified using the callers own balance
+        fn transfer_tx(
+            &self,
+            token: AccountId,
+            to: AccountId,
+            amount: Balance,
+        ) -> Result<Result<(), PSP22Error>, InkEnvError> {
+            build_call::<DefaultEnvironment>()
```
pmikolajczyk41, line 553: I'm afraid, that you need at least think about reentrancy here - see my short report on Slack

in brief - contract `A` can call this dex and in return it should be given some PSP22 token, but PSP22 mechanics will need to ask `A` whether it is ok with receiving money

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,660 @@
...
+                        .push_arg(vec![0x0]),
+                )
+                .returns::<Result<(), PSP22Error>>()
+                .fire()
+        }
+
+        /// Transfers a given amount of a PSP22 token on behalf of a specified account to another account
+        ///
+        /// Will revert if not enough allowance was given to the caller prior to executing this tx
+        fn transfer_from_tx(
```
pmikolajczyk41, line 568: okay, now I see your point from the other PR that we'd better extract these common cross-contract calls (mostly to PSP22)

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,660 @@
...
+                )
+                .returns::<Balance>()
+                .fire()
+        }
+
+        fn out_given_in(
+            amount_token_in: u128,
+            balance_token_in: u128,
+            balance_token_out: u128,
+            swap_fee: u128,
```
pmikolajczyk41, line 622: `Balance`? sorry :P

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,660 @@
...
+        ) -> Result<u128, DexError> {
+            let op1 = 1u128.checked_sub(swap_fee).ok_or(DexError::Arithmethic)?;
```
pmikolajczyk41, line 624: something is no yes with that...

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,660 @@
...
+    }
+
+    #[ink(storage)]
+    #[derive(SpreadAllocate)]
+    pub struct SimpleDex {
+        /// total supply of pool shares
+        pub total_liquidity: u128,
+        /// tracks pool shares per account
+        pub liquidity: Mapping<AccountId, u128>,
+        pub swap_fee: u128,
```
pmikolajczyk41, line 121: I'm just thinking... is there really a need for any fees here?

contracts/simple_dex/lib.rs
pmikolajczyk41: oh, I take it back - `A` should call dex with reentrancy switched on, sorry

contracts/simple_dex/lib.rs
pmikolajczyk41: and if you are eager to move to `Balance` from `u128`, then it would be super to use `Balance::one()` for constant :P

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,660 @@
...
+
+        /// Swaps the transferred amount of native token to one of the pools PSP22 tokens
+        #[ink(message, payable)]
+        pub fn native_to_token(&mut self, token_out: AccountId) -> Result<(), DexError> {
+            let amount_token_in = self.env().transferred_value();
+            // we use the balance of the native token just before the exchange tx
+            let balance_token_in = self
+                .env()
+                .balance()
+                .checked_sub(amount_token_in)
```
pmikolajczyk41, line 166: we are alredy given this `payable`, right? or is it somehow hangin

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,660 @@
...
+            );
+
+            Ok(())
+        }
+
+        /// Swap the specified amount of one of the pools PSP22 tokens to the native token
+        /// Calling account needs to give allowance to the DEX contract to spend amount_token_in of token_in on it's behalf
+        /// before executing this tx.
+        #[ink(message)]
+        pub fn token_to_native(
```
DamianStraszak, line 203: I can feel your pain, needing to have tons of "ifology" because of the `native` case :P

The standard way to avoid it, is to say: OK let's just have a pool of `4` PSP22 tokens, we don't even care which is which. And we go from native to `PSP22` outside of this contract by "wrapping" `AZERO` into `wAZERO` being a simple PSP22 contract, which allows minting `wAZERO` by depositing `AZERO` and getting back `AZERO` by depositing `wAZERO`. 

It adds some mess to the frontend then, and sometimes requires the user to make a manual wrap/unwrap. But I think it's still state of the art.  

Wrapped AZERO will be probably used all over different contracts anyway. So we could probably deploy the official version, I guess. 

Treat it as a discussion starter, not as a request for changes. The current version is 100% better from UX perspective. Just more complicated when it comes to writing the contract/s, making sure they are correct.

DamianStraszak (commented):
Mostly one, large, general comment, but I'm still reading it.

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,660 @@
...
+                    amount_in: amount_token_in,
+                    amount_out: amount_token_out,
+                }),
+            );
+
+            Ok(())
+        }
+
+        /// Swap the specified amount of one of the pools PSP22 tokens to the native token
+        /// Calling account needs to give allowance to the DEX contract to spend amount_token_in of token_in on it's behalf
```
deuszx, line 200:
```suggestion
        /// Calling account needs to give allowance to the DEX contract to spend `amount_token_in` of `token_in` on its behalf
```

contracts/simple_dex/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+num = { version = "0.4.0", default-features = false } 
+access_control = { path = "../access_control", default-features = false, features = ["ink-as-dependency"] }
+game_token = { path = "../game_token", default-features = false, features = ["ink-as-dependency"] }
+
+[lib]
+name = "simple_dex"
+path = "lib.rs"
+crate-type = [
+  # Used for normal contract Wasm blobs.
+  "cdylib",
```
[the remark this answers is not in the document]
fbielejec: this is only needed if you're linking contracts

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,660 @@
...
+        swap_fee: Balance,
+    }
+
+    #[ink(storage)]
+    #[derive(SpreadAllocate)]
+    pub struct SimpleDex {
+        /// total supply of pool shares
+        pub total_liquidity: u128,
+        /// tracks pool shares per account
+        pub liquidity: Mapping<AccountId, u128>,
```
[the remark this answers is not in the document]
fbielejec: I geenrally have a question:

contracts/simple_dex/lib.rs
fbielejec: not sure I follow this comment, could you elaborate pls?

contracts/simple_dex/lib.rs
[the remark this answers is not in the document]
deuszx: But I think normally pool would issue yet another token - liquidity token - rather than track it internally?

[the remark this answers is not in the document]
DamianStraszak: I think it simplifies a lot if the admin can arbitrarily put liquidity in and arbitrarily pull liquidity out, regardless of the balance of tokens (it should be able to specify: I add [a, b, c, d] of tokens (A, B, C, D), or I withdraw [a, b, c, d] of tokens (A, B, C, D)).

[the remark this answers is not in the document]
DamianStraszak: What I'm saying is only under the assumption that there is only one account that can do that. If we want to allow users to be LPs, then things change, but I thought that the main point is that we don't allow that and hence can simplify things.

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,659 @@
...
+        }
+
+        /// Swap the specified amount of one of the pools PSP22 tokens to the native token
+        /// Calling account needs to give allowance to the DEX contract to spend `amount_token_in` of `token_in` on its behalf
+        /// before executing this tx.
+        #[ink(message)]
+        pub fn token_to_native(
+            &mut self,
+            token_in: AccountId,
+            amount_token_in: u128,
```
DamianStraszak, line 206: Normally there should also be a third argument here `min_amount_token_out` that is to deal with slippage. Maybe it is not necessary, but I guess MASSIVE sandwich attacks are possible without that :P

contracts/simple_dex/lib.rs
pmikolajczyk41: this `message` (method) will be accompanied by some amount of native token (since it is payable); my question is whether these funds will be already available in this place

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,659 @@
...
+    }
+
+    #[ink(storage)]
+    #[derive(SpreadAllocate)]
+    pub struct SimpleDex {
+        /// total supply of pool shares
+        pub total_liquidity: u128,
+        /// tracks pool shares per account
+        pub liquidity: Mapping<AccountId, u128>,
+        pub swap_fee: u128,
```
deuszx, line 121: What is it being set to by default? I don't see it being set in the constructor of the contract and if no one calls `set_swap_fee` before the first deal is made then it will use the default?

deuszx:
I don't want to be The Party Pooper but I think we should extract the [wAZERO commit](https://github.com/Cardinal-Cryptography/aleph-node/pull/612/commits/1e31342f89e900f98ffc047925339d5c224eb8fd) to a separate PR and give it the review it deserves, unit tests etc.

Also, since it will almost certainly be present in every execution path (i.e. people will have to use wAZERO to interact with other dApps), should we consider making it a pallet instead? Would that improve the performance?

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,472 @@
...
+        }
+
+        /// Alters the swap_fee parameter
+        ///
+        /// Can only be called by the contract's Admin.
+        #[ink(message)]
+        pub fn set_swap_fee(&mut self, swap_fee: u128) -> Result<(), DexError> {
+            if swap_fee.gt(&100) || swap_fee.lt(&0) {
+                return Err(DexError::WrongParameterValue);
+            }
```
deuszx, line 267: (Porting from a commit-level comment):

I don't think it can be `swap_fee.lt(&0)` b/c it's an unsigned integer type - it will always be false.

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,472 @@
...
+
+        fn out_given_in(
+            amount_token_in: Balance,
+            balance_token_in: Balance,
+            balance_token_out: Balance,
+            swap_fee: u128,
+        ) -> Result<Balance, DexError> {
+            let op0 = amount_token_in
+                .checked_mul(swap_fee)
+                .and_then(|result| result.checked_div(100))
```
deuszx, line 438: Since swap_fee can only be within {0,100} range, this division will always yield 0.

deuszx: For a moment I thought that maybe my mistake is caused by the fact that I'm assuming the `Balance` type to beheave like `u128`, after all - it could be some composite type capable of representing fractions (possibly via two unsigned integers) but I've checked `ink` and it seems to be just `u128` by default:
https://github.com/paritytech/ink/blob/master/crates/env/src/types.rs#L186

Do we change it anywhere?

contracts/wrapped_azero/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+
+            Ok(())
+        }
+
+        /// Unwraps a specified amount
+        #[ink(message, payable)]
+        pub fn unwrap(&mut self, amount: Balance) -> Result<()> {
+            let caller = self.env().caller();
+
+            if !amount.eq(&Balance::zero()) {
```
pmikolajczyk41, line 173: I would simplify this to 
```rust
if amount.eq(&Balance::zero()) {
  Ok(())
}
```

contracts/wrapped_azero/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+                self._mint(caller, amount)?;
+                Self::emit_event(self.env(), Event::Wrapped(Wrapped { caller, amount }));
+            }
+
+            Ok(())
+        }
+
+        /// Unwraps a specified amount
+        #[ink(message, payable)]
```
pmikolajczyk41, line 169: why payable?

pmikolajczyk41 (commented):
generally it looks better and better with every commit, however, I would **strongly** insist on tests: both for calculation methods and for the contract as whole; ofc they can be implemented in further PRs, but imho they are necessary

also, I would like to pass my vote/opinion to @h4nsu for further changes

fbielejec:
> I don't want to be The Party Pooper but I think we should extract the [wAZERO commit](https://github.com/Cardinal-Cryptography/aleph-node/pull/612/commits/1e31342f89e900f98ffc047925339d5c224eb8fd) to a separate PR and give it the review it deserves, unit tests etc.
> 
> Also, since it will almost certainly be present in every execution path (i.e. people will have to use wAZERO to interact with other dApps), should we consider making it a pallet instead? Would that improve the performance?

IMHO overkill - this token differs in 2 functions (maybe 20 lines) from a vanilla PSP22, all that needs to be reviewed is `wrap` and `unwrap`.

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+                amount_token_in,
+                balance_token_in,
+                balance_token_out,
+                self.swap_fee_percentage,
+            )?;
+
+            if balance_token_out < amount_token_out {
+                // liquidity too low
+                return Err(DexError::NotEnoughLiquidityOf(token_out));
+            }
```
deuszx, line 163: This isn't necessary - we've already checked if `balance_token_out < min_amount_token_out` and the latter has to be less than `amount_token_out`: we check that as well just below.

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+            let caller = self.env().caller();
+
+            // check role, only designated account can add liquidity
+            <Self as AccessControlled>::check_role(
+                self.access_control,
+                caller,
+                Role::LiquidityProvider(this),
+                Self::cross_contract_call_error_handler,
+                Self::access_control_error_handler,
+            )?;
```
deuszx, line 208: All of the calls to `check_role` look almost exactly the same:
```rust
<Self as AccessControlled>::check_role(
                self.access_control,
                caller,
                <expected_role>,
                Self::cross_contract_call_error_handler,
                Self::access_control_error_handler,
            )?;
```
We could extract it to a method on `Self`:
```rust
fn check_role(expected: Role) {
    let caller = self.env().caller();
   <Self as AccessControlled>::check_role(
                self.access_control,
                caller,
                <expected_role>,
                Self::cross_contract_call_error_handler,
                Self::access_control_error_handler,
            )?;
}
```

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+                    ExecutionInput::new(Selector::new(BALANCE_OF_SELECTOR)).push_arg(account),
+                )
+                .returns::<Balance>()
+                .fire()
+        }
+
+        /// Swap trade output given a curve with equal token weights
+        ///
+        /// swap_fee_percentage (integer) is a percentage of the trade that goes towards the pool
+        /// B_0 - (100 * B_0 * B_i) / (100 * (B_i + A_i) -A_i * fee)
```
deuszx, line 438: Just FYI, apparently Balancer deals with 0.0001 fee precision: https://docs.balancer.fi/getting-started/faqs/fees#what-are-the-fees-for-a-trade

contracts/wrapped_azero/lib.rs
```diff
@@ -0,0 +1,252 @@
...
+    pub struct UnWrapped {
+        caller: AccountId,
+        amount: Balance,
+    }
+
+    impl WrappedAzero {
+        /// Creates a new token
+        ///
+        /// The token will have its name and symbol set in metadata to the specified values.
+        /// Decimals are fixed at 18.
```
deuszx, line 125:
```suggestion
        /// Decimals are fixed at 12.
```
as per https://github.com/Cardinal-Cryptography/aleph-node/pull/612/files#diff-40606916260c70af8efd0229d64825968cce2842ab0eee83fa92eeb42646bc24R147

h4nsu (changes requested):
We already know we won't be able to use AZERO in this DEX? Do you think we should remove it now in this PR or make it a subject of the next one?

h4nsu (approved):

contracts/simple_dex/lib.rs
```diff
@@ -82,8 +105,8 @@ mod simple_dex {
     pub struct SimpleDex {
         pub swap_fee_percentage: u128,
         pub access_control: AccountId,
-        // pool tokens
-        pub tokens: [AccountId; 4],
+        // a set of pairs that are availiable for swapping between
+        pub swap_pairs: Mapping<SwapPair, ()>,
```
deuszx, line 110: This is equivalent to `Mapping<AccountID, Vec<AccountID>>`, correct? 

I'm just wondering whether it wouldn't be cheaper to use the contract later on if instead of initializing a `SwapPair` we would do 
```rust
swap_pairs
    .get(&token_in)
    .map(|tokens_out| tokens_out.find(&token_out).is_some())
    .or_else(false)
```

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+            self.check_role(caller, Role::Admin(this))?;
+
+            // emit event
+            Self::emit_event(
+                self.env(),
+                Event::SwapFeeSet(SwapFeeSet {
+                    caller,
+                    swap_fee_percentage,
+                }),
+            );
```
deuszx, line 278: nit: I would think that event is emitted after the action has happened. In practice it wouldn't matter probably but it's a matter of personal style more than anything.

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+
+            self.swap_fee_percentage = swap_fee_percentage;
+            Ok(())
+        }
+
+        /// Returns current value of the swap_fee_percentage parameter
+        #[ink(message)]
+        pub fn swap_fee_percentage(&mut self) -> Balance {
```
deuszx, line 286: Does it have to be `&mut`?

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+                    ExecutionInput::new(Selector::new(TRANSFER_FROM_SELECTOR))
+                        .push_arg(from)
+                        .push_arg(to)
+                        .push_arg(amount)
+                        .push_arg(vec![0x0]),
+                )
+                .call_flags(CallFlags::default().set_allow_reentry(true)) // needed for checking allowance before the actual tx
+                .returns::<Result<(), PSP22Error>>()
+                .fire()
+        }
```
deuszx, line 414: nit: This seems like a very often-used util, maybe we should extract it to our own "lib".

contracts/simple_dex/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+        /// Swap trade output given a curve with equal token weights
+        ///
+        /// swap_fee_percentage (integer) is a percentage of the trade that goes towards the pool
+        /// B_0 - (100 * B_0 * B_i) / (100 * (B_i + A_i) -A_i * fee)
+        fn out_given_in(
+            amount_token_in: Balance,
+            balance_token_in: Balance,
+            balance_token_out: Balance,
+            swap_fee_percentage: u128,
+        ) -> Result<Balance, DexError> {
```
deuszx, line 454: It would be good to have unit tests for this function. It's standalone (i.e. not hanging off of the contract) so should be relatively easy.

contracts/wrapped_azero/lib.rs
```diff
@@ -0,0 +1,252 @@
...
+    use ink_env::Error as InkEnvError;
+    use ink_lang::{
+        codegen::{EmitEvent, Env},
+        reflect::ContractEventBase,
+    };
+    use ink_prelude::{format, string::String};
+    use ink_storage::traits::SpreadAllocate;
+    use num_traits::identities::Zero;
+    use openbrush::{
+        contracts::psp22::{extensions::metadata::*, Internal},
```
deuszx, line 20:
```suggestion
        contracts::psp22::{extensions::metadata::*, Internal as PSP22Internal},
```
for readability

deuszx (approved):
I would like to see some tests before we deploy it but code seems fine.

DamianStraszak (approved):

pmikolajczyk41 (approved):

PR #613, opened 2022-09-08
A0-1238: Marketplace for auctioning tickets

obrok:
# Description

Creates a new contract in `contrats/marketplace` that implements ticket auctioning.

## Type of change

# Checklist:


contracts/marketplace/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+    pub struct Marketplace {
+        total_proceeds: Balance,
+        tickets_sold: Balance,
+        min_price: Balance,
+        current_start_block: BlockNumber,
+        auction_length: BlockNumber,
+        sale_multiplier: Balance,
+        ticket_token: AccountId,
+        reward_token: AccountId,
+    }
```
deuszx, line 40: Since it's supposed to be an examplary contract, let's comment it generously.

contracts/marketplace/lib.rs
```diff
@@ -0,0 +1,299 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
deuszx, line 1: I'd leave a crate-level comment here explaining what the contract it is, what are the mechanics etc.

contracts/marketplace/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+    pub enum Error {
+        MissingRole(String),
+        ContractCall(String),
+        PSP22TokenCall(String),
+        MarketplaceEmpty,
+    }
+
+    #[ink(event)]
+    #[derive(Clone, Eq, PartialEq, Debug)]
+    pub struct Buy {
```
deuszx, line 53: In Event Sourcing world, events would be in past tense - _bought_.

contracts/marketplace/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+                |reason| reason.into(),
+                |role| Error::missing_role(role),
+            )
+        }
+
+        fn initializer() -> Role {
+            let code_hash = Self::env()
+                .own_code_hash()
+                .expect("Failure to retrieve code hash.");
+            Initializer(code_hash)
```
deuszx, line 284: Here we're calling the `Role` enum's variant directly, I'd prefer to be consistent and call it like the rest (like `admin` below)  - `Role::Initializer`.

contracts/marketplace/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+        }
+
+        fn admin(&self) -> Role {
+            Role::Admin(self.this())
+        }
+
+        fn this(&self) -> AccountId {
```
deuszx, line 291: This what :)

contracts/marketplace/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+                tickets_sold: 1,
+            }
+        }
+
+        #[ink(message)]
+        /// The length of each auction of a single ticket in blocks.
+        ///
+        /// The contract will decrease the price linearly from `average_price() * sale_multiplier()`
+        /// to `min_price()` over this period. The auction doesn't end after the period elapses -
+        /// the ticket remains available for purchase at `min_price()`.
```
deuszx, line 115: I'm not sure if it's how `ink!` does it but in Rust, comments come before the macros.

contracts/marketplace/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+        /// The average price over all sales the contract made.
+        pub fn average_price(&self) -> Balance {
+            self.total_proceeds.saturating_div(self.tickets_sold)
+        }
+
+        #[ink(message)]
+        /// The multiplier applied to the average price after each sale.
+        ///
+        /// The contract tracks the average price of all sold tickets and starts off each new
+        /// auction at a price that is this multiplier times the average.
```
deuszx, line 142:
```suggestion
        /// auction at a price that is `self.average_price() * self.sale_multiplier()`
```

Personally, that's cleaner to me.

contracts/marketplace/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+        #[ink(message)]
+        /// The block at which the auction of the current ticket started.
+        pub fn current_start_block(&self) -> BlockNumber {
+            self.current_start_block
+        }
+
+        #[ink(message)]
+        /// The price the contract would charge when buying at the current block.
+        pub fn price(&self) -> Balance {
+            self.current_price()
```
deuszx, line 129: Since `current_price()` can be quite "expensive" and this `price()` method is available in the public API, I would try to cache the `current_price()` computation and reuse it in here.

contracts/marketplace/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+
+        fn take_payment(&self, from: AccountId, amount: Balance) -> Result<(), Error> {
+            build_call::<Environment>()
+                .call_type(Call::new().callee(self.reward_token))
+                .exec_input(
+                    ExecutionInput::new(Selector::new(TRANSFER_FROM_GAME_TOKEN_SELECTOR))
+                        .push_arg(from)
+                        .push_arg(self.this())
+                        .push_arg(amount)
+                        .push_arg(DUMMY_DATA),
```
deuszx, line 233: Can you explain why we need this `DUMMY_DATA` arg?

deuszx (commented):
I don't see any blockers. Most (if not all) my comments are more about adding documentation and code clarity. 

That being said, I think there's a benefit to caching the `current_price()` computations, as it's quite "expensive" (the parts that could be cached are `per_block_reduction`, `average_price`) and would probably be called often by other dApps/accounts to discover the price they would have to pay (although, as per my experience, if something can be computed off-chain it will be eventually b/c it's cheaper). We could also note when the price of the current ticket auction is lower than `self.min_price` and cache that the price of the ticket is `self.min_price` from now on and not recompute the `current_price` for it.

contracts/marketplace/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+ink_storage = { version = "~3.3.0", default-features = false }
+ink_lang = { version = "~3.3.0", default-features = false }
+ink_prelude = { version = "~3.3.0", default-features = false }
+ink_engine = { version = "~3.3.0", default-features = false, optional = true }
+
+scale = { package = "parity-scale-codec", version = "3", default-features = false, features = ["derive"] }
+scale-info = { version = "2", default-features = false, features = ["derive"], optional = true }
+
+openbrush = { git = "https://github.com/Supercolony-net/openbrush-contracts.git", rev = "8a20f95", default-features = false, features = ["psp22"] }
+access_control = { path = "../access_control", default-features = false, features = ["ink-as-dependency"] }
```
fbielejec, line 21: make sure all these puppies are mentioned in the std section as well

contracts/marketplace/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+        auction_length: BlockNumber,
+        sale_multiplier: Balance,
+        ticket_token: AccountId,
+        reward_token: AccountId,
+    }
+
+    #[derive(Clone, Eq, PartialEq, Debug, scale::Encode, scale::Decode)]
+    #[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
+    pub enum Error {
+        MissingRole(String),
```
fbielejec, line 45: why not use the actual type here? MissingRole(Role)

contracts/marketplace/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    impl From<ink_env::Error> for Error {
+        fn from(inner: ink_env::Error) -> Self {
+            Error::ContractCall(format!("{:?}", inner))
+        }
+    }
+
+    impl From<PSP22Error> for Error {
+        fn from(inner: PSP22Error) -> Self {
```
fbielejec, line 76: PSP22Error is Decode/Encode right? So you can just:
```
    impl From<PSP22Error> for Error {
        fn from(inner: PSP22Error) -> Self {
            Error::PSP22TokenCall(inner)
        }
    }
```

contracts/marketplace/lib.rs
fbielejec: there's an argument in `psp22::transfer` that allows for passing some data along with a transfer - for messages, or callbacks

contracts/marketplace/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+        }
+
+        fn per_block_reduction(&self) -> Balance {
+            self.average_price()
+                .saturating_div(self.auction_length.into())
+                .max(1u128)
+        }
+
+        fn take_payment(&self, from: AccountId, amount: Balance) -> Result<(), Error> {
+            build_call::<Environment>()
```
fbielejec, line 226: maybe this could check the allowance before and fail with a meaningful error ?

contracts/marketplace/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+        /// Number of tickets available for sale.
+        ///
+        /// The tickets will be auctioned off one by one.
+        pub fn available_tickets(&self) -> Result<Balance, Error> {
+            self.ticket_balance()
+        }
+
+        #[ink(message)]
+        /// The minimal price the contract allows.
+        pub fn min_price(&self) -> Balance {
```
fbielejec, line 157: add a setter pls

contracts/scripts/deploy.sh
```diff
@@ -105,6 +105,43 @@ function deploy_and_instrument_game {
...
+
+  echo "Marketplace for $contract_name instance address: $contract_address"
+
+  # --- GRANT PRIVILEGES ON THE CONTRACT
+
+  cd "$CONTRACTS_PATH"/access_control
+
+  cargo contract call --url "$NODE" --contract "$ACCESS_CONTROL" --message grant_role --args "$AUTHORITY" 'Owner('"$contract_address"')' --suri "$AUTHORITY_SEED"
+  cargo contract call --url "$NODE" --contract "$ACCESS_CONTROL" --message grant_role --args "$AUTHORITY" 'Admin('"$contract_address"')' --suri "$AUTHORITY_SEED"
+  cargo contract call --url "$NODE" --contract "$ACCESS_CONTROL" --message grant_role --args "$game" 'Admin('"$contract_address"')' --suri "$AUTHORITY_SEED"
```
fbielejec, line 140: why is the game contract made an Admin?

fbielejec (commented):
looks really good, just a few comments

contracts/marketplace/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+    use crate::marketplace::Error::MissingRole;
+
+    type Event = <Marketplace as ContractEventBase>::Type;
+
+    const DUMMY_DATA: &[u8] = &[0x0];
+
+    #[ink(storage)]
+    pub struct Marketplace {
+        total_proceeds: Balance,
+        tickets_sold: Balance,
```
fbielejec, line 33: I know it's just an alias, but it seems weird to have Balance here, for sth that is clearly a counter - maybe use u128 explicitly

contracts/marketplace/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+        ticket_token: AccountId,
+        reward_token: AccountId,
+    }
+
+    #[derive(Clone, Eq, PartialEq, Debug, scale::Encode, scale::Decode)]
+    #[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
+    pub enum Error {
+        MissingRole(String),
+        ContractCall(String),
+        PSP22TokenCall(String),
```
fbielejec, line 47: same here, just wrap the other error if you can

contracts/marketplace/lib.rs
deuszx: But do we need to pass it if it's unused? Can't the `psp22::transfer` make it optional?

contracts/marketplace/lib.rs
obrok: I was basically expecting the other contract to fail with a meaningful variant in that case... In `polkadot.js` all these errors look the same, where do they even show up?

contracts/scripts/deploy.sh
obrok: My idea here was that the game contract is going to call `reset()` after it puts in the tokens.

contracts/marketplace/lib.rs
obrok: On the one hand... sure. On the other the tickets are a PSP22, so a number of tickets should have type balance... I went back and forth on this, don't have an opinion, so if this doesn't convince you the other way, I'll change it.

contracts/marketplace/lib.rs
obrok: I started out like that, but the other errors didn't implement something that was needed... I'll check it again on Monday.

obrok:
> As per code organization - modelling the contract as a state machine, without any direct dependence on the other on-chain components (like calling contracts, emitting events), would make the testing much easier. There would be a component that doesn't use any on-chain functions (basically a struct with methods on it where everything "from chain" is passed in as arguments) and the outer layer that feeds expected data into it and interprets the results.

I also thought about doing it like that. However... A lot of what this contract does is interacting with the tickets and game tokens and it seemed to me like that would have to be extracted out into traits and mocked out to test the thing. That seemed like it would be way more effort than it was worth. The price calculation itself could be extracted and tested like that, but then we're only testing a small part, feels like an unsatisfying solution. What I wanted most was to instantiate all three contracts off-chain and test their interaction like that (plus use the mocking tools for managing chain state offered by ink), but that seems impossible at the moment?

contracts/marketplace/lib.rs
obrok: I guess we control this type, so we could make it Encode/Decode, etc.

contracts/marketplace/lib.rs
obrok: Do you mean make this `pub fn price(&mut self)`? That would make people pay gas for calling this function, wouldn't it?

contracts/marketplace/lib.rs
obrok: It's an elixir idiom, I just didn't think twice it might not be obvious to a rustacean. I'll change it.

contracts/marketplace/lib.rs
obrok: From your other comment, it seems you more so mean pre-computing the parts of the expression that don't depend on `current_block`?

obrok:
> That being said, I think there's a benefit to caching the current_price() computations, as it's quite "expensive" (the parts that could be cached are per_block_reduction, average_price) and would probably be called often by other dApps/accounts to discover the price they would have to pay (although, as per my experience, if something can be computed off-chain it will be eventually b/c it's cheaper).

How would one measure that something is "expensive" in this context? Since premature optimization is the root of all evil, and all.

contracts/marketplace/lib.rs
obrok: Are you proposing commenting these private fields? I feel like the necessary info should be exposed in the comments/docstrings on the API of the contract. I'm very much against commenting private details unless it's some unclear hack, etc.

I'm definitely pro adding a top-level docstring of some sort, describing the contract at a high level, and perhaps docstrings on the error variants, although these seem less useful, as they don't seem to be displayed anywhere anyway.

contracts/marketplace/lib.rs
fbielejec: likely because the import wasn't in the std (there is one place where std is used, when using off-chain environment aka in unit tests

contracts/marketplace/lib.rs
fbielejec: it's in an external crate :(

contracts/marketplace/lib.rs
fbielejec: to be completely honest I'm not sure how such failure looks like, i.e. whether you get something beyond a revert, possibly not.

contracts/marketplace/lib.rs
fbielejec: it already is afaik

contracts/marketplace/lib.rs
deuszx: Thanks.

contracts/marketplace/lib.rs
deuszx: I would probably have a field internal to the struct that is set a tuple of block height and the price computed at that height. On every call to `price` I would check if `env::block()` is equal to the cached height and if so return the cached price.

obrok: Hm, that doesn't answer my question. I'm more worried when that field would be set, not how it would be computed. As far as I understand it's not a good idea to set it in `price` itself?

deuszx: You would set it just where you do it now. If the check returns that `env::block() > cached.block_height` then you would recompute and update the cache.

obrok:
@deuszx I ran some experiments by changing the `price` function to `price(&mut self)` and comparing the transaction fee I'm charged when calling it with caching and without. If anything, the version with caching is more expensive - there's maybe a 10% difference. For now, I'm unconvinced there's any benefit to it and it complicates the code, so I'm against introducing the cache.

obrok:
@fbielejec @deuszx I added some docs, the `set_min_price()` message and a `max_price` to `buy()`. Please take another look, I think I've addressed everything.

contracts/marketplace/lib.rs
deuszx: As stated in the https://github.com/Cardinal-Cryptography/aleph-node/pull/613#issuecomment-1243959223, this optimization isn't really optimizing anything (especially that apparently ink! doesn't allow for mutable methonds in the RPC API).

contracts/marketplace/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+
+        /// Buy one ticket at the current_price.
+        ///
+        /// The caller should make an approval for at least `price()` reward tokens to make sure the
+        /// call will succeed. The caller can specify a `max_price` - the call will fail if the
+        /// current price is greater than that.
+        #[ink(message)]
+        pub fn buy(&mut self, max_price: Option<Balance>) -> Result<(), Error> {
+            if self.ticket_balance()? <= 0 {
+                return Err(Error::MarketplaceEmpty);
```
deuszx, line 200: Maybe we should emit an event when the marketplace gets emptied? i.e. after the last ticket is sold. What do you think?

deuszx:
> > As per code organization - modelling the contract as a state machine, without any direct dependence on the other on-chain components (like calling contracts, emitting events), would make the testing much easier. There would be a component that doesn't use any on-chain functions (basically a struct with methods on it where everything "from chain" is passed in as arguments) and the outer layer that feeds expected data into it and interprets the results.
> 
> I also thought about doing it like that. However... A lot of what this contract does is interacting with the tickets and game tokens and it seemed to me like that would have to be extracted out into traits and mocked out to test the thing. That seemed like it would be way more effort than it was worth. The price calculation itself could be extracted and tested like that, but then we're only testing a small part, feels like an unsatisfying solution. What I wanted most was to instantiate all three contracts off-chain and test their interaction using the mocking tools for managing chain state offered by ink, but that seems impossible at the moment?

I can imagine abstracting away chain interaction via (simplest) a trait with methods like `call_contract`, `get_block_number`, `get_caller`, etc. (unless there's a different way to stub these) and having different implementations for on-chain use and production use where the latter would allow for registering things and then asserting them (akin a writer/state monad). That would be bigger than just for this (otherwise relatively simpler) task and could definitely use input from others when desining the API/approach.

fbielejec (approved):
lgtm

deuszx (approved):
LGTM. Ideally, there would be some tests - although apparently it's difficult to test cross-contract calls in Substrate.

I would also consider adding an event for when the marketplace is emptied.

obrok:
> I can imagine abstracting away chain interaction via (simplest) a trait with methods like call_contract, get_block_number, get_caller, etc. (unless there's a different way to stub these) and having different implementations for on-chain use and production use where the latter would allow for registering things and then asserting them (akin a writer/state monad). That would be bigger than just for this (otherwise relatively simpler) task and could definitely use input from others when desining the API/approach.

Yeah, I think that would be pretty useful overall. However, I'd propose doing it separately. From what I understood the plan of testing the game and related contracts is to interact with a local chain via shell scripts - parts of those scripts are already written in `contracts/scripts`.

contracts/marketplace/lib.rs
obrok: It's easy enough to do, although I don't have a good feel yet of what types of things should generally be announced by events, so can't really say if it's a good idea or not.

deuszx: Otherwise, how would an observer of the chain find out that the system has run out of the tickets? One would have to call `buy()` to fail with an error. That has at least two cons compared with the event:
1. Someone has to call `buy()` and pay for the txn
2. The fact marketplace is empty is returned as an error which (to me) is coincidental

I think event here (and in general) would be more explicit and visible to any observer of the contract (since that's how events work, they're "tied" to a contract) rather than watching *all* txns, filtering down to ones that call the contract and checking the result of the call.

obrok: You can use `psp22::balanceOf` to check the balance of tickets available to the market. Secondly, the market exposes `ticketsAvailable`, which is basically just a wrapper around `balanceOf`.

deuszx: Right, in that case I might be enough.

PR #616, opened 2022-09-12
A0-1340: add info level logs to finality-aleph components

maciejnems:
# Description

This PR adds new logs for finality-aleph components. Specificaly to:

**Network status report**
Example report for 2 connected authorities plus 2 connected archivists:
```
2022-09-12 16:02:59 Network status report: validator connected peers - 2 [12D3KooWCmz8XwEief5Z9Dt42MqzSJcjzD48wEoNGXxBmnodAv5S, 12D3KooWQ49SdyE3fM3vw8VupK5GvzGfvVkJMoM9UeaUQD5FvTAX]; generic connected peers - 4;
```

**Connection Manager Service status report** 
Example report for 1 missing undetected authority:
```
2022-09-12 16:02:59 Connection Manager Service status report: authenticated authorities: authenticated authorities: 1: 3/4 {2: 12D3KooWQ49SdyE3fM3vw8VupK5GvzGfvVkJMoM9UeaUQD5FvTAX, 3: 12D3KooWCmz8XwEief5Z9Dt42MqzSJcjzD48wEoNGXxBmnodAv5S}, 0: 3/4 {3: 12D3KooWCmz8XwEief5Z9Dt42MqzSJcjzD48wEoNGXxBmnodAv5S, 2: 12D3KooWQ49SdyE3fM3vw8VupK5GvzGfvVkJMoM9UeaUQD5FvTAX}; missing authorities: 1: [0], 0: [0];
```
Example report of all authorities authenticated:
```
2022-09-12 16:18:13 Connection Manager Service status report: authenticated authorities: 1: 4/4 {2: 12D3KooWQ49SdyE3fM3vw8VupK5GvzGfvVkJMoM9UeaUQD5FvTAX, 1: 12D3KooWSPrwNJc6WM7F9MTEtBfy5k2VrTgYkxUFuyNimjJmhM9h, 3: 12D3KooWCmz8XwEief5Z9Dt42MqzSJcjzD48wEoNGXxBmnodAv5S}, 0: 4/4 {1: 12D3KooWSPrwNJc6WM7F9MTEtBfy5k2VrTgYkxUFuyNimjJmhM9h, 3: 12D3KooWCmz8XwEief5Z9Dt42MqzSJcjzD48wEoNGXxBmnodAv5S, 2: 12D3KooWQ49SdyE3fM3vw8VupK5GvzGfvVkJMoM9UeaUQD5FvTAX};
```

**Block Signature Aggregator status report**
Example report:
```
Block Signature Aggregator status report: started hashes - 7; collected sginatures - 0; hashes in queue - 0;
```
Example report when hash queue is not empty:
```
2022-09-12 16:24:38 Block Signature Aggregator status report: started hashes - 7; collected sginatures - 0; hashes in queue - 7; front of hash queue - 0xd5b8…ee85 for - 2.144170386s;
```

**Justification request queue**
When clearing justification request queue, log about it was changed from debug to info log level,

## Type of change

- Add info logs

### DIFF
--- a/finality-aleph/src/aggregation/aggregator.rs
+++ b/finality-aleph/src/aggregation/aggregator.rs
@@ -1,12 +1,13 @@
 use std::{
     collections::{HashMap, HashSet, VecDeque},
     fmt::Debug,
+    time::Instant,
 };
 
 use aleph_bft::Recipient;
 use codec::Codec;
 use futures::{channel::mpsc, StreamExt};
-use log::{debug, trace, warn};
+use log::{debug, info, trace, warn};
 
 use crate::{
     aggregation::multicast::{Hash, Multicast, SignableHash},
@@ -32,6 +33,7 @@ pub struct BlockSignatureAggregator<H: Hash + Copy, PMS> {
     signatures: HashMap<H, PMS>,
     hash_queue: VecDeque<H>,
     started_hashes: HashSet<H>,
+    front_instant: Instant,
     metrics: Option<Metrics<H>>,
 }
 
@@ -41,6 +43,7 @@ impl<H: Copy + Hash, PMS> BlockSignatureAggregator<H, PMS> {
             signatures: HashMap::new(),
             hash_queue: VecDeque::new(),
             started_hashes: HashSet::new(),
+            front_instant: Instant::now(),
             metrics,
         }
     }
@@ -51,6 +54,9 @@ impl<H: Copy + Hash, PMS> BlockSignatureAggregator<H, PMS> {
         }
         if let Some(metrics) = &self.metrics {
             metrics.report_block(hash, std::time::Instant::now(), Checkpoint::Aggregating);
+        }
+        if self.hash_queue.is_empty() {
+            self.front_instant = Instant::now();
         }
         self.hash_queue.push_back(hash);
 
@@ -66,6 +72,7 @@ impl<H: Copy + Hash, PMS> BlockSignatureAggregator<H, PMS> {
         match self.hash_queue.pop_front() {
             Some(hash) => {
                 if let Some(multisignature) = self.signatures.remove(&hash) {
+                    self.front_instant = Instant::now();
                     Ok((hash, multisignature))
                 } else {
                     self.hash_queue.push_front(hash);
@@ -74,6 +81,32 @@ impl<H: Copy + Hash, PMS> BlockSignatureAggregator<H, PMS> {
             }
             None => Err(AggregatorError::NoHashFound),
         }
+    }
+
+    pub fn status_report(&self) {
+        let mut status = String::from("Block Signature Aggregator status report: ");
+
+        status.push_str(&format!(
+            "started hashes - {:?}; ",
+            self.started_hashes.len()
+        ));
+
+        status.push_str(&format!(
+            "collected sginatures - {:?}; ",
+            self.signatures.len()
+        ));
+
+        status.push_str(&format!("hashes in queue - {:?}; ", self.hash_queue.len()));
+
+        if let Some(hash) = self.hash_queue.front() {
+            status.push_str(&format!(
+                "front of hash queue - {} for - {:?}; ",
+                hash,
+                Instant::now() - self.front_instant
+            ));
+        }
+
+        info!(target: "aleph-aggregator", "{}", status);
     }
 }
 
@@ -113,6 +146,10 @@ impl<
             multicast,
             aggregator,
         }
+    }
+
+    pub fn status_report(&self) {
+        self.aggregator.status_report()
     }
 
     pub(crate) async fn start_aggregation(&mut self, hash: H) {
--- a/finality-aleph/src/aggregation/aggregator.rs
+++ b/finality-aleph/src/aggregation/aggregator.rs
@@ -33,7 +33,7 @@ pub struct BlockSignatureAggregator<H: Hash + Copy, PMS> {
     signatures: HashMap<H, PMS>,
     hash_queue: VecDeque<H>,
     started_hashes: HashSet<H>,
-    front_instant: Instant,
+    last_change: Instant,
     metrics: Option<Metrics<H>>,
 }
 
@@ -43,7 +43,7 @@ impl<H: Copy + Hash, PMS> BlockSignatureAggregator<H, PMS> {
             signatures: HashMap::new(),
             hash_queue: VecDeque::new(),
             started_hashes: HashSet::new(),
-            front_instant: Instant::now(),
+            last_change: Instant::now(),
             metrics,
         }
     }
@@ -56,7 +56,7 @@ impl<H: Copy + Hash, PMS> BlockSignatureAggregator<H, PMS> {
             metrics.report_block(hash, std::time::Instant::now(), Checkpoint::Aggregating);
         }
         if self.hash_queue.is_empty() {
-            self.front_instant = Instant::now();
+            self.last_change = Instant::now();
         }
         self.hash_queue.push_back(hash);
 
@@ -72,7 +72,7 @@ impl<H: Copy + Hash, PMS> BlockSignatureAggregator<H, PMS> {
         match self.hash_queue.pop_front() {
             Some(hash) => {
                 if let Some(multisignature) = self.signatures.remove(&hash) {
-                    self.front_instant = Instant::now();
+                    self.last_change = Instant::now();
                     Ok((hash, multisignature))
                 } else {
                     self.hash_queue.push_front(hash);
@@ -92,7 +92,7 @@ impl<H: Copy + Hash, PMS> BlockSignatureAggregator<H, PMS> {
         ));
 
         status.push_str(&format!(
-            "collected sginatures - {:?}; ",
+            "collected signatures - {:?}; ",
             self.signatures.len()
         ));
 
@@ -100,9 +100,11 @@ impl<H: Copy + Hash, PMS> BlockSignatureAggregator<H, PMS> {
 
         if let Some(hash) = self.hash_queue.front() {
             status.push_str(&format!(
-                "front of hash queue - {} for - {:?}; ",
+                "front of hash queue - {} for - {:.2} s; ",
                 hash,
-                Instant::now() - self.front_instant
+                Instant::now()
+                    .saturating_duration_since(self.last_change)
+                    .as_secs_f64()
             ));
         }
 
--- a/finality-aleph/src/justification/requester.rs
+++ b/finality-aleph/src/justification/requester.rs
@@ -1,7 +1,7 @@
 use std::{marker::PhantomData, sync::Arc, time::Instant};
 
 use aleph_primitives::ALEPH_ENGINE_ID;
-use log::{debug, error, warn};
+use log::{debug, error, info, warn};
 use sc_client_api::HeaderBackend;
 use sp_api::{BlockId, BlockT, NumberFor};
 use sp_runtime::traits::Header;
@@ -128,7 +128,7 @@ where
                 }
             }
             SchedulerActions::ClearQueue => {
-                debug!(target: "aleph-justification", "Clearing queue");
+                info!(target: "aleph-justification", "Clearing justification request queue");
                 self.block_requester.clear_justification_requests();
             }
             SchedulerActions::Wait => (),
--- a/finality-aleph/src/justification/requester.rs
+++ b/finality-aleph/src/justification/requester.rs
@@ -1,4 +1,4 @@
-use std::{marker::PhantomData, sync::Arc, time::Instant};
+use std::{fmt, marker::PhantomData, sync::Arc, time::Instant};
 
 use aleph_primitives::ALEPH_ENGINE_ID;
 use log::{debug, error, info, warn};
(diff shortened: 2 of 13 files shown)


finality-aleph/src/aggregation/aggregator.rs
```diff
@@ -75,6 +82,32 @@ impl<H: Copy + Hash, PMS> BlockSignatureAggregator<H, PMS> {
...
+    pub fn status_report(&self) {
+        let mut status = String::from("Block Signature Aggregator status report: ");
+
+        status.push_str(&format!(
+            "started hashes - {:?}; ",
+            self.started_hashes.len()
+        ));
+
+        status.push_str(&format!(
+            "collected sginatures - {:?}; ",
```
DamianStraszak, line 95: typo `sginatures`

finality-aleph/src/aggregation/aggregator.rs
```diff
@@ -75,6 +82,32 @@ impl<H: Copy + Hash, PMS> BlockSignatureAggregator<H, PMS> {
...
+            self.signatures.len()
+        ));
+
+        status.push_str(&format!("hashes in queue - {:?}; ", self.hash_queue.len()));
+
+        if let Some(hash) = self.hash_queue.front() {
+            status.push_str(&format!(
+                "front of hash queue - {} for - {:?}; ",
+                hash,
+                Instant::now() - self.front_instant
```
DamianStraszak, line 105: This code might panic -- avoid unchecked arithmetic on instants, because monotonicity might fail. Use `saturating_duration_since().as_f64`.  Please, print it with precision of `2` decimal places.

finality-aleph/src/aggregation/aggregator.rs
```diff
@@ -41,6 +43,7 @@ impl<H: Copy + Hash, PMS> BlockSignatureAggregator<H, PMS> {
             signatures: HashMap::new(),
             hash_queue: VecDeque::new(),
             started_hashes: HashSet::new(),
+            front_instant: Instant::now(),
```
DamianStraszak, line 46: Maybe better `last_change` or something suggesting what this stands for?

finality-aleph/src/substrate_network.rs
```diff
@@ -69,6 +69,12 @@ impl Decode for PeerId {
     }
 }
 
+impl fmt::Display for PeerId {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.0)
```
DamianStraszak, line 74: I was thinking maybe to print this in the form of `prefix...sufix`, where both prefix and sufix are, say `8` characters only. Because currently this is quite verbose. 

@timorl what do you think?

finality-aleph/src/justification/requester.rs
```diff
@@ -128,7 +128,7 @@ where
                 }
             }
             SchedulerActions::ClearQueue => {
-                debug!(target: "aleph-justification", "Clearing queue");
+                info!(target: "aleph-justification", "Clearing justification request queue");
```
DamianStraszak, line 131: This is not a good message to have on level `info`. It doesn't give any useful information about the internal state of the component. What we would like to catch here at least, is that we keep requesting the same block, and fail for a prolonged period of time. In such a case we want to learn the number and hash of the requested block.

DamianStraszak (commented):
I have left some comments for everything except network-related logs.

I don't know enough about the inner workings of network to give useful feedback about the relevant logs here -- @timorl should be a better person to ask. I will be looking at the network code for some time on Tuesday, so I might be able to give some comments later. For now, a little comment I can give is that it is useful to display list of peers, etc. sorted by `NodeIndex`, not in arbitrary order.

finality-aleph/src/substrate_network.rs
timorl: Technically this would usually be enough to find the related PeerId in other logs, but I'm worried that using it for different purposes would be harder (e.g. iirc there is an RPC call to mark a node as reserved and it requires the whole PeerId). So weak (since the alternative uses are unlikely) preference for the full thing being printed, but if you think this hurts the logs too much I'm fine with printing just the version you suggested.

finality-aleph/src/lib.rs
```diff
@@ -38,6 +38,7 @@ mod nodes;
 mod party;
 mod session;
 mod session_map;
+pub mod status;
```
timorl, line 41: I'm not convinced this is necessary as a separate module, since it just very thinly pack a `ticker`. Perhaps just move the `STATUS_TICKER_DELAY` (maybe rename to `STATUS_REPORT_INTERVAL`?) to this `lib.rs` file and crate plain tickers everywhere?

If not, then at least remove the `pub` here, it's unnecessary.

timorl (approved):
LGTM, just one comment on top of Damian's.

finality-aleph/src/network/manager/service.rs
```diff
@@ -544,6 +544,77 @@ impl<NI: NetworkIdentity, D: Data> Service<NI, D> {
             }
         }
     }
+
+    pub fn status_report(&self) {
+        let mut status = String::from("Connection Manager Service status report: ");
```
DamianStraszak, line 549: Let's shorten this to "Connection Manager status:"

timorl:
Ah, one additional request – could you bump the two first (not the last one) logs in `network/service.rs` in the `peer_sender` function to `info`? It turns out they don't really happen in usual operation and they should be a symptom of something going quite wrong usually.

finality-aleph/src/substrate_network.rs
maciejnems: I guess the first `8` characters are always the same in PeerId. So maybe we could print prefix of `4` and sufix of `8`?

finality-aleph/src/lib.rs
maciejnems: removed it

timorl (approved):
Some nits.

finality-aleph/src/network/service.rs
```diff
@@ -112,13 +116,13 @@ impl<N: Network, D: Data> Service<N, D> {
                         match network.sender(peer_id, protocol) {
                             Ok(sender) => senders.entry(protocol).or_insert(sender),
                             Err(e) => {
-                                debug!(target: "aleph-network", "Failed creating sender. Dropping message: {:?}", e);
+                                info!(target: "aleph-network", "Failed creating sender. Dropping message: {}", e);
```
DamianStraszak, line 119: Are we sure this will not create spam? I definitely want to avoid a situation where the log is spammed with this. 
Each type of message should be upper bounded to appear once per 10s or some such. 

Similarly for `Failed sending data to peer`.

Just create some aggregate statistics on failures and print them every 10 sec.

maciejnems: After some research this will only log in case `StreamClosed` notification did not happen before trying to send here... or if there is some strange bug in substrate. So seems like this log might be undeterministic ://  Also after some thinking this does not sound like info log. Aggregation could be the way to go.

maciejnems: There is one problem though. This function is called per peer, so this is not that easy to aggregate statistics from all those at one place. I don't think sending such data over channels is worth it. That said we could add this aggregation to `send_to_peer` function. Everything that is sent to senders goes through this function. I think there we can aggregate each failure to send.

Still, I'm wondering, what exactly should we aggregate:
*  Should we aggregate number of unsent messages per peer? (I guess there might be a lot of peers missing, and this could cause a lot of spam) 
* Should we aggregate just what peers we failed to send? (same nit against as in above point)
* Just aggregating number of unsent messages is not that useful. 

And also for how long should we aggregate them? Should we print peers/messages lost since last status report? I think something like this would be the way to go, but 10 seconds seems like a really short time to reset such aggregation.

So to sum up, I would personally do this aggregation in `send_to_peer`, aggregate unsent messages per user and reset it once every 5-10 status reports?

finality-aleph/src/justification/requester.rs
maciejnems: I guess we could save last num that we requested, save how many times and save a block hash associated with this num. Status report could be called every 10 seconds in component above.

I implemented this in a way, that if there is no pending status report, it should log
```
Justification requester status report: tries - 0;
```
and otherwise, something like that
```
Justification requester status report: tries - 1; requested block number - 5; hash - 0x5d6d…614a;
```

finality-aleph/src/network/service.rs
timorl: For context, I suggested this change because there were _no_ such logs (at the appropriate level) even when there were problems with connections – as far as I can tell, these are really rare. Sending failures usually turn up in a completely different part of the code.

On the other hand maybe this is not good enough reasoning for level bumping, so perhaps we shouldn't be doing that. I'm kinda skeptical the setup needed to aggregate failure logs here is worth it, especially since, as @maciejnems explained it would have to add some centralization and cross-thread communication, which is usually more complex than one would hope for in logging code.

finality-aleph/src/justification/requester.rs
```diff
@@ -16,6 +16,64 @@ use crate::{
...
+            self.block_hash = None;
+            self.tries = 1;
+        }
+    }
+
+    fn insert_hash(&mut self, hash: B::Hash) {
+        self.block_hash = Some(hash);
+    }
+
+    fn try_reset(&mut self, number: NumberFor<B>) {
```
DamianStraszak, line 49: What's the purpose of this reset?

finality-aleph/src/justification/requester.rs
```diff
@@ -16,6 +16,64 @@ use crate::{
     network, Metrics,
 };
 
+/// This structure is created for keeping and reporting status of BlockRequester
+pub struct LastRequest<B: BlockT> {
```
DamianStraszak, line 20: Why the strange name? Shouldn't it be called something like `JustificationRequestStatus`?

finality-aleph/src/justification/requester.rs
DamianStraszak: Thanks. This looks good now, but could you also make sure the default (i.e. healthy node) is printing nothing, instead of printing some generic message? This could be done by printing only if `tries` crosses some threshold.

finality-aleph/src/network/service.rs
DamianStraszak: I cannot really provide any useful insight here, on top of what @timorl already said. Aggregation per peer does not make much sense to me, as it does not guarantee satisfaction of the no-spam condition. 

If the only satisfying solution is complex, maybe you should just move it out of scope of this PR, and create a new task for it?

finality-aleph/src/justification/requester.rs
maciejnems: So that we don't log for example 3 retries for block after we received it, but before we request another block number.

finality-aleph/src/justification/requester.rs
maciejnems: Is 3 tries a good threshold?

finality-aleph/src/justification/requester.rs
maciejnems: It was supposed to be only status of LastRequest, but `JustificationRequestStatus` sounds like a better name.

finality-aleph/src/justification/requester.rs
DamianStraszak: But why would that be wrong? 
We are logging info about this particular component, and don't care what happened outside. Actually the logs are here mostly because there might be bugs when interacting with outside components. In particular, if we are logging info about requests, let's focus on requests. Whether we received something is a different matter.

maciejnems: Ok, good point. I will remove it then.

finality-aleph/src/justification/requester.rs
DamianStraszak: Hard to say -- this would be something to figure out. Does it happen to have 3 tries if everything is healthy? What's the typical behavior of this component when things are healthy?

Alternatively, if you don't want to play with that, it's probably fine to always print, but with lower frequency, something like 60 sec should be fine. Also generally I would recommend making separate constants, not share the 10s one between all. It might make sense to increase/reduce them later individually.

maciejnems: When things are healthy nothing should be requested more then once.

Typical behavior is such that after three retries without any finalization in mean time, justification queue is cleared. So usually if this is 3 we know, that something might be going on, and we cleared the queue to fix that. 

On topic of frequency, actually 10 seconds might be too frequently. Some of the logs we have here are quite big. Maybe we could just change all of them to 20/30 seconds

DamianStraszak: Yes, you can change others to 20-30 seconds.

And regarding this particular one -- you can make it print only if, say, `tries>=2`. Alternatively, just make it rare, like 60sec, and you can ten print every time.

finality-aleph/src/network/service.rs
timorl: Oh yeah, this specific change will become mostly irrelevant with the separate validator network, duh. >.<

DamianStraszak: Ok, then if you think this will not create much spam, you can leave it on `info`, otherwise better to change to `debug` and leave it for the future.

maciejnems: I changed it back to debug

DamianStraszak (approved):

PR #627, opened 2022-09-20
A0-1407 Validator Network trait Network

lesniak43:


finality-aleph/src/validator_network.rs
```diff
@@ -0,0 +1,37 @@
...
+/// always delivered, nor the established connections must be secure in any way. If the Network
+/// implementation fails to deliver a message, it may assume that the send method will be called
+/// again.
+#[async_trait::async_trait]
+pub trait Network<A, D>: Send {
+    /// Add the peer to the set of connected peers.
+    fn add_connection(&mut self, peer: AuthorityId, addresses: Vec<A>);
+
+    /// Remove the peer from the set of connected peers and close the connection.
+    fn del_connection(&mut self, peer: AuthorityId);
```
maciejnems, line 16: Maybe `remove_connection`? Sounds better for me.

lesniak43: sure

finality-aleph/src/validator_network.rs
```diff
@@ -0,0 +1,41 @@
...
+/// always delivered, nor the established connections must be secure in any way. If the Network
+/// implementation fails to deliver a message, it may assume that the send method will be called
+/// again.
+#[async_trait::async_trait]
+pub trait Network<A, D>: Send {
+    /// Add the peer to the set of connected peers.
+    fn add_connection(&mut self, peer: AuthorityId, addresses: Vec<A>);
+
+    /// Remove the peer from the set of connected peers and close the connection.
+    fn remove_connection(&mut self, peer: AuthorityId);
```
kostekIV, line 17: do we assume that add[remove]_functions always succeed?  If not change return type to smth like Result<(), NetworkError>

kostekIV: And add errors to reflect  such cases: peer not in the connections, peer already in the connections etc

finality-aleph/src/validator_network.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    /// Send a message to a single peer.
+    /// Note on the implementation: this function should be implemented in a non-blocking manner.
+    fn send(&self, data: D, recipient: AuthorityId);
+
+    /// Receive a message from the network.
+    async fn next(&mut self) -> D;
```
kostekIV, line 24: maybe `-> Option<D>` where None symbolizes end of the network

finality-aleph/src/validator_network.rs
```diff
@@ -0,0 +1,41 @@
...
+pub trait Network<A, D>: Send {
+    /// Add the peer to the set of connected peers.
+    fn add_connection(&mut self, peer: AuthorityId, addresses: Vec<A>);
+
+    /// Remove the peer from the set of connected peers and close the connection.
+    fn remove_connection(&mut self, peer: AuthorityId);
+
+    /// Send a message to a single peer.
+    /// Note on the implementation: this function should be implemented in a non-blocking manner.
+    fn send(&self, data: D, recipient: AuthorityId);
```
kostekIV, line 21: could we also have here return type of the Result to indicate if something was wrong with sending the message (in cases where it is possible)

finality-aleph/src/validator_network.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    /// Receive a message from the network.
+    async fn next(&mut self) -> D;
+}
+
+pub struct MockNetwork;
+
+#[async_trait::async_trait]
+impl Network<(), ()> for MockNetwork {
```
kostekIV, line 30: feels like this does not belong here

finality-aleph/src/validator_network.rs
```diff
@@ -0,0 +1,41 @@
...
+
+/// Network represents an interface for opening and closing connections with other Validators,
+/// and sending direct messages between them.
+///
+/// Note on Network reliability and security: it is neither assumed that the sent messages must be
+/// always delivered, nor the established connections must be secure in any way. If the Network
+/// implementation fails to deliver a message, it may assume that the send method will be called
+/// again.
+#[async_trait::async_trait]
+pub trait Network<A, D>: Send {
```
kostekIV, line 12: Not sure what they exactly should be but some constraints on A and D would be nice here

kostekIV (changes requested):

finality-aleph/src/validator_network.rs
timorl: The main implementation will likely just use channels, so it's very hard to return meaningful errors here, unfortunately. :/

finality-aleph/src/validator_network.rs
timorl: Same comment as above about channels, plus we cannot trust in it succeeding anyway.

finality-aleph/src/validator_network.rs
timorl: This is for development purposes, probably should be commented as such though.

finality-aleph/src/validator_network.rs
```diff
@@ -0,0 +1,41 @@
+use aleph_primitives::AuthorityId;
+use futures::future::pending;
+
+/// Network represents an interface for opening and closing connections with other Validators,
+/// and sending direct messages between them.
+///
+/// Note on Network reliability and security: it is neither assumed that the sent messages must be
+/// always delivered, nor the established connections must be secure in any way. If the Network
+/// implementation fails to deliver a message, it may assume that the send method will be called
+/// again.
```
timorl, line 10: Replace the last sentence with "The Network implementation might fail to deliver any specific message, so messages have to be resend while they still should be delivered."

finality-aleph/src/validator_network.rs
```diff
@@ -0,0 +1,41 @@
...
+#[async_trait::async_trait]
+pub trait Network<A, D>: Send {
+    /// Add the peer to the set of connected peers.
+    fn add_connection(&mut self, peer: AuthorityId, addresses: Vec<A>);
+
+    /// Remove the peer from the set of connected peers and close the connection.
+    fn remove_connection(&mut self, peer: AuthorityId);
+
+    /// Send a message to a single peer.
+    /// Note on the implementation: this function should be implemented in a non-blocking manner.
```
timorl, line 20: Remove the prefix before ':', it's unnecessary.

timorl (approved):
Some comment nits, and fix the thing @kostekIV plus add the missing pieces we discussed.

finality-aleph/src/validator_network.rs
timorl: They will probably appear in implementations/uses naturally, I don't think they are necessary in the trait itself.

finality-aleph/src/validator_network.rs
lesniak43: added a comment

finality-aleph/src/validator_network.rs
lesniak43: fixed

finality-aleph/src/validator_network.rs
lesniak43: fixed

finality-aleph/src/validator_network.rs
kostekIV: still imho trait should at least try to make it possible even though main implementation may ignore it

lesniak43: > do we assume that add[remove]_functions always succeed?
I think we might assume that, because the intention is not to have an established connection after the method returns, but to only add/remove the address to an internal hashmap
> If not change return type to smth like Result<(), NetworkError>
In this case we'd have to make these methods `async`, and we'd rather not...

kostekIV: Would it need to be an async though? Feel free to resolve this (and send) conversation but I still feel like here and in case of send we can make the trait a little more open to possible fails even though the implementation may just not bother with 'fail' execution path.

kostekIV (approved):
Leaving my comments for you to resolve if you feel like they are not necessary. Otherwise lgtm

finality-aleph/src/validator_network.rs
lesniak43: yeah, but in the end it would be more misleading to have this if we do not even plan to implement returning `Err`, so I think it's best to be honest :)

PR #628, opened 2022-09-20
MC: extract aggregator to crate

kostekIV:
Changes required to also have different versions of aggregator (based on aleph-bft version)

### DIFF
--- a/aggregator/Cargo.toml
+++ b/aggregator/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "aggregator"
+version = "0.1.0"
+authors = ["Cardinal Cryptography"]
+edition = "2021"
+license = "Apache 2.0"
+
+
+[dependencies]
+aleph-bft = "0.18.1"
+aleph-bft-rmc = "0.4.0"
+
+async-trait = "0.1"
+futures = "0.3"
+codec = { package = "parity-scale-codec", version = "3.0", default-features = false, features = ["derive"] }
+tokio = { version = "1.17", features = [ "sync", "macros", "time", "rt-multi-thread" ] }
+log = "0.4"
--- a/aggregator/Cargo.toml
+++ b/aggregator/Cargo.toml
@@ -7,8 +7,8 @@ license = "Apache 2.0"
 
 
 [dependencies]
-aleph-bft = "0.18.1"
 aleph-bft-rmc = "0.4.0"
+aleph-bft-types = "0.6.0"
 
 async-trait = "0.1"
 futures = "0.3"
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -11,7 +11,7 @@ use log::{debug, info, trace, warn};
 
 use crate::{
     multicast::{Hash, Multicast, SignableHash},
-    DataNetwork, Metrics,
+    ProtocolSink, Metrics,
 };
 
 #[derive(Debug, PartialEq, Eq)]
@@ -114,7 +114,7 @@ impl<H: Copy + Hash, PMS, M: Metrics<H>> BlockSignatureAggregator<H, PMS, M> {
 pub struct IO<
     H: Hash + Copy,
     D: Clone + Codec + Debug + Send + Sync + 'static,
-    N: DataNetwork<D>,
+    N: ProtocolSink<D>,
     PMS,
     RMC: Multicast<H, PMS>,
     M: Metrics<H>,
@@ -129,7 +129,7 @@ pub struct IO<
 impl<
         H: Copy + Hash,
         D: Clone + Codec + Debug + Send + Sync,
-        N: DataNetwork<D>,
+        N: ProtocolSink<D>,
         PMS,
         RMC: Multicast<H, PMS>,
         M: Metrics<H>,
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -9,9 +9,8 @@ use futures::{channel::mpsc, StreamExt};
 use log::{debug, trace, warn};
 
 use crate::{
-    aggregation::multicast::{Hash, Multicast, SignableHash},
-    metrics::{Checkpoint, Metrics},
-    network::DataNetwork,
+    multicast::{Hash, Multicast, SignableHash},
+    DataNetwork, Metrics,
 };
 
 #[derive(Debug, PartialEq, Eq)]
@@ -28,15 +27,15 @@ pub type AggregatorResult<R> = Result<R, AggregatorError>;
 pub type IOResult = Result<(), IOError>;
 
 /// A wrapper around an `rmc::Multicast` returning the signed hashes in the order of the [`Multicast::start_multicast`] calls.
-pub struct BlockSignatureAggregator<H: Hash + Copy, PMS> {
+pub struct BlockSignatureAggregator<H: Hash + Copy, PMS, M: Metrics<H>> {
     signatures: HashMap<H, PMS>,
     hash_queue: VecDeque<H>,
     started_hashes: HashSet<H>,
-    metrics: Option<Metrics<H>>,
+    metrics: Option<M>,
 }
 
-impl<H: Copy + Hash, PMS> BlockSignatureAggregator<H, PMS> {
-    pub(crate) fn new(metrics: Option<Metrics<H>>) -> Self {
+impl<H: Copy + Hash, PMS, M: Metrics<H>> BlockSignatureAggregator<H, PMS, M> {
+    pub fn new(metrics: Option<M>) -> Self {
         BlockSignatureAggregator {
             signatures: HashMap::new(),
             hash_queue: VecDeque::new(),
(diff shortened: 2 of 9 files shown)


aggregator/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[dependencies]
+aleph-bft = "0.18.1"
+aleph-bft-rmc = "0.4.0"
+
+async-trait = "0.1"
+futures = "0.3"
+codec = { package = "parity-scale-codec", version = "3.0", default-features = false, features = ["derive"] }
+tokio = { version = "1.17", features = [ "sync", "macros", "time", "rt-multi-thread" ] }
+log = "0.4"
```
timorl, line 17: Newline. :<

aggregator/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "aggregator"
+version = "0.1.0"
+authors = ["Cardinal Cryptography"]
+edition = "2021"
+license = "Apache 2.0"
+
+
+[dependencies]
+aleph-bft = "0.18.1"
```
timorl, line 10: I'm pretty sure you are only using this for `Recipient`, so you probably want `aleph-bft-types` instead.

aggregator/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+
+pub type RmcNetworkData<H, S, SS> = Message<SignableHash<H>, S, SS>;
+
+#[derive(Debug)]
+pub enum NetworkError {
+    SendFail,
+}
+
+#[async_trait::async_trait]
+pub trait DataNetwork<D>: Send + Sync {
```
timorl, line 22: Welp, I really don't like that this is another instance of the same trait. Could we just use this instead of the one defined in `finality-aleph` everywhere?

timorl: Alternatively at least fix `D = RmcNetworkData`? Although this will likely look uglier in code, since then this would have 3 type parameters...

aggregator/src/aggregator.rs
```diff
@@ -226,21 +227,30 @@ impl<
...
+    };
 
     type TestMultisignature = usize;
     const TEST_SIGNATURE: TestMultisignature = 42;
 
-    fn build_aggregator() -> BlockSignatureAggregator<THash, TestMultisignature> {
+    struct MockMetrics;
+    impl Metrics<MockHash> for MockMetrics {
+        fn report_aggregation_complete(&mut self, _h: MockHash) {
+            todo!()
```
timorl, line 242: Then do. :>

timorl: Unless you don't need it because of the `Option`,, in which case plx delete.

aggregator/src/multicast.rs
```diff
@@ -4,21 +4,33 @@
...
 
-use crate::crypto::{Keychain, Signature};
-
 /// A convenience trait for gathering all of the desired hash characteristics.
 pub trait Hash: AsRef<[u8]> + StdHash + Eq + Clone + Codec + Debug + Display + Send + Sync {}
 
 impl<T: AsRef<[u8]> + StdHash + Eq + Clone + Codec + Debug + Display + Send + Sync> Hash for T {}
 
+#[derive(StdHash, PartialEq, Eq, Clone, Copy, Encode, Decode, Debug)]
+pub struct MockHash(pub [u8; 32]);
```
timorl, line 20: Hide this thing under a `cfg(test)` plx, perhaps in a separate file even.

finality-aleph/Cargo.toml
```diff
@@ -10,6 +10,8 @@ aleph-bft = "0.18.1"
 aleph-bft-rmc = "0.4.0"
 aleph-primitives = { package = "primitives", path = "../primitives" }
 
+aleph-aggregator = { path = "../aggregator", package = "aggregator" }
```
timorl, line 13: Hmm, if we want this to work as advertised `aleph-aggregator` should get published to crates.io and we have to depend on a given version, not pathwise. I'm pretty sure you can add an explicit version in this specification here already, even if the rest of the infrastructure is not yet done.

Also, figure out how to automatically publish? Probably as a separate task.

finality-aleph/Cargo.toml
```diff
@@ -10,6 +10,8 @@ aleph-bft = "0.18.1"
...
+
```
timorl, line 14: I think this line is fine, but the one two lines above should be deleted, no?

finality-aleph/src/aggregation/mod.rs
```diff
@@ -1,14 +1,40 @@
...
+};
 
-mod aggregator;
-mod multicast;
+pub type RmcNetworkData<B> =
+    aleph_aggregator::RmcNetworkData<<B as Block>::Hash, Signature, SignatureSet<Signature>>;
 
-pub use aggregator::{BlockSignatureAggregator, IO};
-pub use multicast::SignableHash;
+pub struct NetworkWrapper<D: Data, N: DataNetwork<D>>(pub N, pub PhantomData<D>);
```
timorl, line 17: Why is the `PhantomData` `pub`? :D

finality-aleph/src/aggregation/mod.rs
```diff
@@ -1,14 +1,40 @@
...
-pub type RmcNetworkData<B> =
-    Message<SignableHash<<B as Block>::Hash>, Signature, SignatureSet<Signature>>;
+impl<H: Debug + Hash + Eq + Debug + Copy> aleph_aggregator::Metrics<H> for Metrics<H> {
+    fn report_aggregation_complete(&mut self, h: H) {
+        self.report_block(h, Instant::now(), Checkpoint::Aggregating);
+    }
+}
+
+#[async_trait::async_trait]
+impl<T, D> aleph_aggregator::DataNetwork<D> for NetworkWrapper<D, T>
```
timorl, line 26: Yeah, this implementation is kinda a farce. :/ See comment by the trait definition.

timorl (changes requested):
Lots of comments, but good work.

finality-aleph/Cargo.toml
kostekIV: I was thinking of just using commit/tag at first then you don't need to publish it yet to crate.io and can select version from github repository

aggregator/src/lib.rs
fixxxedpoint: What about removing that trait at all from this crate and provide similar mechanism to that  of `rmc` from aleph_bft - API with some channels of data and consumer must do something with them. Then you need to move that network forwarding from aggregatorIO.

aggregator/src/aggregator.rs
fixxxedpoint: Suggestion: remove that `Option`, provide some default empty trait implementation and use it instead.

finality-aleph/Cargo.toml
timorl: Makes some sense, although crates.io would likely be preferable.

Also, even then you probably want to start immediately and include the tag here.

kostekIV: I would mark crates.io as a necessary/desired improvement. I will do the tag version first, after this pr is merged because pushing tag to GitHub with not yet finalized version of the pr will be a mess imo

aggregator/src/aggregator.rs
kostekIV: left with empty implementation

aggregator/src/multicast.rs
kostekIV: moved to tests

finality-aleph/src/aggregation/mod.rs
kostekIV: because I was lazy and didnt want to implement constructor for this :P fixed

aggregator/Cargo.toml
kostekIV: fixed :>

aggregator/Cargo.toml
kostekIV: done

aggregator/src/lib.rs
fixxxedpoint: I am ok(-ish) with copy-pasted impl in aggregator's crate. Suggestion to remove `<D>` is good, but has also weird cons: from type level perspective requiring it to be for `forall D` means it can't be specialized for specific `D=RmcNetworkData`. If we fix `D`, then aggregator itself, kind of, doesn't define a self-contained versioned protocol - finality-aleph might provide DataNetwork impl that do some `additional` things (change that doesn't depend on aggregator's api at all).

fixxxedpoint: So, actually channel based api doesn't fix that `problem` either. Maybe just change it's name from `DataNetwork` to something like `ProtocolSink` and everyone will be happy :)

kostekIV: for now, I have renamed the trait to ProtocolSink but without fixing the D. in the aggregation code we do not need to know what is the D and trying to fix it to RmcNetworkData would result in introducing more type parameters that are actually not relevant to the code

finality-aleph/Cargo.toml
timorl: Yeah, essentially why is `aggregator` in its own very separate line, shouldn't it be right next to the ones above?

aggregator/src/aggregator.rs
timorl: Huh, I just noticed that you could provide a blanked implementation for `Option<T>` when `T: Metrics` that does nothing on `None` and marginally simplify the interface by requiring `M: Metrics` instead of `Option<M>`. But that is just something to think about, I'm not even sure it's actually reasonable.

aggregator/src/lib.rs
timorl: Ugh, this is still disgusting but at least swept under the rug by renaming. xd

I think that maybe the actual answer is to have a generic `DataNetwork` in AlephBFT and then use that everywhere, including changing the interface of `aleph-rmc`. I'll create a task...

timorl (approved):
Just that one nit left from me.

fixxxedpoint (approved):
lgtm

PR #630, opened 2022-09-22
A0-1238: Put tickets into marketplace when button dies

obrok:
# Description

Moves all held ticket to the marketplace on `button.reset()` and calls `marketplace.reset()` to restart the auction as needed. Note that this removes the limitation that only one ticket can be transferred at a time.

## Type of change

- New feature (non-breaking change which adds

### DIFF
--- a/contracts/button/lib.rs
+++ b/contracts/button/lib.rs
@@ -6,7 +6,7 @@ use ink_lang as ink;
 
 #[ink::contract]
 mod button_game {
-    use access_control::{traits::AccessControlled, Role, ACCESS_CONTROL_PUBKEY};
+    use access_control::{roles::Role, traits::AccessControlled, ACCESS_CONTROL_PUBKEY};
     use game_token::MINT_TO_SELECTOR;
     use ink_env::{
         call::{build_call, Call, ExecutionInput, Selector},
--- a/contracts/button/lib.rs
+++ b/contracts/button/lib.rs
@@ -330,8 +330,7 @@ mod button_game {
 
         fn dump_tickets_in_marketplace(&self) -> ButtonResult<()> {
             self.transfer_tickets_to_marketplace()?;
-            self.reset_marketplace()?;
-            Ok(())
+            self.reset_marketplace()
         }
 
         fn transfer_tickets_to_marketplace(&self) -> ButtonResult<()> {
--- a/contracts/button/lib.rs
+++ b/contracts/button/lib.rs
@@ -231,7 +231,8 @@ mod button_game {
             self.ensure_dead()?;
             self.reward_pressiah()?;
             self.reset_state();
-            self.dump_tickets_in_marketplace()
+            self.transfer_tickets_to_marketplace()?;
+            self.reset_marketplace()
         }
 
         /// Sets new access control contract address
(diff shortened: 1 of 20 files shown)


contracts/button/lib.rs
```diff
@@ -295,6 +300,81 @@ mod button_game {
...
+            Self::emit_event(self.env(), Event::GameReset(GameReset { when: now }));
+        }
+
+        fn reward_pressiah(&self) -> ButtonResult<()> {
+            if let Some(pressiah) = self.last_presser {
+                let reward = self.pressiah_score();
+                self.mint_reward(pressiah, reward)??;
+            };
+
+            Ok(())
```
kostekIV, line 320: I see it was like that in the previous implementation but here is philosophical question: why fn reward_pressiah ends in Ok when there is no pressiah?

contracts/button/lib.rs
```diff
@@ -295,6 +300,81 @@ mod button_game {
...
+                return Err(GameError::BeforeDeadline);
+            } else {
+                Ok(())
+            }
+        }
+
+        fn dump_tickets_in_marketplace(&self) -> ButtonResult<()> {
+            self.transfer_tickets_to_marketplace()?;
+            self.reset_marketplace()?;
+            Ok(())
```
kostekIV, line 334:
```suggestion
            self.reset_marketplace()
```

contracts/button/lib.rs
```diff
@@ -212,24 +228,11 @@ mod button_game {
...
         pub fn reset(&mut self) -> ButtonResult<()> {
-            if !self.is_dead() {
-                return Err(GameError::BeforeDeadline);
-            }
-
-            let now = self.env().block_number();
+            self.ensure_dead()?;
+            self.reward_pressiah()?;
+            self.reset_state();
+            self.dump_tickets_in_marketplace()?;
```
kostekIV, line 234: no need to return Ok(())
```suggestion
            self.dump_tickets_in_marketplace()
```

contracts/button/lib.rs
obrok: I'd say it's still `Ok(_)` in that case, because that's the expected outcome - no pressiah, no reward. An unexpected outcome would be something like a call failure due to another contract not being on chain or whatever.

kostekIV (approved):

contracts/button/lib.rs
```diff
@@ -295,6 +298,80 @@ mod button_game {
...
+
+        fn ensure_dead(&self) -> ButtonResult<()> {
+            if !self.is_dead() {
+                return Err(GameError::BeforeDeadline);
+            } else {
+                Ok(())
+            }
+        }
+
+        fn dump_tickets_in_marketplace(&self) -> ButtonResult<()> {
```
fbielejec, line 329: putting these two into one function seems like an overkill, and it is "semantically incorrect" : it transfers (aka dumps) tickets but it also re-sets the contract state

obrok: I disagree, but not strongly enough to argue :P

obrok:
@fbielejec PTAL

fbielejec (approved):

PR #632, opened 2022-09-23
MC: Bump rmc version in aggregator

kostekIV:


finality-aleph/Cargo.toml
```diff
@@ -9,7 +9,7 @@ license = "Apache 2.0"
 aleph-bft = "0.18.1"
 aleph-bft-rmc = "0.4.0"
 aleph-primitives = { package = "primitives", path = "../primitives" }
-aleph-aggregator = { package = "aggregator", path = "../aggregator" }
+aleph-aggregator = { package = "aggregator", git = "https://github.com/Cardinal-Cryptography/aleph-node.git", tag = "aggregator-rmcv0.4" }
```
timorl, line 12: Uh, the tag should probably be versioned by the version of aggregator itself, rather than its dependencies.

timorl (approved):
Looks good, does it work? Also, is aggregator being compiled and tested as part of the pipeline currently?

finality-aleph/Cargo.toml
kostekIV: will rename

kostekIV:
> Looks good, does it work? Also, is aggregator being compiled and tested as part of the pipeline currently?

let me double check it, it should be as as it is part of the workspace

finality-aleph/Cargo.toml
kostekIV: done

kostekIV:
> > Looks good, does it work? Also, is aggregator being compiled and tested as part of the pipeline currently?
> 
> let me double check it, it should be as as it is part of the workspace

yep it is compiled and tested in pipelines

Marcin-Radecki (approved):

PR #635, opened 2022-09-23
A0-1262: Fix env not wrapped and vars not escaped in deploy-feature-envs

mikogs:


.github/workflows/deploy-feature-envs.yaml
```diff
@@ -306,7 +308,7 @@ jobs:
           CREATE_HOOK: true
         run: |
           # Set up envs
-          PR_IMAGE="${{ env.FE_ALEPHNODE_REGISTRY }}:${{ env.IMAGE_TAG }}"
+          PR_IMAGE="${{ env.FE_ALEPHNODE_REGISTRY_ESCAPED }}\:${{ env.IMAGE_TAG }}"
```
Kosiare, line 311: Is `\` supposed to be here?

mikogs: yes, it's supposed to be there

Kosiare (approved):

PR #636, opened 2022-09-26
`aleph-client`: private multisig

pmikolajczyk41:
# Description

The multisig module in `aleph-client` requires `KeyPair`s of all members right away when creating `MultisigParty` object. This is too restrictive, since it doesn't allow to interact with multisig accounts without knowing all private keys (while just public keys with a single private one should be sufficient to do something).

This PR lifts this requirement. Firstly, `MultisigParty` keeps only public keys of its members. Secondly, every multisig action requires the connection to be already signed.

**Note for reviewers**: If you want to play with the new code, I left you a commit `playground` (it's reverted in the final PR version) which adds a simple command do `cliain` which performs a simple scenario using the new API.

## Type of change

# Checklist:

### DIFF
--- a/aleph-client/src/multisig.rs
+++ b/aleph-client/src/multisig.rs
@@ -262,7 +262,7 @@ impl MultisigParty {
     }
 
     /// Checks whether `connection` is signed by some member and if so, returns their index.
-    fn map_signer_to_member(&self, connection: &SignedConnection) -> Result<usize> {
+    fn map_signer_to_member_index(&self, connection: &SignedConnection) -> Result<usize> {
         self.members
             .binary_search(&account_from_keypair(&connection.signer))
             .map_err(|_| MultisigError::IncorrectSignature.into())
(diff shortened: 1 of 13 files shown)


aleph-client/src/multisig.rs
```diff
@@ -277,36 +261,35 @@ impl MultisigParty {
...
 
-    /// Checks whether `member_idx` is a proper position for `self.members`.
-    fn ensure_index(&self, member_idx: usize) -> Result<()> {
-        ensure!(
-            member_idx < self.members.len(),
-            MultisigError::IncorrectMemberIndex
-        );
-        Ok(())
+    /// Checks whether `connection` is signed by some member and if so, returns their index.
+    fn map_signer_to_member(&self, connection: &SignedConnection) -> Result<usize> {
```
kostekIV, line 265: nit: map_signer_to_index or return the accountId

aleph-client/src/multisig.rs
```diff
@@ -44,6 +42,8 @@ pub enum MultisigError {
     CallConflict,
     #[error("👪❌ Only the author can cancel aggregation.")]
     NotAuthor,
+    #[error("👪❌ The connection is signed by an account that doesn't match to any member.")]
+    IncorrectSignature,
```
kostekIV, line 46: nit: rename since IncorrectSignature means at least to me that something is bad with the signature itself not that it was signed by not related to Multisignature account

kostekIV (approved):
nice rework

Marcin-Radecki (approved):

aleph-client/src/multisig.rs
pmikolajczyk41: renamed

aleph-client/src/multisig.rs
pmikolajczyk41: sure, renamed

PR #637, opened 2022-09-29
Add metrics trait

kostekIV:
Small improvement for metrics: rework metrics into trait + implement it for type Option<Metrics>.


Marcin-Radecki (approved):
Change looks good.

Have some general observations to Rust code and conventions in this language: while this change indeed it makes the code on the caller side easier, it makes it more complex on the "lib" side (ie in file `finality-aleph/src/metrics.rs`), which is good in general given we won't change frequently file `finality-aleph/src/metrics.rs`. Also, I'm not sure if adding one more template parameter increases code compilation time? Not saying this is an issue here, but in general, I wonder ham many generics parameters we can afford and still say code is readable enough  & compiles fast.

maciejzelaszczyk (approved):

PR #638, opened 2022-09-30
A0 1359 kickout funcionality

Marcin-Radecki:
# Description

This PR introduces two changes: 1. Automatic committee kick-out logic due to bad uptime and 2) minor change for calculating rewards for new elections.
  
1. Kick out logic

In case of insufficient validator's uptime, we need to make such validators are removed from the committee, so that the network is as healthy as possible. This is achieved by calculating a number of underperformance sessions, which means that number of blocks produced by the validator is less than some predefined threshold. 

In other words, if a validator:
* produced less or equal blocks to a `CurrentEraCommitteeKickOutThresholds::block_count_threshold`, and,
* it happened at least `CurrentEraCommitteeKickOutThresholds::underperformed_session_count_threshold` times, 
then the validator is considered an underperformer and hence removed (ie kicked out) from the committee.

2. minor change for calculating rewards for new elections.

Previously, when a validator produced more blocks than expected in a session, e.g. due to other validators not producing blocks, then it could be rewarded more than 100% of the total possible reward. This change caps this to 100% always when number of produced blocks or greater or equal to 90%. 

## Type of change

# Checklist:

# Testing done
Manual tests on local chain:
* adding validator manually to kick out the list,
* stopping validator to test situation it does not produce blocks

# Future PRs
There would be at least one additional PR that contains e2e tests for new kick-out logic.


pallets/elections/src/impls.rs
```diff
@@ -293,6 +299,45 @@ where
...
+
+    fn mark_validator_underperformance(
+        thresholds: &CommitteeKickOutThresholds,
+        validator: &T::AccountId,
+    ) {
+        UnderperformedValidatorSessionCount::<T>::mutate(&validator, |count| {
+            *count += 1;
+        });
+        let counter = UnderperformedValidatorSessionCount::<T>::get(&validator);
+        if counter >= thresholds.underperformed_session_count_threshold {
```
deuszx, line 325: We should probably "reset" this counter if validator performed correctly in `n out of M` recent sessions. I.e. I don't want to keep being worried whether I get kicked out for missing a block b/c months ago I had a bad streak.

deuszx, line 325: Is `UnderperfromedValidatorSessionCount` resetting its values after automatically? I.e. do we initialize a new instance of it every few sessions? I'm worried that if we don't then it will be too strict long-term. We should probably "reset" this counter if validator performed correctly in `n out of M` recent sessions. I.e. I don't want to keep being worried whether I get kicked out for missing a block b/c months ago I had a bad streak.

pallets/elections/src/impls.rs
```diff
@@ -293,6 +299,45 @@ where
...
+    }
+
+    fn mark_validator_underperformance(
+        thresholds: &CommitteeKickOutThresholds,
+        validator: &T::AccountId,
+    ) {
+        UnderperformedValidatorSessionCount::<T>::mutate(&validator, |count| {
+            *count += 1;
+        });
+        let counter = UnderperformedValidatorSessionCount::<T>::get(&validator);
```
mike1729, line 324: `mutate` returns the result of the closure passed as its argument, so you can save it in the line above.

pallets/elections/src/lib.rs
```diff
@@ -279,6 +371,20 @@ pub mod pallet {
 
             Ok(())
         }
+
+        fn kick_out_underperformed_non_reserved_validators() {
+            let to_be_kicked_validators =
+                ToBeKickedOutFromCommittee::<T>::iter_keys().collect::<BTreeSet<_>>();
```
mike1729, line 377: Log this list if non-empty.

pallets/elections/src/lib.rs
```diff
@@ -305,6 +418,8 @@ pub mod pallet {
         ///
         /// We calculate the supports for them for the sake of eras payouts.
         fn elect() -> Result<Supports<T::AccountId>, Self::Error> {
+            Self::kick_out_underperformed_non_reserved_validators();
```
kostekIV, line 421: could you add this to the comment about the order of things that happens (it is at the start of the file impls.rs)

pallets/elections/src/lib.rs
kostekIV: or/and emit an chain event

pallets/elections/src/lib.rs
```diff
@@ -2,11 +2,31 @@
...
-//! - Committee: Set of nodes that produce and finalize blocks in the session.
-//! - Validator: Node that can become a member of committee (or already is) via rotation.
-//! - ReservedValidators: Validators that are chosen to be in committee every single session.
+//! - committee ([`EraValidators`]): Set of nodes that produce and finalize blocks in the session.
+//! - validator: Node that can become a member of committee (or already is) via rotation.
+//! - `EraValidators::reserved`: immutable validators, ie they cannot be removed from that list.
+//! - `EraValidators::non_reserved`: validators that can be kicked out from that list
+//!
+//! # Kick out logic
+//! In case of insufficient validator's uptime, we need to make such validators are removed from
```
DamianStraszak, line 13: "we need to make such validators are removed " -> "we need to remove such validators"

pallets/elections/src/lib.rs
```diff
@@ -73,12 +92,23 @@ pub mod pallet {
...
     #[pallet::generate_deposit(pub(super) fn deposit_event)]
     pub enum Event<T: Config> {
+        /// Emitted via [`Pallet<T>::change_validators`]
         ChangeValidators(Vec<T::AccountId>, Vec<T::AccountId>, CommitteeSeats),
+
+        /// Emitted via [`Pallet<T>::set_kick_out_thresholds`]
+        SetCommitteeKickOutThresholds(CommitteeKickOutThresholds),
+
+        /// Emitted during elections when there are some underperforming validators
+        KickOutValidators(BTreeSet<T::AccountId>),
```
DamianStraszak, line 111: Hmm, why not `Vec` instead of `BTreeSet`?

pallets/elections/src/impls.rs
```diff
@@ -293,6 +299,45 @@ where
...
+        if counter >= thresholds.underperformed_session_count_threshold {
+            // ideally we'd include here information about how many sessions a validator missed,
+            // yet in no_std world String does not exist. Instead BoundedVec, we should use some
+            // encoded error struct
+            let reason: Vec<u8> = "Insufficient uptime".into();
```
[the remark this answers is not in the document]
DamianStraszak: This is possible, but we would need one variant to be `Custom(BoundedVec<...>)` anyway, because we sometimes want custom kicks by sudo. From the perspective of frontend it's better to just have strings here, to avoid complication...

pallets/elections/src/impls.rs
```diff
@@ -293,6 +300,45 @@ where
...
+
+    fn mark_validator_underperformance(
+        thresholds: &CommitteeKickOutThresholds,
+        validator: &T::AccountId,
+    ) {
+        UnderperformedValidatorSessionCount::<T>::mutate(&validator, |count| {
+            *count += 1;
+        });
+        let counter = UnderperformedValidatorSessionCount::<T>::get(&validator);
+        if counter >= thresholds.underperformed_session_count_threshold {
```
DamianStraszak, line 326: Is my interpretation correct that `UnderperformedValidatorSessionCount` is never reset?
This is quite harsh then :P

primitives/src/lib.rs
```diff
@@ -80,9 +88,31 @@ impl Default for CommitteeSeats {
     }
 }
 
+/// Configurable parameters for kick-out validator mechanism
+#[derive(Decode, Encode, TypeInfo, Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct CommitteeKickOutThresholds {
+    /// in a session, what is a limit for produced blocks under which the session counts as bad uptime
+    pub block_count_threshold: BlockCount,
```
DamianStraszak, line 96: I think this is not the ideal way of defining a bad session. It would be better to make this a minimum fraction (percentage) of performance that is acceptable. Like `70%` means you must have >=70% performance in a session.

Absolute block counts are problematic, because they are sensitive to committee size etc. For instance, imagine we set the threshold to 50, forget about it and then increase committee size to 20 nodes. In which case we would ban all nodes :/

primitives/src/lib.rs
```diff
@@ -80,9 +88,31 @@ impl Default for CommitteeSeats {
...
+    /// how many bad uptime sessions force validator to be removed from the committee
+    pub underperformed_session_count_threshold: SessionCount,
```
DamianStraszak, line 98: I know this is hard technically, but we should think how to do the counting "per period of time" and not "for ever". If the threshold is say `3` (a very reasonable number btw.), then a validator which missed 3 sessions in 2 years of time, will be kicked the same as if it would miss 3 sessions in one day. I don't think it's desired.

One solution I see is that we could have a sudo extrinsic `reset_underperformed_session_counts` that just zeros it for everyone, and maybe reset the thing every month or so. That's a bad solution though, looking for better ones :)

DamianStraszak (changes requested):

primitives/src/lib.rs
deuszx: Yes! That's what I meant about "not resetting"...

> That's a bad solution though, looking for better ones :)

Why not fix to `3` like you said. If Substrate runtime updates are as good as people say, that value could get changed with a `sudo`/update later on?

pallets/elections/src/impls.rs
Marcin-Radecki: good point, done

pallets/elections/src/lib.rs
Marcin-Radecki: done

bin/runtime/src/lib.rs
```diff
@@ -107,7 +107,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("aleph-node"),
     impl_name: create_runtime_str!("aleph-node"),
     authoring_version: 1,
-    spec_version: 34,
+    spec_version: 35,
```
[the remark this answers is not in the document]
Marcin-Radecki: agree. but minor's version is better choice - we're adding new feature anyway

pallets/elections/src/impls.rs
```diff
@@ -293,6 +299,45 @@ where
 
         T::ValidatorRewardsHandler::add_rewards(rewards);
     }
+
+    fn calculate_underperforming_validators() {
```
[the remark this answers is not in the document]
Marcin-Radecki: done

pallets/elections/src/impls.rs
```diff
@@ -293,6 +299,45 @@ where
...
+        let thresholds = CurrentEraCommitteeKickOutThresholds::<T>::get();
+        let current_committee = T::SessionInfoProvider::current_committee();
+        for validator in current_committee {
+            let underperformance = match SessionValidatorBlockCount::<T>::try_get(&validator) {
+                Ok(block_count) => block_count <= thresholds.block_count_threshold,
+                Err(_) => true,
+            };
+            if underperformance {
+                Self::mark_validator_underperformance(&thresholds, &validator)
```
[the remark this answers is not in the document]
Marcin-Radecki: done

pallets/elections/src/impls.rs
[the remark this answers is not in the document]
Marcin-Radecki: Actually I merged both remarks - now `KickOutReason` is an `enum`

pallets/elections/src/lib.rs
Marcin-Radecki: done

primitives/src/lib.rs
Marcin-Radecki: On the other hand, having 0 as the block limit makes not harsh at all in my opinion, and this is because making 0 blocks in a session really means in practice two situations A) keys are misconfigured or B) the validator process is down. In both cases, it's on the validator to fix those, so having such 3 sessions, which in practice in the current committee size means a whole era, it's not big deal IMHO. Agree that setting to a non-zero block threshold can be harsh since it implies minor network issues which could make the validator to miss some slots and hence count the session as bad performance. 

My proposal is then: for the initial version, which was tested already, let's make block count hardcoded to 0 (ie, not configurable), and keep session count configurable as it is now and write 3 here as the default. Also for the initial version, don't add any resetting `underperformed_session_count_threshold` for now since we need to figure out a better way to do this, and we don't want to block this PR on this. If some validator would be kicked out by 3 sessions, which I doubt it would be **not** due to the validator's fault, we can always add it back in the next era.

primitives/src/lib.rs
Marcin-Radecki: See my proposal below

pallets/elections/src/impls.rs
Marcin-Radecki: See my proposal below

pallets/elections/src/lib.rs
Marcin-Radecki: This was due to the fact in the code where this is emitted I have already `BTreeSet` and I want to avoid another O(N) iteration. On the UI side, it looks very similar, let me get the screenshot.

pallets/elections/src/lib.rs
Marcin-Radecki: Ok let me fix this

pallets/elections/src/impls.rs
Marcin-Radecki: See my proposal below to @DamianStraszak 's post.

bin/runtime/src/lib.rs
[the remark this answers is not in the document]
DamianStraszak: You should also bump transaction_version

pallets/elections/src/lib.rs
DamianStraszak: Ok, fine, let's leave it then.

Marcin-Radecki: image

primitives/src/lib.rs
deuszx: > Fix what to 3 exactly :P?

Width of the sliding window over the last 	`N` (=3) sessions which we would take into consideration when deciding whether validator underperformed. i.e. it would reset automatically after `N`.

DamianStraszak: Yes. This automatic reset is complicated because you need to add items to storage for that. The more storage you add, the less efficient it is, and also the more complex and the harder to change in the future. We found a workaround to make it stateless by resetting if session number `=0 (mod N)` -- this has some disadvantages, but keeps it simple and does not require additional storage.

pallets/elections/src/impls.rs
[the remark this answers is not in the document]
kostekIV: cool, leaving this conversation for @DamianStraszak to resolve :)

pallets/elections/src/impls.rs
```diff
@@ -21,6 +22,7 @@ pub const LENIENT_THRESHOLD: Perquintill = Perquintill::from_percent(90);
 ///
 /// 1. Block `B` initialized
 /// 2. `end_session(S)` is called
+/// -  Based on block count we might mark the session for a given validator as undeperformed
```
kostekIV, line 25: could you also add a point when the kick might happen

Marcin-Radecki: will do

pallets/elections/src/impls.rs
Marcin-Radecki: Actually, now done

bin/runtime/src/lib.rs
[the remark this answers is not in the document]
Marcin-Radecki: done

pallets/elections/src/lib.rs
Marcin-Radecki: done

primitives/src/lib.rs
Marcin-Radecki: done

pallets/elections/src/impls.rs
Marcin-Radecki: done

pallets/elections/src/lib.rs
```diff
@@ -202,13 +248,87 @@ pub mod pallet {
...
+
+            NextEraCommitteeKickOutConfig::<T>::put(current_committee_kick_out_config.clone());
+            Self::deposit_event(Event::SetCommitteeKickOutConfig(
+                current_committee_kick_out_config,
+            ));
+
+            Ok(())
+        }
+
+        /// Remove node from the committee.
```
DamianStraszak, line 299: The comment could be more precise, for instance: "Schedule a non-reserved node to be kicked out from the committee at the end of the era."

pallets/elections/src/lib.rs
```diff
@@ -305,6 +463,8 @@ pub mod pallet {
         ///
         /// We calculate the supports for them for the sake of eras payouts.
         fn elect() -> Result<Supports<T::AccountId>, Self::Error> {
+            Self::kick_out_underperformed_non_reserved_validators();
```
DamianStraszak, line 466: It would be better if we run it outside of `elect()` but with the guarantee that it is run before `elect()`. As you mentioned -- `elect()` currently does not mutate the state, would be great if it stayed that way...

pallets/elections/src/lib.rs
```diff
@@ -130,44 +160,60 @@ pub mod pallet {
...
 
+    /// Default value for kick out config, see [`CurrentEraCommitteeKickOutConfig`]
+    #[pallet::type_value]
+    pub fn DefaultCommitteeKickOutConfig<T: Config>() -> CommitteeKickOutConfig {
+        CommitteeKickOutConfig::default()
+    }
+
+    /// Current era config for kick-out functionality, see [`CommitteeKickOutConfig`]
+    #[pallet::storage]
+    pub type CurrentEraCommitteeKickOutConfig<T> =
```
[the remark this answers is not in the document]
Marcin-Radecki: This is because we follow a convention in pallet elections so and also in pallet staking so that each user action performed in the current era (such as an intent to become a validator or adding a new node to non-reserved for example) actually applies in the next era. We could have made a single `CommitteeKickOutConfig` but then we'll lose that functionality.

pallets/elections/src/lib.rs
Marcin-Radecki: Sure, done. The reason I've kept it concise is that there's a limit of characters next to the extrinsic description in UI, but this time it fits. Otherwise, it cuts the sentence displaying this  `...` which might be annoying to the user.  

image

pallets/elections/src/lib.rs
Marcin-Radecki: The reason it was implemented like this is that we need to do it once per era in a way that supports and targets calculated in `elect()` including kick outs. 

Reasons why it cannot be done in `start_session`:
* we cannot do it after `SessionManager::new_session` is called since this is just _after_ `elect()` is done. 
* we cannot do it before `SessionManager::new_session` since `pallet_staking::ErasStartSessionIndex` is not yet populated 
with the next era info - `SessionManager::new_session` does it
Then it gets more error-prone to get it done in other functions such as `end_session` or `start_session` since we're away (in time) from elections. Also performance-wise we're not bloating the election block as its complexity is `O(non_reserved_validators.len())`.

 With all that above, I think for this version it's fine to leave it as it is, not sure if spending more time is worth results.

pallets/elections/src/lib.rs
```diff
@@ -279,6 +402,28 @@ pub mod pallet {
...
+            if !to_be_kicked_validators.is_empty() {
+                let non_reserved_validators = NextEraNonReservedValidators::<T>::get()
+                    .into_iter()
+                    .collect::<BTreeSet<_>>();
+                let non_reserved_validators_len = non_reserved_validators.len();
+                let non_reserved_in_order = non_reserved_validators
+                    .into_iter()
+                    .filter(|v| !to_be_kicked_validators.contains(v))
+                    .collect::<Vec<_>>();
+                if non_reserved_in_order.len() < non_reserved_validators_len {
```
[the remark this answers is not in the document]
Marcin-Radecki: In this way, as it is coded I think it's correct because when we add someone to `ToBeKickedOutFromCommittee` (either manually in `kick_out_from_committee` tx or via automatic logic) we check whether node is in _next_ era non_reserved node set. The version you suggest was actually my first choice indeed, but then I've come to the conclusion there's `reserved` nodes that might eventually get into `ToBeKickedOutFromCommittee`, yet they won't be kicked actually and they will "magically" disappear from that list with no reason, which might be confusing to the user on UI side (we're going to display contents of `ToBeKickedOutFromCommittee`. 

I'll try to include information about the reason in the event though which is a good suggestion.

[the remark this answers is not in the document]
Marcin-Radecki: But you're right it's not clear with the code for the reader - I don't need to check therefore intersection of `to_be_kicked_validators` and next era non_reserved in kick_out_underperformed_non_reserved_validators, let me simplify that.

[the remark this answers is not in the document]
Marcin-Radecki: Ah no, I need to do that check as well. This is because there's a corner case scenario (not that unlikely though) in which someone modifies next era non_reserved via `change_validators`, so that for example remove validator `V` from that list which is also in `ToBeKickedOutFromCommittee`. Without a current check, we'll emit a false positive event that we're kicking `V` from the committee.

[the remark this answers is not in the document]
Marcin-Radecki: I've included reason in the event

pallets/elections/src/lib.rs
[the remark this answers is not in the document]
DamianStraszak: This "convention" makes sense only for very specific cases. We must use it for consensus-critical data that we query in the host one session ahead, such as session authorities or emergency finalizer. Otherwise it is not necessary. Here it's just a config that is not queried anywhere in consensus code.

pallets/elections/src/impls.rs
```diff
@@ -293,6 +304,58 @@ where
...
+    }
+
+    fn clear_underperformance_session_counter(session: SessionIndex) {
+        let clean_session_counter_delay =
+            CurrentEraCommitteeKickOutConfig::<T>::get().clean_session_counter_delay;
+        if session % clean_session_counter_delay == 0 {
+            info!(target: "pallet_elections", "Clearing UnderperformedValidatorSessionCount");
+            let _result = UnderperformedValidatorSessionCount::<T>::clear(u32::MAX, None);
+        }
+    }
```
deuszx, line 358: Shouldn't this be set _per-validator_ ? Unless this is done this way on purpose.

pallets/elections/src/impls.rs
```diff
@@ -293,6 +304,58 @@ where
...
+        let expected_blocks_per_validator = Self::blocks_to_produce_per_session();
+        for validator in non_reserved_nodes_in_next_era {
+            let underperformance = match SessionValidatorBlockCount::<T>::try_get(&validator) {
+                Ok(block_count) => {
+                    let block_performance_ratio = match block_count <= expected_blocks_per_validator
+                    {
+                        true => Perbill::from_rational(block_count, expected_blocks_per_validator),
+                        false => Perbill::one(),
+                    };
+                    block_performance_ratio <= thresholds.minimal_expected_performance
```
deuszx, line 324: Isn't this equivalent to simply:
```rust
Perbill::from_rational(block_count, expected_block_per_validator) <= thresholds.minimal_expected_performance
```
If I'm reading [the code](https://github.com/paritytech/substrate/blob/e2fc0430ab3bcd644cded0353d4514611256ad3c/core/sr-primitives/src/lib.rs#L201) correctly, `n / m` (where n > m) will simplify to `Perbill::one()`.

Unless, again, this is on purpose to improve performance.

pallets/elections/src/impls.rs
Marcin-Radecki: Do you mean that each validator needs to have its specific counter? Can be, but it requires additional `StorageMap`, so I don't think it is worth it given profits. This is too much corner case anyway imho, since we set this `clean_session_counter_delay` to quite long value anyway (960 sessions == 10 eras)

deuszx: Yes, I imagine validators will underperform with different "schedules" and some might be just on the verge of being kicked out but they luck out and the counter gets cleared, starts over and they continue to underperfrom without any consequences. In the worst case scenario, a validator would underperform 959+960 sessions in a row.

If this is done purposefully then that's fine, I can understand the explanation that it's overengineered without much practical benefit.

pallets/elections/src/impls.rs
Marcin-Radecki: Good point, let me change this

pallets/elections/src/impls.rs
Marcin-Radecki: I think the main point here is additional storage - rule of thumb is, we should not be adding more storage unless having strong reason to do so

pallets/elections/src/lib.rs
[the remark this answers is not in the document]
Marcin-Radecki: Discussed offline, I'll change this to not check non reserved nodes when added to kick out list

pallets/elections/src/lib.rs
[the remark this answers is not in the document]
Marcin-Radecki: Discussed offline, I'll rework to have only one instance

DamianStraszak (approved):
Looks awesome now. 

Some e2e test would be great to have though :)

pallets/elections/src/impls.rs
DamianStraszak: Another benefit of this being stateless is that changing the config does not require any changes to the counters, and thus it's overall simpler. In general, whenever you add some storage, then you make it more complicated to change the logic in the future, because you might then need to write migrations etc.

pallets/elections/src/impls.rs
Marcin-Radecki: It's done

Marcin-Radecki:
> Looks awesome now.
> 
> Some e2e test would be great to have though :)

List of e2e tests to be done: https://cardinal-cryptography.atlassian.net/jira/software/projects/A0/boards/1?assignee=603e52ed333ff40070d045e6&selectedIssue=A0-1432

PR #640, opened 2022-10-04
Sacrifice build time for execution speed

mike1729:
# Description

Adds `production` profile to `Cargo.toml` that enables lto [1] and disables parallel compilation in LLVM [2]. Those changes trade compilation time for execution speed.

Bibliography
1. https://doc.rust-lang.org/rustc/linker-plugin-lto.html
2. https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units

## Type of change

# Checklist:


.github/workflows/nightly-pipeline.yaml
```diff
@@ -83,28 +83,28 @@ jobs:
...
           target-key: e2e
           cargo-key: e2e
           cache-version: v1
           cargo-targets: e2e-tests/target/
-      
+
       - name: Build binary and docker image
         run: |
           cd e2e-tests/
-          cargo build --release
```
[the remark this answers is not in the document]
mike1729: That was my initial idea but @bartoszjedrzejewski insited on having uniform builds across our pipelines so that if there is a bug due to the production profile we would catch it here. However, I would prefer faster e2e pipeline since I find it very improbable to bump into such a bug.

Cargo.toml
```diff
@@ -19,3 +19,12 @@ exclude = [
...
 ]
+
+[profile.release]
+# Substrate runtime requires unwinding.
+panic = "unwind"
+
+[profile.production]
+inherits = "release"
+lto = true
+codegen-units = 1
```
Marcin-Radecki, line 30: On a quick look, I could not find in offical LLVM docs https://llvm.org/docs/CodeGenerator.html that indeed setting this to one can improve performance. Shall we back up on it till next task, unless some benches have been made that prove performance improvement with that setting enabled?

.github/workflows/nightly-pipeline.yaml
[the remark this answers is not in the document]
Marcin-Radecki: I think integration tests (ie e2e tests) must be compiled using the same build settings as the production build, from the reason @mike1729 stated above.

Cargo.toml
Marcin-Radecki: E.g. this says it does not help much https://github.com/paritytech/polkadot/issues/4311#issuecomment-972403714

Cargo.toml
```diff
@@ -19,3 +19,12 @@ exclude = [
...
   "bin/cliain",
 ]
+
+[profile.release]
+# Substrate runtime requires unwinding.
+panic = "unwind"
+
+[profile.production]
+inherits = "release"
+lto = true
```
Marcin-Radecki, line 29: We should explore if `thin` LTO gives us similar performances at some point http://blog.llvm.org/2016/06/thinlto-scalable-and-incremental-lto.html , as some sources https://blog.rust-lang.org/inside-rust/2020/06/29/lto-improvements.html say it can have similar perf to fat LTO

Cargo.toml
```diff
@@ -19,3 +19,12 @@ exclude = [
   "benches/payout-stakers",
   "bin/cliain",
 ]
+
+[profile.release]
+# Substrate runtime requires unwinding.
+panic = "unwind"
```
Marcin-Radecki, line 25: If we go with the solution above, we don't need to specify this setting explicitly, as it will be present in the default `release` profile

Marcin-Radecki (changes requested):
While enabling LTO is a project is generally non-controversial, as it is a mature enough feature that is used in LLVM for quite some time (not only by Rust, but also before 2015 by `clang` in C++), I think enabling that is fine in this PR. However, to achieve full effect of such change has to our system, more work is required:
* most importantly, we need some figures before and after this change is introduced to compare what the speedup really is, as all those resources about LTO talk about theoretical speedup. Ideally, we should have some metrics that can measure that, e.g. CPU load would be the first one comes to my mind. Feature envs can help us measure so. If we don't rush with this PR, I'd say it's essential to be done before it's merged.
* Having such a number then can be advertised on socials which makes PR effect, but we need to measure it first as point above states very carefully, it's easy to claim a lot of speedup on the first look  
* Then we should update this repo's README.md that we're using LTO, and docs for external validators that they should build their node, if there are any `cargo build --release` steps baked there

.github/workflows/nightly-pipeline.yaml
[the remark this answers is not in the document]
Marcin-Radecki: Ah but this build e2e client, not `aleph-node` test binary - in that case we can even (and I think now better) leave with non `--release`  , which is called `dev`: https://doc.rust-lang.org/cargo/reference/profiles.html#dev

Cargo.toml
Marcin-Radecki: After discussing offline, no harm to include this

Cargo.toml
Marcin-Radecki: Binary size can be some metric here

Cargo.toml
Marcin-Radecki: I was wrong , this just extends profile

.github/workflows/nightly-pipeline.yaml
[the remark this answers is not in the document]
mike1729: Changed to dev @fbielejec

Marcin-Radecki (approved):

PR #641, opened 2022-10-04
Basic parts of the validator network

timorl:
# Description

First of about 5 PRs for the validator network. Includes some basic mocks for testing and how we will send data. The other parts are already prepared, we just didn't want to drop them all in one PR. Obviously nothing connected yet.

## Type of change

# Checklist:

- I have added tests


finality-aleph/src/validator_network/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// Send a message to a single peer.
+    /// This function should be implemented in a non-blocking manner.
+    fn send(&self, data: D, recipient: AuthorityId);
+
+    /// Receive a message from the network.
+    async fn next(&mut self) -> Option<D>;
+}
+
+/// A stream that can be split into a sending and receiving part.
+pub trait Splittable: AsyncWrite + AsyncRead + Unpin + Send {
```
fixxxedpoint, line 41: Do we need this? What's wrong with [split](https://docs.rs/tokio/latest/tokio/io/fn.split.html) since we are just wrapping tokio's types?

finality-aleph/src/validator_network/io.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    #[tokio::test]
+    async fn fails_to_decode_empty_data() {
+        let (mut sender, receiver) = duplex(4096);
+        let payload = 0u32.to_le_bytes().to_vec();
+        sender
+            .write_all(&payload)
+            .await
+            .expect("sending should work");
+        match receive_data::<_, i32>(receiver).await {
```
DamianStraszak, line 171: Are you absolutely sure there are no valid objects that encode to an empty byte vec? Because this kind of assumes it is the case. I imagine that an empty array could probably encode to `vec![]`.

fixxxedpoint (commented):
left few comments

finality-aleph/src/validator_network/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+pub trait Data: Clone + Codec + Send + Sync + 'static {}
+
+impl<D: Clone + Codec + Send + Sync + 'static> Data for D {}
+
+/// Network represents an interface for opening and closing connections with other Validators,
+/// and sending direct messages between them.
+///
+/// Note on Network reliability and security: it is neither assumed that the sent messages must be
+/// always delivered, nor the established connections must be secure in any way. The Network
+/// implementation might fail to deliver any specific message, so messages have to be resend while
```
DamianStraszak, line 22: resend -> resent

finality-aleph/src/validator_network/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    /// Remove the peer from the set of connected peers and close the connection.
+    fn remove_connection(&mut self, peer: AuthorityId);
+
+    /// Send a message to a single peer.
+    /// This function should be implemented in a non-blocking manner.
+    fn send(&self, data: D, recipient: AuthorityId);
+
+    /// Receive a message from the network.
+    async fn next(&mut self) -> Option<D>;
```
DamianStraszak, line 37: Why doesn't it output the sender as well?

DamianStraszak (changes requested):
First round of comments.

finality-aleph/src/validator_network/io.rs
timorl: Yup, but in this case we are explicitly using `i32` as the target type, and this does not encode into empty byte vecs.

finality-aleph/src/validator_network/mod.rs
timorl: We don't need it, if it turns out we actually need it for something in the future we can add it then.

finality-aleph/src/validator_network/io.rs
```diff
@@ -0,0 +1,176 @@
...
+    mut stream: S,
+    data: D,
+) -> Result<S, SendError> {
+    let mut encoded = data.encode();
+    let len = u32::try_from(encoded.len()).map_err(|_| Error::DataTooLong(u32::MAX))?;
+    if len > MAX_DATA_SIZE {
+        return Err(Error::DataTooLong(len).into());
+    }
+    let mut payload = len.to_le_bytes().to_vec();
+    payload.append(&mut encoded);
```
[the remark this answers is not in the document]
timorl: We explicitly wanted this to work on `Data`, because keeping typing information everywhere is convenient.

How would we avoid allocations if this was declared for a type? Types often don't have constant encoding length.

finality-aleph/src/validator_network/io.rs
```diff
@@ -0,0 +1,176 @@
...
+        .map_err(|_| Error::ConnectionClosed)?;
+    let len = u32::from_le_bytes(buf);
+    if len > MAX_DATA_SIZE {
+        return Err(Error::DataTooLong(len).into());
+    }
+    let mut buf: Vec<u8> = vec![0; len as usize];
+    stream
+        .read_exact(&mut buf[..])
+        .await
+        .map_err(|_| Error::ConnectionClosed)?;
```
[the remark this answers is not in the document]
timorl: What do you mean?

finality-aleph/src/validator_network/io.rs
```diff
@@ -0,0 +1,176 @@
...
+        sender
+            .write_all(&payload)
+            .await
+            .expect("sending should work");
+        match receive_data::<_, i32>(receiver).await {
+            Err(e) => assert_eq!(ReceiveError::DataCorrupted, e),
+            _ => panic!("decoded no data into something?!"),
+        }
+    }
+}
```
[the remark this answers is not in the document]
timorl: We fail early, so if we get bad data once we drop the connection and thus the latter situation doesn't need to be tested.

The former I did not want to test as it requires allocating 16MB of memory for no reason. :< Should we?

finality-aleph/src/validator_network/io.rs
DamianStraszak: Oh, right, I got confused, I think it is correct. Will resolve in a minute after I convince myself :P

finality-aleph/src/validator_network/mod.rs
timorl: Fixed.

finality-aleph/src/validator_network/io.rs
```diff
@@ -0,0 +1,176 @@
...
+    let len = u32::from_le_bytes(buf);
+    if len > MAX_DATA_SIZE {
+        return Err(Error::DataTooLong(len).into());
+    }
+    let mut buf: Vec<u8> = vec![0; len as usize];
+    stream
+        .read_exact(&mut buf[..])
+        .await
+        .map_err(|_| Error::ConnectionClosed)?;
+    let data = D::decode(&mut &buf[..]).map_err(|_| ReceiveError::DataCorrupted)?;
```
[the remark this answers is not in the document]
timorl: Yup, very good point, fixed.

finality-aleph/src/validator_network/mod.rs
DamianStraszak: Yeah, we don't need it because the API of the AlephBFT network is suboptimal (does not require the sender). And it is suboptimal because we initially assumed the information about the sender will be hard to get. It is actually easy to get, and here we are throwing it away for free.

What I'm proposing is to break the cycle, and actually start improving the network by making the right choices here.

DamianStraszak: Ok fine. The only remaining thing I'm worried about currently is that this will hinder the ability to log suitably precise diagnostic information. Feel free to resolve if you think it should not be a problem.

timorl: Logging will mostly happen on a different level which will have access to peer information anyway.

finality-aleph/src/validator_network/io.rs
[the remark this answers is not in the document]
fixxxedpoint: It looks like it masks all possible network errors by mapping them to `ConnectionClosed`. It looks like one of first spots for debugging in case there are some connectivity problems.

finality-aleph/src/validator_network/io.rs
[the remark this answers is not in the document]
fixxxedpoint: Making low level network code depend on so many (unnecessary from network's perspective) traits from `Data` doesn't look very convenient. If that would be a method for some type, then it could have some internal buffer (potentially even shared between connections). The  `Codec` trait was invented partially around idea of limiting number of necessary allocations (writing to `Output` buffer instead of transforming into Vec<u8>). In this approach it could be simply switched with `From<[u8]>` and `Into<[u8]>'.

finality-aleph/src/validator_network/io.rs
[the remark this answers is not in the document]
fixxxedpoint: It's for free :) And writing such test can also provide some improvements, i.e. you can consider calling `size_hint` first instead of trying to allocate potentially too big message first. It would also justify usage of `codec` here.

finality-aleph/src/validator_network/mod.rs
fixxxedpoint: I am a bit confused by your answer. [split](https://docs.rs/tokio/latest/tokio/io/fn.split.html) is implemented for anything that implements `AsyncWrite + AsyncRead` which is a superset of types required by this trait. Are this `Unpin` and `Send` also necessary - they just make it harder to implement and can be moved to clients of that api (if the need them at all).

DamianStraszak (approved):
It's good to go, although I generally agree with @fixxxedpoint regarding efficiency -- Codec is a highly optimized library that avoids allocation, and we don't use it here. 

HOWEVER, since we aim for something that works ASAP, I think it's good.

finality-aleph/src/validator_network/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+
+/// Network represents an interface for opening and closing connections with other Validators,
+/// and sending direct messages between them.
+///
+/// Note on Network reliability and security: it is neither assumed that the sent messages must be
+/// always delivered, nor the established connections must be secure in any way. The Network
+/// implementation might fail to deliver any specific message, so messages have to be resent while
+/// they still should be delivered.
+#[async_trait::async_trait]
+pub trait Network<A: Data, D: Data>: Send {
```
fixxxedpoint, line 25: Can we move those `D` and `A` parameters on per method basis. A single instance of network is able to send anything fullfiling `D`. It can simplify things like protocol versioning - you can just start sending new type of messages, without the need to wrap your types, or some other hackery.

finality-aleph/src/validator_network/io.rs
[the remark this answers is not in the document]
timorl: Welp, good point, for some reason I thought the underlying errors would be generic to `io`, but `io::Error` allows for custom errors, so they are probably informative. Changed. Annoyingly now we do not have `Eq` anymore, but that's a small price to pay.

finality-aleph/src/validator_network/io.rs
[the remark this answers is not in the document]
timorl: Added test, but I did not do the thing with `size_hint` – it would result in strangely repeated code, and allocating a big vector for encoding implies we already allocated a lot of space for the message itself, plus it will never happen in practice (if we actually cannot send a message we created we are in far bigger trouble).

finality-aleph/src/validator_network/mod.rs
timorl: Unfortunately that is not the case, this interface will be implemented as a set of channels that pass the messages and other calls to a running service. These channels have to be parametrized by `D`. :/

fixxxedpoint (approved):
lgtm

PR #645, opened 2022-10-07
add score

fbielejec:
# Description

add score to the event, needed for indexing purposes

### DIFF
--- a/contracts/button/lib.rs
+++ b/contracts/button/lib.rs
@@ -47,6 +47,7 @@ mod button_game {
         #[ink(topic)]
         by: AccountId,
         when: BlockNumber,
+        score: Balance,
     }
 
     /// Event emitted when the finished game is reset and pressiah is rewarded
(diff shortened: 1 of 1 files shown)


contracts/button/lib.rs
```diff
@@ -47,6 +47,7 @@ mod button_game {
         #[ink(topic)]
         by: AccountId,
         when: BlockNumber,
+        score: Balance,
```
pmikolajczyk41, line 50: should we make it `#[ink(topic)]`?

pmikolajczyk41 (approved):

contracts/button/lib.rs
fbielejec: nah I doubt it's interesting to index by scores, they will be all over the place no?

h4nsu (approved):

contracts/button/lib.rs
pmikolajczyk41: I was misled by 'needed for indexing purposes' :P

PR #646, opened 2022-10-07
A0-1408: Protocol negotiation and connection handling

timorl:
# Description

PR 3/5 for the new validator network. This one adds protocol negotiation and full connection handling once the connections have been established as streams.

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/finality-aleph/src/validator_network/protocol_negotiation.rs
+++ b/finality-aleph/src/validator_network/protocol_negotiation.rs
@@ -1,0 +1,267 @@
+use std::{
+    cmp::{max, min},
+    fmt::{Display, Error as FmtError, Formatter},
+};
+
+use tokio::{
+    io::{AsyncReadExt, AsyncWriteExt},
+    time::{timeout, Duration},
+};
+
+use crate::validator_network::protocols::Protocol;
+
+pub type ProtocolVersion = u32;
+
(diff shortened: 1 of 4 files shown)


finality-aleph/src/validator_network/protocol_negotiation.rs
```diff
@@ -0,0 +1,267 @@
...
+        for _ in 0..2 {
+            tokio::select! {
+                result = &mut negotiation1 => incorrect_negotiation(result, ProtocolNegotiationError::InvalidRange(invalid_range.clone())),
+                result = &mut negotiation2 => incorrect_negotiation(result, ProtocolNegotiationError::InvalidRange(invalid_range.clone())),
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn fails_when_connection_droped() {
```
DamianStraszak, line 260: droped -> dropped

DamianStraszak (approved):
Looks good, although I will be able to say more on correctness after I see the full code.

The logging here seems sufficient, because this is still per-peer code. But at some layer we must print aggregated logs from all peers.

finality-aleph/src/validator_network/protocol_negotiation.rs
timorl: Fixed, plus added a small trait requirement which we will actually need later.

maciejnems (approved):
LGTM

PR #650, opened 2022-10-10
A0-1408: Connection manager and service

timorl:
# Description

PR 4/5 in the validator network. Connection manager and service that has to be run to have the network actually do net work.

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/finality-aleph/src/validator_network/manager.rs
+++ b/finality-aleph/src/validator_network/manager.rs
@@ -94,10 +94,12 @@ impl<A: Data, D: Data> Manager<A, D> {
 
     /// Add an established incoming connection with a known peer,
     /// but only if the peer is on the list of peers that we want to stay connected with.
-    pub fn add_incoming(&mut self, peer_id: AuthorityId, exit: oneshot::Sender<()>) {
-        if self.addresses.contains_key(&peer_id) {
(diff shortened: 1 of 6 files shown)


finality-aleph/src/validator_network/manager.rs
```diff
@@ -0,0 +1,237 @@
...
+        if self.addresses.contains_key(&peer_id) {
+            self.outgoing.insert(peer_id, data_for_network);
+        };
+    }
+
+    /// Add an established incoming connection with a known peer,
+    /// but only if the peer is on the list of peers that we want to stay connected with.
+    pub fn add_incoming(&mut self, peer_id: AuthorityId, exit: oneshot::Sender<()>) {
+        if self.addresses.contains_key(&peer_id) {
+            self.incoming.insert(peer_id, exit);
```
DamianStraszak, line 99: I think it would be good to log something in case we overwrite an incoming connection -- it is something unexpected that we can miss.

DamianStraszak (approved):
GOOD.

maciejnems (approved):
Very Nice

finality-aleph/src/validator_network/manager.rs
timorl: Added.

PR #651, opened 2022-10-10
PeerId is no longer Copy

timorl:
# Description

Remove the requirement that `PeerId` is `Copy`. It's not strictly needed and we will soon be using ones that are not.

## Type of change

????

# Checklist:

?!?


maciejnems (approved):
Very nice and important change <3

DamianStraszak (approved):

PR #653, opened 2022-10-10
A0-1410: add mock implementation of validator network

maciejnems:
# Description
This PR introduces mock implementation of validator network. This is the first PR in series of integrating validator network with aleph node.

## Type of change

- Mock implementation

### DIFF
--- a/finality-aleph/src/testing/mocks/validator_network.rs
+++ b/finality-aleph/src/testing/mocks/validator_network.rs
@@ -1,4 +1,3 @@
-#![allow(dead_code)]
 use std::sync::Arc;
 
 use aleph_primitives::{AuthorityId, KEY_TYPE};
@@ -68,7 +67,7 @@ impl<D: Data> NetworkIdentity for MockNetwork<D> {
 }
 
 impl<D: Data> MockNetwork<D> {
-    pub async fn new(address: &str) -> Self {
+    pub async fn _new(address: &str) -> Self {
         let key_store = Arc::new(KeyStore::new());
         let id: AuthorityId = key_store
             .ed25519_generate_new(KEY_TYPE, None)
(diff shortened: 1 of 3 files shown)


finality-aleph/src/testing/mocks/validator_network.rs
```diff
@@ -0,0 +1,96 @@
...
+
+use aleph_primitives::{AuthorityId, KEY_TYPE};
+use sp_keystore::{testing::KeyStore, CryptoStore};
+
+use crate::{
+    network::{mock::Channel, Data, Multiaddress, NetworkIdentity, PeerId},
+    validator_network::Network,
+};
+
+type MockPeerId = AuthorityId;
```
kostekIV, line 12: if possible use some primitive type like u32,string or smth like this

timorl, line 12: I think you should use unaliased `AuthorityId` everywhere here – it is inherently connected to `validator_network`'s interfaces anyway.

kostekIV (approved):
lgtm

timorl (approved):
Just one nit that I'm not even that sure about.

finality-aleph/src/testing/mocks/validator_network.rs
kostekIV: not blocking for this pr, but wouldn't we like this new network part to be not dependent on any extern types (Especially ones from substrate)? @timorl

timorl: That might be nice, but it would require lots of traits for keys which we don't yet have, and writing them would take a lot of time. We probably should have a parkinglot task for improving that.

timorl: Added as 1454. @lesniak43 since you were also a proponent of that iirc and probably want to know.

PR #655, opened 2022-10-11
Initial benchmarking

pmikolajczyk41:
This PR adds some code to actually benchmark pallet snarcos. 

These benchmarks are simplified to the only 2 relations that we already have (from HouseSnark repo). In terms of proper weight calculation, there is still some work to do (like including proof and input size to the costs).

I have problems with running these benches with `--execution=wasm` (https://substrate.stackexchange.com/questions/5287/cannot-run-benchmarking-did-not-find-the-benchmarking-metadata), so it's possible that the code in `bin/runtime|node` should be somehow adjusted.

Further work on this matter should definitely include testing those benchmarks. However, it's probably non-trivial (combining some macros and configuration).


pallets/snarcos/src/benchmarking/suite.rs
```diff
@@ -0,0 +1,51 @@
...
+benchmarks! {
+
+    store_key {
+        let caller = account("caller", 0, SEED);
+        let identifier = [0u8; 4];
+        let l in 1 .. T::MaximumVerificationKeyLength::get();
+        let key = vec![0u8; l as usize];
+    } : _(RawOrigin::Signed(caller), identifier, key.clone())
+
+    verify_xor {
```
fbielejec, line 21: these two puppies below  are not really "benchmarked" but after the call I understand that the PR is just for building some benchmarks scaffolding. What I could suggest is pre-generating  keys of various sizes (say exponentially  increasing or in a fibonacci series) and running benchmarks for those as inputs?

bin/runtime/src/lib.rs
```diff
@@ -966,6 +976,51 @@ impl_runtime_apis! {
...
+    impl frame_benchmarking::Benchmark<Block> for Runtime {
+        fn benchmark_metadata(extra: bool) -> (
+            Vec<frame_benchmarking::BenchmarkList>,
+            Vec<frame_support::traits::StorageInfo>,
+        ) {
+            use frame_benchmarking::{Benchmarking, BenchmarkList};
+            use frame_support::traits::StorageInfoTrait;
+
+            let mut list = Vec::<BenchmarkList>::new();
+            list_benchmarks!(list, extra);
```
fbielejec, line 990: can you pass just a vec![] here, since the list is empty anway?

bin/node/src/main.rs
```diff
@@ -105,6 +107,23 @@ fn main() -> sc_cli::Result<()> {
         Some(Subcommand::TryRuntime) => Err("TryRuntime wasn't enabled when building the node. \
         You can enable it with `--features try-runtime`."
             .into()),
+        #[cfg(feature = "runtime-benchmarks")]
```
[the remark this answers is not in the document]
pmikolajczyk41: that's right, but I would prefer to leave this as it is, because:
 - `Subcommand::TryRuntime` is handled in the same way already
 - checking `cfg!(feature = "runtime-benchmarks")` inside increases nesting
 - when `runtime-benchmarks` is off, then `--help` doesn't display unnecessary arguments (of `BenchmarkCmd`)

pallets/snarcos/src/benchmarking/suite.rs
pmikolajczyk41: I agree, but I would suggest postponing it to the next benchmarking PR, when we have https://github.com/Cardinal-Cryptography/HouseSnark/pull/4 merged; then I will be able to benchmark against more relations, proof systems and key sizes. WDYT?

bin/runtime/src/lib.rs
pmikolajczyk41: no, unfortunately the macro is expecting this way :P

bin/runtime/src/lib.rs
```diff
@@ -966,6 +976,51 @@ impl_runtime_apis! {
...
+                hex_literal::hex!("26aa394eea5630e07c48ae0c9558cef7ff553b5a9862a516939d82b3d3d8661a").to_vec().into(),
+                // Event Count
+                hex_literal::hex!("26aa394eea5630e07c48ae0c9558cef70a98fdbe9ce6c55837576c60c7af3850").to_vec().into(),
+                // System Events
+                hex_literal::hex!("26aa394eea5630e07c48ae0c9558cef780d41e5e16056765bc8461851072c9d7").to_vec().into(),
+            ];
+
+            let params = (&config, &whitelist);
+            let mut batches = Vec::<BenchmarkBatch>::new();
+            add_benchmarks!(params, batches);
```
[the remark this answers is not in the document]
pmikolajczyk41: the same answer as above; additionally, notice, that we are returning `Ok(batches)`; I suppose, that `add_benchmarks!` takes `mut Vec` and populates it according to:
```rust
#[cfg(feature = "runtime-benchmarks")]
mod benches {
    define_benchmarks!([pallet_snarcos, Snarcos]);
}
```
so we have to do it this way

fbielejec (approved):
benchmark run in wasm now, good work!

PR #660, opened 2022-10-12
A0 1410 integrate validator network pretty

maciejnems:
# Description

This PR integrates new validator network with network service. For now it only uses dummy implementation of network that satisfies the trait but does nothing. After this PR, network service uses two connection managers and they are both connected to two different networks.

1. Legacy Network - uses Substrate network for authentication broadcast and sending AlephBFT data
2. New Network - uses Substrate network (the same one) for authentication broadcast and validator network for sending AlephBFT data

This code is still compatible with older nodes due to NetworkData compatibility implemented in this PR. They can communicate through legacy network. Legacy network and its handling should be removed after we update mainnet to use new validator network and we make sure validator network works correctly.

## Type of change

# Tests:
- Integration tests were updated, so that they still test legacy network.
- New tests should be added in separate for testing new network.
- I have tested this change running together with older nodes. In result they still communicate with them and finalizing blocks.
- I have tested by hand that the nodes do use authentication for new network and the add other nodes to validator network.

### DIFF
--- a/finality-aleph/src/network/manager/compatibility.rs
+++ b/finality-aleph/src/network/manager/compatibility.rs
@@ -48,6 +48,23 @@ impl<D: Data, M: Multiaddress, LM: Multiaddress> Encode for VersionedNetworkData
         use VersionedNetworkData::*;
         match self {
             Other(version, payload) => encode_with_version(*version, payload.clone()),
+            Legacy(data) => encode_with_version(0, data.encode()),
+            V1(data) => encode_with_version(1, data.encode()),
+        }
+    }
+}
+
+impl<D: Data, M: Multiaddress, LM: Multiaddress> VersionedNetworkData<D, M, LM> {
+    /// Encodes Network Data, to legacy encoding for the Legacy type. For other types
+    /// it encodes to versioned encoding. This is reverse a funtion to `backwards_compatible_decode`.
+    /// It is needed as encode needs to be a reverse function to decode, which is not possible without
+    /// data having `ByteCount` and `Version` encoded. This is something pre-compatibility
+    /// nodes will not understand.
+    /// This should be removed, after rolling update with new network is completed
+    pub fn backwards_compatible_encode(&self) -> Vec<u8> {
+        use VersionedNetworkData::*;
+        match self {
+            Other(version, payload) => encode_with_version(*version, payload.clone()),
             Legacy(data) => data.encode(),
             V1(data) => encode_with_version(1, data.encode()),
         }
@@ -60,6 +77,7 @@ impl<D: Data, M: Multiaddress, LM: Multiaddress> Decode for VersionedNetworkData
         let version = Version::decode(input)?;
         let num_bytes = ByteCount::decode(input)?;
         match version {
+            0 => Ok(Legacy(NetworkData::decode(input)?)),
             1 => Ok(V1(DiscoveryMessage::decode(input)?)),
             _ => {
                 let mut payload = vec![
@@ -108,7 +126,7 @@ fn decode_pre_compatibility_network_data<D: Data, M: Multiaddress, LM: Multiaddr
 }
 
 /// Decodes Network Data, even if it was produced by ancient code which does not conform to our
-/// backwards compatibility style.
+/// backwards compatibility style. This is reverse a funtion to `backwards_compatible_encode`.
 /// This should be removed, after rolling update with new network is completed
 pub fn backwards_compatible_decode<D: Data, M: Multiaddress, LM: Multiaddress>(
     data_raw: Vec<u8>,
--- a/finality-aleph/src/network/manager/compatibility.rs
+++ b/finality-aleph/src/network/manager/compatibility.rs
@@ -1,0 +1,136 @@
+use std::{
+    fmt::{Display, Error as FmtError, Formatter},
+    mem::size_of,
+};
+
+use codec::{Decode, DecodeAll, Encode, Error as CodecError, Input as CodecInput};
+
+use crate::network::{
+    manager::{DiscoveryMessage, NetworkData},
+    Data, Multiaddress,
+};
+
+type Version = u16;
+type ByteCount = u32;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum VersionedNetworkData<D: Data, M: Multiaddress, LM: Multiaddress> {
+    // Most likely from the future.
+    Other(Version, Vec<u8>),
+    Legacy(NetworkData<D, LM>),
+    V1(DiscoveryMessage<M>),
+}
+
+fn encode_with_version(version: Version, mut payload: Vec<u8>) -> Vec<u8> {
+    let mut result = version.encode();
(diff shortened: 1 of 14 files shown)


finality-aleph/src/network/manager/compatibility.rs
```diff
@@ -0,0 +1,153 @@
...
+                    // so that network data is false positively recognized  as from the future
+                    // therefore we should try to decode formats
+                    decode_pre_compatibility_network_data(data_raw)
+                        .map_err(|_| UnknownVersion(version))
+                }
+            }
+        }
+        Err(_) => decode_pre_compatibility_network_data(data_raw),
+    }
+}
```
deuszx, line 153: We should write roundtrip tests for `Encode`/`Decode` instances of `VersionedNetworkData` -- i.e. make sure that the (de)serialization logic matches each other.

maciejnems: done

timorl (approved):
Nice work, mostly comments about outdated comments, with only one about possibly changing code.

finality-aleph/src/network/service.rs
```diff
@@ -64,66 +90,85 @@ enum SendError {
...
+                            Ok(s) => sender.insert(s),
                             Err(e) => {
                                 debug!(target: "aleph-network", "Failed creating sender. Dropping message: {}", e);
                                 continue;
                             }
                         }
                     };
-                    if let Err(e) = sender.send(data.encode()).await {
+                    // Right now we need to use backward compatible encode here.
+                    // In the future we can change it back to normal encode.
```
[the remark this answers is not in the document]
maciejnems: Oh yeah, this should not be here at all

finality-aleph/src/network/service.rs
```diff
@@ -393,11 +594,23 @@ mod tests {
         }
     }
 
+    fn message(i: u8) -> NetworkData<MockData, LegacyMockMultiaddress> {
+        NetworkData::Data(vec![i, i + 1, i + 2], SessionId(1))
+    }
+
+    /// This is a dummy implementation so that VersionedNetworkData can be put in HashSet.
+    /// It will inserted with other data, so this can always return the same thing.
```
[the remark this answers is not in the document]
maciejnems: In the tests it is sometime inserted with more data. And we need HashSet, as we do not have control over order of peers during broadcast. (technically we could order expected messages so that the test pass,  but I would call that a bad idea)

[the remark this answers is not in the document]
maciejnems: Ok, nevermind. I derived hash for those as described in https://www.notion.so/cardinalcryptography/Basic-traits-derivation-and-field-access-for-public-structs-0aa22c4887024547b613db5754e0664a

DamianStraszak (changes requested):
Do we have logs that allow to extract the following pieces of information:
1. In session `x` we are connected to node indices `a, b, c ...` in the legacy network
2. In session `x` we are connected to node indices `a, b, c ...` in the new validator network.
?

If I remember correctly `1.` is there, correct? What about `2.`?

finality-aleph/src/network/manager/compatibility.rs
```diff
@@ -77,12 +80,10 @@ impl<M: Multiaddress> Decode for VersionedAuthentication<M> {
...
             1 => Ok(V1(DiscoveryMessage::decode(input)?)),
             _ => {
-                let mut payload = vec![
-                    0;
-                    num_bytes
-                        .try_into()
-                        .map_err(|_| "input too big to decode")?
-                ];
+                if num_bytes > MAX_AUTHENTICATION_SIZE {
+                    Err("Authentication too big to decode. Is over 16 KiB")?;
```
timorl, line 84: Perhaps "Authentication has unknown version and is encoded as more than 16KiB." would be better?

timorl (approved):
But please add a unit test for the last change.

maciejnems:
> Do we have logs that allow to extract the following pieces of information:
> 
> 1. In session `x` we are connected to node indices `a, b, c ...` in the legacy network
> 2. In session `x` we are connected to node indices `a, b, c ...` in the new validator network.
>    ?
> 
> If I remember correctly `1.` is there, correct? What about `2.`?

We have `1.` and `2.` in a sense, that we have authenticated and found nodes `a,b,c ...` for both networks. This happens in connection manager and it is this log:
```
Connection Manager status report: authenticated authorities: 0: 4/4 {0: 12D3…vg4jUoRN, 2: 12D3…fDBm9SMS, 3: 12D3…GY3RVRic}, 1: 4/4 {0: 12D3…vg4jUoRN, 2: 12D3…fDBm9SMS, 3: 12D3…GY3RVRic};     
Connection Manager status report: authenticated authorities: 0: 4/4 {0: 5DhQv6hwhQ9QMaPYvLFVwfKnT4pdVeQRhSVjucUF45A6wqef, 2: 5F6hcf6QxRbATMLVR7bnbQKX46T8aWpZZr7wQaQwxv6Dycoc, 3: 5EEUggGTcDrE9c5JHLK2wihm3mhELNPUexuRuQ6xpBoB7Lq1}, 1: 4/4 {0: 5DhQv6hwhQ9QMaPYvLFVwfKnT4pdVeQRhSVjucUF45A6wqef, 2: 5F6hcf6QxRbATMLVR7bnbQKX46T8aWpZZr7wQaQwxv6Dycoc, 3: 5EEUggGTcDrE9c5JHLK2wihm3mhELNPUexuRuQ6xpBoB7Lq1};    
```
We also have a status of how many nodes are connected in legacy protocols and new authority protocol. That said we do not have a log about exact number of TCP connections open, as this is not in scope of service, but in scope of implementation of TCP network.

DamianStraszak (approved):

finality-aleph/src/network/manager/compatibility.rs
```diff
@@ -0,0 +1,182 @@
...
+        VersionedAuthentication::V1(message)
+    }
+}
+
+fn encode_with_version(version: Version, mut payload: Vec<u8>) -> Vec<u8> {
+    let mut result = version.encode();
+    // This will produce rubbish if we ever try encodings that have more than u32::MAX bytes.
+    let num_bytes = payload.len() as ByteCount;
+    result.append(&mut num_bytes.encode());
+    result.append(&mut payload);
```
deuszx, line 49: You have 3 allocations here and multiple cloning of the underlying vectors:
* allocate vector of length `size_of(Version)` for `result` 
* allocation for `num_bytes.encode()` and then cloning it into `result` -- which requires extending the `result` vector
* cloning `payload` into `result` which, again, requires extendning the `result` vector.


Since this will be happening for every protocol message, it has a huge impact on the overall performance.

A better approach would be something serde-like where you pass in a buffer (`&mut Vec<u8>`) as an argument from the top-level call to `encode` and pre-allocate the vector with the expected length of the data.

maciejnems: You are right, I will create an `encode_to` implementation in seperate PR. Also now I noticed we have `as u16` here, which is very stupid and will also be removed in this PR

PR #662, opened 2022-10-14
Chain extension

pmikolajczyk41:
This PR includes:
 1. `pallet_contracts::ChainExtension` which exposes a single function for calling `pallet_snarcos::bare_store_key` with mocked arguments
 2. `ink::chain_extension` definition (reusable in any contract)
 3. sample ink! contract that uses the extension

Reading:
 - https://github.com/Cardinal-Cryptography/substrate/blob/aleph-v0.9.26/frame/contracts/src/chain_extension.rs
 - https://github.com/paritytech/ink/tree/master/examples/rand-extension
 - https://use.ink/macros-attributes/chain-extension/


bin/runtime/src/lib.rs
```diff
@@ -640,6 +646,49 @@ impl pallet_utility::Config for Runtime {
...
 }
 
+pub struct AlephChainExtension;
+impl ChainExtension<Runtime> for AlephChainExtension {
+    fn call<E: Ext>(func_id: u32, env: Environment<E, InitState>) -> Result<RetVal, DispatchError>
+    where
+        <E::T as SysConfig>::AccountId: UncheckedFrom<<E::T as SysConfig>::Hash> + AsRef<[u8]>,
+    {
+        match func_id {
+            41 => {
```
deuszx, line 656: Consider extracting to a const for better readability:
```rust
const SNARCOS_CHAIN_EXT: usize = 41;

match func_id {
  SNARCOS_CHAIN_EXT => { ... }
```

pmikolajczyk41: surely, done

deuszx: I edited original post and suggested using an enum -- bit more futureproof if you add more extensions.

deuszx: Ah yes, I was looking at my older code but it had a `TryFrom` impl that used an external crate for `u32 -> enum` transformation.

h4nsu (approved):

bin/runtime/src/chain_extension/snarcos-extension/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "snarcos-extension"
```
h4nsu, line 2: I find it kinda weird this crate is under `runtime` folder. `contracts` would be much better place in my opinion

h4nsu (approved):
One more thing, confused the hell out of me

bin/runtime/src/chain_extension/snarcos-extension/Cargo.toml
deuszx: I have to plead guilty here as I made @pmikolajczyk41 move it from `/contracts` to here. 

My reasoning was (and still is) that the chain extension is part of the core chain (you can see it being defined in [`runtime/src/lib.rs` ](https://github.com/Cardinal-Cryptography/aleph-node/blob/8d56f701304fd43e1a018bcf9c5671d9ef55f5ab/bin/runtime/src/lib.rs#L680) ) and not living in the client space like contracts. Now it lives closer to the place where it's actually used and defined - we wouldn't be adding a chain extension written by clients (`/contracts`, even though now live in the main repo are not part of the core chain). 

We can still make it "publishable" and have contracts refer to it in their dependencies https://github.com/Cardinal-Cryptography/aleph-node/blob/liminal/chain-extension/contracts/snarcos-contract/Cargo.toml#L26 - making it all type-safe.

pmikolajczyk41: I see both points and I really cannot decide. There is, lets say, some solution with using https://github.com/Supercolony-net/obce, but for now it would be an overkill. 

I'll leave this as it is now, and in case we find it inconvenient, we'll move it somewhere else (I suppose/hope somebody will play with it more and thus have some experience-based feedback).

CCing @Yithis to make him aware of these doubts (since he'll be extending this extension (today?))

bin/runtime/src/chain_extension/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+            _ => {
+                error!("Called an unregistered `func_id`: {}", func_id);
+                Err(DispatchError::Other("Unimplemented func_id"))
+            }
+        }
+    }
+
+    fn enabled() -> bool {
+        true
+    }
```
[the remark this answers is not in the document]
pmikolajczyk41: sure, removed

PR #663, opened 2022-10-14
A0-1411: Turn on validator TCP network

timorl:
# Description

Actually uses the new validator network.

## Type of change

# Checklist:

- [TODO] I have made neccessary updates to the Infrastructure


finality-aleph/src/nodes/validator_node.rs
```diff
@@ -42,10 +48,40 @@ where
...
 
-    let (validator_network, validator_network_identity) = tcp_network::new_noop().await;
+    // We generate the phrase manually to only save the key in RAM.
+    let peer_id = keystore
+        .ed25519_generate_new(
+            KEY_TYPE,
+            Some(Mnemonic::new(MnemonicType::Words12, Language::English).phrase()),
+        )
+        .await
+        .expect("generating a key should work");
```
DamianStraszak, line 63: Why do you think this keeps the key in RAM only?

finality-aleph/src/nodes/validator_node.rs
```diff
@@ -42,10 +48,40 @@ where
...
+        external_addresses,
+        validator_port,
         ..
     } = aleph_config;
 
-    let (validator_network, validator_network_identity) = tcp_network::new_noop().await;
+    // We generate the phrase manually to only save the key in RAM.
+    let peer_id = keystore
+        .ed25519_generate_new(
+            KEY_TYPE,
```
DamianStraszak, line 59: If this is the network key, then we should definitely use a different `KEY_TYPE`. The current one is dedicated for aleph-bft consensus.

finality-aleph/src/validator_network/protocols.rs
```diff
@@ -96,15 +97,18 @@ async fn v0_outgoing<D: Data, S: Splittable>(
     peer_id: AuthorityId,
     result_for_parent: mpsc::UnboundedSender<(AuthorityId, Option<mpsc::UnboundedSender<D>>)>,
 ) -> Result<(), ProtocolError> {
+    trace!(target: "validator-network", "Extending hand to {}.", peer_id);
     let (sender, receiver) = v0_handshake_outgoing(stream, authority_pen, peer_id.clone()).await?;
+    info!(target: "validator-network", "Handshake with {} finished successfully.", peer_id);
```
DamianStraszak, line 102: `Handshake` -> `Outgoing handshake` ?

And similarly below for `incoming`

finality-aleph/src/validator_network/service.rs
```diff
@@ -168,8 +171,9 @@ impl<D: Data, A: Data, ND: Dialer<A>, NL: Listener> Service<D, A, ND, NL> {
                     },
                     // pass the data to the manager
                     SendData(data, peer_id) => {
-                        if let Err(e) = self.manager.send_to(&peer_id, data) {
-                            info!(target: "validator-network", "Failed sending to {}: {}", peer_id, e);
+                        match self.manager.send_to(&peer_id, data) {
+                            Ok(_) => trace!(target: "validator-network", "Sending data to {}.", peer_id),
+                            Err(e) => info!(target: "validator-network", "Failed sending to {}: {}", peer_id, e),
```
DamianStraszak, line 176: Just to make sure: are we sure there is no potential for spam with this log?

docker/docker-compose.yml
```diff
@@ -10,6 +10,8 @@ services:
       - RPC_PORT=9933
       - WS_PORT=9943
       - PORT=30333
+      - VALIDATOR_PORT=30342
+      - PUBLIC_VALIDATOR_ADDRESS=127.0.0.1:30342
```
DamianStraszak, line 14: Why not start the numbering from `30343` which looks like the default otherwise?

local-tests/test_multiple_restarts.py
```diff
@@ -24,13 +24,15 @@ def printt(s): print(ctime() + ' | ' + s)
 
 chain.set_flags('no-mdns',
                 port=Seq(30334),
+                validator_port=Seq(3043),
```
DamianStraszak, line 27: why not default 30343?

local-tests/test_catch_up.py
```diff
@@ -26,13 +26,15 @@ def printt(s): print(ctime() + ' | ' + s)
 
 chain.set_flags('no-mdns',
                 port=Seq(30334),
+                validator_port=Seq(3043),
```
DamianStraszak, line 29: why not default 30343?

timorl: Fixed.

local-tests/test_multiple_restarts.py
timorl: Typo. I blame python, based on my personal prejudice.

Fixed.

docker/docker-compose.yml
timorl: Started out as a typo, and then I rolled with it. You are right though, this is just confusing, fixed.

finality-aleph/src/validator_network/service.rs
timorl: Oh, it's _incredibly_ spammy if we fail a connection, good catch. I wanted to make sure that we know when we are not connected despite trying, but the manager status report should now be sufficient for that.

Downgraded to `trace`.

finality-aleph/src/nodes/validator_node.rs
```diff
@@ -42,10 +48,40 @@ where
...
         justification_rx,
         backup_saving_path,
+        external_addresses,
+        validator_port,
         ..
     } = aleph_config;
 
-    let (validator_network, validator_network_identity) = tcp_network::new_noop().await;
+    // We generate the phrase manually to only save the key in RAM.
+    let peer_id = keystore
```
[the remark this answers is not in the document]
timorl: Changed as suggested. Note that deeper in the code we are still calling it peer_id, but we also have a trait for that, so I think there it should either be clearer or irrelevant what we are referring to, depending on the specific place in the code.

docker/docker_entrypoint.sh
```diff
@@ -48,6 +50,7 @@ ARGS=(
   --node-key-file "${NODE_KEY_PATH}"
   --backup-path "${BACKUP_PATH}"
   --rpc-port "${RPC_PORT}" --ws-port "${WS_PORT}" --port "${PORT}"
+  --validator-port ${VALIDATOR_PORT} --public-validator-addresses ${PUBLIC_VALIDATOR_ADDRESS}
```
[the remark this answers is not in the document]
timorl: I just wanted to be consistent with the previous line. ;.;

Split this into two, and the above into three.

bin/node/src/aleph_cli.rs
```diff
@@ -34,6 +43,14 @@ impl AlephCli {
         )
     }
 
+    pub fn external_addresses(&self) -> Vec<String> {
+        self.public_validator_addresses.clone()
+    }
+
+    pub fn validator_port(&self) -> u16 {
+        self.validator_port.unwrap_or(30343)
```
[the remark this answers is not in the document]
timorl: The only other case in this file does it this way. :D But I think you are right, native clap should be better, changed.

scripts/run_nodes.sh
```diff
@@ -96,8 +96,11 @@ run_node() {
     --execution Native \
     --rpc-cors=all \
     --no-mdns \
+    --public-validator-addresses 127.0.0.1:$((30343 + i)) \
+    --validator-port $((30343 + i)) \
```
kostekIV, line 100: based on that it looks like the ort of `public-validator-addresses` should match with `validator-port`. Consider removing port part from the `public-validator-addresses` and get full address by combining it with `validator-port`

local-tests/test_catch_up.py
```diff
@@ -26,13 +26,15 @@ def printt(s): print(ctime() + ' | ' + s)
...
+                validator_port=Seq(30343),
                 ws_port=Seq(9944),
                 rpc_port=Seq(9933),
                 unit_creation_delay=200,
                 execution='Native',
                 pruning='archive')
 addresses = [n.address() for n in chain]
-chain.set_flags(bootnodes=addresses[0], public_addr=addresses)
+validator_addresses = [n.validator_address() for n in chain]
+chain.set_flags(bootnodes=addresses[0], public_addr=addresses, public_validator_addresses=validator_addresses)
```
kostekIV, line 37: nit: if this is smt that only validators use consider using `chain.set_flags_validator` to only set it for validator nodes

finality-aleph/src/nodes/validator_node.rs
```diff
@@ -42,10 +48,40 @@ where
         millisecs_per_block,
         justification_rx,
         backup_saving_path,
+        external_addresses,
+        validator_port,
         ..
     } = aleph_config;
 
-    let (validator_network, validator_network_identity) = tcp_network::new_noop().await;
+    // We generate the phrase manually to only save the key in RAM.
```
kostekIV, line 56: how will this work in case a node crashes and rejoin the network during the session?

kostekIV (approved):

finality-aleph/src/nodes/validator_node.rs
timorl: A new authentication will be generated and broadcast. These keys changing was already taken into account when writing validator node connection management, because it technically could have happened with libp2p peer addresses anyway, now it's only more likely.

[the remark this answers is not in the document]
timorl: Mostly to avoid spamming relatively unimportant files on disk right next to the absolutely crucial aleph validator keys. Expanded the comment to make this clearer.

finality-aleph/src/nodes/validator_node.rs
timorl: Created a specific one for the validator network and used it here.

local-tests/test_catch_up.py
timorl: Changed, although hm, now `public_addr` is also only present in validators, I hope this does not break anything...

This also required some additional changes in accepting the lack of this parameter and panicking when it's not present on a validator.

maciejnems (approved):

DamianStraszak (approved):

PR #665, opened 2022-10-14
A0-1140: Add utils for contracts to aleph-client

obrok:
# Description

Adds utility code helpful for interacting with deployed contracts. See the end of this description for a usage example. I'm using the currently forbidden `pub mod` in this PR, please ignore for now, I'll make sure to either change that or argue for a change in the accepted practices before merging.

I went with a simplistic method of setting `MAX_GAS` and `MAX_READ_GAS`. Perhaps that's good enough for tests, but not good enough for merging into the client proper? Maybe it would be enough for now to set those in the contract constructor instead of having them constant?

## Type of change

# Checklist:

- I have added tests - in a way - the code is not tested directly, but will be used in e2e-tests. However, that's still on a branch, so not part of this PR.

```rust
// This is a wrapper around (some of) the functionality of openbrush PSP22 contracts
// that uses the building blocks added in this PR:

#[derive(Debug)]
struct PSP22TokenInstance {
    contract: ContractInstance,
}

impl PSP22TokenInstance {
    fn new(address: AccountId32, metadata_path: &Option<String>) -> Result<Self> {
        let metadata_path = metadata_path
            .as_ref()
            .context("PSP22Token metadata not set.")?;
        Ok(Self {
            contract: ContractInstance::new(address, metadata_path)?,
        })
    }

    fn transfer(&self, conn: SignedConnection, to: AccountId32, amount: Balance) -> Result<()> {
        self.contract.contract_exec(
            conn,
            "PSP22::transfer",
            vec![to.to_string().as_str(), amount.to_string().as_str(), "0x00"].as_slice(),
        )
    }

    fn balance_of(&self, conn: &Connection, account: AccountId32) -> Result<Balance> {
        to_u128(self.contract.contract_read(
            conn,
            "PSP22::balance_of",
            &vec![account.to_string().as_str()],
        )?)
    }
}
```

The usage of the wrapper is as simple as:

```rust
    let token = PSP22TokenInstance::new(address, ticket_token_metadata_path);

    token.transfer(sign(&conn, authority.clone()), player.public().into(), 1)?;
    assert_eq!(token.balance_of(&conn, player.public().into())?, 1);
```

### DIFF
--- a/aleph-client/Cargo.toml
+++ b/aleph-client/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aleph_client"
-version = "1.7.0"
+version = "1.8.0"
 edition = "2021"
 license = "Apache 2.0"
 
@@ -12,6 +12,9 @@ log = "0.4"
 rayon = "1.5"
 serde_json = { version = "1.0" }
 thiserror = "1.0"
+contract-metadata = "1.5"
+contract-transcode = "0.1"
+ink_metadata = "3.3"
 
 ac-primitives = { git = "https://github.com/Cardinal-Cryptography/substrate-api-client.git", branch = "aleph-v0.9.26" }
 substrate-api-client = { git = "https://github.com/Cardinal-Cryptography/substrate-api-client.git", branch = "aleph-v0.9.26", features = ["staking-xt"] }
@@ -24,6 +27,7 @@ pallet-staking = { git = "https://github.com/Cardinal-Cryptography/substrate.git
 pallet-treasury = { git = "https://github.com/Cardinal-Cryptography/substrate.git", branch = "aleph-v0.9.26" }
 pallet-balances = { git = "https://github.com/Cardinal-Cryptography/substrate.git", branch = "aleph-v0.9.26" }
 pallet-vesting = { git = "https://github.com/Cardinal-Cryptography/substrate.git", branch = "aleph-v0.9.26" }
+ac-node-api = { git = "https://github.com/Cardinal-Cryptography/substrate-api-client.git", branch = "aleph-v0.9.26" }
 
 pallet-aleph = { path = "../pallets/aleph" }
 pallet-elections = { path = "../pallets/elections" }
--- a/aleph-client/Cargo.toml
+++ b/aleph-client/Cargo.toml
@@ -18,6 +18,7 @@ ink_metadata = "3.3"
 
 ac-primitives = { git = "https://github.com/Cardinal-Cryptography/substrate-api-client.git", branch = "aleph-v0.9.26" }
 substrate-api-client = { git = "https://github.com/Cardinal-Cryptography/substrate-api-client.git", branch = "aleph-v0.9.26", features = ["staking-xt"] }
+ac-node-api = { git = "https://github.com/Cardinal-Cryptography/substrate-api-client.git", branch = "aleph-v0.9.26" }
 
 frame-support = { git = "https://github.com/Cardinal-Cryptography/substrate.git", branch = "aleph-v0.9.26" }
 sp-core = { git = "https://github.com/Cardinal-Cryptography/substrate.git", branch = "aleph-v0.9.26", features = ["full_crypto"] }
@@ -27,7 +28,6 @@ pallet-staking = { git = "https://github.com/Cardinal-Cryptography/substrate.git
 pallet-treasury = { git = "https://github.com/Cardinal-Cryptography/substrate.git", branch = "aleph-v0.9.26" }
 pallet-balances = { git = "https://github.com/Cardinal-Cryptography/substrate.git", branch = "aleph-v0.9.26" }
 pallet-vesting = { git = "https://github.com/Cardinal-Cryptography/substrate.git", branch = "aleph-v0.9.26" }
-ac-node-api = { git = "https://github.com/Cardinal-Cryptography/substrate-api-client.git", branch = "aleph-v0.9.26" }
 
 pallet-aleph = { path = "../pallets/aleph" }
 pallet-elections = { path = "../pallets/elections" }
--- a/aleph-client/src/contract/mod.rs
+++ b/aleph-client/src/contract/mod.rs
@@ -119,7 +119,7 @@ impl ContractInstance {
 
     /// Executes a 0-argument contract call.
     pub fn contract_exec0(&self, conn: &SignedConnection, message: &str) -> Result<()> {
-        self.contract_exec(conn, message, &vec![])
+        self.contract_exec(conn, message, &[])
     }
 
     /// Executes a contract call.
--- a/aleph-client/src/contract/mod.rs
+++ b/aleph-client/src/contract/mod.rs
@@ -20,6 +20,49 @@ use substrate_api_client::{compose_extrinsic, GenericAddress, XtStatus};
 use crate::{try_send_xt, AnyConnection, Connection, SignedConnection};
 
 /// Represents a contract instantiated on the chain.
+///
+/// For example, you could write this wrapper around (some of) the functionality of openbrush PSP22
+/// contracts:
+///
+/// ```no_run
+/// # use anyhow::{Result, Context};
+/// # use sp_core::crypto::AccountId32;
+/// # use aleph_client::{Connection, SignedConnection};
+/// # use aleph_client::contract::ContractInstance;
+/// # use aleph_client::contract::util::to_u128;
+///
+/// #[derive(Debug)]
+/// struct PSP22TokenInstance {
+///     contract: ContractInstance,
+/// }
+///
+/// impl PSP22TokenInstance {
+///     fn new(address: AccountId32, metadata_path: &Option<String>) -> Result<Self> {
+///         let metadata_path = metadata_path
+///             .as_ref()
+///             .context("PSP22Token metadata not set.")?;
+///         Ok(Self {
+///             contract: ContractInstance::new(address, metadata_path)?,
+///         })
+///     }
+///
+///     fn transfer(&self, conn: SignedConnection, to: AccountId32, amount: u128) -> Result<()> {
+///         self.contract.contract_exec(
+///             conn,
+///             "PSP22::transfer",
+///             vec![to.to_string().as_str(), amount.to_string().as_str(), "0x00"].as_slice(),
+///         )
+///     }
+///
+///     fn balance_of(&self, conn: &Connection, account: AccountId32) -> Result<u128> {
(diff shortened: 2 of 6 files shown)


aleph-client/src/contract/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+
+impl ContractInstance {
+    const MAX_READ_GAS: u64 = 500000000000u64;
+    const MAX_GAS: u64 = 10000000000u64;
+
+    /// Creates a new contract instance under `address` with metadata read from `metadata_path`.
+    pub fn new(address: AccountId32, metadata_path: &str) -> Result<Self> {
+        Ok(Self {
+            address,
+            ink_project: load_metadata(metadata_path)?,
```
pmikolajczyk41, line 36: My scout rules are protesting against reading a file in a constructor, and instead I would expect `new` to have a signature `pub fn new(address: AccountId32, metadata_path: &InkProject) -> Self` (or possibly just `InkProject`).

aleph-client/src/contract/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+            .field("ink_project", &self.ink_project)
+            .finish()
+    }
+}
+
+/// Helper for loading contract metadata from a file.
+///
+/// The contract-metadata lib contains a similar function starting with version 0.2. It seems that
+/// version conflicts with some of our other dependencies, however, if we upgrade in the future we
+/// can drop this function in favour of their implementation.
```
pmikolajczyk41, line 133: we don't have any policy about such situations, though it's not the first time we encounter such problem; I would loosely propose to add a short comment in `Cargo.toml`, above the mentioned dependency

aleph-client/src/contract/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+        &self.address
+    }
+
+    /// The metadata of this contract instance.
+    pub fn ink_project(&self) -> &InkProject {
+        &self.ink_project
+    }
+
+    /// Reads the value of a read-only, 0-argument call via RPC.
+    pub fn contract_read0(&self, conn: &Connection, message: &str) -> Result<Value> {
```
pmikolajczyk41, line 51: I would say, that all these methods should be generic over `AnyConnection` instead of receiving concrete type.

aleph-client/src/contract/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+        message: &str,
+        args: &[&str],
+    ) -> Result<()> {
+        let data = self.encode(message, args)?;
+        let xt = compose_extrinsic!(
+            conn.as_connection(),
+            "Contracts",
+            "call",
+            GenericAddress::Id(self.address.clone()),
+            Compact(0u64),
```
pmikolajczyk41, line 82: I would keep this `0u64` as some constant or at least add a short comment here to indicate what it refers to

aleph-client/src/contract/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+    fn contract_read_request(&self, payload: &[u8]) -> serde_json::Value {
+        let payload = hex::encode(payload);
+        json!({
+            "jsonrpc": "2.0",
+            "method": "contracts_call",
+            "params": [{
+                "origin": self.address,
+                "dest": self.address,
+                "value": 0,
+                "gasLimit": Self::MAX_READ_GAS,
```
pmikolajczyk41, line 102: is this true: since it is an RPC call, we could pass here infinity?

aleph-client/src/contract/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+        let hex_data = response_data["result"]["Ok"]["data"]
+            .as_str()
+            .context("Contract response data not found - the contract address might be invalid.")?;
+        self.decode_response(message, hex_data)
+    }
+
+    /// Executes a contract call.
+    pub fn contract_exec(
+        &self,
+        conn: SignedConnection,
```
pmikolajczyk41, line 72: `&SignedConnection`, you don't need ownership

aleph-client/src/contract/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+            .context("RPC request error - there may be more info in node logs.")?;
+        let response_data = from_str::<serde_json::Value>(&response)?;
+        let hex_data = response_data["result"]["Ok"]["data"]
+            .as_str()
+            .context("Contract response data not found - the contract address might be invalid.")?;
+        self.decode_response(message, hex_data)
+    }
+
+    /// Executes a contract call.
+    pub fn contract_exec(
```
pmikolajczyk41, line 70: very weak nit: we have only `contract_exec`, but regarding reading, we have `contract_read` and `contract_read0`; I would encourage you to add `contract_exec0`, since in tests we are very likely to have simple non-arg messages to trigger (like `flipper`)

pmikolajczyk41 (approved):
I have left some comments, but none of them is a hard blocker, hence approve from me.

Generally, it is a solid work and I'm really happy to see finally some good and handy support for the tests.

pmikolajczyk41:
One thought about gas limits (and possible transferred money as payable): taking some inspiration from `substrate-api-client`, I would consider parametrizing this API (it makes more sense to do it at the method-level, but it'd be more convenient to do it at the struct-level) with something like:
```
pub trait ContractCallingParams {
    const GasLimit: Balance;
    const GasReadLimit: Balance;
    ...
}

pub struct DefaultContractCallingParams;
impl ContractCallingParams for DefaultContractCallingParams {
    // inline your constants
}

pub fn contract_exec<CCP: ContractCallingParams>(...) {
...
}
```

aleph-client/Cargo.toml
```diff
@@ -24,6 +27,7 @@ pallet-staking = { git = "https://github.com/Cardinal-Cryptography/substrate.git
 pallet-treasury = { git = "https://github.com/Cardinal-Cryptography/substrate.git", branch = "aleph-v0.9.26" }
 pallet-balances = { git = "https://github.com/Cardinal-Cryptography/substrate.git", branch = "aleph-v0.9.26" }
 pallet-vesting = { git = "https://github.com/Cardinal-Cryptography/substrate.git", branch = "aleph-v0.9.26" }
+ac-node-api = { git = "https://github.com/Cardinal-Cryptography/substrate-api-client.git", branch = "aleph-v0.9.26" }
```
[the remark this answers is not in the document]
Marcin-Radecki: For the future, ideally, we should automate this in CI (in C++ world it was done by `clang-format`, so part of automatic code formatting), so we just need to find some tool

aleph-client/src/contract/mod.rs
Marcin-Radecki: I think it's perfectly file to do operations that can fail in the constructors. I think this rule comes from the old C-days when there was not any exceptional safety of code so it looked like `FileHandle x = someClassConstructor(file); x.load(); if x.okay() {...} x.close();`  etc  Now we have `Result`s so it's fine for a constructor to fail

Marcin-Radecki (approved):

aleph-client/Cargo.toml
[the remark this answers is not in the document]
pmikolajczyk41: since we are already using dylint (in contracts), we can check this one out

aleph-client/src/contract/mod.rs
pmikolajczyk41: yeah, but you pass the whole responsibility of checking for file and reading it to a constructor of a contract representation, which seems to me a bit bad

aleph-client/src/contract/mod.rs
obrok: Dunno about infinity, but numbers larger than what's currently set "don't work". I took the value from polkadot.js and never looked back.

aleph-client/src/contract/mod.rs
pmikolajczyk41: "where to get such a thing" - under the path that they are passing to this constructor xD

nevertheless, I won't be protesting longer, go ahead, but know that I'm not so convinced :P

obrok: Under that path there is a file, not an `InkProject` :P

I don't think it matters, and we can refactor later if we need it.

aleph-client/src/contract/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+
+/// Represents a contract instantiated on the chain.
+pub struct ContractInstance {
+    address: AccountId32,
+    ink_project: InkProject,
+}
+
+impl ContractInstance {
+    const MAX_READ_GAS: u64 = 500000000000u64;
+    const MAX_GAS: u64 = 10000000000u64;
```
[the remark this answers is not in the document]
obrok: Yeah, we might need to have an `Option<GasLimit>` or some other mechanism... I think I would prefer to delay this decision until we actually need it.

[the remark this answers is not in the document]
pmikolajczyk41: sure, resolving

aleph-client/src/contract/mod.rs
pmikolajczyk41: no, it's not a strong request; resolving

aleph-client/src/contract/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+            "id": 1
+        })
+    }
+
+    fn encode(&self, message: &str, args: &[&str]) -> Result<Vec<u8>> {
+        ContractMessageTranscoder::new(&self.ink_project).encode(message, args)
+    }
+
+    fn decode_response(&self, from: &str, contract_response: &str) -> Result<Value> {
+        let contract_response = contract_response.replacen("0x", "", 1);
```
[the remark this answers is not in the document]
obrok: :+1: I only found `strip_prefix`, but that's experimental, I think.

aleph-client/src/contract/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+        let data = self.encode(message, args)?;
+        let xt = compose_extrinsic!(
+            conn.as_connection(),
+            "Contracts",
+            "call",
+            GenericAddress::Id(self.address.clone()),
+            Compact(Self::PAYABLE_VALUE),
+            Compact(Self::MAX_GAS),
+            Self::STORAGE_FEE_LIMIT,
+            None::<Option<u128>>,
```
pmikolajczyk41, line 141: choose one of them :P otherwise there are errors for sending a tx

Marcin-Radecki (approved):
Decent rework.

PR #666, opened 2022-10-17
A0-1478: Cleanup compatibility code

maciejnems:
# Description

This removes `as` casting from compatibility code in `VersionedJustification` and `VersionedAuthentication`.

## Type of change

- Bug fix

# Checklist:

### DIFF
--- a/finality-aleph/src/justification/compatibility.rs
+++ b/finality-aleph/src/justification/compatibility.rs
@@ -5,6 +5,7 @@ use std::{
 
 use aleph_bft::{PartialMultisignature, SignatureSet};
 use codec::{Decode, DecodeAll, Encode, Error as CodecError, Input as CodecInput};
+use log::warn;
 
 use crate::{
     crypto::{Signature, SignatureV1},
@@ -62,7 +63,16 @@ fn encode_with_version(version: Version, payload: &[u8]) -> Vec<u8> {
     // that big.
     // We do not have a guarantee that size_hint is implemented for AlephJustification, so we need
     // to compute actual size to place it in the encoded data.
-    let size = payload.len().try_into().unwrap_or(ByteCount::MAX);
+    let size = payload.len().try_into().unwrap_or_else(|_| {
+        if payload.len() > ByteCount::MAX.into() {
+            warn!(
+                "Versioned Justification too big during Encode. Size is approximately {:?} KiB. Should be {:?} KiB at max.",
+                payload.len() / 1024,
+                ByteCount::MAX / 1024
+            );
+        }
+        ByteCount::MAX
+    });
 
     let mut result = Vec::with_capacity(version.size_hint() + size.size_hint() + payload.len());
 
--- a/finality-aleph/src/justification/compatibility.rs
+++ b/finality-aleph/src/justification/compatibility.rs
@@ -57,12 +57,17 @@ enum VersionedAlephJustification {
 }
 
 fn encode_with_version(version: Version, mut payload: Vec<u8>) -> Vec<u8> {
-    let mut result = version.encode();
+    let mut result =
+        Vec::with_capacity(size_of::<Version>() + size_of::<ByteCount>() + payload.len());
+    version.encode_to(&mut result);
     // This will produce rubbish if we ever try encodings that have more than u16::MAX bytes. We
     // expect this won't happen, since we will switch to proper multisignatures before proofs get
     // that big.
-    let num_bytes = payload.len() as ByteCount;
-    result.append(&mut num_bytes.encode());
+    payload
+        .len()
+        .try_into()
+        .unwrap_or(u16::MAX)
+        .encode_to(&mut result);
     result.append(&mut payload);
     result
 }
--- a/finality-aleph/src/justification/compatibility.rs
+++ b/finality-aleph/src/justification/compatibility.rs
@@ -56,19 +56,20 @@ enum VersionedAlephJustification {
     V3(AlephJustification),
 }
 
-fn encode_with_version(version: Version, mut payload: Vec<u8>) -> Vec<u8> {
-    let mut result =
-        Vec::with_capacity(size_of::<Version>() + size_of::<ByteCount>() + payload.len());
-    version.encode_to(&mut result);
+fn encode_with_version(version: Version, payload: &[u8]) -> Vec<u8> {
     // This will produce rubbish if we ever try encodings that have more than u16::MAX bytes. We
     // expect this won't happen, since we will switch to proper multisignatures before proofs get
     // that big.
-    payload
-        .len()
-        .try_into()
-        .unwrap_or(u16::MAX)
-        .encode_to(&mut result);
-    result.append(&mut payload);
+    // We do not have a guarantee that size_hint is implemented for AlephJustification, so we need
+    // to compute actual size to place it in the encoded data.
+    let size = payload.len().try_into().unwrap_or(ByteCount::MAX);
+
+    let mut result = Vec::with_capacity(version.size_hint() + size.size_hint() + payload.len());
+
+    version.encode_to(&mut result);
+    size.encode_to(&mut result);
+    result.extend_from_slice(payload);
+
     result
 }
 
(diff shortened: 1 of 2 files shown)


finality-aleph/src/justification/compatibility.rs
```diff
@@ -57,12 +57,17 @@ enum VersionedAlephJustification {
 }
 
 fn encode_with_version(version: Version, mut payload: Vec<u8>) -> Vec<u8> {
-    let mut result = version.encode();
+    let mut result =
```
kostekIV, line 60: nit: extract this function to deduplicate code

timorl (approved):

kostekIV (approved):
lgtm

finality-aleph/src/justification/compatibility.rs
```diff
@@ -57,12 +57,17 @@ enum VersionedAlephJustification {
...
+        Vec::with_capacity(size_of::<Version>() + size_of::<ByteCount>() + payload.len());
```
fixxxedpoint, line 61: `size_of` and size of encoded `Version` might be different. Use `size_hint` instead?

finality-aleph/src/justification/compatibility.rs
```diff
@@ -57,12 +57,17 @@ enum VersionedAlephJustification {
...
+    version.encode_to(&mut result);
     // This will produce rubbish if we ever try encodings that have more than u16::MAX bytes. We
     // expect this won't happen, since we will switch to proper multisignatures before proofs get
     // that big.
-    let num_bytes = payload.len() as ByteCount;
-    result.append(&mut num_bytes.encode());
+    payload
+        .len()
+        .try_into()
+        .unwrap_or(u16::MAX)
```
fixxxedpoint, line 69: Use `BytesCount` for consistency.

finality-aleph/src/justification/compatibility.rs
```diff
@@ -57,12 +57,17 @@ enum VersionedAlephJustification {
...
+        .encode_to(&mut result);
     result.append(&mut payload);
```
deuszx, line 71: In one case you're cloning the `payload` and then, `append` does it again. If you had passed it in via reference (`payload: &Vec<u8>`) then here you could use `result.extend_from_slice(&payload)` and get away without the `clone()`.

finality-aleph/src/justification/compatibility.rs
```diff
@@ -57,12 +57,17 @@ enum VersionedAlephJustification {
 }
 
 fn encode_with_version(version: Version, mut payload: Vec<u8>) -> Vec<u8> {
```
fixxxedpoint, line 59: Perhaps you can change `payload`s type to `&Vec...` and avoid cloning?

finality-aleph/src/justification/compatibility.rs
fixxxedpoint, line 69: It's a bit confusing to return data that is potentialy corrupt if it's larger than `MAX`.

finality-aleph/src/justification/compatibility.rs
maciejnems: From what I understand `extend_from_slice` iterates over whole slice, clones each element and pushes it to vec while append moves all data at once. I'm not sure if using it is any better than cloning whole vec and after that moving everything with append.

deuszx: That's a good point -- although you're looking at needlessly copying up to 16kB of data each time. We would have to do some benchmarks but I think allocations that big are more expensive than iteration over a slice where elements live next to each other already.

maciejnems: On the other hand we usually have the `V1` here, not `Other`, so optimizing for this case does not change much. Better optimization would be to use `encode_to` instead of `encode`.

finality-aleph/src/justification/compatibility.rs
maciejnems: I do not see any better way to do this. We do not want to panic here and we also cannot return an error here. This way it will work for everyone who knows the version, and people who do not know the version will also be safe.

Besides that, situation when this exceeds `u16::MAX` will likely never happen.

We have this maximum, so that no one spams a node not knowing the version with creation of huge buffers.

finality-aleph/src/justification/compatibility.rs
maciejnems: I would have to create one common file or folder for compatibility, I'm not sure it is worth it now with only two places that use this code. I would leave it as it is right now.

finality-aleph/src/network/manager/compatibility.rs
```diff
@@ -41,21 +41,25 @@ impl<M: Multiaddress> From<DiscoveryMessage<M>> for VersionedAuthentication<M> {
...
-    version.encode_to(&mut result);
+fn encode_with_version(version: Version, payload: &[u8]) -> Vec<u8> {
     // If size is bigger then u16 we set it to MAX_AUTHENTICATION_SIZE.
     // This should never happen but in case it does we will not panic.
     // Also for other users if they have this version of protocol, authentication
     // will be decoded. If they do not know the protocol, authentication will result
     // in decoding error.
-    payload
+    // We do not have a guarantee that size_hint is implemented for DiscoveryMessage, so we need
+    // to compute actual size to place it in the encoded data.
```
deuszx, line 51: But in the body of `fn size_hint` for `VersionedAuthentication<M>` you do have `V1(data) => data.size_hint()` (where `data` is `DiscoveryMessage`). How are these two different?

maciejnems: Well, unfortunatelly this is said in parity_scale_codec docs

> size_hint(&self) -> usize: Gets the capacity (in bytes) required for the encoded data. This is to avoid double-allocation of memory needed for the encoding. It can be an estimate and does not need to be an exact number. If the size is not known, even no good maximum, then we can skip this function from the trait implementation. This is required to be a cheap operation, so should not involve iterations etc.

So we cannot use it here as this can be an estimate. I checked what happens in this example, for DiscoveryMessage and it returns `0`. Here we want exact size, so we can't use size_hint.

deuszx: Looking at https://github.com/paritytech/parity-scale-codec/blob/master/src/codec.rs#L246 -- we probably don't even need to implement `size_hint` if we're overwriting the `fn encode` -- `size_hint` is used only in the default implementation of `Encode::encode` and since we're overwriting it, `size_hint` will never be used.

maciejnems: True, we also thought about creating a task to rewrite our code to use `size_hint` together with `encode_to` as it is faster. Still it depends how much work would be needed to have this optimization.

deuszx: I would start first with running some benchmarks and looking at the flamegraphs to see how much time is spent doing the decoding/encoding. I can tell you from experience that it can be A LOT.

maciejnems: https://cardinal-cryptography.atlassian.net/browse/A0-1481

finality-aleph/src/justification/compatibility.rs
fixxxedpoint: What about checking this condition first and returning empty `Vec` then?

maciejnems: This could potentially break our code if we incorrectly estimate max size of justification. We do not want to return empty vector from this function in case if something is a correct justification

fixxxedpoint: Current version changes format if payload::size() > u16::MAX. Anyway, assumption here is that it shouldn't happen. But if, returned vec might be corrupted in that case (truncated). Leave as it is now, or don't pretend that returned data has any meaning (corrupted) and return empty or error. I would prefer early exit and returning error or actually not allowing to build payload that is potentially too big in the first place (but it is too late for that design). Decision is yours. You can resolve this comment.

fixxxedpoint (approved):
lgtm. `as` shouldn't panic anyway but this improves overall quality of our encoding.

finality-aleph/src/justification/compatibility.rs
deuszx: Actually the `extend_from_slice` will expand the underlying vector if it's too small. Meaning, we would create a message that is bigger than the message limit.

maciejnems: Yep, and we want to create a message that is too big in this case. So that it works for everyone who has this version. In decode we do not check size if we know the version.

finality-aleph/src/justification/compatibility.rs
```diff
@@ -56,14 +57,29 @@ enum VersionedAlephJustification {
...
     // that big.
-    let num_bytes = payload.len() as ByteCount;
-    result.append(&mut num_bytes.encode());
-    result.append(&mut payload);
+    // We do not have a guarantee that size_hint is implemented for AlephJustification, so we need
+    // to compute actual size to place it in the encoded data.
+    let size = payload.len().try_into().unwrap_or_else(|_| {
+        if payload.len() > ByteCount::MAX.into() {
+            warn!(
+                "Versioned Justification too big during Encode. Size is approximately {:?} KiB. Should be {:?} KiB at max.",
```
deuszx, line 69: How about we also log the version?

deuszx (approved):

PR #668, opened 2022-10-18
A0-1140: Contract event utils

obrok:
# Description

Adds utilities to `aleph_client` that allow for listening to events from multiple contracts.

This is part 2 of #665, so I'm targeting that branch, until that one gets merged, but I think you can already look at it and comment.

## Type of change

# Checklist:

- [x] I have added tests - the code will be used in e2e-tests
- [x] I have created new documentation
- [x] I have bumped aleph-client version if relevant

### DIFF
--- a/aleph-client/src/contract/event.rs
+++ b/aleph-client/src/contract/event.rs
@@ -4,24 +4,30 @@
 //! then run the listen loop. You might want to run the loop in a separate thread.
 //!
 //! ```no_run
+//! # use std::sync::Arc;
 //! # use std::sync::mpsc::channel;
 //! # use std::thread;
 //! # use std::time::Duration;
-//! # use aleph_client::{Connection};
+//! # use aleph_client::{Connection, SignedConnection};
 //! # use aleph_client::contract::ContractInstance;
 //! # use aleph_client::contract::event::{listen_contract_events, subscribe_events};
 //! # use anyhow::Result;
 //! # use sp_core::crypto::AccountId32;
-//! # fn example(conn: Connection, address1: AccountId32, address2: AccountId32, path1: &str, path2: &str) -> Result<()> {
+//! # fn example(conn: SignedConnection, address1: AccountId32, address2: AccountId32, path1: &str, path2: &str) -> Result<()> {
 //!     let subscription = subscribe_events(&conn)?;
-//!     let contract1 = ContractInstance::new(address1, path1)?;
-//!     let contract2 = ContractInstance::new(address2, path2)?;
+//!
+//!     // The `Arc` makes it possible to pass a reference to the contract to another thread
+//!     let contract1 = Arc::new(ContractInstance::new(address1, path1)?);
+//!     let contract2 = Arc::new(ContractInstance::new(address2, path2)?);
 //!     let (cancel_tx, cancel_rx) = channel();
+//!
+//!     let contract1_copy = contract1.clone();
+//!     let contract2_copy = contract2.clone();
 //!
 //!     thread::spawn(move || {
 //!         listen_contract_events(
 //!             subscription,
-//!             &[&contract1, &contract2],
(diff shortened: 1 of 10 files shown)


aleph-client/src/contract/event.rs
```diff
@@ -0,0 +1,220 @@
...
+//! # fn example(conn: Connection, address1: AccountId32, address2: AccountId32, path1: &str, path2: &str) -> Result<()> {
+//!     let subscription = subscribe_events(&conn)?;
+//!     let contract1 = ContractInstance::new(address1, path1)?;
+//!     let contract2 = ContractInstance::new(address2, path2)?;
+//!     let (cancel_tx, cancel_rx) = channel();
+//!
+//!     thread::spawn(move || {
+//!         listen_contract_events(
+//!             subscription,
+//!             &[&contract1, &contract2],
```
pmikolajczyk41, line 24: I find it a bit inconvenient to 'lose' my contract instance for just event listening. Imagine a scenario, where I want to:
 1. execute some contract action
 2. wait for a confirmation event
 3. execute another contract action (for the same contract)
 4. wait for a confirmation event

With the current API I have to create 4 contract instances, because spawning listening requires ownership of the contract instance. I see that `InkProject` doesn't implement `Clone`, but I think we could capture and save the metadata path in `ContractInstance` and expose a custom cloning method.

aleph-client/src/contract/event.rs
```diff
@@ -0,0 +1,220 @@
...
+//!             Some(cancel_rx),
+//!             |event_or_error| { println!("{:?}", event_or_error) }
+//!         );
+//!     });
+//!
+//!     thread::sleep(Duration::from_secs(20));
+//!     cancel_tx.send(()).unwrap();
```
pmikolajczyk41, line 31: I would encourage you to write a new helper method that gets a collection of `&ContractInstance`, connection, event handler and timeout and does: subscription, spawning and utilizing `cancel_tx` as a timeout

pmikolajczyk41 (approved):
Works for me, evidently a lot of work was put here, well done :muscle:

aleph-client/src/contract/event.rs
Marcin-Radecki: We can't avoid passing ownership of `ContractInstance` since we need `InkProject` to decode `Raw` into `ContractEvent`. But what we can do is to split the interface provided by this module so that (note that `decode_contract_event` does two below things actually)
* we get all `Raw` events filtered by contract account id , ie `Vec<Raw>` 
* and then we apply conversion on that vector, in the same thread
Then for listening to itself we just need to pass AccountId
The disadvantage of the above proposal is that current version is more concise

Marcin-Radecki (approved):

aleph-client/src/contract/event.rs
obrok: You don't have to lose it! I made `listen` take `&[&ContractInstance]` on purpose. It's a bit fiddly, but what you can do is pass the contracts to the listening thread via an `Arc` and then extract `&ContractInstance` from the `Arc` for listening.

Perhaps I should add this to the example?

aleph-client/src/contract/util.rs
```diff
@@ -1,3 +1,7 @@
+//! Utilities for writing contract wrappers.
+//!
+//! The functions in this module simplify parsing the data types returned by contract calls.
```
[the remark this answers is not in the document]
obrok: I'm not yet sure what kind of utils we might need, but I'm thinking any such utils _should_ end up here. Perhaps we would need some unparsing utils as well, so I prefer `util` as a name for the module. I'll make the comment more generic, though.

aleph-client/src/contract/event.rs
pmikolajczyk41: yes, definitely, please

PR #670, opened 2022-10-20
A0-1477: extend validator network status report

maciejnems:
# Description
Extend validator network status report to contain info about what peers we have connected and in which way. Examples are:

## Examples of logs
If we do not maintain any connections:
```
Validator Network Manager status report: not maintaining any 
```

If 3 nodes are fully connected:
```
Validator Network Manager status report: target - 3 connections; both ways: 3, [5ExN…G5GBK9tk, 5FCY…mukzBY9p, 5D6W…tjuhjhjP];  
```
If 3 nodes are not fully connected:
```
 Validator Network Manager status report: target - 3 connections; both ways: 1, [5ExN…G5GBK9tk]; incoming connections: 1, [5FCY…mukzBY9p]; outgoing connections: 1, [5D6W…tjuhjhjP];   
```
If noone is connecting to us:
```
Validator Network Manager status report: WARNING! No incoming peers even though we expected tham, maybe connecting to us is impossible; target - 3 connections; both ways: 0, []; outgoing connections: 3, [5D6W…tjuhjhjP, 5ExN…G5GBK9tk, 5GiZ…1UzWnDHn];
```

### DIFF
--- a/finality-aleph/src/network/mod.rs
+++ b/finality-aleph/src/network/mod.rs
@@ -43,6 +43,21 @@ pub mod testing {
     };
 }
 
+/// This trait is needed for logging. It implements a shorter version of PeerId for ids implementing display.
+pub trait ShortId: Display {
+    fn to_short_id(&self) -> String {
+        let id = format!("{}", self);
+
+        let prefix: String = id.chars().take(4).collect();
+
+        let suffix: String = id.chars().skip(id.len().saturating_sub(8)).collect();
+
+        format!("{}…{}", &prefix, &suffix)
+    }
+}
+
+impl<P: PeerId> ShortId for P {}
+
 /// Represents the id of an arbitrary node.
 pub trait PeerId: PartialEq + Eq + Clone + Debug + Display + Hash + Codec + Send {}
 
--- a/finality-aleph/src/validator_network/manager.rs
+++ b/finality-aleph/src/validator_network/manager.rs
@@ -1,12 +1,12 @@
 use std::{
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     fmt::{Display, Error as FmtError, Formatter},
 };
 
 use aleph_primitives::AuthorityId;
 use futures::channel::{mpsc, oneshot};
 
-use crate::validator_network::Data;
+use crate::{network::ShortId, validator_network::Data};
 
 /// Network component responsible for holding the list of peers that we
 /// want to connect to, and managing the established connections.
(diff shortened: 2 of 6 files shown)


finality-aleph/src/validator_network/manager.rs
```diff
@@ -35,25 +35,88 @@ impl Display for SendError {
...
+            .filter(|(_, exit)| !exit.is_closed())
+            .map(|(k, _)| k)
+            .collect();
+
+        let both_ways = incoming.intersection(&outgoing).cloned().collect();
+        let incoming: HashSet<_> = incoming.difference(&both_ways).cloned().collect();
+        let outgoing: HashSet<_> = outgoing.difference(&both_ways).cloned().collect();
+
+        if !self.wanted_peers.is_empty() && both_ways.is_empty() && incoming.is_empty() {
+            status.push_str("WARNING! No incoming peers even though we expected tham, maybe connecting to us is impossible; ");
```
timorl, line 71: Just `write!`. Everywhere, unless I am missing something massively.

finality-aleph/src/validator_network/manager.rs
```diff
@@ -35,25 +35,88 @@ impl Display for SendError {
...
+        let both_ways_status = (
+            both_ways.len(),
+            both_ways
+                .into_iter()
+                .map(|authority_id| authority_id.to_short_id())
+                .collect::<Vec<_>>()
+                .join(", "),
+        );
+        status.push_str(&format!(
+            "target - {:?} connections; both ways: {:?}, [{}]; ",
```
timorl, line 83: You can `write!` the first part even before the "WARNING" above, I think it might be more readable.

finality-aleph/src/network/mod.rs
```diff
@@ -43,6 +43,21 @@ pub mod testing {
     };
 }
 
+/// This trait is needed for logging. It implements a shorter version of PeerId for ids implementing display.
```
timorl, line 46: This has... very weird assumptions that are not at all reflected in the type.

I would suggest something like adding this method to the `PeerId` trait explicitly as `short_id` and in the implementations perhaps use anothe function from `crypto` which would only be defined for specific length slices (which we could get from both kinds of `PeerId` in our case).

And then we could have a `struct Shortened<P: PeerId>(P)` wrapper with it's own implementation of `Display`, which would use this function. You would still have to wrap this at some point, but you could do that in the constructor of the status report, making printing simpler.

finality-aleph/src/validator_network/manager.rs
```diff
@@ -35,25 +35,88 @@ impl Display for SendError {
...
-            self.wanted_peers, self.incoming_peers, self.outgoing_peers,
-        )
+
+        let both_ways_status = (
+            both_ways.len(),
+            both_ways
+                .into_iter()
+                .map(|authority_id| authority_id.to_short_id())
+                .collect::<Vec<_>>()
+                .join(", "),
```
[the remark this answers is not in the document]
maciejnems: Unfotunatelly both `{:?}` and `{:#?}` display it with `"` or in a wrong way. That's why I used join.

finality-aleph/src/network/mod.rs
maciejnems: I'm not so sure about wrapping here just for logging sake. This trait only needs `PeerId` to implement display and nothing else. If you worry about a constraint on it having some length we can add a check here if PeerId is shorter then 13 characters and print it whole in this case.

 We can also just move `short_id` inside `PeerId` trait. This another trait is completely unnecessary.

finality-aleph/src/validator_network/manager.rs
```diff
@@ -37,23 +37,99 @@ impl Display for SendError {
...
+                self.both_ways_peers
+                    .iter()
+                    .map(|authority_id| authority_id.to_short_id())
+                    .collect::<Vec<_>>()
+                    .join(", "),
+            );
+            write!(
+                f,
+                "both ways - {:?} [{}]; ",
+                both_ways_status.0, both_ways_status.1,
```
DamianStraszak, line 97: Would be cleaner to just have two variables with suitable names :)

finality-aleph/src/validator_network/service.rs
```diff
@@ -206,7 +206,7 @@ impl<D: Data, A: Data, ND: Dialer<A>, NL: Listener> Service<D, A, ND, NL> {
                 },
                 // periodically reporting what we are trying to do
                 _ = status_ticker.tick() => {
-                    info!(target: "validator-network", "Manager status report: {}.", self.manager.status_report())
+                    info!(target: "validator-network", "Validator Network Manager status report: {}", self.manager.status_report())
```
DamianStraszak, line 209: Let's try to cut on unnecessary words here:
"Validator Network Manager status: ... "

Is there any other log that is not from the manager? If not, you can also omit "Manager"

DamianStraszak (approved):
There was this issue with the old network that sometimes the connections to session `N+1` (with current session `N`) would be established just before `N+1` started (or even after that) instead of during session `N`. Will this problem be easy to spot with the current logging if it appears again?

finality-aleph/src/validator_network/service.rs
timorl: If we ensure the target is always displayed we can even omit the "Validator Network", leaving just "Status" (half-joking – I'm not sure what triggers this, but targets are often not logged).

finality-aleph/src/validator_network/manager.rs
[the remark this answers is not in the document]
timorl: Oh, right, if you print a `HashSet` using `{:?}` it likely uses `Debug` on the elements... Annoying and the code looks really weird here imo, but I havo no better idea. :<

finality-aleph/src/network/mod.rs
timorl: Adding `short_id` to `PeerId` with a default implementation as now, plus a check for 13 chars sounds good to me. The name could be better, to make clear it's for printing/logging, but I don't have any great ideas for that – `pretty` is too generic, `pretty_short` just sounds weird, `short_n_sweet` is too much of a joke even for me, `short_display` uncovers how terribly we are mangling our usage of `Display`...

Another related idea, although I haven't checked it works with `AuthorityId` – maybe instead of `Display` demand that `PeerId` implements `ToString` and then use that in the default implementation described above. This is kinda what we are already doing for substrate peer ids, and it feels cleaner than overriding `Display` behaviour. This is a very soft suggestion though, just doing what the first sentence of this comment says should be enough.

finality-aleph/src/validator_network/service.rs
maciejnems: There is no other status log from Validator Network. I will remove "Manager"

finality-aleph/src/network/mod.rs
maciejnems: I think in rust `ToString` Is be default implemented using `Display` and documentation says Display should be implemented instead because of that.

I would leave Display here and maybe rename it to `to_short_string`? Technically in is `to_string` but another `to_string` which is shorter...

timorl (approved):

DamianStraszak (approved):

timorl (approved):

DamianStraszak (approved):

PR #675, opened 2022-10-24
cleanup version and data with session

kostekIV:
# Description 
Cleanup of the version usage. 
Add wrapper for tuple (data, session) to ensure how it is encoded.

## Type of change

# Checklist:

- I have made corresponding changes to the existing documentat

### DIFF
--- a/finality-aleph/src/compatibility.rs
+++ b/finality-aleph/src/compatibility.rs
@@ -1,38 +1,8 @@
-use codec::{Decode, Encode, Error, Input, Output};
-
-use crate::{network::Data, SessionId};
+use codec::{Decode, Encode};
 
 #[derive(Encode, Eq, Decode, PartialEq, Debug, Copy, Clone)]
 pub struct Version(pub u16);
 
 pub trait Versioned {
     const VERSION: Version;
-}
-
-/// Wrapper for data send over network. We need it to ensure compatibility.
-/// The order of the data and session_id is fixed in encode and the decode expects it to be data, session_id.
-#[derive(Clone)]
-pub struct NetworkDataInSession<D: Data> {
-    pub data: D,
-    pub session_id: SessionId,
-}
-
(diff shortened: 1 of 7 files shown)


finality-aleph/src/compatibility.rs
```diff
@@ -0,0 +1,38 @@
...
+
+#[derive(Encode, Eq, Decode, PartialEq, Debug, Copy, Clone)]
+pub struct Version(pub u16);
+
+pub trait Versioned {
+    const VERSION: Version;
+}
+
+/// Wrapper for data send over network. We need it to ensure compatibility.
+/// The order of the data and session_id is fixed in encode and the decode expects it to be data, session_id.
```
timorl, line 13: Add "Since data is versioned, i.e. it's encoding starts with a version number in the standardized way, this will allow us to retrofit versioning here if we ever need to change this structure." or something like that.

finality-aleph/src/network/manager/mod.rs
```diff
@@ -47,11 +50,11 @@ impl<M: Multiaddress> AuthData<M> {
 pub type Authentication<M> = (AuthData<M>, Signature);
 
 /// Data inside session, sent to validator network.
-pub type DataInSession<D> = (D, SessionId);
+pub type DataInSession<D> = NetworkDataInSession<D>;
```
timorl, line 53: `NetworkDataInSession` should probably just be defined here (likely as `DataInSession`), with only `Version` remaining in `compatibility.rs`.

timorl (approved):
Nice.

It's a bit annoying that there is some compatibility code left in `lib.rs`, but maybe it can be moved over in some follow-up task, together with the other changes we speculated about earlier.

finality-aleph/src/network/manager/mod.rs
kostekIV: done

Marcin-Radecki (approved):

PR #683, opened 2022-10-25
simple e2e-test for message_compatibility and version upgrade

fixxxedpoint:
# Description

Simple e2e-test for interaction with the new message-compatibility feature. It sets version-upgrade, awaits it, and verify if finalization is ok. WARNING: dummy/fake implementation of version message-compatibiliy should pass this test as well. Next there will be `negative` test verifying if an upgrade, where not enough nodes decides to switch version, will fail.

## Type of change

new e2e-test

### DIFF
--- a/.github/actions/run-e2e-test/action.yml
+++ b/.github/actions/run-e2e-test/action.yml
@@ -26,6 +26,15 @@ inputs:
   follow-up-finalization-check:
     description: 'Whether to run a follow-up finalization check.'
     required: false
+  upgrade_version:
+    description: 'Version which will be used during the VersionUpgrade test.'
+    required: false
+  upgrade_session:
+    description: 'Session for which an upgrade should be scheduled.'
+    required: false
+  upgrade_finalization_wait_sessions:
+    description: 'Number of sessions we should wait after performing an upgrade.'
+    required: false
 
 runs:
   using: 'composite'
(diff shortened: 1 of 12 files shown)


.github/actions/run-e2e-test/action.yml
```diff
@@ -26,6 +26,15 @@ inputs:
...
     required: false
+  upgrade_version:
+    description: 'Version which will be used during the VersionUpgrade test.'
+    required: false
+  upgrade_session:
+    description: 'Session for which an upgrade should be scheduled.'
+    required: false
+  upgrade_finalization_wait_sessions:
+    description: 'Number of sessions we should wait after performing an upgrade.'
+    required: false
```
pmikolajczyk41, line 37: I have a feeling that this is not the best way of introducing parametrization... if I'm not mistaken, these parameters will be used in only just a few tests that focus solely on upgrading. Attaching them to the common parameters set seems questionable.

I'm afraid that it the only alternative is to pass such values as env variables directly in the job specification. Neither pretty nor elegant, but still, that would be my suggestion.

aleph-client/src/lib.rs
```diff
@@ -258,6 +260,17 @@ pub trait CallSystem {
...
 
+pub trait VersionUpgrade {
+    type Version;
+    type Error: StdError;
+
+    fn schedule_upgrade(
+        &self,
+        version: Self::Version,
+        session: SessionIndex,
+    ) -> anyhow::Result<(), Self::Error>;
```
pmikolajczyk41, line 271: nit: `anyhow::Result<(), Self::Error>` === `Result<(), Self::Error>`

e2e-tests/src/cases.rs
```diff
@@ -25,6 +26,7 @@ pub type PossibleTestCases = Vec<(&'static str, TestCase)>;
 /// This comes up in local tests.
 pub fn possible_test_cases() -> PossibleTestCases {
     vec![
+        ("version_upgrade", test_schedule_version_change),
```
pmikolajczyk41, line 29: `finalization` testcase should remain first - if finalization is gone, nothing else matters

aleph-client/src/version_upgrade.rs
```diff
@@ -0,0 +1,39 @@
+use primitives::SessionIndex;
+use sp_core::Pair;
+use substrate_api_client::{compose_call, compose_extrinsic, ExtrinsicParams, XtStatus};
+
+use crate::{try_send_xt, AnyConnection, RootConnection, VersionUpgrade};
+
+impl VersionUpgrade for RootConnection {
```
pmikolajczyk41, line 7: tbh I don't see any gains from traiting here... why not just:
```rust
pub fn schedule_upgrade<Version>(connection: &RootConnection, version: Version, session: SessionIndex) -> anyhow::Result<()> { ... }
```

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -402,6 +402,23 @@ jobs:
...
+        uses: actions/checkout@v2
+
+      - name: Run e2e test
+        uses: ./.github/actions/run-e2e-test
+        with:
+          test-case: version_upgrade
+          upgrade_to_version: 1,
+          upgrade_session: 3,
+          upgrade_finalization_wait_sessions: 2,
+        timeout-minutes: 10
```
pmikolajczyk41, line 420: I have a feeling that `follow-up-finalization-check: true` is a kind of check, that we want to run always, unless a testcase destroys the chain or smth like this. Even though your e2e test performs finalization check on its own, I think we can add this follow-up anyway

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -402,6 +402,23 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@v2
+
+      - name: Run e2e test
+        uses: ./.github/actions/run-e2e-test
+        with:
+          test-case: version_upgrade
+          upgrade_to_version: 1,
```
kostekIV, line 417:
```suggestion
          upgrade_version: 1,
```

kostekIV (approved):
lgtm :)

pmikolajczyk41 (approved):

PR #693, opened 2022-10-27
Decrease sleep in the devnet deployment workflow from 15m to 8m

mikogs:


Kosiare (approved):

mikogs:
Looks like it won't be needed anymore. Closing it.

PR #694, opened 2022-10-28
Passing arguments to `SnarcosChainExtension::store_key`

pmikolajczyk41:
This PR enables passing arguments to the chain extension (`store_key` method). Also some formatting is introduced to ease adding `verify` in the incoming PR.

image


contracts/snarcos-contract/lib.rs
```diff
@@ -5,25 +5,26 @@ use ink_lang as ink;
...
         #[ink(constructor)]
         pub fn new() -> Self {
             Self {}
         }
 
         #[ink(message)]
-        pub fn trigger(&mut self) -> Result<(), StoreKeyError> {
-            self.env().extension().store_key()?;
-            self.env().emit_event(KeyStored {});
+        pub fn trigger(
```
deuszx, line 22: Why not call this `store_key` ?

pmikolajczyk41: idk, changed

bin/runtime/src/chain_extension/mod.rs
```diff
@@ -1,49 +1,81 @@
...
+
+        // Pass the arguments to the pallet and interpret the result.
+        let return_status = match Snarcos::<Runtime>::bare_store_key(identifier, key) {
+            Ok(_) => SNARCOS_STORE_KEY_OK,
+            // In case `DispatchResultWithPostInfo` was returned (or some simpler equivalent for
+            // `bare_store_key`), we could adjust weight. However, for the storing key action it
+            // doesn't make sense.
+            Err(Error::<Runtime>::VerificationKeyTooLong) => SNARCOS_STORE_KEY_TOO_LONG_KEY,
+            Err(Error::<Runtime>::IdentifierAlreadyInUse) => SNARCOS_STORE_KEY_IN_USE,
+            _ => SNARCOS_STORE_KEY_UNKNOWN,
```
deuszx, line 77: Does it mean the key is unknown or that the error is unknown?

bin/runtime/src/chain_extension/mod.rs
```diff
@@ -1,49 +1,81 @@
...
+            return Ok(RetVal::Converging(SNARCOS_STORE_KEY_TOO_LONG_KEY));
+        }
+
+        // We can decode identifier without charging yet - it is cheap and doesn't interact with
+        // any storage.
+        let identifier = env.read_as::<VerificationKeyIdentifier>()?;
+
+        // Now we charge - even if decoding fails and we shouldn't touch storage, we have to incur
+        // fee for reading memory.
+        env.charge_weight(Self::store_key_weight(key_length))?;
```
deuszx, line 65: Reading is probably cheaper than storing so we should revisit in the future?

bin/runtime/src/chain_extension/mod.rs
pmikolajczyk41: the error is unknown; the constant prefix is built like: `<pallet>_<dispatchable>_<error id>`

bin/runtime/src/chain_extension/mod.rs
pmikolajczyk41: well yes, but it will be quite weird to benchmark; to be rethought

bin/runtime/src/chain_extension/mod.rs
pmikolajczyk41: renamed

at the moment, `bare_store_key` cannot fail in any other way than the 2 well-specified errors; so this is rather future-compatible :P

Yithis (approved):
LGTM

PR #697, opened 2022-10-28
A0-1140: e2e tests for contracts

obrok:
# Description

* Adds e2e tests for button and marketplace
* Changes `contracts/script/test.sh` to run the new tests through `e2e-tests` on CI
* Still missing - tests for `simple-dex` - next PR

## Type of change

- Tests (non-breaking change which tests existing functionality)

# Checklist:

- I have added tests


aleph-client/src/contract/util.rs
```diff
@@ -46,3 +46,21 @@ pub fn to_account_id(value: Value) -> Result<AccountId> {
...
+/// # #![feature(assert_matches)]
+/// # use std::assert_matches::assert_matches;
+/// # use anyhow::anyhow;
+/// # use aleph_client::contract::util::to_bool;
+/// use contract_transcode::Value;
+///
+/// assert_matches!(to_bool(Value::Bool(true)), Ok(true));
+/// assert_matches!(to_bool(Value::UInt(42)), Err(_));
+/// ```
+pub fn to_bool(value: Value) -> Result<bool> {
```
fbielejec, line 61: It'd be more "rusty" to implement [TryFrom](https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html) fro such a conversion and then use .into() for a short & sweet syntax

contracts/scripts/deploy.sh
```diff
@@ -114,6 +114,7 @@ function deploy_button_game {
   cd "$CONTRACTS_PATH"/access_control
 
   cargo contract call --url "$NODE" --contract "$ACCESS_CONTROL" --message grant_role --args "$AUTHORITY" 'Owner('"$contract_address"')' --suri "$AUTHORITY_SEED" --skip-confirm
+  cargo contract call --url "$NODE" --contract "$ACCESS_CONTROL" --message grant_role --args "$AUTHORITY" 'Minter('"$game_token"')' --suri "$AUTHORITY_SEED" --skip-confirm
```
fbielejec, line 117: needed? Only the game contract should be a minter?

e2e-tests/src/test/button_game/contracts.rs
```diff
@@ -0,0 +1,222 @@
...
+};
+use anyhow::{Context, Result};
+use sp_core::{
+    crypto::{AccountId32, AccountId32 as AccountId, Ss58Codec},
+    Pair,
+};
+
+use crate::Config;
+
+pub trait AsContractInstance {
```
fbielejec, line 16: instead of all these three trait's I'd suggest implementing `impl From<Foo> for Bar`, as it is more customary and gives you both from/into for free

e2e-tests/src/test/button_game/contracts.rs
```diff
@@ -0,0 +1,222 @@
...
+    pub fn press(&self, conn: &SignedConnection) -> Result<()> {
+        self.contract.contract_exec0(conn, "press")
+    }
+
+    pub fn reset(&self, conn: &SignedConnection) -> Result<()> {
+        self.contract.contract_exec0(conn, "reset")
+    }
+}
+
+impl AsContractInstance for ButtonInstance {
```
fbielejec, line 91: impl From<ButtonInstance> for ContractInstance

e2e-tests/src/test/button_game/contracts.rs
```diff
@@ -0,0 +1,222 @@
...
+    fn as_contract(&self) -> &ContractInstance {
+        &self.contract
+    }
+}
+
+impl ToAccount for ButtonInstance {
```
fbielejec, line 97: impl From<ButtonInstance> for AccountId

e2e-tests/src/test/button_game/contracts.rs
```diff
@@ -0,0 +1,222 @@
...
+    pub fn balance_of(&self, conn: &Connection, account: &AccountId32) -> Result<Balance> {
+        to_u128(self.contract.contract_read(
+            conn,
+            "PSP22::balance_of",
+            &[account.to_string().as_str()],
+        )?)
+    }
+}
+
+impl AsContractInstance for PSP22TokenInstance {
```
fbielejec, line 164: impl From

e2e-tests/src/test/button_game/contracts.rs
```diff
@@ -0,0 +1,222 @@
...
+    fn as_contract(&self) -> &ContractInstance {
+        &self.contract
+    }
+}
+
+impl ToAccount for PSP22TokenInstance {
```
fbielejec, line 170: impl From

e2e-tests/src/test/button_game/contracts.rs
```diff
@@ -0,0 +1,222 @@
...
+        self.contract
+            .contract_exec(conn, "buy", &[max_price.as_str()])
+    }
+
+    pub fn price<C: AnyConnection>(&self, conn: &C) -> Result<Balance> {
+        to_u128(self.contract.contract_read0(conn, "price")?)
+    }
+}
+
+impl AsContractInstance for MarketplaceInstance {
```
fbielejec, line 212: impl From

e2e-tests/src/test/button_game/contracts.rs
```diff
@@ -0,0 +1,222 @@
...
+    fn as_contract(&self) -> &ContractInstance {
```
fbielejec, line 213: impl From

e2e-tests/src/test/button_game/contracts.rs
```diff
@@ -0,0 +1,222 @@
...
+        &self.contract
+    }
+}
+
+impl ToAccount for MarketplaceInstance {
```
fbielejec, line 218: impl From

fbielejec (commented):
I left a few comments, generally From/TryFrom is what you'd expect for doing type conversions. `From` for trivial conversions `TryFrom` if it can fail :shrug:  

I'd like to try the tests as well, so will give it a second round of reviews too, but looks solid already!

aleph-client/src/contract/util.rs
obrok: Can I implement `TryFrom` for a type from another crate?

fbielejec: yes, kinda: https://doc.rust-lang.org/rust-by-example/generics/new_types.html

obrok: Yeah, I knew about that, just thought it would be too much cruft. Hm, I guess it could be something like:

```rust
// Instead of
to_bool(result)
/// do
ValueConvert(result).into()
```

at the call site? Honestly, I don't think this is very clean, but can change if you think it's better, let me know.

contracts/scripts/deploy.sh
obrok: It's useful for tests - without it, there's no easy way to get an arbitrary amount of game tokens to an account without going through the full game flow.

fbielejec: So I don't know how we'll deploy these SCs to mainnet - but likely with this script. So maybe you could:  add ENV

obrok: :+1:

maciejzelaszczyk (approved):
Great work! One gripe I have is that this is relatively large for a single PR. In the future, try to split PRs of such length into 2-3 smaller ones, e.g. prep for test (traits, conversions, etc.) and the actual test.

aleph-client/src/contract/util.rs
maciejzelaszczyk: I do not have a problem with these conversions in their current form.

PR #700, opened 2022-10-31
Chain-extension: verify

pmikolajczyk41:
This PR extends current ChainExtension with `verify` method. 

It is built atop #694 (this one is rather ready, just waiting for @Yithis approval).

image


bin/runtime/src/chain_extension/mod.rs
```diff
@@ -27,6 +41,7 @@ impl ChainExtension<Runtime> for SnarcosChainExtension {
     {
         match func_id {
             SNARCOS_STORE_KEY_FUNC_ID => Self::snarcos_store_key(env),
+            SNARCOS_VERIFY_FUNC_ID => Self::snarcos_verify(env),
             _ => {
                 error!("Called an unregistered `func_id`: {}", func_id);
                 Err(DispatchError::Other("Unimplemented func_id"))
```
fbielejec, line 47: nit: maybe you could pass the unimplemented function_id to the error variant (e.g. using format!) ?

pmikolajczyk41: unfortunately, it's not so easy since `DispatchError::Other` expects `&'static str`

fbielejec (approved):

PR #701, opened 2022-10-31
A0-1472 Validator Network log addresses

lesniak43:
# Description

https://cardinal-cryptography.atlassian.net/browse/A0-1472

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/tcp_network.rs
+++ b/finality-aleph/src/tcp_network.rs
@@ -10,8 +10,32 @@ use tokio::net::{
 
 use crate::{
     network::{Multiaddress, NetworkIdentity, PeerId},
-    validator_network::{Dialer, Listener, Splittable},
+    validator_network::{ConnectionInfo, Dialer, Listener, Splittable},
 };
+
+impl ConnectionInfo for TcpStream {
+    type Address = std::net::SocketAddr;
+
+    fn peer_address(&self) -> Result<Self::Address, std::io::Error> {
+        self.peer_addr()
+    }
+}
+
+impl ConnectionInfo for OwnedWriteHalf {
+    type Address = std::net::SocketAddr;
+
+    fn peer_address(&self) -> Result<Self::Address, std::io::Error> {
+        self.peer_addr()
+    }
+}
+
+impl ConnectionInfo for OwnedReadHalf {
+    type Address = std::net::SocketAddr;
+
+    fn peer_address(&self) -> Result<Self::Address, std::io::Error> {
+        self.peer_addr()
+    }
+}
 
 impl Splittable for TcpStream {
     type Sender = OwnedWriteHalf;
--- a/finality-aleph/src/tcp_network.rs
+++ b/finality-aleph/src/tcp_network.rs
@@ -14,26 +14,29 @@ use crate::{
 };
 
 impl ConnectionInfo for TcpStream {
-    type Address = std::net::SocketAddr;
-
-    fn peer_address(&self) -> Result<Self::Address, std::io::Error> {
-        self.peer_addr()
+    fn peer_address_info(&self) -> String {
+        match self.peer_addr() {
+            Ok(addr) => addr.to_string(),
+            Err(e) => e.to_string(),
+        }
     }
 }
 
 impl ConnectionInfo for OwnedWriteHalf {
-    type Address = std::net::SocketAddr;
-
-    fn peer_address(&self) -> Result<Self::Address, std::io::Error> {
-        self.peer_addr()
+    fn peer_address_info(&self) -> String {
+        match self.peer_addr() {
+            Ok(addr) => addr.to_string(),
+            Err(e) => e.to_string(),
+        }
     }
 }
 
 impl ConnectionInfo for OwnedReadHalf {
-    type Address = std::net::SocketAddr;
-
-    fn peer_address(&self) -> Result<Self::Address, std::io::Error> {
-        self.peer_addr()
+    fn peer_address_info(&self) -> String {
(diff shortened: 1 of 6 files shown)


finality-aleph/src/testing/mocks/validator_network.rs
```diff
@@ -160,33 +162,44 @@ impl AsyncRead for UnreliableDuplexStream {
...
+    peer_address: Address,
 }
 
 impl UnreliableSplittable {
     /// Create a pair of mock splittables connected to each other.
-    pub fn new(max_buf_size: usize, ends_after: Option<usize>) -> (Self, Self) {
+    pub fn new(
+        max_buf_size: usize,
+        ends_after: Option<usize>,
+        peer_address: Address,
```
timorl, line 173: Shouldn't this take a pair of `Address`es?

finality-aleph/src/validator_network/incoming.rs
```diff
@@ -65,7 +65,8 @@ pub async fn incoming<D: Data, S: Splittable>(
     result_for_parent: mpsc::UnboundedSender<(AuthorityId, oneshot::Sender<()>)>,
     data_for_user: mpsc::UnboundedSender<D>,
 ) {
+    let addr = stream.peer_address_info();
     if let Err(e) = manage_incoming(authority_pen, stream, result_for_parent, data_for_user).await {
-        info!(target: "validator-network", "Incoming connection failed: {}", e);
+        info!(target: "validator-network", "Incoming connection failed: {}. Peer address info: {}", e, addr);
```
timorl, line 70: I would go with "Incoming connection from {} failed: {}", this will look weird when there are errors as addresses, but much more natural in the overwhelming majority of cases.

finality-aleph/src/tcp_network.rs
```diff
@@ -10,9 +10,36 @@ use tokio::net::{
...
     network::{Multiaddress, NetworkIdentity, PeerId},
-    validator_network::{Dialer, Listener, Splittable},
+    validator_network::{ConnectionInfo, Dialer, Listener, Splittable},
 };
 
+impl ConnectionInfo for TcpStream {
+    fn peer_address_info(&self) -> String {
+        match self.peer_addr() {
+            Ok(addr) => addr.to_string(),
+            Err(e) => e.to_string(),
```
timorl, line 20: I would suggest formatting the error more explicitly as an error here, as in "unknown address: {}" or something.

finality-aleph/src/validator_network/outgoing.rs
```diff
@@ -10,39 +10,31 @@ use crate::{
...
 impl<A: Data, ND: Dialer<A>> Display for OutgoingError<A, ND> {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), FmtError> {
         use OutgoingError::*;
         match self {
             Dial(e) => write!(f, "dial error: {}", e),
-            ProtocolNegotiation(e) => write!(f, "protocol negotiation error: {}", e),
-            Protocol(e) => write!(f, "protocol error: {}", e),
+            ProtocolNegotiation(addr, e) => write!(
+                f,
+                "protocol negotiation error: {}, peer address info: {}",
```
timorl, line 30: Similarly I would go with something like "protocol negotiation with {} error: {}" and for consistency also below.

timorl (approved):
Couple small comments, but looks good.

DamianStraszak (approved):

finality-aleph/src/testing/mocks/validator_network.rs
lesniak43: shouldy's

finality-aleph/src/validator_network/incoming.rs
lesniak43: fixed

finality-aleph/src/tcp_network.rs
lesniak43: fixed

finality-aleph/src/validator_network/outgoing.rs
lesniak43: changed to "Outgoing connection to {} failed, protocol negotiation error: {}.", and "Outgoing connection to {} failed, protocol error: {}.", since I was unable to make the second sentence look good in the proposed style

timorl: Unfortunately this is not supposed to be a whole message, just the part that corresponds to this specific error – in particular whatever you write here will end up formatted in `info!(target: "validator-network", "Outgoing connection to {} failed: {}, will retry after {}s.", peer_id, e, RETRY_DELAY.as_secs());`, so you would get quite some repetition. I think this will mostly be fine if you remove "Outgoing" from these strings?

lesniak43: but then it would become "Outgoing connection to {} failed: connection to {} failed, protocol negotiation error: {}, will retry after {}s.", which doesn't sound good...

I'd say in that case "peer address info: {}, protocol negotiation error: {}" is much better, what do you think?

timorl: If doesn't really flow great, the thing you wrote out doesn't sound that bad once you notice the two `{}` there are different things – the first one is a peer id, the second is an address.

How about "communication with {} failed, protocol[ negotiation] error: {}"?

lesniak43: fixed

lesniak43:
I've also added logging the list of peer addresses in case of failure during establishing an outgoing connection.

timorl (approved):

PR #703, opened 2022-11-02
Support Marlin system

pmikolajczyk41:
This PR introduces Marlin system to `pallet_snarcos`.

Notes:
 - `PairingEngine` config argument was removed (we use only one BLS curve across codebase and abstracting this at this moment just doesn't make sense)
 - we have to implement `VerifyingSystem` for every system separately, because `ark_marlin::Marlin` doesn't implement `ark_snark::SNARK` (or more generally, there is no common snarkish trait for Groth16, Gm17 and Marlin)

### DIFF
--- a/pallets/snarcos/src/systems.rs
+++ b/pallets/snarcos/src/systems.rs
@@ -1,0 +1,54 @@
+use ark_bls12_381::{Bls12_381, Fr};
+use ark_serialize::CanonicalDeserialize;
+use ark_snark::SNARK;
+use codec::{Decode, Encode};
+use scale_info::TypeInfo;
+use sp_std::vec::Vec;
+
+#[derive(Copy, Clone, Eq, PartialEq, Debug, Decode, Encode, TypeInfo)]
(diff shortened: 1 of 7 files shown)


pallets/snarcos/src/systems.rs
```diff
@@ -0,0 +1,82 @@
...
+pub(super) trait VerifyingSystem {
+    type CircuitField: CanonicalDeserialize;
+    type Proof: CanonicalDeserialize;
+    type VerifyingKey: CanonicalDeserialize;
+
+    fn verify(
+        key: &Self::VerifyingKey,
+        input: &[Self::CircuitField],
+        proof: &Self::Proof,
+    ) -> Result<bool, ()>;
```
h4nsu, line 26: I assume empty error here is a placeholder and something should be added in future. Maybe add a TODO here (with a reference to JIRA issue, if already created).

h4nsu (approved):
LGTM

PR #704, opened 2022-11-02
A0-1499: E2e ban test threshold

maciejzelaszczyk:
# Description

This introduces the threshold test for the ban functionality, 4. here:

https://cardinal-cryptography.atlassian.net/browse/A0-1432

## Type of change

# Checklist:

- I have added tests


e2e-tests/src/ban.rs
```diff
@@ -124,3 +138,58 @@ pub fn check_ban_event<C: AnyConnection>(
...
 }
+
+pub fn check_session_count<C: AnyConnection>(
+    connection: &C,
+    seats: &CommitteeSeats,
+    reserved_validators: &[AccountId],
+    non_reserved_validators: &[AccountId],
+    start_session: &SessionIndex,
+    sessions_to_check: &SessionCount,
+    ban_session_threshold: &SessionCount,
```
kostekIV, line 149: no need for borrows here

e2e-tests/src/ban.rs
```diff
@@ -124,3 +138,58 @@ pub fn check_ban_event<C: AnyConnection>(
...
+) -> anyhow::Result<()> {
+    let validators: Vec<_> = reserved_validators
+        .iter()
+        .chain(non_reserved_validators.iter())
+        .collect();
+
+    let mut expected_validator_session_count = HashMap::new();
+
+    for session in *start_session..*start_session + *sessions_to_check {
+        wait_for_at_least_session(connection, session)?;
```
kostekIV, line 159: rather than waiting here, run this function after `start_session + sessions_to_check` have passed

e2e-tests/src/ban.rs
```diff
@@ -124,3 +138,58 @@ pub fn check_ban_event<C: AnyConnection>(
...
+            if members_for_session.contains(&val) {
+                *session_count += 1;
+                *session_count %= ban_session_threshold;
+            }
+
+            let expected_session_count = expected_validator_session_count
+                .get(&val)
+                .unwrap_or_else(|| panic!("Missing expected session count for validator {}", val));
+
+            check_underperformed_validator_session_count(connection, val, expected_session_count);
```
kostekIV, line 190: from task/test description it should be enough to just check if the validator that was in the committee was marked as underperformer - that would get rid of maps, modulos and counter greatly simplifying the test

e2e-tests/src/ban.rs
maciejzelaszczyk: Changed.

e2e-tests/src/ban.rs
maciejzelaszczyk: I have changed the test logic to check the storage values once all the checked sessions have passed.

e2e-tests/src/ban.rs
maciejzelaszczyk: This is also changed. There is still some less-than-pretty conditional wrangling since the diff in counts between sessions depends on whether the particular validator is in committee and, additionally, the counter might be reset.

pallets/elections/src/lib.rs
```diff
@@ -539,6 +538,8 @@ pub mod pallet {
                 }
             }
 
+            Self::emit_fresh_bans_event();
```
kostekIV, line 541: if this works it should be put in separate pr (with versions, spec_version etc bumped)

kostekIV (approved):

pallets/elections/src/lib.rs
maciejzelaszczyk: It is out of this PR, then.

Marcin-Radecki (approved):
Superb!

PR #706, opened 2022-11-03
version upgrade catchup test

fixxxedpoint:


scripts/catchup_version_upgrade_test.sh
```diff
@@ -0,0 +1,208 @@
...
+    done
+    IFS=$tmp
+}
+
+function initialize {
+    wait_for_finalized_block $1 $2 $3
+}
+
+function wait_for_finalized_block() {
+    local init_block=$1
```
kostekIV, line 36: rename to smth like block_to_be_finalized or smth better :P

scripts/catchup_version_upgrade_test.sh
```diff
@@ -0,0 +1,208 @@
...
+    local -n nodes=$1
+
+    for node in ${nodes[@]}; do
+        log "disconnecting node $node..."
+        docker network disconnect main-network $node
+        log "node $node disconnected"
+    done
+}
+
+function wait_for_round {
```
kostekIV, line 113: what is round here?

kostekIV (approved):
amazing work, the bash functions look like smth we would like to use in the future (if we would ever want to write in bash again!)

scripts/catchup_version_upgrade_test.sh
fixxxedpoint: I'll rename it to `block`.

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -485,6 +516,130 @@ jobs:
...
+        with:
+          name: cliain-docker
+
+      - name: Load cliain docker image
+        shell: bash
+        run: docker load -i cliain.tar
+
+      - name: Call catchup_test.sh
+        run: |
+          chmod +x ./scripts/catchup_version_upgrade_test.sh
```
pmikolajczyk41, line 547: can't we just add this permission to the file in git and abandon this step in all 4 job definitions?

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -510,6 +665,10 @@ jobs:
       run-e2e-ban-manual,
       run-e2e-version-upgrade,
       run-e2e-failing-version-upgrade,
+      run-e2e-version-upgrade-catchup-one-node-late,
+      run-e2e-version-upgrade-catchup-two-nodes-late,
+      run-e2e-version-upgrade-early-catchup-one-node-late,
+      run-e2e-version-upgrade-early-catchup-two-nodes-late,
```
pmikolajczyk41, line 671: as far as I can tell, you can collapse all these jobs into a single one by using https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs; this would reduce massively the new changes (to a single job, with the same steps as any old one + matrix definition for env variables (`UPGRADE_BEFORE_DISABLE` x {`NODES`, `PORTS`, `EXT_STATUS`}))

aleph-client/src/version_upgrade.rs
```diff
@@ -10,6 +10,7 @@ pub fn schedule_upgrade(
     connection: &RootConnection,
     version: Version,
     session: SessionIndex,
+    state: XtStatus,
```
pmikolajczyk41, line 13: technically this requires aleph-client version bump, but I don't know what is the current policy from track-core-chain

docker/docker-compose.base.yml
```diff
@@ -0,0 +1,75 @@
...
+      service: AlephNonBootNode
+    container_name: Node1
+    environment:
+      - RPC_PORT=9934
+      - WS_PORT=9944
+      - PORT=30334
+      - VALIDATOR_PORT=30344
+      - PUBLIC_VALIDATOR_ADDRESS=127.0.0.1:30344
+      - NAME=Node1
+      - BASE_PATH=/data/5GBNeWRhZc2jXu7D55rBimKYDk8PGk8itRYFTPfC8RJLKG5o
```
pmikolajczyk41, line 29: just for clarity - please add a comment about seed (just like `Node0` has in `common.yml`)

pmikolajczyk41 (approved):
a lot of good work! it could have been split into 2 PRs (refactoring dockering and tests), but it's okay in that form as well

I agree with @kostekIV about reusing bash functions; and thus I would suggest adding some docs there for faster recognition and following

.github/workflows/e2e-tests-main-devnet.yml
fixxxedpoint: `Dockerfile` for cliain handles it. I'll remove it.

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -54,6 +54,42 @@ jobs:
...
+    name: Build docker image for cliain
+    runs-on: ubuntu-20.04
+    steps:
+      - name: GIT | Checkout source code
+        uses: actions/checkout@v2
+
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+
+      - name: Restore cache
```
pmikolajczyk41, line 67: and probably you want to run also
```
      - name: Cleanup cache
        uses: ./.github/actions/post-cache
```
at the end of the job

fixxxedpoint: Pretty inconvenient design.

pmikolajczyk41: :man_shrugging:

PR #708, opened 2022-11-04
Add missing export

obrok:
# Description

Missed the export for this variable, which makes deployments crash.


mikogs (approved):

contracts/env/fe-benjamin
```diff
@@ -14,4 +14,4 @@ export TICKET_BALANCE=100
 export INITIAL_PRICE=69
 
 # serious environment with no hacks
-ENV_NAME=prod
+export ENV_NAME=prod
```
fbielejec, line 17: not really prod this one, can also be called dev actually

PR #710, opened 2022-11-07
Chain extension: unit tests

pmikolajczyk41:
This PR: 
1) increases testability of `SnarcosChainExtension` through abstracting away `Environment` and `Runtime`, 
1) introduces a suite of 15 unit tests for `SnarcosChainExtension` that ensure correct charging, arguments reading and error passing

There is (at least) one controversial thing here - unstable feature of additional constant generics. It was introduced because all the main calls (`pallet_snarcos::(store_key|verify)`) are invoked as **static** methods of `Runtime`. Therefore, it makes testing different behaviors highly inconvenient (since we cannot change the mock behavior depending on any variable). Being able to introduce enums into generics gives us the possibility of concise defining scenarios.


DamianStraszak (approved):

bin/runtime/src/chain_extension/tests/executor.rs
```diff
@@ -0,0 +1,75 @@
...
+    Errorer(Error),
+}
+
+/// Testing counterpart for `Runtime`.
+///
+/// `STORE_KEY_RESPONDER` instructs how to behave then `store_key` is called.
+/// `VERIFY_RESPONDER` instructs how to behave then `verify` is called.
+pub(super) struct MockedExecutor<
+    const STORE_KEY_RESPONDER: Responder,
+    const VERIFY_RESPONDER: Responder,
```
deuszx, line 25: Would the code be that much longer if you didn't use this unstable, generally unknown to people, complicated feature?

deuszx: You could pass them in as functions to the constructor of `MockedExecutor`:
```rust
impl MockedExecutor{
  fn new(store_key_fn, verify_fn) { ... }
}
```
and use in `impl Executor`
```rust
impl Executor for MockedExecutor {
  fn store_key(args) => self.store_key_fn(args)
}
```
and then in your unit test
```rust
let panicking_executor = MockedExecutor::new(store_key_fn_panic(), verify_fn_panic());
```

I understand you like the feature and may feel using it is smart (it is brilliant I give you that) but it's not stable, it's generally unknown and it makes reading, understanding and hence extending the code harder.

deuszx: I can see now how this approach won't work.

pmikolajczyk41:
I guess that we are looking forward to see https://github.com/rust-lang/rust/issues/95174 stabilized soon...

PR #714, opened 2022-11-07
Add labels so that contracts deployment can be triggered

mikogs:


fbielejec (approved):

.github/workflows/contracts-e2e-tests-and-deploy.yaml
```diff
@@ -279,13 +290,23 @@ jobs:
       - name: Cache contracts' target directories in S3 bucket
```
fbielejec, line 290: but why this step is even here? There is already caching, using our own [cache action](https://github.com/Cardinal-Cryptography/aleph-node/pull/714/files#diff-64b660450dbf52b1f9f9bad88cc7ac7817e727c15d40c74b0738f9f8011bbce4R187)?

fbielejec (changes requested):
There is no need to uplaod to S3 anything more but metadata and addresses.json file. We have a perfectly fine compile caching for rust that can be used. See the e2e job

.github/workflows/contracts-e2e-tests-and-deploy.yaml
mikogs: I have removed caching targets to S3

.github/workflows/contracts-e2e-tests-and-deploy.yaml
```diff
@@ -136,7 +137,7 @@ jobs:
 
   build_and_deploy_contracts:
     name: Deploy contracts on feature environment
-    if: ${{ (github.event_name == 'push') && (github.ref == 'refs/heads/benjamin') }}
+    if: (github.event_name == 'push' && github.ref_name == 'add-labels-for-deploying-contracts') || (github.event_name == 'pull_request' && github.event.action == 'labeled' && github.event.label.name == '[AZERO] DEPLOY-CONTRACTS')
```
fbielejec, line 140: leftover from testing?

.github/workflows/contracts-e2e-tests-and-deploy.yaml
```diff
@@ -294,9 +309,14 @@ jobs:
...
+          version: 'v1.23.6'
+
+      - name: Restart the-button deployment
+        shell: bash
         run: |
-          curl -X POST 'https://api.github.com/repos/Cardinal-Cryptography/the-button/actions/workflows/main.yaml/dispatches' \
-          -H "Accept: application/vnd.github+json" \
-          -H 'Authorization: Bearer ${{ secrets.CI_GH_TOKEN }}' \
-          -d '{ "ref":"main", "inputs": { "buildImage": "false", "deployImage": "true" }}'
+          aws eks --region eu-central-1 update-kubeconfig --name alephzero-devnet-eu-central-1-eks
```
fbielejec, line 320: Will this still build a new image, with say updated metadata and/or addresses ?

k8 is over my head, best if someone else takes a look too (every PR should have two reviewers as a general-ish rule anyway)

fbielejec (commented):
some leftovers from testing & I reckon a second reviewer should take a glance too

.github/workflows/contracts-e2e-tests-and-deploy.yaml
mikogs: this is now changed back to 'benjamin', had to test something

.github/workflows/contracts-e2e-tests-and-deploy.yaml
mikogs: No, it will not build new the-button docker image. JSON files are now mounted into the the-button container. Pod in kubernetes that runs the-button container, has got an additional init container which executes before running the-button, which downloads the JSON files from S3 bucket. This whole thing sits in `aleph-apps` repository (and the init container dockerfile, simirarly to other init containers, is in `aleph-infra`).
For example, we have JSON here https://the-button.dev.azero.dev/metadata/addresses.json (same container, no additional endpoints). For `featurenet` environment, the file is downloaded with `fetch`  at the start of the-button frontend.
With that, the-button workflow is free of any featurenet steps (like downloading JSON files from S3) and the-button docker image does not have to be rebuilt. The running application (its pod) just have to be restarted.

mikogs: Also, the code here triggers the re-deployment of the-button straight away. And waits for its success in the next line.

.github/workflows/contracts-e2e-tests-and-deploy.yaml
```diff
@@ -2,16 +2,17 @@ name: contracts-e2e-tests-and-deploy
...
-    paths:
-      - 'contracts/**'
-      - '.github/**'
+    types:
+      - labeled
+      - opened
+      - created
   push:
     paths:
       - 'contracts/**'
```
Kosiare, line 11: Is this folder correct? I cannot find it anywhere.

Kosiare: My bad, I thought this PR is gonna be merged with `main` branch :+1:

.github/workflows/contracts-e2e-tests-and-deploy.yaml
mikogs: @Kosiare application does not have to be updated. just the init container needs to be restarted.

.github/workflows/contracts-e2e-tests-and-deploy.yaml
Kosiare: @mikogs it is still not changed :)

Kosiare (approved):

PR #716, opened 2022-11-08
A0-1455: Add future protocol and decision process for connection direction

timorl:
# Description

Not used for now, that'll be in the next PR. This is PR 2/3. For more context still see #709 .

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/finality-aleph/src/validator_network/manager/direction.rs
+++ b/finality-aleph/src/validator_network/manager/direction.rs
@@ -4,6 +4,8 @@ use aleph_primitives::AuthorityId;
 
 use crate::validator_network::Data;
 
+/// Data about peers we know and whether we should connect to them or they to us. For the former
+/// case also keeps the peers' addresses.
 pub struct DirectedPeers<A: Data> {
     own_id: AuthorityId,
     outgoing: HashMap<AuthorityId, Vec<A>>,
@@ -18,10 +20,10 @@ fn bit_xor_sum_parity((a, b): (u8, u8)) -> u8 {
     result % 2
 }
 
-// Whether we shold call the remote or the other way around. We xor the peer ids and based on the
-// parity of the sum of bits of the result decide whether the caller should be the smaller or
-// greated lexicographically. They are never equal, because cryptography.
-fn should_call(own_id: &[u8], remote_id: &[u8]) -> bool {
+/// Whether we should call the remote or the other way around. We xor the peer ids and based on the
+/// parity of the sum of bits of the result decide whether the caller should be the smaller or
+/// greated lexicographically. They are never equal, because cryptography.
+fn should_we_call(own_id: &[u8], remote_id: &[u8]) -> bool {
     let xor_sum_parity: u8 = own_id
         .iter()
         .cloned()
@@ -46,11 +48,11 @@ impl<A: Data> DirectedPeers<A> {
 
     /// Add a peer to the list of peers we want to stay connected to, or
     /// update the list of addresses if the peer was already added.
-    /// Returns whether we should start attampts at connecting with the peer, which is the case
+    /// Returns whether we should start attempts at connecting with the peer, which is the case
     /// exactly when the peer is one with which we should attempt connections AND it was added for
     /// the first time.
     pub fn add_peer(&mut self, peer_id: AuthorityId, addresses: Vec<A>) -> bool {
-        match should_call(self.own_id.as_ref(), peer_id.as_ref()) {
+        match should_we_call(self.own_id.as_ref(), peer_id.as_ref()) {
             true => self.outgoing.insert(peer_id, addresses).is_none(),
             false => {
                 // We discard the addresses here, as we will never want to call this peer anyway,
@@ -104,15 +106,19 @@ mod tests {
         (own_container, own_id)
     }
 
+    fn some_addresses() -> Vec<Address> {
+        vec![
+            String::from(""),
+            String::from("a/b/c"),
+            String::from("43.43.43.43:43000"),
+        ]
+    }
+
     #[tokio::test]
     async fn exactly_one_direction_attempts_connections() {
         let (mut own_container, own_id) = container_with_id().await;
         let (mut remote_container, remote_id) = container_with_id().await;
-        let addresses = vec![
-            String::from(""),
-            String::from("a/b/c"),
-            String::from("43.43.43.43:43000"),
-        ];
+        let addresses = some_addresses();
         assert!(
             own_container.add_peer(remote_id, addresses.clone())
                 != remote_container.add_peer(own_id, addresses.clone())
@@ -122,11 +128,7 @@ mod tests {
     async fn container_with_added_connecting_peer() -> (DirectedPeers<Address>, AuthorityId) {
         let (mut own_container, own_id) = container_with_id().await;
         let (mut remote_container, remote_id) = container_with_id().await;
-        let addresses = vec![
-            String::from(""),
-            String::from("a/b/c"),
-            String::from("43.43.43.43:43000"),
-        ];
+        let addresses = some_addresses();
         match own_container.add_peer(remote_id.clone(), addresses.clone()) {
             true => (own_container, remote_id),
             false => {
@@ -139,11 +141,7 @@ mod tests {
     async fn container_with_added_nonconnecting_peer() -> (DirectedPeers<Address>, AuthorityId) {
         let (mut own_container, own_id) = container_with_id().await;
         let (mut remote_container, remote_id) = container_with_id().await;
-        let addresses = vec![
-            String::from(""),
-            String::from("a/b/c"),
-            String::from("43.43.43.43:43000"),
-        ];
+        let addresses = some_addresses();
         match own_container.add_peer(remote_id.clone(), addresses.clone()) {
             false => (own_container, remote_id),
             true => {
@@ -156,11 +154,7 @@ mod tests {
     #[tokio::test]
     async fn no_connecting_on_readd() {
         let (mut own_container, remote_id) = container_with_added_connecting_peer().await;
-        let addresses = vec![
-            String::from(""),
-            String::from("a/b/c"),
-            String::from("43.43.43.43:43000"),
-        ];
+        let addresses = some_addresses();
         assert!(!own_container.add_peer(remote_id, addresses));
     }
 
--- a/finality-aleph/src/validator_network/manager/direction.rs
+++ b/finality-aleph/src/validator_network/manager/direction.rs
@@ -1,0 +1,225 @@
+use std::collections::{HashMap, HashSet};
+
+use aleph_primitives::AuthorityId;
+
+use crate::validator_network::Data;
+
+pub struct DirectedPeers<A: Data> {
+    own_id: AuthorityId,
+    outgoing: HashMap<AuthorityId, Vec<A>>,
+    incoming: HashSet<AuthorityId>,
+}
+
+fn bit_xor_sum_parity((a, b): (u8, u8)) -> u8 {
+    let mut result = 0;
+    for i in 0..8 {
+        result += ((a >> i) ^ (b >> i)) % 2;
+    }
+    result % 2
+}
+
+// Whether we shold call the remote or the other way around. We xor the peer ids and based on the
+// parity of the sum of bits of the result decide whether the caller should be the smaller or
+// greated lexicographically. They are never equal, because cryptography.
+fn should_call(own_id: &[u8], remote_id: &[u8]) -> bool {
+    let xor_sum_parity: u8 = own_id
+        .iter()
+        .cloned()
+        .zip(remote_id.iter().cloned())
+        .map(bit_xor_sum_parity)
+        .fold(0u8, |a, b| (a + b) % 2);
+    match xor_sum_parity == 0 {
+        true => own_id < remote_id,
+        false => own_id > remote_id,
+    }
+}
+
+impl<A: Data> DirectedPeers<A> {
+    /// Create a new set of peers directed using our own peer id.
+    pub fn new(own_id: AuthorityId) -> Self {
+        DirectedPeers {
+            own_id,
+            outgoing: HashMap::new(),
+            incoming: HashSet::new(),
+        }
+    }
+
+    /// Add a peer to the list of peers we want to stay connected to, or
+    /// update the list of addresses if the peer was already added.
+    /// Returns whether we should start attampts at connecting with the peer, which is the case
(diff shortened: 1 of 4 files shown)


pmikolajczyk41 (commented):
I will review `v1/mod.rs` tomorrow morning

finality-aleph/src/validator_network/protocols/mod.rs
```diff
@@ -24,6 +26,7 @@ pub type Version = u32;
 /// protocol. Remove after it's no longer needed.
 #[derive(PartialEq, Debug, Eq, Clone, Copy)]
 pub enum ConnectionType {
+    New,
```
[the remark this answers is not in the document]
timorl: It should be here only for the duration of the compatibility window, it shouldn't be needed in the future at all, so I wouldn't overoptimize the naming here.

finality-aleph/src/validator_network/manager/direction.rs
```diff
@@ -0,0 +1,225 @@
...
+    pub fn interested(&self, peer_id: &AuthorityId) -> bool {
+        self.incoming.contains(peer_id) || self.outgoing.contains_key(peer_id)
+    }
+
+    /// Iterator over the peers we want connections from.
+    pub fn incoming_peers(&self) -> impl Iterator<Item = &AuthorityId> {
+        self.incoming.iter()
+    }
+
+    /// Iterator over the peers we want to connect to.
```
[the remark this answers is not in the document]
timorl: This is actually on purpose, because these are only the ones we want to connect to _actively_, the other iterator is for the ones we want to connect to us – "want to be connected to" would kinda suggest that it should iterate over _both_, no?

finality-aleph/src/validator_network/manager/direction.rs
```diff
@@ -0,0 +1,225 @@
...
+#[cfg(test)]
+mod tests {
+    use aleph_primitives::AuthorityId;
+
+    use super::DirectedPeers;
+    use crate::validator_network::mock::key;
+
+    type Address = String;
+
+    async fn container_with_id() -> (DirectedPeers<Address>, AuthorityId) {
```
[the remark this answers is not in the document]
timorl: Eh, but it's not needed in real code, and we do these kinds of context generation in other tests anyway. In many places in fact.

finality-aleph/src/validator_network/manager/direction.rs
```diff
@@ -0,0 +1,225 @@
+use std::collections::{HashMap, HashSet};
+
+use aleph_primitives::AuthorityId;
+
+use crate::validator_network::Data;
+
+pub struct DirectedPeers<A: Data> {
```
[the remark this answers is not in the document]
timorl: We usually use single letter names for type parameters... I added some more documentation here and this has possibly made this less confusing.

finality-aleph/src/validator_network/manager/direction.rs
```diff
@@ -0,0 +1,225 @@
...
+    own_id: AuthorityId,
+    outgoing: HashMap<AuthorityId, Vec<A>>,
```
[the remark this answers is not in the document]
timorl: Added.

finality-aleph/src/validator_network/manager/direction.rs
```diff
@@ -0,0 +1,225 @@
...
+    let mut result = 0;
+    for i in 0..8 {
+        result += ((a >> i) ^ (b >> i)) % 2;
+    }
+    result % 2
+}
+
+// Whether we shold call the remote or the other way around. We xor the peer ids and based on the
+// parity of the sum of bits of the result decide whether the caller should be the smaller or
+// greated lexicographically. They are never equal, because cryptography.
```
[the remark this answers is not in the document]
timorl: Typo fixed, the number of slashes was on purpose (this is not a public function so it will not generate docs anyway), but I guess it doesn't hurt, so also switched.

finality-aleph/src/validator_network/manager/direction.rs
```diff
@@ -0,0 +1,225 @@
...
+        DirectedPeers {
+            own_id,
+            outgoing: HashMap::new(),
+            incoming: HashSet::new(),
+        }
+    }
+
+    /// Add a peer to the list of peers we want to stay connected to, or
+    /// update the list of addresses if the peer was already added.
+    /// Returns whether we should start attampts at connecting with the peer, which is the case
```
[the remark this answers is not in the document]
timorl: Fixed.

finality-aleph/src/validator_network/manager/direction.rs
```diff
@@ -0,0 +1,225 @@
...
+    for i in 0..8 {
+        result += ((a >> i) ^ (b >> i)) % 2;
+    }
+    result % 2
+}
+
+// Whether we shold call the remote or the other way around. We xor the peer ids and based on the
+// parity of the sum of bits of the result decide whether the caller should be the smaller or
+// greated lexicographically. They are never equal, because cryptography.
+fn should_call(own_id: &[u8], remote_id: &[u8]) -> bool {
```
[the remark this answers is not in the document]
timorl: Changed to `should_we_call`, thanks.

finality-aleph/src/validator_network/manager/direction.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    #[tokio::test]
+    async fn no_connecting_on_readd() {
+        let (mut own_container, remote_id) = container_with_added_connecting_peer().await;
+        let addresses = vec![
+            String::from(""),
+            String::from("a/b/c"),
+            String::from("43.43.43.43:43000"),
+        ];
```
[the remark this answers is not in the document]
timorl: Extracted, although unfortunately we use it in some other tests here and there anyway. <_<"

finality-aleph/src/validator_network/manager/direction.rs
```diff
@@ -0,0 +1,219 @@
...
+    outgoing: HashMap<AuthorityId, Vec<A>>,
+    incoming: HashSet<AuthorityId>,
+}
+
+fn bit_xor_sum_parity((a, b): (u8, u8)) -> u8 {
+    let mut result = 0;
+    for i in 0..8 {
+        result += ((a >> i) ^ (b >> i)) % 2;
+    }
+    result % 2
```
pmikolajczyk41, line 20: `(a ^ b).count_ones() % 2` :)

finality-aleph/src/validator_network/manager/direction.rs
```diff
@@ -0,0 +1,219 @@
...
+/// Whether we should call the remote or the other way around. We xor the peer ids and based on the
+/// parity of the sum of bits of the result decide whether the caller should be the smaller or
+/// greated lexicographically. They are never equal, because cryptography.
+fn should_we_call(own_id: &[u8], remote_id: &[u8]) -> bool {
+    let xor_sum_parity: u8 = own_id
+        .iter()
+        .cloned()
+        .zip(remote_id.iter().cloned())
+        .map(bit_xor_sum_parity)
+        .fold(0u8, |a, b| (a + b) % 2);
```
pmikolajczyk41, line 32: what about
```rust
(x.iter().fold(0, BitXor::bitxor) ^ y.iter().fold(0, BitXor::bitxor)).count_ones() as u8 % 2
```
?

pmikolajczyk41: + you don't need `bit_xor_sum_parity`

finality-aleph/src/validator_network/protocols/v1/mod.rs
```diff
@@ -0,0 +1,446 @@
...
+/// Performs the incoming handshake, and then manages a connection sending and receiving data.
+/// Exits on parent request, or in case of broken or dead network connection.
+pub async fn incoming<D: Data, S: Splittable>(
+    stream: S,
+    authority_pen: AuthorityPen,
+    result_for_parent: mpsc::UnboundedSender<ResultForService<D>>,
+    data_for_user: mpsc::UnboundedSender<D>,
+) -> Result<(), ProtocolError> {
+    trace!(target: "validator-network", "Waiting for extended hand...");
+    let (sender, receiver, peer_id) = v0_handshake_incoming(stream, authority_pen).await?;
```
pmikolajczyk41, line 114: I suppose it's a temp mock, but I would suggest adding a few lines of real mock, because it's really confusing xD

finality-aleph/src/validator_network/protocols/v1/mod.rs
```diff
@@ -0,0 +1,446 @@
...
+    let (data_for_network, data_from_user) = mpsc::unbounded();
+    result_for_parent
+        .unbounded_send((peer_id.clone(), Some(data_for_network), ConnectionType::New))
+        .map_err(|_| ProtocolError::NoParentConnection)?;
+    debug!(target: "validator-network", "Starting worker for communicating with {}.", peer_id);
+    manage_connection(sender, receiver, data_from_user, data_for_user).await
+}
+
+/// Performs the incoming handshake, and then manages a connection sending and receiving data.
+/// Exits on parent request, or in case of broken or dead network connection.
```
pmikolajczyk41, line 106: probably I have some blackout - how do we wait for parent request?

pmikolajczyk41 (approved):
as far as I can tell, everything looks okay (at least quite similar to v0, which was already checked a couple of times :P)

finality-aleph/src/validator_network/protocols/v1/mod.rs
timorl: Huh? Why do you think this is a temp mock, or any mock at all?

pmikolajczyk41: oh, I assumed that vx uses things from vx and `v1_handshake` will be added soon

finality-aleph/src/validator_network/manager/direction.rs
timorl: :o ! <3

finality-aleph/src/validator_network/manager/direction.rs
timorl: + ratio + L + even better than what you put above.

finality-aleph/src/validator_network/protocols/v1/mod.rs
timorl: When the data sending channel is dropped, clarified comment.

finality-aleph/src/validator_network/protocols/v1/mod.rs
timorl: Handshake versioning is separate from protocol versioning, and both v1 and v0 use handshake v0 – there was no reason to write a new handshake. That's also why `handshake.rs` is in `protocol/` directly.

pmikolajczyk41: uhm, then the name `v0_handshake` is a bit confusing :P

finality-aleph/src/validator_network/manager/direction.rs
```diff
@@ -0,0 +1,212 @@
...
+pub struct DirectedPeers<A: Data> {
+    own_id: AuthorityId,
+    outgoing: HashMap<AuthorityId, Vec<A>>,
+    incoming: HashSet<AuthorityId>,
+}
+
+/// Whether we should call the remote or the other way around. We xor the peer ids and based on the
+/// parity of the sum of bits of the result decide whether the caller should be the smaller or
+/// greated lexicographically. They are never equal, because cryptography.
+fn should_we_call(own_id: &[u8], remote_id: &[u8]) -> bool {
```
deuszx, line 21: This looks like a custom implementation of equality operator on `AuthorityId`. I'm pretty sure there should exist a version of it that exits early, without doing the computation over whole thing.

At the same time, optimizing this seems totally unnecessary.

finality-aleph/src/validator_network/protocols/v1/mod.rs
timorl: Kinda, I expect it will become less confusing with time, as it becomes clearly unrelated to the protocol versioning, as it will never match the protocol version again. It might also become clearer when there are multiple handshakes – they would likely end up in places like `handshake::vX`.

finality-aleph/src/validator_network/manager/direction.rs
timorl: Huh, equality? It's a relation, but definitely not equality.

deuszx: Why not? This method could be called for `own_id == remote_id` so all the outcomes are possible: less-than, more-than and equal.

deuszx: Ah, you're correct of course. I was talking about `Ord` but wrote about equality.

timorl: It's also not `Ord`, because it's not transitive, and that is by design. Otherwise there would be a minimal element and someone could control this way whether nodes want to connect to them or they to nodes (to an extent, they cannot actually pick _the_ minimal element, but they could generate keys until they get a sufficiently small/big one). I don't know of any way this control could be abused, but I see no reason for providing it to attackers, since making this nontransitive isn't particularly hard.

maciejnems (approved):
Very nice PR

finality-aleph/src/validator_network/manager/direction.rs
```diff
@@ -0,0 +1,212 @@
...
+        let (mut remote_container, remote_id) = container_with_id().await;
+        let addresses = some_addresses();
+        assert!(
+            own_container.add_peer(remote_id, addresses.clone())
+                != remote_container.add_peer(own_id, addresses.clone())
+        );
+    }
+
+    async fn container_with_added_connecting_peer() -> (DirectedPeers<Address>, AuthorityId) {
+        let (mut own_container, own_id) = container_with_id().await;
```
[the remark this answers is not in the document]
timorl: Yeah, the naming here isn't great, changed to 0 and 1 everywhere – the direction of the connection isn't relevant, but neither is `own` and `remote`, so the names were suggesting more meaning than they had.

finality-aleph/src/validator_network/manager/direction.rs
```diff
@@ -0,0 +1,212 @@
...
+            false => (own_container, remote_id),
+            true => {
+                remote_container.add_peer(own_id.clone(), addresses);
+                (remote_container, own_id)
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn no_connecting_on_readd() {
```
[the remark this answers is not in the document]
timorl: Changed to `subsequent_add` to make this clearer.

PR #717, opened 2022-11-08
A0-1502: Finality version e2e test

maciejzelaszczyk:
# Description

An e2e test for the finality version. The main idea is to wait `2.5` sessions, schedule a finality version change for session `4`, wait for the last finalized block of session `4` and then perform the following checks for each block until the last block of session `4`:

1. Check that the `FinalityVersion` `StorageItem` has values as expected.
2.  Check that the `next_session_finality_version` `state_call` returns values as expected.

You can refer to the original PR introducing the tested change here:
https://github.com/Cardinal-Cryptography/aleph-node/pull/608

This also introduces the possibility to set the default finality version in the chain spec. If a version is not set but a chain spec is used, then we fall back on the legacy version. If no chain spec is used, then we fall back on the default finality version which is always `0`.

## Type of change

# Checklist:

- I have added tests
- I have bumped the `spec_version`
- I have bumped the `aleph-client` version


maciejzelaszczyk:
I have disabled the runtime upgrade check due to consistent errors from `pallet_staking`.

maciejzelaszczyk:
> One, perhaps non-obvious, thing to consider is whether we need a specific code here:
> 
> https://github.com/Cardinal-Cryptography/aleph-node/blob/24c1b1d23711c27077ad96ac347f2d39cdf12825/pallets/aleph/rpc/src/lib.rs#L50

Irrelavant in this iteration.

pallets/aleph/src/impls.rs
```diff
@@ -49,4 +53,15 @@ where
             }
         }
     }
+
+    fn update_next_session_finality_version() {
```
kostekIV, line 57: that looks like some sort of fix? If so should this should halt the abft version upgrade in the testnet probably

kostekIV (approved):

pallets/aleph/src/impls.rs
maciejzelaszczyk: It basically check whether there is a `VersionChange` on the next session and, if so, it swaps out the `NextSessionFinalityVersion` `StorageItem` with the version for the next session. By including this, we do not have to have a separate RPC call for the `NextSessionFinalityVersion`.

primitives/src/lib.rs
```diff
@@ -63,9 +63,12 @@ pub const DEFAULT_COMMITTEE_SIZE: u32 = 4;
 pub const DEFAULT_BAN_MINIMAL_EXPECTED_PERFORMANCE: Perbill = Perbill::from_percent(0);
 pub const DEFAULT_BAN_SESSION_COUNT_THRESHOLD: SessionCount = 3;
 pub const DEFAULT_BAN_REASON_LENGTH: u32 = 300;
+
 pub const DEFAULT_CLEAN_SESSION_COUNTER_DELAY: SessionCount = 960;
 pub const DEFAULT_BAN_PERIOD: EraIndex = 10;
 
+pub const DEFAULT_FINALITY_VERSION: Version = 1;
```
DamianStraszak, line 70: This is kind of controversial, if this should be bumped always or should it stay at 0 forever. Any thoughts @kostekIV ?

DamianStraszak: There is an option to make a policy that this constant in the runtime should be always equal to the legacy version of finality in the host.

kostekIV: the second option is better imo

.github/scripts/run_consensus.sh
```diff
@@ -12,6 +12,9 @@ OVERRIDE_DOCKER_COMPOSE=${OVERRIDE_DOCKER_COMPOSE:-""}
 # default minimum validator count
 MIN_VALIDATOR_COUNT=4
 
+# default minimum validator count
+MIN_VALIDATOR_COUNT=4
```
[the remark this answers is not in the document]
maciejzelaszczyk: This should now be fixed.

primitives/src/lib.rs
DamianStraszak: Ok, let me summon @timorl here -- shall this be `1`? Do we want this synchronized with the legacy finality version? How to guarantee that?

DamianStraszak: I think this matters mostly for e2e tests, because it's only there where there can be a state that does not have this initialized...

maciejzelaszczyk: @DamianStraszak This definitely my reading of this value - only ever important for e2e tests and for setting up new chains. @timorl is there any particular reason that it was changed to `2` and then to `1`? The last change being https://github.com/Cardinal-Cryptography/aleph-node/pull/828

maciejzelaszczyk: @DamianStraszak If we want to not have to specify this in the chain spec, it is definitely possible - I would have to rework the test a bit but not that much. The downside being that we would not test the presence of a value for the finality version before the first change (but this is rather irrelevant as you have mentioned).

DamianStraszak: I think there are generally two options here (this is for new chains, and/or e2e tests).
1) Set the initial value via chainspec, and leave `0` as default, always. This is a quite flexible approach, but the downside is that in all pipelines and whatnot we need to include a correct initial configuration in the chainspec.
2) Don't allow specifying this via chainspec, only specify this via `DEFAULT_FINALITY_VERSION` -- this is not quite flexible, but is simpler for pipelines and stuff.

I think I like 1) more. But let's hear @timorl and @kostekIV

maciejzelaszczyk: @timorl and @kostekIV any preference for either of the options mentioned by @DamianStraszak?

timorl: Wait, I'm confused, I thought the default is only used for generating the chainspec in this PR, not in any other way, so it's already kinda 1) (which I also think is the better approach), isn't it? If my understanding is wrong, then I would support a solution like this anyway – make it chainspec-dependent (as in 1), but make the chainspec generation by default use the latest legacy version, avoiding the need for special setups in (most) tests.

Sorry for the late response, the way I read notifications on github made me miss this probably quite a couple of times. <_<

maciejzelaszczyk: Yes, prior to this PR there was no `GenesisConfig` value at all. The default finality version was used as follows https://github.com/Cardinal-Cryptography/aleph-node/blob/7c1d8220631e75b7830008d021f5f870c346ba80/pallets/aleph/src/lib.rs#L127. However, running e2e tests reveled that reading the value for the `FinalityVersion` immediately after the booting of the chain resulted in `None` values being returned. Hence, the addition of the default to `GenesisConfig`, which solved the issue.

timorl: Ah, that explains why it was working for me...

Anyway, I would then definitely propose 1 + make the default chainspec config that is generated contain the default value (but not the default value being used as the default when the chainspec is unset, that should still be 0). Making chainspec generation produce different values for this setting can happen in a separate PR not to bloat this one (if possible).

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -709,6 +724,7 @@ jobs:
       run-e2e-adder-contract-test,
 #      run-e2e-failing-version-upgrade,
 #      run-e2e-version-upgrade-catchup,
+      run-e2e-finality-version-change,
```
timorl, line 727: Silly question – what about the two tests above? Do they still not work? Is there a task for recreating them?

e2e-tests/src/test/finality_version.rs
```diff
@@ -0,0 +1,222 @@
...
+    finality_version::{
+        check_finality_version_at_block, check_next_session_finality_version_at_block,
+    },
+};
+
+const UPGRADE_TO_VERSION: u32 = 1;
+const UPGRADE_SESSION: SessionIndex = 3;
+const UPGRADE_FINALIZATION_WAIT_SESSIONS: u32 = 3;
+
+const FIRST_INCOMING_FINALITY_VERSION: Version = 1;
```
timorl, line 22: Shouldn't this be equal to `LEGACY_FINALITY_VERSION` rather than a hardcoded `1`? Although TBH I'm kinda confused what this test is exactly doing – I would expect the test to always try switching from `LEGACY_FINALITY_VERSION` to the next one, but... is that the case?

timorl: Wait, shouldn't this be 2 even? Does this test actually change the version now?

timorl (approved):
Approving, but conditional on this actually checking the version upgrade – I left a comment about not being sure that the change actually happens, since it seems it's going between versions `1` and `1`. I kinda suspect I might have misunderstood something though.

.github/workflows/e2e-tests-main-devnet.yml
maciejzelaszczyk: There's that https://cardinal-cryptography.atlassian.net/browse/A0-1644?search_id=5a378e7d-c138-4db0-b19a-e976662cea2d. I pinged @fixxxedpoint since he wrote the initial tests.

e2e-tests/src/test/finality_version.rs
maciejzelaszczyk: You are right - I missed this and it did not actually check the change. The first incoming version should be something different that `LEGACY_FINALITY_VERSION`. For the purpose of the test, it makes sense to set it to `LEGACY_FINALITY_VERSION + 1 = 2`.

maciejzelaszczyk: Should now be fixed.

primitives/src/lib.rs
maciejzelaszczyk: @DamianStraszak I have gone for `1.` as per the suggestions and made it part of this PR. Is now ready to re-read.

finality-aleph/src/abft/legacy.rs
```diff
@@ -1,3 +1,4 @@
+pub use aleph_primitives::LEGACY_FINALITY_VERSION as VERSION;
```
DamianStraszak, line 1: The "same" should happen in the file finality-aleph/src/abft/current.rs

primitives/src/lib.rs
```diff
@@ -63,9 +63,15 @@ pub const DEFAULT_COMMITTEE_SIZE: u32 = 4;
...
 pub const DEFAULT_BAN_REASON_LENGTH: u32 = 300;
+
 pub const DEFAULT_CLEAN_SESSION_COUNTER_DELAY: SessionCount = 960;
 pub const DEFAULT_BAN_PERIOD: EraIndex = 10;
 
+/// Version returned when no version has been set.
+pub const DEFAULT_FINALITY_VERSION: Version = 0;
+/// Version of the legacy abft.
+pub const LEGACY_FINALITY_VERSION: u16 = 1;
+
```
DamianStraszak, line 74: It's perhaps best to define here also `CURRENT_FINALITY_VERSION` and import in `finality-aleph/src/abft/current.rs`

DamianStraszak (changes requested):

bin/node/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aleph-node"
-version = "0.9.0"
+version = "0.10.0"
```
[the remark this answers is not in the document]
maciejzelaszczyk: I have rolled this back to `0.9.0`.

finality-aleph/src/abft/legacy.rs
maciejzelaszczyk: Done.

primitives/src/lib.rs
maciejzelaszczyk: Changed.

primitives/src/lib.rs
```diff
@@ -65,9 +65,17 @@ pub const DEFAULT_COMMITTEE_SIZE: u32 = 4;
...
 pub const DEFAULT_BAN_SESSION_COUNT_THRESHOLD: SessionCount = 3;
 pub const DEFAULT_BAN_REASON_LENGTH: u32 = 300;
+
 pub const DEFAULT_CLEAN_SESSION_COUNTER_DELAY: SessionCount = 960;
 pub const DEFAULT_BAN_PERIOD: EraIndex = 10;
 
+/// Version returned when no version has been set.
+pub const DEFAULT_FINALITY_VERSION: Version = 0;
+/// Current version of abft.
+pub const CURRENT_FINALITY_VERSION: u16 = 2;
```
DamianStraszak, line 75: One nit. Can you define it as `LEGACY_FINALITY_VERSION + 1` ?

That way, we once and for all avoid any possible bugs here.

DamianStraszak (approved):

primitives/src/lib.rs
maciejzelaszczyk: Sure, have made the necessary change.

PR #718, opened 2022-11-08
It's not a mixer (1/3)

pmikolajczyk41:
1. Simple basic PSP22 token (for testing and presentation) - `contracts/public_token/`
2. Initial version of the `Blender` contract (with some simplifications and without `withdraw` method) - `contracts/blender/`
3. Auxiliary script - `contracts/setup_blending.sh`. It:
  a) runs local chain,
  b) builds and deploys both contracts
  c) set some allowance for `Blender`
  d) registers token in `Blender`
  e) registers verifying key in `pallet_snarcos` (through `Blender`)


contracts/blender/src/merkle_tree.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// Array of node values (root is at [1], children are in [2n] and [2n+1]).
+    nodes: Vec<TE>,
+    /// Marker of the first 'non-occupied' leaf.
+    next_free_leaf: u32,
+    #[codec(skip)]
+    _phantom: Box<KB>,
+}
+
+/// Creates an 'empty' tree, i.e. every leaf contains `TE::default()`.
+impl<TE: TreeElement + Default, KB: KinderBlender<TE>, const LEAVES: usize> Default
```
deuszx, line 44: Why do you need the `KinderBlender` abstraction? Do you really see it being changed?

contracts/blender/src/merkle_tree.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    /// Add `elem` to the first 'non-occupied' leaf. Returns `Err(())` iff there are no free leafs.
+    pub fn add(&mut self, elem: TE) -> Result<(), ()> {
+        if self.next_free_leaf as usize == 2 * LEAVES {
+            return Err(());
+        }
+
+        self.nodes[self.next_free_leaf as usize] = elem;
+
+        let mut parent = (self.next_free_leaf / 2) as usize;
```
deuszx, line 79: Don't you need to update the `next_free_leaf` somewhere in here?

pmikolajczyk41: this is why unit tests are essential step :P on my way to update it

contracts/blender/src/merkle_tree.rs
deuszx: And why not use the same hasher as in real contract? You need to test what you're gonna ship anyway.

contracts/blender/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+        Withdraw,
+    }
+
+    #[derive(Eq, PartialEq, Debug, Decode, Encode)]
+    #[cfg_attr(feature = "std", derive(TypeInfo))]
+    pub enum Error {
+        /// Caller is missing some permission.
+        InsufficientPermission,
+        /// Merkle tree is full - no new notes can be created.
+        TooMuchNotes,
```
deuszx, line 70:
```suggestion
        TooManyNotes,
```

contracts/blender/src/merkle_tree.rs
```diff
@@ -0,0 +1,90 @@
...
+///
+/// It has `LEAVES` leaves (and thus `2 * LEAVES - 1` nodes in general). `KB` is used for computing
+/// values in the inner nodes. `TE` represents *both* leaves' values and inner nodes' values.
+///
+/// `LEAVES` must be power of `2`.
+#[derive(Clone, Eq, PartialEq, Decode, Encode, PackedLayout, SpreadLayout, SpreadAllocate)]
+#[cfg_attr(feature = "std", derive(TypeInfo, ink_storage::traits::StorageLayout))]
+pub struct MerkleTree<TE: TreeElement, KB: KinderBlender<TE>, const LEAVES: usize> {
+    /// Array of node values (root is at [1], children are in [2n] and [2n+1]).
+    nodes: Vec<TE>,
```
deuszx, line 36: This is of course just temporary and more mature implementation will use `Mapping`, correct?

contracts/blender/src/lib.rs
pmikolajczyk41: :see_no_evil:

contracts/blender/src/merkle_tree.rs
deuszx: I was just thinking that this additional complexity may not be worth the gains here but if you disagree I'm fine with that as well.

contracts/blender/src/merkle_tree.rs
pmikolajczyk41: why? perfectly I would keep here an array `[TE; 2 * LEAVES]`, but this doesn't work for 2 reasons:
 - expressions on const generics are not supported in Rust yet (still unstable grrr)
 - big arrays don't work with storage spreading/packing/layouting and it's much easier to have a vector

mapping doesn't give us anything valuable I think (unless it can optimize read access for index sequences like `n`/`n/2`/`n/4`/...

I will also summon @DamianStraszak here

contracts/blender/src/merkle_tree.rs
pmikolajczyk41: let's wait for other reviewers to settle this one :)

contracts/blender/src/merkle_tree.rs
deuszx: If my understanding is correct, `Vec<_>` will be stored as a single element always so for reading it will read the whole thing into Wasm's VM. What's the limit of that? Are you sure we will never store more than the system allows for and basically brick the contract?

pmikolajczyk41: in `Spread` fashion every element will have its own cell (and `Blender` is `SpreadAllocate`) - that's my belief

DamianStraszak (changes requested):

contracts/blender/src/merkle_tree.rs
deuszx: To me this seems like an instance of YAGNI.

contracts/blender/src/merkle_tree.rs
pmikolajczyk41: I will go for the mapping; but I will just leave it one thought: initially I was trying to use array (since the size is already known), but some of the memory-related traits are not supported for big arrays

contracts/blender/src/merkle_tree.rs
pmikolajczyk41: image

contracts/blender/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+            token_id: TokenId,
+            value: TokenAmount,
+            note: Note,
+            proof: Vec<u8>,
+        ) -> Result<()> {
+            self.acquire_deposit(token_id, value)?;
+            self.verify_deposit(token_id, value, note, proof)?;
+            self.notes.add(note).map_err(|_| Error::TooManyNotes)?;
+            self.merkle_roots.insert(self.notes.root(), &());
+            Ok(())
```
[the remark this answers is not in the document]
pmikolajczyk41: event emitted

contracts/blender/src/merkle_tree.rs
pmikolajczyk41: done

contracts/blender/src/contract.rs
```diff
@@ -0,0 +1,199 @@
...
+                        .push_arg(self.env().caller())
+                        .push_arg(self.env().account_id())
+                        .push_arg(deposit as Balance)
+                        .push_arg::<Vec<u8>>(vec![]),
+                )
+                .call_flags(CallFlags::default().set_allow_reentry(true))
+                .returns::<core::result::Result<(), PSP22Error>>()
+                .fire()
+                .map_err(|e| BlenderError::InkEnv(format!("{:?}", e)))?
+                .map_err(BlenderError::Psp22)
```
deuszx, line 120: nitpick: if you implement `TryFrom`s you can replace both with `??`.

contracts/blender/src/merkle_tree.rs
```diff
@@ -0,0 +1,79 @@
...
+        }
+    }
+}
+
+impl<const LEAVES: u32> MerkleTree<LEAVES> {
+    /// Get the value from the root node.
+    ///
+    /// Returns `None` if the tree is empty (`Self::add` has not been called yet).
+    pub fn root(&self) -> Option<Hash> {
+        self.nodes.get(1)
```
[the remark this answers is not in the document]
pmikolajczyk41: changed in https://github.com/Cardinal-Cryptography/aleph-node/pull/718/commits/734984bf47c95db20bb14c64b92bad4b683600cb

contracts/blender/src/contract.rs
pmikolajczyk41: but then I have to return `Ok(())` explicitly :P anyway, I guess that's the correct way; changed in https://github.com/Cardinal-Cryptography/aleph-node/pull/718/commits/12dd4b0db57d5c3024727067216aa9f5839e7bc4

contracts/blender/src/contract.rs
```diff
@@ -0,0 +1,199 @@
...
+            &mut self,
+            token_id: TokenId,
+            token_address: AccountId,
+        ) -> Result<()> {
+            self.ensure_mr_blendermaster()?;
+            self.registered_tokens
+                .contains(token_id)
+                .not()
+                .then(|| self.registered_tokens.insert(token_id, &token_address))
+                .ok_or(BlenderError::TokenIdAlreadyRegistered)
```
deuszx, line 190: A slightly shorter alternative is to use [`insert_return_size`](https://docs.rs/ink_storage/3.3.1/ink_storage/struct.Mapping.html#method.insert_return_size)
```rust
self
  .registered_tokens
  .insert_return_size(token_id, &token_address)
  .ok_or(BlenderError::TokenIdAlreadyRegistered)
```

This should also be slightly betteer in terms of performance as it doesn't do additional FFI call for `.not()` (which I assume calls to the Runtime to check if the keys exists in the `Mapping`).

contracts/blender/src/merkle_tree.rs
deuszx: Ready to resolve when you are 👍

contracts/blender/src/contract.rs
deuszx: Oh right. Definitely 🙈

deuszx (approved):

DamianStraszak (approved):

PR #719, opened 2022-11-08
A0-1413: remove legacy network

maciejnems:
# Description

This PR removes legacy substrate network protocols that were previously used for authentication and data sending. They were running in parallel with new Validator Network. Now only thing left is Authentication protocol in substrate network and TCP validator network. 

I left room for adding new protocols in substrate network even thought right now we have only one (Authentication).

## Type of change

- Breaking change - old network is no longer existing

# Checklist:

- I have added tests for service using validator netwok
- I have modified existing tests


timorl (approved):
<3

Just to make sure I haven't missed something – in this version messages to the validator network still go through `network::service`, right?

maciejnems:
Yes, this will be changed in another PR.

docker/docker-compose.bridged.yml
```diff
@@ -15,6 +17,7 @@ services:
       - main
       - Node1
     environment:
+      - PUBLIC_VALIDATOR_ADDRESS=Node1:30344
```
kostekIV, line 20: nit: does port number need to increase for every node since it is prefixed anyway with Node{Id}?

kostekIV (approved):
lgtm, just making sure: removed `does_not_rebroadcast_quickly_but_still_responds` test has some equivalent in the new network code?

docker/docker-compose.bridged.yml
maciejnems: It does not, but all nodes listen on different ports, so we would have to override that too. I only intended to fix deployment, not organize our docker-compose.

maciejnems:
> lgtm, just making sure: removed `does_not_rebroadcast_quickly_but_still_responds` test has some equivalent in the new network code?

This should not have an equivalent as we probably will not be bringing `Authentication` back. We will only leave `AuthenticationBroadcast`. It turns out this optimization is not needed for tests

PR #720, opened 2022-11-09
A0-1140: Deploy simple dex

obrok:
# Description

* Adds `simple_dex` to `deploy.sh`
* Exposes `out_given_in` as a contract call, will be needed for frontend and e2e tests.

## Type of change

### DIFF
--- a/contracts/simple_dex/lib.rs
+++ b/contracts/simple_dex/lib.rs
@@ -162,19 +162,11 @@ mod simple_dex {
                 return Err(DexError::InsufficientAllowanceOf(token_in));
             }
 
-            let balance_token_in = self.balance_of(token_in, this)?;
-            let balance_token_out = self.balance_of(token_out, this)?;
-
-            if balance_token_out < min_amount_token_out {
-                // throw early if we cannot support this swap anyway due to liquidity being too low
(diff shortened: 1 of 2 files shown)


contracts/simple_dex/lib.rs
```diff
@@ -365,6 +357,50 @@ mod simple_dex {
                 .map_err(|why| DexError::InkEnv(format!("Can't retrieve own code hash: {:?}", why)))
         }
 
+        /// Swap trade output given a curve with equal token weights
+        ///
+        /// swap_fee_percentage (integer) is a percentage of the trade that goes towards the pool
+        /// B_0 - (100 * B_0 * B_i) / (100 * (B_i + A_i) -A_i * fee)
+        #[ink(message)]
+        pub fn out_given_in(
```
fbielejec, line 365: this could call into a stateless _out_given_in (or  similar name) that we could unit test, if we narrow down the precision problems to it (likely to happen)? Else LGTM

obrok: :+1: But I propose a new PR for that.

fbielejec: I hold you to that

fbielejec (approved):

maciejnems (approved):

PR #721, opened 2022-11-09
A0-1455: Enable two-way connections

timorl:
# Description

Finally, this actually turns two-way connections on.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/validator_network/manager/legacy.rs
+++ b/finality-aleph/src/validator_network/manager/legacy.rs
@@ -1,0 +1,322 @@
+use std::{
+    collections::{HashMap, HashSet},
+    fmt::{Display, Error as FmtError, Formatter},
+};
+
+use aleph_primitives::AuthorityId;
+use futures::channel::mpsc;
+
+use crate::{
+    network::PeerId,
+    validator_network::{
+        manager::{AddResult, SendError},
+        Data,
+    },
+};
+
+/// Network component responsible for holding the list of peers that we
+/// want to connect to, and managing the established connections.
+pub struct Manager<A: Data, D: Data> {
+    addresses: HashMap<AuthorityId, Vec<A>>,
+    outgoing: HashMap<AuthorityId, mpsc::UnboundedSender<D>>,
+    incoming: HashMap<AuthorityId, mpsc::UnboundedSender<D>>,
+}
+
+struct ManagerStatus {
+    wanted_peers: usize,
+    both_ways_peers: HashSet<AuthorityId>,
+    outgoing_peers: HashSet<AuthorityId>,
+    incoming_peers: HashSet<AuthorityId>,
+    missing_peers: HashSet<AuthorityId>,
+}
+
+impl ManagerStatus {
+    fn new<A: Data, D: Data>(manager: &Manager<A, D>) -> Self {
+        let incoming: HashSet<_> = manager
+            .incoming
+            .iter()
+            .filter(|(_, exit)| !exit.is_closed())
+            .map(|(k, _)| k.clone())
+            .collect();
+        let outgoing: HashSet<_> = manager
+            .outgoing
+            .iter()
+            .filter(|(_, exit)| !exit.is_closed())
+            .map(|(k, _)| k.clone())
+            .collect();
+
+        let both_ways = incoming.intersection(&outgoing).cloned().collect();
+        let incoming: HashSet<_> = incoming.difference(&both_ways).cloned().collect();
+        let outgoing: HashSet<_> = outgoing.difference(&both_ways).cloned().collect();
+        let missing = manager
+            .addresses
+            .keys()
+            .filter(|a| !both_ways.contains(a) && !incoming.contains(a) && !outgoing.contains(a))
+            .cloned()
+            .collect();
+
+        ManagerStatus {
+            wanted_peers: manager.addresses.len(),
+            both_ways_peers: both_ways,
+            incoming_peers: incoming,
+            outgoing_peers: outgoing,
+            missing_peers: missing,
+        }
+    }
+}
+
+impl Display for ManagerStatus {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), FmtError> {
+        if self.wanted_peers == 0 {
+            return write!(f, "not maintaining any connections; ");
+        }
+
+        write!(f, "target - {:?} connections; ", self.wanted_peers)?;
+
+        if self.both_ways_peers.is_empty() && self.incoming_peers.is_empty() {
+            write!(f, "WARNING! No incoming peers even though we expected tham, maybe connecting to us is impossible; ")?;
+        }
+
+        if !self.both_ways_peers.is_empty() {
+            let peers = self
+                .both_ways_peers
+                .iter()
+                .map(|authority_id| authority_id.to_short_string())
+                .collect::<Vec<_>>()
+                .join(", ");
+            write!(
+                f,
+                "both ways - {:?} [{}]; ",
+                self.both_ways_peers.len(),
+                peers,
+            )?;
+        }
+
+        if !self.incoming_peers.is_empty() {
+            let peers = self
+                .incoming_peers
+                .iter()
+                .map(|authority_id| authority_id.to_short_string())
+                .collect::<Vec<_>>()
+                .join(", ");
+            write!(
+                f,
+                "incoming only - {:?} [{}]; ",
+                self.incoming_peers.len(),
+                peers
+            )?;
+        }
+
+        if !self.outgoing_peers.is_empty() {
+            let peers = self
+                .outgoing_peers
+                .iter()
+                .map(|authority_id| authority_id.to_short_string())
+                .collect::<Vec<_>>()
+                .join(", ");
+            write!(
+                f,
+                "outgoing only - {:?} [{}];",
+                self.outgoing_peers.len(),
+                peers
+            )?;
+        }
+
+        if !self.missing_peers.is_empty() {
+            let peers = self
+                .missing_peers
+                .iter()
+                .map(|authority_id| authority_id.to_short_string())
+                .collect::<Vec<_>>()
+                .join(", ");
+            write!(f, "missing - {:?} [{}];", self.missing_peers.len(), peers)?;
+        }
+
+        Ok(())
+    }
+}
+
+impl<A: Data, D: Data> Manager<A, D> {
+    /// Create a new Manager with empty list of peers.
+    pub fn new() -> Self {
+        Manager {
+            addresses: HashMap::new(),
+            outgoing: HashMap::new(),
+            incoming: HashMap::new(),
+        }
+    }
+
+    /// Add a peer to the list of peers we want to stay connected to, or
+    /// update the list of addresses if the peer was already added.
+    /// Returns whether this peer is a new peer.
+    pub fn add_peer(&mut self, peer_id: AuthorityId, addresses: Vec<A>) -> bool {
+        self.addresses.insert(peer_id, addresses).is_none()
+    }
+
+    /// Return Option containing addresses of the given peer, or None if
+    /// the peer is unknown.
+    pub fn peer_addresses(&self, peer_id: &AuthorityId) -> Option<Vec<A>> {
+        self.addresses.get(peer_id).cloned()
+    }
+
+    /// Add an established outgoing connection with a known peer,
+    /// but only if the peer is on the list of peers that we want to stay connected with.
+    pub fn add_outgoing(
+        &mut self,
+        peer_id: AuthorityId,
+        data_for_network: mpsc::UnboundedSender<D>,
+    ) -> AddResult {
+        use AddResult::*;
+        if !self.addresses.contains_key(&peer_id) {
+            return Uninterested;
+        }
+        match self.outgoing.insert(peer_id, data_for_network) {
+            Some(_) => Replaced,
+            None => Added,
+        }
+    }
+
+    /// Add an established incoming connection with a known peer,
+    /// but only if the peer is on the list of peers that we want to stay connected with.
+    pub fn add_incoming(
+        &mut self,
+        peer_id: AuthorityId,
+        exit: mpsc::UnboundedSender<D>,
+    ) -> AddResult {
+        use AddResult::*;
+        if !self.addresses.contains_key(&peer_id) {
+            return Uninterested;
+        };
+        match self.incoming.insert(peer_id, exit) {
+            Some(_) => Replaced,
+            None => Added,
+        }
+    }
+
+    /// Remove a peer from the list of peers that we want to stay connected with.
+    /// Close any incoming and outgoing connections that were established.
+    pub fn remove_peer(&mut self, peer_id: &AuthorityId) {
+        self.addresses.remove(peer_id);
+        self.incoming.remove(peer_id);
+        self.outgoing.remove(peer_id);
+    }
+
+    /// Send data to a peer.
+    /// Returns error if there is no outgoing connection to the peer,
+    /// or if the connection is dead.
+    pub fn send_to(&mut self, peer_id: &AuthorityId, data: D) -> Result<(), SendError> {
+        self.outgoing
+            .get(peer_id)
+            .ok_or(SendError::PeerNotFound)?
+            .unbounded_send(data)
+            .map_err(|_| SendError::ConnectionClosed)
+    }
+
+    /// A status of the manager, to be displayed somewhere.
+    pub fn status_report(&self) -> impl Display {
+        ManagerStatus::new(self)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use futures::{channel::mpsc, StreamExt};
+
+    use super::{AddResult::*, Manager, SendError};
+    use crate::validator_network::mock::key;
+
+    type Data = String;
+    type Address = String;
+
+    #[tokio::test]
+    async fn add_remove() {
+        let mut manager = Manager::<Address, Data>::new();
+        let (peer_id, _) = key().await;
+        let (peer_id_b, _) = key().await;
+        let addresses = vec![
+            String::from(""),
+            String::from("a/b/c"),
+            String::from("43.43.43.43:43000"),
+        ];
+        // add new peer - returns true
+        assert!(manager.add_peer(peer_id.clone(), addresses.clone()));
+        // add known peer - returns false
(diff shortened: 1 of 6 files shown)


finality-aleph/src/validator_network/manager/mod.rs
```diff
@@ -30,191 +41,165 @@ impl Display for SendError {
...
+        let mut outgoing_peers = HashSet::new();
+        let mut missing_outgoing = HashSet::new();
+
+        for peer in manager.wanted.incoming_peers() {
+            match manager.active_connection(peer) {
+                true => incoming_peers.insert(peer.clone()),
+                false => missing_incoming.insert(peer.clone()),
+            };
+        }
+        for peer in manager.wanted.outgoing_peers() {
```
pmikolajczyk41, line 64: I'm not sure yet, but what about `manager.have`? Is it subset of `wanted`? If so, I would suggest mentioning about it in `Manager`'s fields documentation

finality-aleph/src/validator_network/manager/mod.rs
```diff
@@ -30,191 +41,165 @@ impl Display for SendError {
...
-                .map(|authority_id| authority_id.to_short_string())
-                .collect::<Vec<_>>()
-                .join(", ");
-            write!(
-                f,
-                "outgoing only - {:?} [{}];",
-                self.outgoing_peers.len(),
-                peers
-            )?;
+        match wanted_incoming {
```
pmikolajczyk41, line 102: weak suggestion: merge both merges and the `if` above to `match (wanted_incoming, wanted_outgoing)`

finality-aleph/src/validator_network/manager/mod.rs
```diff
@@ -30,191 +41,165 @@ impl Display for SendError {
...
+            0 => write!(f, "not expecting any incoming connections; ")?,
+            _ => {
+                write!(f, "expecting {:?} incoming connections; ", wanted_incoming)?;
+                match self.incoming_peers.is_empty() {
+                    true => write!(f, "WARNING! No incoming peers even though we expected tham, maybe connecting to us is impossible; ")?,
```
pmikolajczyk41, line 107: typo `them`

finality-aleph/src/validator_network/manager/mod.rs
```diff
@@ -30,191 +41,165 @@ impl Display for SendError {
...
-                .iter()
-                .map(|authority_id| authority_id.to_short_string())
-                .collect::<Vec<_>>()
-                .join(", ");
-            write!(f, "missing - {:?} [{}];", self.missing_peers.len(), peers)?;
+        match wanted_outgoing {
+            0 => write!(f, "not attempting any outgoing connections; ")?,
+            _ => {
+                write!(f, "attempting {:?} outgoing connections; ", wanted_outgoing)?;
+                if !self.outgoing_peers.is_empty() {
```
pmikolajczyk41, line 130: aren't we concerned if we cannot connect to any node? (similarly to the lack of incoming connections)

finality-aleph/src/validator_network/manager/mod.rs
```diff
@@ -30,191 +41,165 @@ impl Display for SendError {
...
+pub enum AddResult {
+    /// We do not want to maintain a connection with this peer.
+    Uninterested,
+    /// Connection added.
+    Added,
+    /// Old connection replaced with new one.
+    Replaced,
+}
+
 impl<A: Data, D: Data> Manager<A, D> {
```
pmikolajczyk41, line 164: weak suggestion: I would keep `impl Manager` closer to `struct Manager`, in particular, I wouldn't separate them with other big blocks

finality-aleph/src/validator_network/manager/mod.rs
```diff
@@ -30,191 +41,165 @@ impl Display for SendError {
...
+            .unwrap_or(false)
+    }
+
     /// Add a peer to the list of peers we want to stay connected to, or
     /// update the list of addresses if the peer was already added.
-    /// Returns whether this peer is a new peer.
+    /// Returns whether we should start attempts at connecting with the peer.
     pub fn add_peer(&mut self, peer_id: AuthorityId, addresses: Vec<A>) -> bool {
-        self.addresses.insert(peer_id, addresses).is_none()
+        self.wanted.add_peer(peer_id, addresses)
```
pmikolajczyk41, line 184: What if old `addresses` is equal to the old set and we already hold connection in `self.have`? We will reconnect unnecessarily, right?

finality-aleph/src/validator_network/manager/mod.rs
```diff
@@ -223,16 +208,15 @@ impl<A: Data, D: Data> Manager<A, D> {
...
 
     /// Send data to a peer.
     /// Returns error if there is no outgoing connection to the peer,
     /// or if the connection is dead.
     pub fn send_to(&mut self, peer_id: &AuthorityId, data: D) -> Result<(), SendError> {
-        self.outgoing
+        self.have
             .get(peer_id)
             .ok_or(SendError::PeerNotFound)?
             .unbounded_send(data)
```
pmikolajczyk41, line 222: shouldn't we also remove this connection from `self.have` in case of error?

finality-aleph/src/validator_network/manager/mod.rs
pmikolajczyk41: oh, I see in the testcase that it is checked - within `self.wanted.add_peer()`, true? then maybe please add a word about it in docs

finality-aleph/src/validator_network/manager/mod.rs
```diff
@@ -284,66 +272,67 @@ mod tests {
...
+        } else {
+            // We need to switch the names around, because the connection was randomly the
+            // other way around.
+            let temp_id = connecting_id;
+            connecting_id = listening_id;
+            listening_id = temp_id;
+            let temp_manager = connecting_manager;
+            connecting_manager = listening_manager;
+            listening_manager = temp_manager;
+            assert!(connecting_manager.add_peer(listening_id.clone(), addresses.clone()));
```
pmikolajczyk41, line 309: \+ `assert!(!listening_manager.add_peer(connecting_id.clone(), addresses.clone()))`? :)

finality-aleph/src/validator_network/service.rs
```diff
@@ -150,76 +154,122 @@ impl<D: Data, A: Data + Debug, ND: Dialer<A>, NL: Listener> Service<D, A, ND, NL
             });
     }
 
+    fn peer_addresses(&self, peer_id: &AuthorityId) -> Option<Vec<A>> {
+        match self.legacy_connected.contains(peer_id) {
```
pmikolajczyk41, line 158: shouldn't we prefer connection with new protocol?

finality-aleph/src/validator_network/service.rs
```diff
@@ -150,76 +154,122 @@ impl<D: Data, A: Data + Debug, ND: Dialer<A>, NL: Listener> Service<D, A, ND, NL
...
+                            self.legacy_connected.insert(peer_id.clone())
+                        },
+                        LegacyOutgoing => {
+                            self.manager.remove_peer(&peer_id);
+                            self.legacy_connected.insert(peer_id.clone());
+                            false
                         }
+                        New => {
+                            // We always return New if connecting fails, so only New+Some means we
+                            // actually negotiated the new protocol, otherwise we should stay
```
pmikolajczyk41, line 241: do we expect some tests for service?

pmikolajczyk41 (approved):
looks cool, however service seems now really not-trivial

finality-aleph/src/validator_network/manager/mod.rs
timorl: Hmm, maybe I don't understand the suggestion, but I don't think this simplifies anything – if I match on the pair I have to duplicate the nonempty printing code between `(0, _)` & `(_,0)` vs `(_,_)` cases, no?

I considered extracting separate methods for formatting incoming and outgoing outputs (essentially the two `match`es here), but I don't think this actually helps much either and, more importantly, seems relatively unrelated to your comment. ;p

pmikolajczyk41: yeah, let's leave it as it is

finality-aleph/src/validator_network/manager/mod.rs
timorl: Fair, I usually structure code in files in a way that makes more sense when read from beginning to end, and then the initial declaration of the struct serves as an introduction, but that's probably not really a good practice. Moved some of the code around to keep related stuff closer to each other.

finality-aleph/src/validator_network/manager/mod.rs
timorl: It will be overrriden by any successful connection attempt, until then the only difference removing would make is that the returned errors would be `PeerNotFound` rather than `ConnectionClosed` whenever we attempt to send something. I think the latter is marginally more informative, so I would leave it as is.

finality-aleph/src/validator_network/manager/mod.rs
timorl: We already did that in the `if` earlier, I don't think we have to check that again, it would be unnecessarily confusing.

finality-aleph/src/validator_network/service.rs
timorl: We do, but the logic for that is in other parts of the code – if `legacy_connected` contains a `peer_id` this means we are now convinced that we have to use the legacy protocol to communicate with this peer until further notice.

finality-aleph/src/validator_network/service.rs
timorl: There are the validator network integration tests in `testing`, although they do not check compatibility. Checking compatibility would be quite a lot of work and only for temporary gain, as the compat code should be removed soon (after the next update). It's probably better to just check compatibility manually before the update (which we should do anyway).

finality-aleph/src/validator_network/manager/legacy.rs
```diff
@@ -0,0 +1,322 @@
...
+        assert_eq!(manager.peer_addresses(&peer_id_b), None);
+        // remove peer
+        manager.remove_peer(&peer_id);
+        // try to get address of removed peer
+        assert_eq!(manager.peer_addresses(&peer_id), None);
+        // remove again
+        manager.remove_peer(&peer_id);
+        // remove unknown peer
+        manager.remove_peer(&peer_id_b);
+    }
```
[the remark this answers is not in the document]
timorl: Excellent question, I did not touch this code other than moving it so I have no idea... Removed.

finality-aleph/src/validator_network/manager/mod.rs
timorl: Fixed, thanks.

finality-aleph/src/validator_network/manager/mod.rs
timorl: The warning about incoming connections is there only because we expect this to be a common misconfiguration and it's _much_, **much**, I cannot stress _how_ much (seriously, I tried imagining it and it would probably require actually trying **hard**) harder to misconfigure network in a way that allows incoming connections, but breaks outgoing.

Added a comment as to why we warn about this and not the other way around, because you are probably right it's not obvious what our reasoning is _a priori_.

finality-aleph/src/validator_network/manager/mod.rs
timorl: Expanded the last sentence in the docs of this method, but I'm not sure that's what you meant, please let me know if more clarification is required.

finality-aleph/src/validator_network/service.rs
```diff
@@ -150,76 +154,122 @@ impl<D: Data, A: Data + Debug, ND: Dialer<A>, NL: Listener> Service<D, A, ND, NL
...
-                                Uninterested => warn!(target: "validator-network", "We connected to peer {} for unknown reasons.", peer_id),
-                                Added => info!(target: "validator-network", "New outgoing connection to peer {}.", peer_id),
-                                Replaced => info!(target: "validator-network", "Replaced outgoing connection to peer {}.", peer_id),
-                            },
-                            None => self.spawn_new_outgoing(peer_id, addresses, outgoing_result_for_parent.clone()),
+                Some((peer_id, maybe_data_for_network, connection_type)) = worker_results.next() => {
+                    use ConnectionType::*;
+                    let spawn_new_legacy_connection = match connection_type {
+                        LegacyIncoming => {
+                            self.manager.remove_peer(&peer_id);
```
maciejnems, line 231: Shouldn't all peer removal be based on whether `maybe_data_for_network.is_some()`? This would simplify a lot and seems like good logic.

finality-aleph/src/validator_network/service.rs
```diff
@@ -150,76 +154,122 @@ impl<D: Data, A: Data + Debug, ND: Dialer<A>, NL: Listener> Service<D, A, ND, NL
...
+                            // actually negotiated the new protocol, otherwise we should stay
+                            // with our previous guess.
+                            if maybe_data_for_network.is_some() {
+                                self.legacy_connected.remove(&peer_id);
+                            }
+                            false
+                        },
                     };
+                    if spawn_new_legacy_connection {
+                        match self.legacy_manager.peer_addresses(&peer_id) {
```
maciejnems, line 250: Maybe this could be moved and changed to
```
if self.legacy_connected.insert(peer_id.clone()) {
    Some(addresses) => self.spawn_new_outgoing(peer_id.clone(), addresses, result_for_parent.clone()),
    None => {
        self.legacy_connected.remove(&peer_id);
    },
}
```
This match with setting `spawn_new_legacy_connection` seems unnecessary

finality-aleph/src/validator_network/service.rs
```diff
@@ -150,76 +154,122 @@ impl<D: Data, A: Data + Debug, ND: Dialer<A>, NL: Listener> Service<D, A, ND, NL
...
-                        match maybe_data_for_network {
-                            Some(data_for_network) => match self.manager.add_outgoing(peer_id.clone(), data_for_network) {
-                                Uninterested => warn!(target: "validator-network", "We connected to peer {} for unknown reasons.", peer_id),
-                                Added => info!(target: "validator-network", "New outgoing connection to peer {}.", peer_id),
-                                Replaced => info!(target: "validator-network", "Replaced outgoing connection to peer {}.", peer_id),
-                            },
-                            None => self.spawn_new_outgoing(peer_id, addresses, outgoing_result_for_parent.clone()),
+                Some((peer_id, maybe_data_for_network, connection_type)) = worker_results.next() => {
+                    use ConnectionType::*;
+                    let spawn_new_legacy_connection = match connection_type {
```
maciejnems, line 229: This part of code is kinda confusing to me. If all peers connections would be changed only if `maybe_data_for_network.is_some()` we could move those removals and additions to `add_connection` inside the match there. And rename it to something like `handle_connection` Now there are two matches next to each other.

maciejnems (commented):
Looks Great, just some suggestions about logs and complexity of code.

finality-aleph/src/validator_network/manager/mod.rs
```diff
@@ -41,180 +44,168 @@ pub enum AddResult {
...
 impl Display for ManagerStatus {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), FmtError> {
-        if self.wanted_peers == 0 {
+        let wanted_incoming = self.wanted_incoming();
+        let wanted_outgoing = self.wanted_outgoing();
+        if wanted_incoming + wanted_outgoing == 0 {
             return write!(f, "not maintaining any connections; ");
         }
 
-        write!(f, "target - {:?} connections; ", self.wanted_peers)?;
```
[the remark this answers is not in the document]
timorl: Blah, true, created a task for that.

finality-aleph/src/validator_network/service.rs
```diff
@@ -150,76 +154,122 @@ impl<D: Data, A: Data + Debug, ND: Dialer<A>, NL: Listener> Service<D, A, ND, NL
...
+                        None => if let Some(addresses) = self.peer_addresses(&peer_id) {
+                            self.spawn_new_outgoing(peer_id, addresses, result_for_parent.clone());
+                        }
+                    }
                 },
                 // periodically reporting what we are trying to do
                 _ = status_ticker.tick() => {
-                    info!(target: "validator-network", "Validator Network status: {}", self.manager.status_report())
+                    info!(target: "validator-network", "Validator Network status: {}", self.manager.status_report());
+                    debug!(target: "validator-network", "Validator Network legacy status: {}", self.legacy_manager.status_report());
```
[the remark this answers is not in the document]
timorl: I don't like there would be a constant log of "not maintaining any connections" through most of the lifetime of this update, which would make people ask unnecessary questions. This log will be mostly relevant early during the rolling update, and for that time we can turn `debug` on.

finality-aleph/src/validator_network/service.rs
```diff
@@ -150,76 +154,122 @@ impl<D: Data, A: Data + Debug, ND: Dialer<A>, NL: Listener> Service<D, A, ND, NL
...
+                                self.legacy_connected.remove(&peer_id);
+                            },
+                        }
+                    }
+                    use AddResult::*;
+                    match maybe_data_for_network {
+                        Some(data_for_network) => match self.add_connection(peer_id.clone(), data_for_network, connection_type) {
+                            Uninterested => warn!(target: "validator-network", "Established connection with peer {} for unknown reasons.", peer_id),
+                            Added => info!(target: "validator-network", "New connection with peer {}.", peer_id),
+                            Replaced => info!(target: "validator-network", "Replaced connection with peer {}.", peer_id),
```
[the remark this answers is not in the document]
timorl: Eh~, I guess we could but I think we can get this information from other sources sufficiently well.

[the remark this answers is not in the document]
maciejnems: But not on info level, right?

finality-aleph/src/validator_network/service.rs
timorl: This first match only handles the marking of something as legacy, see changes mentioned in previous comments.

finality-aleph/src/validator_network/service.rs
timorl: No, because the spawning of new threads depends on this status, and that is in the `None` branch. :/ I moved some of the logic around and added some comments, perhaps now it's more readable. I also think I fixed a (minor but embarrassing) bug on the way.

maciejnems (approved):

PR #723, opened 2022-11-10
A0-1140: e2e tests for simple_dex

obrok:
# Description

Adds an e2e test for the simple dex. 

## Type of change

- Tests (non-breaking change which tests existing functionality)

### DIFF
--- a/contracts/scripts/test.sh
+++ b/contracts/scripts/test.sh
@@ -7,10 +7,12 @@ CONTRACTS_PATH=$(pwd)/contracts
 EARLY_BIRD_SPECIAL=$(jq --raw-output ".early_bird_special" < "$CONTRACTS_PATH"/addresses.json)
 THE_PRESSIAH_COMETH=$(jq --raw-output ".the_pressiah_cometh" < "$CONTRACTS_PATH"/addresses.json)
 BACK_TO_THE_FUTURE=$(jq --raw-output ".back_to_the_future" < "$CONTRACTS_PATH"/addresses.json)
+SIMPLE_DEX=$(jq --raw-output ".simple_dex" < "$CONTRACTS_PATH"/addresses.json)
 
 pushd "$E2E_PATH"
 
 RUST_LOG="aleph_e2e_client=info" cargo run --release -- \
+  --test-cases simple_dex \
   --test-cases marketplace \
   --test-cases button_game_reset \
   --test-cases early_bird_special \
@@ -19,9 +21,11 @@ RUST_LOG="aleph_e2e_client=info" cargo run --release -- \
   --early-bird-special "$EARLY_BIRD_SPECIAL" \
   --the-pressiah-cometh "$THE_PRESSIAH_COMETH" \
   --back-to-the-future "$BACK_TO_THE_FUTURE" \
+  --simple-dex "$SIMPLE_DEX" \
   --button-game-metadata ../contracts/button/target/ink/metadata.json \
   --ticket-token-metadata ../contracts/ticket_token/target/ink/metadata.json \
   --reward-token-metadata ../contracts/game_token/target/ink/metadata.json \
-  --marketplace-metadata ../contracts/marketplace/target/ink/metadata.json
+  --marketplace-metadata ../contracts/marketplace/target/ink/metadata.json \
+  --simple-dex-metadata ../contracts/simple_dex/target/ink/metadata.json
 
 exit $?
--- a/e2e-tests/src/test/button_game/helpers.rs
+++ b/e2e-tests/src/test/button_game/helpers.rs
@@ -19,8 +19,10 @@ use log::{info, warn};
 use rand::Rng;
 use sp_core::Pair;
 
-use super::contracts::{ButtonInstance, PSP22TokenInstance};
-use crate::{test::button_game::contracts::MarketplaceInstance, Config};
+use super::contracts::{
+    ButtonInstance, MarketplaceInstance, PSP22TokenInstance, SimpleDexInstance,
+};
+use crate::Config;
 
 /// A wrapper around a KeyPair for purposes of converting to an account id in tests.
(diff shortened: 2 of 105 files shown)


contracts/scripts/test.sh
```diff
@@ -19,9 +21,11 @@ RUST_LOG="aleph_e2e_client=info" cargo run --release -- \
   --early-bird-special "$EARLY_BIRD_SPECIAL" \
   --the-pressiah-cometh "$THE_PRESSIAH_COMETH" \
   --back-to-the-future "$BACK_TO_THE_FUTURE" \
```
fbielejec, line 23: can we add wAzero contract here too?

e2e-tests/src/test/button_game/mod.rs
```diff
@@ -5,18 +5,115 @@ use anyhow::Result;
...
+    let expected_output = dex.out_given_in(account_conn, token1, token2, initial_amount, None)?;
+    assert!(expected_output > 0);
+
+    token1.approve(account_conn, &dex.into(), initial_amount)?;
+    dex.swap(
+        account_conn,
+        token1,
+        initial_amount,
+        token2,
+        expected_output * 9 / 10,
```
fbielejec, line 89: this could explicitly be called amount_with_slippage_10 or something to that effect

e2e-tests/src/test/button_game/mod.rs
```diff
@@ -5,18 +5,115 @@ use anyhow::Result;
...
+        authority_conn,
+        &[
+            (token1, mega(3000)),
+            (token2, mega(5000)),
+            (token3, mega(10000)),
+        ],
+    )?;
+
+    let more_than_liquidity = mega(1_000_000);
+    assert!(dex
```
fbielejec, line 74: would be nice to add custom messages by assertions (eg "should fail due to not enough liquidity")

e2e-tests/src/test/button_game/mod.rs
```diff
@@ -5,18 +5,115 @@ use anyhow::Result;
...
+            <= 1
+    );
+
+    token1.approve(account_conn, &dex.into(), balance_after)?;
+    dex.swap(account_conn, token1, balance_after, token3, mega(90))?;
+    assert_recv_id(&mut events, "Swapped");
+    let balance_token3 = token3.balance_of(&conn, &account.public().into())?;
+    token3.approve(account_conn, &dex.into(), balance_token3)?;
+    dex.swap(account_conn, token3, balance_token3, token1, mega(90))?;
+    refute_recv_id(&mut events, "Swapped");
```
fbielejec, line 112: similarly, can you pass a reason to assert(not())

e2e-tests/src/test/button_game/mod.rs
```diff
@@ -5,18 +5,115 @@ use anyhow::Result;
...
+    token1.approve(account_conn, &dex.into(), initial_amount)?;
+    dex.swap(
+        account_conn,
+        token1,
+        initial_amount,
+        token2,
+        expected_output * 9 / 10,
+    )?;
+    assert_recv_id(&mut events, "Swapped");
+    assert!(token2.balance_of(&conn, &account.public().into())? == expected_output);
```
fbielejec, line 92: you could also add a test case where swap tx is rejected due to an excessive slippage

e2e-tests/src/test/button_game/mod.rs
```diff
@@ -5,18 +5,115 @@ use anyhow::Result;
...
+        mut events,
+    } = setup_dex_test(config)?;
+    let authority_conn = &sign(&conn, &authority);
+    let account_conn = &sign(&conn, &account);
+    let token1 = token1.as_ref();
+    let token2 = token2.as_ref();
+    let token3 = token3.as_ref();
+    let dex = dex.as_ref();
+
+    dex.add_swap_pair(authority_conn, token1.into(), token2.into())?;
```
fbielejec, line 48: pairs with wAzero shoudl be covered too, I reckon this is what will see most traffic

Marcin-Radecki (approved):

e2e-tests/src/test/button_game/helpers.rs
```diff
@@ -118,6 +125,78 @@ pub(super) struct ButtonTestContext {
...
+    let account = random_account();
+
+    let dex = Arc::new(SimpleDexInstance::new(config)?);
+    let token1 =
+        reward_token_for_button(config, &conn, &config.test_case_params.early_bird_special)?;
+    let token2 =
+        reward_token_for_button(config, &conn, &config.test_case_params.the_pressiah_cometh)?;
+    let token3 =
+        reward_token_for_button(config, &conn, &config.test_case_params.back_to_the_future)?;
+
```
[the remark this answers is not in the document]
obrok: The way the test is written right now, it wouldn't really add any value - the tokens 1-3 are treated as any generic PSP22.

e2e-tests/src/test/button_game/mod.rs
obrok: :+1:

e2e-tests/src/test/button_game/mod.rs
obrok: :+1:

e2e-tests/src/test/button_game/mod.rs
obrok: As I said in another comment - I don't think it matters at all for these tests, because the tokens are treated generically. I guess you're more worried about testing wAzero itself? I would say that should be a separate test scenario, where we mostly test wrapping and unwrapping.

fbielejec: we can do it like that - it would boil down to testing wrap/unwrap so not too involved I suppose

obrok: I'll do it in another PR

PR #724, opened 2022-11-10
A0-1140: Leave dust in the dex

obrok:
# Description

Makes sure dust caused by rounding is not paid out when making dex swaps.

I added property tests for this. Not sure how the team feels about property tests in general, but they felt very appropriate for this.

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/contracts/simple_dex/lib.rs
+++ b/contracts/simple_dex/lib.rs
@@ -153,6 +153,12 @@ mod simple_dex {
             let this = self.env().account_id();
             let caller = self.env().caller();
 
+            let balance_token_out = self.balance_of(token_out, this)?;
+            if balance_token_out < min_amount_token_out {
+                // throw early if we cannot support this swap anyway due to liquidity being too low
+                return Err(DexError::NotEnoughLiquidityOf(token_out));
+            }
+
             let swap_pair = SwapPair::new(token_in, token_out);
             if !self.swap_pairs.contains(&swap_pair) {
                 return Err(DexError::UnsupportedSwapPair(swap_pair));
@@ -163,12 +169,7 @@ mod simple_dex {
                 return Err(DexError::InsufficientAllowanceOf(token_in));
             }
 
-            let amount_token_out = self.out_given_in(
-                token_in,
-                token_out,
-                amount_token_in,
-                Some(min_amount_token_out),
-            )?;
+            let amount_token_out = self.out_given_in(token_in, token_out, amount_token_in)?;
 
             if amount_token_out < min_amount_token_out {
                 // thrown if too much slippage occured before this tx gets executed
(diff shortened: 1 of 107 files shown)


deuszx:
> Not sure how the team feels about property tests in general,

This team member loves them.

contracts/simple_dex/lib.rs
```diff
@@ -373,14 +373,27 @@ mod simple_dex {
...
 
+            Self::_out_given_in(
+                amount_token_in,
+                balance_token_in,
+                balance_token_out,
+                self.swap_fee_percentage,
+            )
+        }
+
+        fn _out_given_in(
```
deuszx, line 389: Cab we move the comment with the formula from the public `fn out_given_in` to here?

contracts/simple_dex/lib.rs
```diff
@@ -513,4 +528,29 @@ mod simple_dex {
...
+
+        proptest! {
+            #[test]
+            fn rounding_benefits_dex(
+                balance_token_a in 1..1000u128,
+                balance_token_b in 1..1000u128,
+                pay_token_a in 1..1000u128
+            ) {
+                let get_token_b =
+                    SimpleDex::_out_given_in(pay_token_a, balance_token_a, balance_token_b, 0).unwrap();
```
deuszx, line 546: I think we shoud play with different `swap_fee_percentage` as that may affect the results.

contracts/simple_dex/lib.rs
```diff
@@ -399,6 +412,8 @@ mod simple_dex {
 
             balance_token_out
                 .checked_sub(op4)
+                // If the division is not even, leave the 1 unit of dust in the exchange instead of paying it out.
+                .and_then(|result| result.checked_sub(if op3 % op2 > 0 { 1 } else { 0 }))
```
deuszx, line 416: Why are we always subtracting 1, even if the remainder of `op3 / op2` divitions (`op3 % op2`) may be higher than that.

contracts/simple_dex/lib.rs
```diff
@@ -373,14 +373,27 @@ mod simple_dex {
...
+                self.swap_fee_percentage,
+            )
+        }
+
+        fn _out_given_in(
+            amount_token_in: Balance,
+            balance_token_in: Balance,
+            balance_token_out: Balance,
+            swap_fee_percentage: Balance,
+        ) -> Result<Balance, DexError> {
```
deuszx, line 394: I wonder - can it be the case that the returned `Balance` here (the amount of tokens the caller will be paid out in exchange for his `amount_token_in`) is less than `min_amount_token_out` he requested? We check only whether `balance_token_out > min_amount_token_in` but can it be the case that our formula `_out_given_in` will give less than that in the end? After the swap fee, etc.

contracts/simple_dex/lib.rs
fbielejec: mo nor necessary - there will most likely never be a swap fee charged. Swap fee in most DEXes goes to LPs and there is only one here (us)

contracts/simple_dex/lib.rs
fbielejec: wont be a part of the interface docstring if moved

contracts/simple_dex/lib.rs
fbielejec: +1, maybe indeed subtract the remainder

contracts/simple_dex/lib.rs
fbielejec: as you wish., but I doubt it will ever be non-zero

deuszx: It may be the case for us but people may decide to use it as starting point for their own contracts where the fee will be non-zero and that will influence the "dust", etc.

contracts/simple_dex/lib.rs
deuszx: @fbielejec pointed me to https://github.com/Cardinal-Cryptography/aleph-node/blob/7e5d524caa933dae6331465d146d1e40b86ccb7f/contracts/simple_dex/lib.rs#L173 which I totally missed.

contracts/simple_dex/lib.rs
deuszx: OK, it's just annoying to jump between public to private methods when verifying it.

contracts/simple_dex/lib.rs
obrok: So my thinking here was:

1. The "real" value of `op3 / op3` is between `op4` and `op4 + 1`.
2. The whole number division always gives us `op4`.
3. So to be more conservative, but as close to the "real" value as possible we use `op4 + 1` when the division is not even.

I don't think `op3 % op2` factors into this at all. Am I completely missing something?

fbielejec (approved):

PR #728, opened 2022-11-15
A0-1535 Revert pallet contracts from runtime and rpc endpoint

Marcin-Radecki:
# Description

This change reverts `pallet-contracts` from Mainnet branch `release-8-mainnet`. It contains changes to both runtime and host because the compilation of those seemingly separated code pieces is coupled.  `cliain` and `aleph-node` was not touched as it does not affect the compilation of `aleph-node` and `aleph-runtime` artifacts.

Original commit: https://github.com/Cardinal-Cryptography/aleph-node/commit/731bba3e95ba2a4f2e4a98eddf3d04cce8f65998

Testing done:
* start nodes locally, connect polkadot wallet -> no Contracts in Developers tab,
* `./scripts/run_checks_on_excluded_packages.sh` passed
* ` ./scripts/run_checks_on_aleph_node.sh` passed

## Type of change

# Checklist:

### DIFF
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -20,16 +20,12 @@ pub use frame_support::{
 };
 use frame_support::{
     sp_runtime::Perquintill,
-    traits::{ConstU32, ConstU64, EqualPrivilegeOnly, SortedMembers, U128CurrencyToVote},
+    traits::{ConstU32, EqualPrivilegeOnly, SortedMembers, U128CurrencyToVote},
     weights::constants::WEIGHT_PER_MILLIS,
     PalletId,
 };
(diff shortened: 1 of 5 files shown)


bin/runtime/src/lib.rs
```diff
@@ -108,10 +104,10 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("aleph-node"),
     impl_name: create_runtime_str!("aleph-node"),
     authoring_version: 1,
-    spec_version: 38,
+    spec_version: 39,
```
mike1729, line 107: I'd leave it at 38 so that we cannot send this runtime by mistake to testnet.

mike1729 (changes requested):

PR #729, opened 2022-11-15
It's not a mixer (2/3)

pmikolajczyk41:
We add `withdraw` method.


contracts/blender/src/contract.rs
```diff
@@ -33,7 +34,15 @@ mod blender {
...
-        note: Note,
+        leaf_idx: u32,
+    }
+
+    #[ink(event)]
+    pub struct Withdrawn {
+        #[ink(topic)]
+        token_id: TokenId,
+        value: TokenAmount,
+        recipient: AccountId,
```
deuszx, line 45: we could index recipient as well.

pmikolajczyk41: done

contracts/blender/src/contract.rs
```diff
@@ -82,22 +94,114 @@ mod blender {
...
+            self.verify_merkle_root(merkle_root)?;
+            self.verify_nullifier(nullifier)?;
+            self.verify_withdrawal(token_id, value, merkle_root, nullifier, new_note, proof)?;
+
+            let leaf_idx = self.append_note(new_note)?;
+            self.nullifiers.insert(nullifier, &());
+
+            self.withdraw_funds(token_id, value, fee_for_caller, recipient)?;
+
+            self.env().emit_event(Withdrawn {
```
deuszx, line 133: I suggest using the approach described [here](https://github.com/deuszx/bulletin-board-example/blob/d6f00049f6c07a113d0e46b6c66825f3f382f236/contracts/README.md#creating-and-emitting-events).

pmikolajczyk41: done

contracts/blender/src/contract.rs
```diff
@@ -185,41 +287,98 @@ mod blender {
...
+            }
+        }
+
+        fn transfer(
+            &self,
+            token_contract: AccountId,
+            value: TokenAmount,
+            recipient: AccountId,
+        ) -> Result<()> {
+            build_call::<super::blender::Environment>()
```
deuszx, line 371: You could use the *Ref pattern here instead. Calls this simple are supported.

PR #732, opened 2022-11-16
Inline MerkleTree

pmikolajczyk41:
We avoid nested struct (we don't have any guarantees/understanding whether it will be initialized correctly). KISS


DamianStraszak (approved):
I love it :)

fbielejec (approved):
OK I GUESS

PR #734, opened 2022-11-16
A0-1586: At most one discovery message

timorl:
# Description

We always send at most one discovery message at a time in many places, so make that official by making them a `Option` instead of `Vec`.

## Type of change

# Checklist:


kostekIV (approved):
Nice!

pmikolajczyk41 (approved):

PR #735, opened 2022-11-17
Reading data from contract

pmikolajczyk41:
We introduce two new messages for:
 - reading merkle path from specified leaf index
 - checking whether a nullifier has been already used

image

### DIFF
--- a/contracts/blender/src/contract.rs
+++ b/contracts/blender/src/contract.rs
@@ -53,11 +53,15 @@ mod blender {
 
     /// Describes a path from a leaf to the root.
     ///
-    /// It is represented as sibling hashes list (bottom to top). The pairs are always of form
-    /// `(left hash, right hash)` with the exception of the last pair, which is just
-    /// `(root hash, root hash)`. The requested leaf is put in the first pair (check index parity
-    /// to find out the coordinate).
-    pub type MerklePath = Vec<(Hash, Hash)>;
+    /// It is represented as a hash sequence from bottom to top:
+    /// - the first will be the 'lefter' of our leaf and its sibling
+    /// - the second element will be righter of our leaf and its sibling
+    /// - the third element will be lefter of their parent and its sibling
+    /// - the fourth element will be righter of their parent and its sibling
+    /// - ...
+    ///
+    /// Path does not contain the root.
+    pub type MerklePath = Vec<Hash>;
 
     #[ink(storage)]
     #[derive(SpreadAllocate)]
@@ -179,14 +183,11 @@ mod blender {
             let mut path: MerklePath = vec![];
             while current_idx > 1 {
                 let (left_sibling_idx, right_sibling_idx) = Self::get_sibling_indices(current_idx);
-                let left_hash = self.tree_value(left_sibling_idx);
-                let right_hash = self.tree_value(right_sibling_idx);
-                path.push((left_hash, right_hash));
+                path.push(self.tree_value(left_sibling_idx));
+                path.push(self.tree_value(right_sibling_idx));
                 current_idx /= 2;
             }
 
-            let root_hash = self.current_root();
-            path.push((root_hash, root_hash));
             Some(path)
         }
 
--- a/contracts/blender/src/contract.rs
+++ b/contracts/blender/src/contract.rs
@@ -203,8 +203,8 @@ mod blender {
             Some(MerklePath {
                 leaf_is_right_child: leaf_idx & 1 == 1,
                 leaf_sibling: self.tree_value(leaf_idx ^ 1),
-                auth_path: vec![],
-                path: vec![],
+                auth_path,
+                path,
             })
         }
 
--- a/contracts/blender/src/contract.rs
+++ b/contracts/blender/src/contract.rs
@@ -36,6 +36,7 @@ mod blender {
         token_id: TokenId,
         value: TokenAmount,
         leaf_idx: u32,
+        note: Note,
     }
 
     #[ink(event)]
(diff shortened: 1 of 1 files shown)


contracts/blender/src/contract.rs
```diff
@@ -160,6 +168,34 @@ mod blender {
             self.current_root()
         }
 
+        /// Retrieve the path from the leaf to the root. `None` if the leaf does not exist.
+        #[ink(message, selector = 4)]
+        pub fn merkle_path(&self, leaf_idx: u32) -> Option<MerklePath> {
+            if self.max_leaves > leaf_idx || leaf_idx >= self.next_free_leaf {
```
deuszx, line 174: `self.max_leaves > leaf_idx` shouldn't be necessary as the second condition already covers it, right?

pmikolajczyk41: nope, we want to ensure, that `leaf_idx` belongs to the interval `[self.max_leaves; self.next_free_leaf)`, so we have to check both boundaries

deuszx: Ah, of course. `leaf_idx` has to be less than `self.max_leaves` b/c at the position `0..self.max_leaves` we have intermediate nodes of the trie.

The smaller value (`leaf_idx`) being on the right side threw me off. My bad.

contracts/blender/src/contract.rs
```diff
@@ -51,6 +51,14 @@ mod blender {
     type Result<T> = core::result::Result<T, BlenderError>;
     type Event = <Blender as ContractEventBase>::Type;
 
+    /// Describes a path from a leaf to the root.
+    ///
+    /// It is represented as sibling hashes list (bottom to top). The pairs are always of form
+    /// `(left hash, right hash)` with the exception of the last pair, which is just
+    /// `(root hash, root hash)`. The requested leaf is put in the first pair (check index parity
+    /// to find out the coordinate).
+    pub type MerklePath = Vec<(Hash, Hash)>;
```
deuszx, line 60: It should be possible to represent this as just `Vec<Hash>` since leaves/nodes are at deterministic positions:
* head will be our starting leaf at `leaf_idx`
* second element would be its sibiling
* third element is sibiling of their parent 
* fourth element sibiling of the previous' elements parent
* etc.

I also wouldn't send the `root_hash` as it's not necessarily required - it is recoverable from the merkle path and the leaf.

And for root computation we would [`reduce`](https://doc.rust-lang.org/nightly/core/iter/trait.Iterator.html#method.reduce) over the collection with `f` being the hashing function.

Or am I wrong again?

pmikolajczyk41: generally fair enough - I will flatten this struct

two notes however:
 - head will be the 'lefter' of our leaf and its sibling
 - second element will be `righter` of our leaf and its sibling
 - third element will be `lefter` of their parent and its sibling
 - fourth element will be `righter` of their parent and its sibling
 
and `reduce` would truly be cute and elegant, but eventually we won't use it (this list, maybe somehow transformed, will be passed to the circuit and there Jedrzej will use some ready code from arkworks)

pmikolajczyk41: done

deuszx: 🙈

contracts/blender/src/contract.rs
```diff
@@ -46,11 +47,24 @@ mod blender {
...
+        new_note: Note,
     }
 
     type Result<T> = core::result::Result<T, BlenderError>;
     type Event = <Blender as ContractEventBase>::Type;
 
+    /// Describes a path from a leaf to the root.
+    ///
+    /// It is represented as a hash sequence from bottom to top:
+    /// - the first will be the 'lefter' of our leaf and its sibling
```
deuszx, line 59: Does _lefter_ here is meant to mean "the left element of the two - our leaf and its sibiling"?

Similarly, _righter_ below.

contracts/blender/src/contract.rs
```diff
@@ -160,6 +176,31 @@ mod blender {
...
+                path.push(self.tree_value(right_sibling_idx));
+                current_idx /= 2;
+            }
+
+            Some(path)
+        }
+
+        /// Check whether `nullifier` has been already used.
+        #[ink(message, selector = 5)]
+        pub fn observed_nullifier(&self, nullifier: Nullifier) -> bool {
```
deuszx, line 200: Personally, I would use `contains_*` instead since it's a standard for map-like types.

deuszx (approved):

contracts/blender/src/contract.rs
pmikolajczyk41: our parent has two children - left child and right child; so the 'lefter of `a` and `b`' is the one who is the left child

deuszx: Yes, that's what I meant 👍

contracts/blender/src/contract.rs
```diff
@@ -160,6 +176,31 @@ mod blender {
...
+                return None;
+            }
+
+            let mut current_idx = leaf_idx;
+            let mut path: MerklePath = vec![];
+            while current_idx > 1 {
+                let (left_sibling_idx, right_sibling_idx) = Self::get_sibling_indices(current_idx);
+                path.push(self.tree_value(left_sibling_idx));
+                path.push(self.tree_value(right_sibling_idx));
+                current_idx /= 2;
```
DamianStraszak, line 192: You don't need to add both to the path, just the one which does not equal `current_idx`. Given the leaf, and the ones I propose, you can reconstruct the path.

DamianStraszak (changes requested):

contracts/blender/src/contract.rs
pmikolajczyk41: optimized

contracts/blender/src/contract.rs
pmikolajczyk41: changed

DamianStraszak (approved):

PR #736, opened 2022-11-17
check liquidity in swaps only

fbielejec:
# Description

Refactored out_given_in to be a pure query call

### DIFF
--- a/e2e-tests/src/test/button_game/mod.rs
+++ b/e2e-tests/src/test/button_game/mod.rs
@@ -76,6 +76,10 @@ pub fn simple_dex(config: &Config) -> Result<()> {
         "out_given_in should always return"
     );
 
+    let more_than_liquidity = mega(1_000_000);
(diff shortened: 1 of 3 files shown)


deuszx (approved):
Simple yet effective :+1:

Can't comment on the e2e tests so leave that to @obrok

e2e-tests/src/test/button_game/mod.rs
```diff
@@ -71,14 +71,18 @@ pub fn simple_dex(config: &Config) -> Result<()> {
         ],
     )?;
 
+    assert!(
+        dex.out_given_in(account_conn, token1, token2, 100).is_ok(),
+        "out_given_in should always return"
+    );
```
obrok, line 77: This doesn't really make sense - the `Err` variant from this call signifies some connection or serialization error, not an `Err` variant received from the contract call. If the contract returns a `Result`, then a `contract_read` will return `Result<Result<_>>`.

```suggestion
```

obrok (approved):

PR #737, opened 2022-11-17
A0-1140: e2e tests for wrapped_azero

obrok:
# Description

Added an e2e scenario for wrapped_azero.

# Checklist:

- I have added tests

### DIFF
--- a/e2e-tests/src/test/button_game/contracts.rs
+++ b/e2e-tests/src/test/button_game/contracts.rs
@@ -303,3 +303,55 @@ impl From<&MarketplaceInstance> for AccountId {
         marketplace.contract.address().clone()
     }
 }
+
+#[derive(Debug)]
+pub struct WAzeroInstance {
+    contract: ContractInstance,
+}
+
+impl WAzeroInstance {
+    pub fn new(config: &Config) -> Result<Self> {
+        let wazero_address = config
+            .test_case_params
+            .wrapped_azero
+            .clone()
+            .context("Wrapped AZERO address not set.")?;
+        let wazero_address = AccountId::from_string(&wazero_address)?;
+        let metadata_path = config
+            .test_case_params
+            .wrapped_azero_metadata
+            .clone()
+            .context("Wrapped AZERO metadata path not set.")?;
+
+        Ok(Self {
+            contract: ContractInstance::new(wazero_address, &metadata_path)?,
+        })
+    }
+
+    pub fn wrap(&self, conn: &SignedConnection, value: Balance) -> Result<()> {
+        self.contract.contract_exec_value0(conn, "wrap", value)
+    }
+
+    pub fn unwrap(&self, conn: &SignedConnection, amount: Balance) -> Result<()> {
+        self.contract
+            .contract_exec(conn, "unwrap", &[amount.to_string()])
+    }
+
+    pub fn balance_of<C: AnyConnection>(&self, conn: &C, account: AccountId) -> Result<Balance> {
+        self.contract
+            .contract_read(conn, "PSP22::balance_of", &[account.to_string()])?
+            .try_into()
+    }
+}
+
+impl<'a> From<&'a WAzeroInstance> for &'a ContractInstance {
+    fn from(dex: &'a WAzeroInstance) -> Self {
+        &dex.contract
+    }
+}
+
+impl<'a> From<&'a WAzeroInstance> for AccountId {
+    fn from(dex: &'a WAzeroInstance) -> Self {
+        dex.contract.address().clone()
+    }
+}
--- a/e2e-tests/src/test/button_game/contracts.rs
+++ b/e2e-tests/src/test/button_game/contracts.rs
@@ -345,13 +345,13 @@ impl WAzeroInstance {
 }
 
 impl<'a> From<&'a WAzeroInstance> for &'a ContractInstance {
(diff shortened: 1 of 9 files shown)


Marcin-Radecki (approved):

e2e-tests/src/test/button_game/contracts.rs
```diff
@@ -303,3 +303,55 @@ impl From<&MarketplaceInstance> for AccountId {
...
+    }
+
+    pub fn unwrap(&self, conn: &SignedConnection, amount: Balance) -> Result<()> {
+        self.contract
+            .contract_exec(conn, "unwrap", &[amount.to_string()])
+    }
+
+    pub fn balance_of<C: AnyConnection>(&self, conn: &C, account: AccountId) -> Result<Balance> {
+        self.contract
+            .contract_read(conn, "PSP22::balance_of", &[account.to_string()])?
```
deuszx, line 342: This is something I was wondering about just yesterday - how does it work that for openbrush's contracts we have to prefix it with the `trait` name while for our contracts we don't (see `"wrap"`/`"unwrap"`)?

e2e-tests/src/test/button_game/contracts.rs
```diff
@@ -303,3 +303,55 @@ impl From<&MarketplaceInstance> for AccountId {
...
+            .try_into()
+    }
+}
+
+impl<'a> From<&'a WAzeroInstance> for &'a ContractInstance {
+    fn from(dex: &'a WAzeroInstance) -> Self {
+        &dex.contract
```
deuszx, line 349: This isn't a `dex` though - copy/paste error?

e2e-tests/src/test/button_game/mod.rs
```diff
@@ -9,15 +9,57 @@ use sp_core::Pair;
...
 mod contracts;
 mod helpers;
 
+/// Test wrapped azero
+///
+/// The scenario:
+///
+/// 1. Wraps some azero and checks that the PSP22 balance increased accordingly.
+/// 2. Unwraps half of the amount, checks that some wrapped funds remained while the rest has been returned to azero,
+///    less fees.
```
deuszx, line 28:
```suggestion
///    minus fees.
```

deuszx (approved):

e2e-tests/src/test/button_game/contracts.rs
obrok: In general, you can set the selector to anything you want for these calls, the name of the function is just the default. I'm not seeing the `PSP22::` prefix explicitly mentioned anywhere in the openbrush sources... I'm guessing that's how ink handles `impl Trait`.

deuszx: My understanding was that we will take this string and iterate over all available methods in the `metadata.json` and check if there's any matching. This isn't really hashing and comparing selectors b/c I saw usages of this functionality of `aleph_client` where a selector was set to sth like `#[ink(selector = 7)]` and then we called it like `contract.read("withdraw")`.

obrok: I guess it's not really the selector, but how the method is named in `metadata.json`. The rest of my comment stands, though.

obrok: The other option is that it comes from macro magic in openbrush. I don't think it matters that much either way.

obrok: 1. What do you mean by "of this tool" here?
2. IMO the answer is: users should look at `metadata.json` (or more likely use some tool to look at it, like `polkadot.js`) and call the methods by whatever name appears in there. As far as I understand it, someone can craft any `metadata.json` they want for any contract, there can even be multiple valid `metadata.json` with different names for a single contract. At the point when the contract is deployed, it is somewhat divorced from its source code, so talking about concepts like method names and traits and so on is meaningless in that context.
3. We already talked about it on some dailies, and I think the future is: a different tool that generates a typed lib for interacting with a contract based on `metadata.json`. Or in fact a lib in any language of your choosing via plugins or somesuch. In that case it's again based on whatever is written in a given `metadata.json`, so in a sense it can have little relation to the source code of the contract.

deuszx: "This tool" == `aleph_client` and its features to integrate with contracts. 

> At the point when the contract is deployed, it is somewhat divorced from its source code, so talking about concepts like method names and traits and so on is meaningless in that context.

Exactly. That's why I think specifying methods on a contract via method names is not very future-proof. We should be using selectors as those are the "ultimate" identifiers of the contract's methods and having to look into `metadata.json` is very cumbersome, requires a rather in-depth knowledge of the system.

And finally, about your third point, I totally agree. It solves the previous two.

obrok: > Exactly. That's why I think specifying methods on a contract via method names is not very future-proof. We should be using selectors as those are the "ultimate" identifiers of the contract's methods and having to look into metadata.json is very cumbersome, requires a rather in-depth knowledge of the system.

Perhaps a final point on this is that there is space in `aleph_client` for both kinds of invocation: `contract_exec_by_selector` and `contract_exec_by_method_name`. I feel the current version is convenient for the thing I was doing, especially since it maps 1-to-1 to what you see in `polkadot.js` as you're experimenting with a given contract. However, if you want a more robust integration, you might want to make all calls via selector.

PR #739, opened 2022-11-18
Blender: ownable

pmikolajczyk41:
Use `Openbrush::Ownable` instead of manual control

image


deuszx (approved):
If you don't add
```rust
impl Ownable for Blender { }
```
you won't be able to control the permissions.

pmikolajczyk41:
@deuszx 
image

deuszx:
@pmikolajczyk41 you need to add
```rust
#![feature(min_specialization)]
```
at the top of your contract file 🙈 
Don't even ask me how many hours I lost on that.

PR #740, opened 2022-11-18
A0-1543: Enhance interfaces

maciejzelaszczyk:
# Description

Small PR to extract common code from `pallet_snarcos` and `snarcos-extension`.

## Type of change

# Checklist:


deuszx:
> Small PR

👀 

image

pmikolajczyk41 (changes requested):
My greatest concern is about not having error enum in the pallet, which is handled by substrate macros as any other one. I suppose we can experiment with this approach and see what's going to happen.

PR #741, opened 2022-11-18
A0-1591: Remove unnecessary abstraction layers in network

timorl:
# Description

Another refactor to make incoming changes to multiaddresses less painful. This one removes some abstraction layers that are no longer useful and somewhat decouples `network` from `network::manager` – we should probably decouple them more, but that's better left for another task.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/network/io.rs
+++ b/finality-aleph/src/network/io.rs
@@ -1,33 +1,42 @@
 use futures::channel::mpsc;
 
 use crate::network::{
-    manager::NetworkData, ConnectionManagerIO, Data, Multiaddress, NetworkServiceIO as NetworkIo,
-    SessionManagerIO,
+    manager::{DataInSession, VersionedAuthentication},
+    ConnectionManagerIO, Data, Multiaddress, NetworkServiceIO as NetworkIO, SessionManagerIO,
 };
 
-type NetworkServiceIO<D, M> = NetworkIo<NetworkData<D, M>, M>;
+type AuthenticationNetworkIO<D, M> = NetworkIO<VersionedAuthentication<M>, DataInSession<D>, M>;
 
 pub fn setup<D: Data, M: Multiaddress + 'static>() -> (
     ConnectionManagerIO<D, M>,
-    NetworkServiceIO<D, M>,
+    AuthenticationNetworkIO<D, M>,
     SessionManagerIO<D>,
 ) {
     // Prepare and start the network
     let (commands_for_network, commands_from_io) = mpsc::unbounded();
+    let (data_for_network, data_from_user) = mpsc::unbounded();
     let (messages_for_network, messages_from_user) = mpsc::unbounded();
     let (commands_for_service, commands_from_user) = mpsc::unbounded();
     let (messages_for_service, commands_from_manager) = mpsc::unbounded();
+    let (data_for_user, data_from_network) = mpsc::unbounded();
     let (messages_for_user, messages_from_network) = mpsc::unbounded();
 
     let connection_io = ConnectionManagerIO::new(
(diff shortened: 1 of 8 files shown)


finality-aleph/src/network/io.rs
```diff
@@ -1,33 +1,42 @@
...
 pub fn setup<D: Data, M: Multiaddress + 'static>() -> (
     ConnectionManagerIO<D, M>,
-    NetworkServiceIO<D, M>,
+    AuthenticationNetworkIO<D, M>,
     SessionManagerIO<D>,
 ) {
     // Prepare and start the network
     let (commands_for_network, commands_from_io) = mpsc::unbounded();
+    let (data_for_network, data_from_user) = mpsc::unbounded();
     let (messages_for_network, messages_from_user) = mpsc::unbounded();
```
maciejnems, line 18: nit: Maybe now this name should indicate it is discovery messages? Not sure if there is any good name for that though. `discovery_for_network`/`authentications_for_network`? `auth_data_for_network` is definitely too long. That said, a lot of places would have to be renamed and just `messages` is close enough, so up to you.

Also you already used `authentications_from_network` in ConnectionManager, so maybe this is the way.

finality-aleph/src/network/io.rs
```diff
@@ -1,33 +1,42 @@
...
-    manager::NetworkData, ConnectionManagerIO, Data, Multiaddress, NetworkServiceIO as NetworkIo,
-    SessionManagerIO,
+    manager::{DataInSession, VersionedAuthentication},
+    ConnectionManagerIO, Data, Multiaddress, NetworkServiceIO as NetworkIO, SessionManagerIO,
 };
 
-type NetworkServiceIO<D, M> = NetworkIo<NetworkData<D, M>, M>;
+type AuthenticationNetworkIO<D, M> = NetworkIO<VersionedAuthentication<M>, DataInSession<D>, M>;
 
 pub fn setup<D: Data, M: Multiaddress + 'static>() -> (
```
maciejnems, line 10: I think a good thing to add somewhere, maybe here, would be a comment about what is `network`, `io`, `user`, `service`. It could explain a lot for someone reading this code. Maybe not necessarily in this PR, but should be added.

Yithis (approved):

finality-aleph/src/network/service.rs
```diff
@@ -641,8 +605,8 @@ mod tests {
 
         test_data
             .mock_io
-            .messages_for_user
-            .unbounded_send((message.clone().into(), DataCommand::SendTo(peer_id.clone())))
+            .data_for_network
```
maciejnems, line 608: Now this test should probably be named `test_sends_validator_data`

maciejnems (approved):
Very nice refactor <3

finality-aleph/src/network/io.rs
timorl: I think we should be able to get rid of this entirely in an upcoming refactor of the whole `[validator_]network` system, when we make all the submodules have proper interfaces rather than relay on channels, so I definitely don't want to touch it now.

finality-aleph/src/network/io.rs
timorl: Yeah, the reason I did not make this consistently `authentications_` is as in the comment below, I suspect a lot of this will be abstracted away using proper interfaces soon-ish, so adding a lot of changes just to have temporarily simpler names doesn't seem necessary.

finality-aleph/src/network/service.rs
timorl: Ha, that definitely would have been confusing, changed.

PR #742, opened 2022-11-21
Harness tangling

pmikolajczyk41:
Use tangling procedure instead of old dummy kinder blending. Also, increase number of leaves in the local dev script.


fbielejec (approved):
OK, as i Understood we will simply refactor out a crate to re-use the code duplicated here, time-permitting

PR #743, opened 2022-11-21
Blender setup++

pmikolajczyk41:
Improve setting up the local environment:
 - two token contracts
 - two blender users

### DIFF
--- a/contracts/setup_blending.sh
+++ b/contracts/setup_blending.sh
@@ -2,24 +2,53 @@
 
 set -euo pipefail
 
-SURI=//Alice
 ROOT_DIR=$(pwd)
 
-TOKENS=1000
+# Actors
+CONTRACTS_ADMIN=//Alice
+DAMIAN=//0
+HANS=//1
+DAMIAN_ACCOUNT=5D34dL5prEUaGNQtPPZ3yN5Y6BnkfXunKXXz6fo7ZJbLwRRH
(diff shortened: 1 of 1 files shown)


pmikolajczyk41:
inb4: we cannot use `//Bob`, because `run_nodes.sh` endows only `//Alice` and `//0|1|2|3` with money

contracts/setup_blending.sh
```diff
@@ -2,24 +2,53 @@
...
-TOKENS=1000
+# Actors
+CONTRACTS_ADMIN=//Alice
+DAMIAN=//0
+HANS=//1
+DAMIAN_ACCOUNT=5D34dL5prEUaGNQtPPZ3yN5Y6BnkfXunKXXz6fo7ZJbLwRRH
+HANS_ACCOUNT=5GBNeWRhZc2jXu7D55rBimKYDk8PGk8itRYFTPfC8RJLKG5o
+
+# Token economics
+TOTAL_TOKEN_ISSUANCE=2000
```
maciejzelaszczyk, line 15: Would change this to `TOKEN_ISSUANCE_PER_CONTRACT`.

pmikolajczyk41: sure, changed

maciejzelaszczyk (approved):
One nit. Apart from that, great effort :)

PR #745, opened 2022-11-21
Blender: public input serialization

pmikolajczyk41:
For 90% a correct public input serialization. To be merged after we test it with relation keys


Yithis (approved):
LGTM

PR #749, opened 2022-11-22
Chillout

mike1729:
Remove a validator from pallet staking when they get a ban.

# Checklist:


Marcin-Radecki (approved):

Marcin-Radecki:
Actually one more comment: since it is possible that `reserved` validator ends up in `BanValidator` event, so we should not remove validators that are in the reserved nodes list.

Marcin-Radecki (approved):
Great contribution.

pallets/elections/src/traits.rs
```diff
@@ -86,3 +86,21 @@ where
         pallet_staking::ErasStakers::<T>::iter_key_prefix(era).collect()
     }
 }
+
+pub trait ValidatorExtractor {
+    type AccountId;
+
+    /// Removes given validator from pallet's staking valiators list
```
DamianStraszak, line 93: typo, should be -> Removes given validator from pallet's staking validator list

DamianStraszak (approved):

pallets/elections/src/traits.rs
mike1729: Fixed.

PR #761, opened 2022-11-22
Remove `only_owner` access modifier from `register_new_tokens`

deuszx:
Remove `#[modifiers(only_owner)]` from `register_new_tokens` so that anyone can do it.


fbielejec (approved):
i approve

PR #763, opened 2022-11-23
Marketplace upgradability

krzysztofziobro:
# Description

An example of how we could perform upgrades for stateful contracts + a simple upgradability test.

## Type of change

- Documentation (?)

# Checklist:

- I have added tests


obrok:
Contract interaction is currently broken by the subxt migration in `main`, I don't think we should add any more code to `benjamin` until that's resolved.

e2e-tests/src/test/button_game/mod.rs
```diff
@@ -55,6 +60,56 @@ pub fn marketplace(config: &Config) -> Result<()> {
     )?;
     marketplace.buy(&sign(&conn, player), None)?;
 
+    // Upgrade begins
```
obrok, line 63: Request: I think this whole upgrade should be a separate test instead of embedded in the other test.

e2e-tests/src/test/button_game/mod.rs
```diff
@@ -55,6 +60,56 @@ pub fn marketplace(config: &Config) -> Result<()> {
...
+        config
+            .test_case_params
+            .marketplace_v2_code_hash
+            .as_ref()
+            .expect("New code's code_hash must be specified."),
+        None, // Alternative for performing "atomic" upgrade + migration (make sure that selector to migrate method is correct)
+              // Some("0x060d3f50".to_string())
+    );
+    info!("Trying to set code actual hash_code: {:?}", set_code_result);
+    assert!(set_code_result.is_ok());
```
obrok, line 77: Suggestion: I think inline comments like this really reduce readability. If you feel like this code needs more explanation, extract it to a function and add a docstring to that function.

e2e-tests/src/test/button_game/mod.rs
```diff
@@ -55,6 +60,56 @@ pub fn marketplace(config: &Config) -> Result<()> {
...
+            &Some(
+                config
+                    .test_case_params
+                    .marketplace_v2_metadata
+                    .clone()
+                    .expect("New code's metadata path must be specified."),
+            ),
+        )
+        .expect("Adding existing contract should succeed."),
+    );
```
obrok, line 92: Suggestion: Same here, I think this should be a single function call that hides the gory details.

e2e-tests/src/test/button_game/mod.rs
```diff
@@ -55,6 +60,56 @@ pub fn marketplace(config: &Config) -> Result<()> {
...
+        "Trying to perform migration after changing the metadata: {:?}",
+        migration_result
+    );
+    assert!(migration_result.is_ok());
+
+    // Check if migration was actually performed
+    //
+    // Will fail when:
+    // - Upgrade was not successful/not performed, or
+    // - Migration was not successful/not performed
```
obrok, line 105: Suggestion: Same here, I think this comment should be a docstring for `migration_performed`.

contracts/marketplace/lib.rs
```diff
@@ -249,24 +264,57 @@ pub mod marketplace {
...
+            };
+
+            if let Some(selector_bytes) = migration_selector {
+                let selector = Selector::from(selector_bytes);
+                let code_hash = Hash::from(code_hash);
+                build_call::<Environment>()
+                    .call_type(DelegateCall::new().code_hash(code_hash))
+                    .exec_input(ExecutionInput::new(selector))
+                    .call_flags(CallFlags::default().set_tail_call(true))
+                    // placeholder, it's a tail call anyway
```
obrok, line 292: Suggestion: remove this comment, it doesn't explain anything (to me).

e2e-tests/src/test/button_game/mod.rs
```diff
@@ -7,29 +7,34 @@ use helpers::sign;
...
+/// Tests trading on the marketplace (with update).
 ///
 /// The scenario:
 ///
 /// 1. Buys a ticket without setting the max price (this should succeed).
 /// 2. Tries to buy a ticket with setting the max price too low (this should fail).
 /// 3. Tries to buy a ticket with setting the max price appropriately (this should succeed).
-pub fn marketplace(config: &Config) -> Result<()> {
+///
+/// Additionally there is an update performed amidst contract's operation.
```
obrok, line 31: Suggestion: if you decide to keep the structure of the test you're proposing, then update the comment with the actual sequence of steps instead of a mysterious "there is an update performed at an unknown point".

obrok (changes requested):

PR #764, opened 2022-11-23
Request justification from forks

mike1729:
Fixes (hopefully) testnet bug.

Todo: fix tests.


timorl (changes requested):
Supposedly you are already rewriting this, but dropping the review anyway, since I essentially had it ready. <_<

finality-aleph/src/justification/mod.rs
```diff
@@ -55,36 +55,28 @@ pub struct JustificationNotification<Block: BlockT> {
...
 }
 
-impl<B: BlockT> Default for JustificationHandlerConfig<B> {
+impl Default for JustificationHandlerConfig {
     fn default() -> Self {
         Self {
             verifier_timeout: Duration::from_millis(500),
-            notification_timeout: Duration::from_millis(1000),
-            min_allowed_delay: 3u32.into(),
+            notification_timeout: Duration::from_millis(800),
```
timorl, line 69: Huh, why did you lower this?

Perhaps add some comments here about the choice of the constants, so it's clearer in the future when anything changes.

finality-aleph/src/justification/requester.rs
```diff
@@ -202,6 +182,55 @@ where
...
+            .children(finalized_hash)
+            .unwrap_or_default();
+        let best_number = blockchain_info.best_number;
+        if best_number < num {
+            // most probably block best_number is not yet finalized
+            num = best_number - 1;
+        }
+        match blockchain_backend.hash(num) {
+            Ok(Some(hash)) => {
+                targets.push(hash);
```
timorl, line 224: Is it a problem that this might be duplicated?

finality-aleph/src/justification/requester.rs
```diff
@@ -202,6 +182,55 @@ where
...
+            }
+            Ok(None) => {
+                debug!(target: "aleph-justification", "Cancelling request, because we don't have block {:?}.", num);
+            }
+            Err(err) => {
+                debug!(target: "aleph-justification", "Cancelling request, because fetching block {:?} failed {:?}.", num, err);
```
timorl, line 230: This and the one above sound like quite unexpected occurrences – the backend claims we have a block with this number but refuses to show it? Mighty suspicious... I would `warn` in both these situations.

finality-aleph/src/justification/requester.rs
```diff
@@ -202,6 +182,55 @@ where
...
+    // somewhere in between them.
+    fn request_targets(&self, mut num: NumberFor<B>) -> Vec<<B as BlockT>::Hash> {
+        let blockchain_backend = self.backend.blockchain();
+        let blockchain_info = blockchain_backend.info();
+        let finalized_hash = blockchain_info.finalized_hash;
+
+        let mut targets = blockchain_backend
+            .children(finalized_hash)
+            .unwrap_or_default();
+        let best_number = blockchain_info.best_number;
```
timorl, line 217: I would split off at least this optimistic block request into a separate function, the logic is completely independent of the one for children and slightly unexpected, so it would be better to parse it in isolation.

finality-aleph/src/justification/requester.rs
```diff
@@ -202,6 +182,55 @@ where
...
+        if best_number < num {
```
timorl, line 218: Needs a `-1` (or `<=`). Or, better yet, rename variable to `top_wanted_number` (or something) and put the `-1` on declaration.

finality-aleph/src/justification/requester.rs
timorl: Oh, and then ofc call that method here, having a `request_targets` method is great.

timorl (changes requested):

finality-aleph/src/justification/mod.rs
mike1729: Done.

finality-aleph/src/justification/requester.rs
mike1729: I've planned to check that but forgot. Added.

finality-aleph/src/justification/requester.rs
```diff
@@ -72,17 +72,16 @@ impl<B: BlockT> fmt::Display for JustificationRequestStatus<B> {
     }
 }
 
-pub struct BlockRequester<B, RB, C, S, F, V>
+pub struct BlockRequester<B, RB, S, F, V, BE>
```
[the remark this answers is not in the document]
mike1729: Added.

finality-aleph/src/justification/requester.rs
```diff
@@ -202,6 +184,71 @@ where
...
+                .flatten()
+                .map(|header| {
+                    let number = *header.number();
+                    let mut ancestor = header;
+                    let mut delay = if self.min_allowed_delay < (number - finalized_number) {
+                        self.min_allowed_delay
+                    } else {
+                        number - finalized_number
+                    };
+                    while delay > 0u32.into() {
```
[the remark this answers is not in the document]
mike1729: Fixed.

finality-aleph/src/justification/requester.rs
mike1729: changed.

finality-aleph/src/justification/requester.rs
mike1729: Not sure if I understand correctly: you want sth like

```suggestion
 request_targets() -> Vec<H> {
        let mut targets = children();
        targets.push(top_wanted(number));
        targets
}
```

?

maciejnems: I would split this into two functions, one for `best_number` and one for `children` (not sure about the names). And after that we could modify status report to report:
* highest block and hash from `best_number` function result
* number of possible alternative branches from `children` function result if `n > 1`

finality-aleph/src/justification/requester.rs
```diff
@@ -72,49 +72,46 @@ impl<B: BlockT> fmt::Display for JustificationRequestStatus<B> {
     }
 }
 
-pub struct BlockRequester<B, RB, C, S, F, V>
+pub struct BlockRequester<B, RB, S, F, V, BB>
```
maciejnems, line 75: Not sure if it collides with anything, but maybe `BE` is better that `BB` for backend?

finality-aleph/src/justification/requester.rs
```diff
@@ -202,6 +182,54 @@ where
...
+            self.block_requester.request_justification(&hash, number);
+        } else {
+            debug!(target: "aleph-justification", "Cancelling request, because we don't have block {:?}.", hash);
+        }
+    }
+
+    // We request justifications for all the children of last finalized block and a justification
+    // for a block of number num on longest branch.
+    // Assuming that we request at the same pace that finalization is progressing, the former ensures
+    // that we are up to date with finalization. On the other hand, the former enables fast catch up
```
maciejnems, line 205: > On the other hand, the `latter` enables fast catch up

finality-aleph/src/justification/mod.rs
timorl: Huh, if this is the reasoning, do we even need any requests for non-children?

finality-aleph/src/justification/requester.rs
timorl: That was what I wanted, yes, but see other comment about possibly dropping `top_wanted` anyway?

If we keep it, then `children` could return a `HashSet`, so we don't need to bother with the `contains` check, and this method in general could return an `IntoIter<Item = H>`.

finality-aleph/src/justification/requester.rs
timorl: HAhah, I meant the other variable, but this rename is helpful as well. xd

finality-aleph/src/justification/requester.rs
```diff
@@ -202,6 +182,54 @@ where
...
 
     pub fn finalized_number(&self) -> NumberFor<B> {
-        self.client.info().finalized_number
+        self.backend.info().finalized_number
+    }
+
+    fn request(&mut self, hash: <B as BlockT>::Hash) {
+        if let Ok(Some(header)) = self.backend.header(BlockId::Hash(hash)) {
+            let number = *header.number();
+            debug!(target: "aleph-justification", "Trying to request block {:?}", number);
```
maciejnems, line 191: I think now this log is unnecessary.

finality-aleph/src/justification/requester.rs
mike1729: There are places where we need both, eg `run_validator_node`.

finality-aleph/src/justification/mod.rs
mike1729: We need it for catch-up sync.

finality-aleph/src/justification/requester.rs
mike1729: changed.

finality-aleph/src/justification/requester.rs
mike1729: Changed the logic due to request status, not sure how to split it now.

finality-aleph/src/justification/requester.rs
mike1729: I'm happy you're happy :)

finality-aleph/src/justification/requester.rs
mike1729: fixed.

finality-aleph/src/justification/requester.rs
```diff
@@ -202,6 +182,54 @@ where
...
+            debug!(target: "aleph-justification", "Cancelling request, because we don't have block {:?}.", hash);
+        }
+    }
+
+    // We request justifications for all the children of last finalized block and a justification
+    // for a block of number num on longest branch.
+    // Assuming that we request at the same pace that finalization is progressing, the former ensures
+    // that we are up to date with finalization. On the other hand, the former enables fast catch up
+    // if we are behind.
+    // We don't remove the child that it's on the same branch as best since a fork may happen
```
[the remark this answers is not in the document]
mike1729: changed.

finality-aleph/src/justification/requester.rs
mike1729: outdated.

finality-aleph/src/justification/mod.rs
timorl: Eh? That shouldn't matter, since if we are catching up we are downloading already finalized blocks, right?

finality-aleph/src/justification/requester.rs
```diff
@@ -32,76 +35,104 @@ impl<B: BlockT> JustificationRequestStatus<B> {
...
+                    status.push_str("hash - unknown; ");
+                }
+
+                write!(f, "{}", status)?;
             }
+        }
+        if self.children_tries >= self.report_threshold {
+            if let Some(parent) = self.parent {
+                let n = self
+                    .n_childred
```
maciejnems, line 92: This could just be usize and by default set to `0` if it is saved with parent. Or if you want to keep option, even:
```
if let (Some(parent), Some(n)) = (self.parent, self.n_children) {
    ...
}
```

finality-aleph/src/justification/requester.rs
```diff
@@ -182,51 +213,64 @@ where
...
+                if let Ok(Some(header)) = blockchain_backend.header(BlockId::Hash(hash)) {
+                    Some(header)
+                } else {
+                    warn!(target: "aleph-justification",
+                      "Cancelling request for child {:?} of {:?}: no block.", hash, finalized_hash);
+                    None
+                }
+            })
+            .collect::<Vec<_>>();
+        self.request_status
```
maciejnems, line 254: This should be only done when `targets.len() > 0`

finality-aleph/src/justification/requester.rs
```diff
@@ -23,7 +23,10 @@ const REPORT_THRESHOLD: u32 = 2;
 pub struct JustificationRequestStatus<B: BlockT> {
     block_number: Option<NumberFor<B>>,
     block_hash: Option<B::Hash>,
-    tries: u32,
+    block_tries: u32,
+    parent: Option<B::Hash>,
+    n_childred: Option<usize>,
```
timorl, line 28: As a fellow parent I'm pretty sure you mean `child_dread`. ;>

finality-aleph/src/justification/requester.rs
```diff
@@ -32,89 +35,114 @@ impl<B: BlockT> JustificationRequestStatus<B> {
...
             report_threshold: REPORT_THRESHOLD,
         }
     }
 
-    fn save_block_number(&mut self, num: NumberFor<B>) {
-        if Some(num) == self.block_number {
-            self.tries += 1;
+    fn save_children(&mut self, hash: B::Hash, n_childred: usize) {
+        if self.parent == Some(hash) {
+            self.children_tries += 1;
```
timorl, line 48: Why do you assume that the same parent has the same number of children on a consecutive attempt?

finality-aleph/src/justification/requester.rs
```diff
@@ -202,6 +213,67 @@ where
...
+        self.request_status
+            .save_children(finalized_hash, targets.len());
+        let best_number = blockchain_info.best_number;
+        if best_number <= top_wanted {
+            // most probably block best_number is not yet finalized
+            top_wanted = best_number - NumberFor::<B>::one();
+        }
+        match blockchain_backend.header(BlockId::Number(top_wanted)) {
+            Ok(Some(header)) => {
+                if !targets.contains(&header) {
```
maciejnems, line 263: I think we don't need this `contains`. We can just push and `save_block` anyway.

finality-aleph/src/justification/requester.rs
timorl: Lines 238 to 255 in one method (and really make it return a `HashSet` plx, especially since you `collect` anyway), lines 256 to 275 in another (moreorless, you want to return not just push :P). I wouldn't worry about double-counting requests for a child in the request status, but if you want to that is also pretty simple, since `insert` returns whether the value was already there.

timorl (approved):

finality-aleph/src/justification/requester.rs
```diff
@@ -202,6 +213,67 @@ where
...
+        let blockchain_backend = self.backend.blockchain();
+        let blockchain_info = blockchain_backend.info();
+        let finalized_hash = blockchain_info.finalized_hash;
+
+        let mut targets = blockchain_backend
+            .children(finalized_hash)
+            .unwrap_or_default()
+            .into_iter()
+            .filter_map(|hash| {
+                if let Ok(Some(header)) = blockchain_backend.header(BlockId::Hash(hash)) {
```
DamianStraszak, line 245: Hmm, if we have the hash of the block, and we know its number `blockchain_info.finalized_number + 1` then we don't need to query the header, I guess. 

There is even a special type for block_hash + number in aleph finality `BlockHashNum<B>` :) It can be used instead of `Header`

finality-aleph/src/justification/requester.rs
```diff
@@ -32,76 +35,104 @@ impl<B: BlockT> JustificationRequestStatus<B> {
...
+            self.parent = Some(hash);
+            self.n_childred = Some(n_childred);
+            self.children_tries = 1;
         }
     }
 
-    fn insert_hash(&mut self, hash: B::Hash) {
-        self.block_hash = Some(hash);
+    fn save_block(&mut self, num: NumberFor<B>, hash: B::Hash) {
+        if self.block_number == Some(num) {
```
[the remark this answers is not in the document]
mike1729: Fixed.

finality-aleph/src/justification/requester.rs
mike1729: Removed Option.

finality-aleph/src/justification/requester.rs
mike1729: Now it's always true.

finality-aleph/src/justification/requester.rs
mike1729: We can but there is an easy `O(1)` check so I used it.

finality-aleph/src/justification/requester.rs
mike1729: Good catch, it simplifies a lot.

finality-aleph/src/justification/requester.rs
mike1729: Fixed ;)

finality-aleph/src/justification/requester.rs
mike1729: Fixed.

finality-aleph/src/justification/requester.rs
```diff
@@ -32,76 +35,104 @@ impl<B: BlockT> JustificationRequestStatus<B> {
...
             report_threshold: REPORT_THRESHOLD,
         }
     }
 
-    fn save_block_number(&mut self, num: NumberFor<B>) {
-        if Some(num) == self.block_number {
-            self.tries += 1;
+    fn save_children(&mut self, hash: B::Hash, n_childred: usize) {
+        if self.parent == Some(hash) {
+            self.children_tries += 1;
```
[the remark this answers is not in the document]
mike1729: Fixed.

finality-aleph/src/justification/requester.rs
mike1729: Huh, it inspired me to further simplify the current version. Check it out ;)

finality-aleph/src/justification/requester.rs
timorl: Oh gods, the `>=` here is so confusing. Do I understand correctly that this can _never_ become lower, so the `>` par is not strictly speaking necessary? If so I would change this to `==`, the fail case is acceptable then anyway.

timorl (approved):
I hope the small spam we will be doing when up to date with finalization will not be a problem – we don't check whether the children, for which we request are not literally the newest block.

But I suspect this either shouldn't really be a problem, or we will find out it is a problem really quickly.

finality-aleph/src/justification/requester.rs
maciejnems: I would still count this as another try for requesting children of `self.parent`, even if `n_children` is different. Now we won't. So I'm still for removing `self.n_children >= n_children ` and adding  `self.n_children = n_children;` here. Not a strong opinion, though, so what timorl said is also fine with me.

maciejnems (approved):

finality-aleph/src/justification/requester.rs
mike1729: Ok, @maciejnems convinced me. Changed.

finality-aleph/src/justification/requester.rs
```diff
@@ -202,6 +203,72 @@ where
...
+        self.justification_request_scheduler.on_request_sent();
+        self.block_requester.request_justification(hash, num);
+    }
+
+    // We request justifications for all the children of last finalized block.
+    // Assuming that we request at the same pace that finalization is progressing, it ensures
+    // that we are up to date with finalization.
+    // We also request the child that it's on the same branch as top_wanted since a fork may happen
+    // somewhere in between them.
+    fn request_children(&mut self, info: &Info<B>) {
```
DamianStraszak, line 221: There is one pretty annoying problem with this fix. 
It assumes that justifications for ALL blocks exist and are generally available via the substrate request.
What is true is that:
All "regular" justifications have been created at some point, but I don't think we have a strict guarantee that many nodes imported them (I think we know that at least one validator did, but not much beyond that). This would require going deep into some unpleasant details do understand it fully.
Also there are also "emergency finalizations" which complicate the picture quite a bit, because the expected way of using them would be to emergency finalize the last block of a session. So this way we would skip many.

DamianStraszak: I think the only way to make sure this works, is to request, once upon a time, ALL descendants/ancestors (one of these, I always confuse them) of the last finalized block, corresponding to the last block in session.

DamianStraszak (changes requested):

finality-aleph/src/justification/requester.rs
```diff
@@ -202,6 +203,72 @@ where
...
+            .children(finalized_hash)
+            .unwrap_or_default();
+
+        if !children.is_empty() {
+            self.request_status
+                .save_children(finalized_hash, children.len());
+        }
+
+        for child in &children {
+            self.do_request(child, finalized_number);
```
maciejnems, line 237: What a bug :O here should be child number, or finalized_number + 1

maciejnems (changes requested):

finality-aleph/src/justification/requester.rs
```diff
@@ -202,6 +203,72 @@ where
...
+        }
+    }
+
+    // This request is important in the case when we are far behind and want to catch up.
+    fn request_wanted(&mut self, mut top_wanted: NumberFor<B>, info: &Info<B>) {
+        let best_number = info.best_number;
+        if best_number <= top_wanted {
+            // most probably block best_number is not yet finalized
+            top_wanted = best_number - NumberFor::<B>::one();
```
DamianStraszak, line 246: This could panic if the chain is empty. We should use `saturating_sub`.

finality-aleph/src/justification/requester.rs
mike1729: Fixed.

finality-aleph/src/justification/requester.rs
mike1729: Fixed.

finality-aleph/src/lib.rs
```diff
@@ -258,3 +259,8 @@ pub struct AlephConfig<B: Block, H: ExHashT, C, SC> {
     pub external_addresses: Vec<String>,
     pub validator_port: u16,
 }
+
+pub trait GetBlockchainBackend<B: Block> {
```
timorl, line 263: Rather than having a trait that only returns another trait, it's likely better to implement some analogue of the original trait on the object you are implementing this one on by calling the returning method internally.

This comment is probably borderline incomprehensible, but we already talked about this, so I'm mostly leaving it here to mark the discussion publicly. <_<"

timorl (changes requested):
Some abstraction overcomplication.

maciejnems (approved):
Looks good and I bump timorls suggestion, to rewrite `GetBlockchainBackend` part

DamianStraszak (approved):

timorl (approved):
<3

PR #767, opened 2022-11-24
[aleph-client]: Add decoders from `Value` into `String` and `Option<T>`.

deuszx:
# Description
Adds decoders from `ConvertibleValue` into `std::String` and `Option<T>`.

## Type of change

# Checklist:

### DIFF
--- a/aleph-client/src/contract/convertible_value.rs
+++ b/aleph-client/src/contract/convertible_value.rs
@@ -1,6 +1,6 @@
 use std::ops::Deref;
 
-use anyhow::{anyhow, bail, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use contract_transcode::Value;
 use sp_core::crypto::Ss58Codec;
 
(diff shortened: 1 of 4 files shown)


aleph-client/src/lib.rs
```diff
@@ -1,3 +1,7 @@
+// Needed for the generic `impl TryFrom<ConvertibleValue> for Option<T>`.
+#![feature(auto_traits)]
+#![feature(negative_impls)]
```
obrok, line 3: Maybe we should hide all of this behind a feature? That way if someone doesn't need the particular conversion they can disable the feature and not be exposed to these dragons. Not sure if that's considered good practice in rustland.

deuszx: That's a good point but I can't imagine people wanting `aleph-client` and, at the same time, not wanting the conversion from `Value` to `Option<T>` 🤷

obrok: I guess it depends on the contract they're interacting with? I don't quite know what's the impact of enabling these features. Is there any chance this blows up someone's compiler? If that chance is low, then probably no point in doing that. Maybe @mike1729 could opine.

aleph-client/src/contract/convertible_value.rs
```diff
@@ -104,3 +104,73 @@ where
...
+where
+    T: TryFrom<ConvertibleValue, Error = anyhow::Error> + Debug,
+    // We will derive this impl only when `T != ConvertibleValue`.
+    // Otherwise we will get a conflict with generic impl in the rust `core` crate.
+    (ConvertibleValue, T): NotEq,
+{
+    type Error = anyhow::Error;
+
+    fn try_from(value: ConvertibleValue) -> std::result::Result<Option<T>, Self::Error> {
+        if let Value::Tuple(tuple) = &value.0 {
```
kostekIV, line 146: to reduce nesting (here and in other similar places
```rust
let tuple = match value.0 {
 Value::Tuple(tuple) => tuple,
 _ => bail!(...),
};
...
```

obrok: I think I tried that, but it makes it so you have to duplicate the `bail!` in the case where we now do `_ => ()`. I decided this was better than having the `bail!` twice.

aleph-client/src/contract/convertible_value.rs
```diff
@@ -104,3 +104,73 @@ where
...
     }
 }
+
+impl TryFrom<ConvertibleValue> for String {
+    type Error = anyhow::Error;
+
+    fn try_from(value: ConvertibleValue) -> std::result::Result<String, Self::Error> {
+        if let Value::Seq(seq) = value.0 {
+            let mut bytes: Vec<u8> = Vec::with_capacity(seq.len());
+            for el in seq.elems() {
```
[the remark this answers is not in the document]
deuszx: In general I agree but on some occasions I find `for ... in ...` more readable. I am not a fan of putting big lambdas into `map` closures, I think Rust syntax makes it very unreadable.

aleph-client/src/contract/convertible_value.rs
deuszx: I went with the suggestion though. It's true that it adds a `bail!` call but it's a different one -- i.e. different error message and I agree with @kostekIV that returning early and decreasing nesting improves readability.

deuszx:
Closing in favor of https://github.com/Cardinal-Cryptography/aleph-node/pull/768

PR #769, opened 2022-11-25
A0-1635: Limit nonfinalized block production

timorl:
# Description

Produce fewer blocks when finalization does not happen, configurable using a flag.

## Type of change

# Checklist:

### DIFF
--- a/bin/node/src/aleph_cli.rs
+++ b/bin/node/src/aleph_cli.rs
@@ -3,6 +3,7 @@ use std::path::PathBuf;
 use aleph_primitives::DEFAULT_UNIT_CREATION_DELAY;
 use clap::{ArgGroup, Parser};
 use finality_aleph::UnitCreationDelay;
+use log::warn;
 
 #[derive(Debug, Parser, Clone)]
 #[clap(group(ArgGroup::new("backup")))]
@@ -33,6 +34,12 @@ pub struct AlephCli {
     /// with `--no-backup`, but note that that limits crash recoverability.
     #[clap(long, value_name = "PATH", group = "backup")]
     backup_path: Option<PathBuf>,
+
+    /// The maximum number of nonfinalized blocks, after which block production should be locally
+    /// stopped. DO NOT CHANGE THIS, PRODUCING MORE OR FEWER BLOCKS MIGHT BE CONSIDERED MALICIOUS
+    /// BEHAVIOUR AND PUNISHED ACCORDINGLY!
+    #[clap(long, default_value_t = 20)]
+    max_nonfinalized_blocks: u32,
 }
 
 impl AlephCli {
@@ -54,5 +61,12 @@ impl AlephCli {
 
     pub fn no_backup(&self) -> bool {
         self.no_backup
+    }
+
+    pub fn max_nonfinalized_blocks(&self) -> u32 {
+        if self.max_nonfinalized_blocks != 20 {
+            warn!("Running block production with a value of max-nonfinalized-blocks {}, which is not the default of 20. THIS MIGHT BE CONSIDERED MALICIOUS BEHAVIOUR AND RESULT IN PENALTIES!", self.max_nonfinalized_blocks);
+        }
+        self.max_nonfinalized_blocks
     }
 }
--- a/bin/node/src/service.rs
+++ b/bin/node/src/service.rs
@@ -1,6 +1,7 @@
 //! Service and ServiceFactory implementation. Specialized wrapper over substrate service.
 
 use std::{
+    ops::Sub,
     path::{Path, PathBuf},
     sync::Arc,
 };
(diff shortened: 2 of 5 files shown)


bin/node/src/service.rs
```diff
@@ -33,6 +35,24 @@ type FullClient = sc_service::TFullClient<Block, RuntimeApi, AlephExecutor>;
...
+        &self,
+        chain_head_number: N,
+        _chain_head_slot: Slot,
+        finalized_number: N,
+        _slow_now: Slot,
+        _logging_target: &str,
+    ) -> bool {
+        let nonfinalized_blocks = (chain_head_number - finalized_number)
+            .try_into()
+            .unwrap_or(u32::MAX);
```
DamianStraszak, line 51: When reading this I wonder what if `chain_head_number - finalized_number` is negative and instead of `0` we output `u32::MAX` there? On the other hand, if `N` is a signed type then this can panic in some strange circumstances?

We should use `saturating_sub` instead of `-`.

DamianStraszak (changes requested):
Should we also change `docker_entrypoint.sh` so that we can then change this in out infra easily?

On the other hand, when there is a dedicated env variable for this, then it's maybe easier to change for other parties :P

bin/node/src/aleph_cli.rs
```diff
@@ -33,6 +34,12 @@ pub struct AlephCli {
     /// with `--no-backup`, but note that that limits crash recoverability.
     #[clap(long, value_name = "PATH", group = "backup")]
     backup_path: Option<PathBuf>,
+
+    /// The maximum number of nonfinalized blocks, after which block production should be locally
+    /// stopped. DO NOT CHANGE THIS, PRODUCING MORE OR FEWER BLOCKS MIGHT BE CONSIDERED MALICIOUS
+    /// BEHAVIOUR AND PUNISHED ACCORDINGLY!
+    #[clap(long, default_value_t = 20)]
```
DamianStraszak, line 41: This I think is still under discussion, right :) ?

bin/node/src/service.rs
timorl: Fixed using Substrate's traits.

bin/node/src/aleph_cli.rs
timorl: Dunno, if it's easily changeable then 20 should be alright, no?

mike1729: According to @maciejnems findings only values less than 128 are, right?

mike1729 (commented):
Approve modulo default slack value.

bin/node/src/aleph_cli.rs
DamianStraszak: As for me, the larger the value is the less likely we ever hit it in normal operation (lag in finalization), and thus better. On the other hand we should be below the strange threshold from the sync protocol, that Maciek discovered.

timorl: The larger it is, the more blocks we have to roll back in an outage like the one on Wednesday. The theoretical chances of reaching a 20 finalization lag as seen through pure AlephBFT are already astronomically small (it's ~60 rounds) and the practical reasons at session startup are why this is already 20, on your suggestion. If we have any other issues with finalization we should fix them rather than masking them. We should probably also fix the session startup thing, this should be possible with the semi-recent changes to how AlephBFT acquires data, then we could drop this back to 6.

bin/node/src/service.rs
```diff
@@ -33,6 +35,24 @@ type FullClient = sc_service::TFullClient<Block, RuntimeApi, AlephExecutor>;
...
+        chain_head_number: N,
+        _chain_head_slot: Slot,
+        finalized_number: N,
+        _slow_now: Slot,
+        _logging_target: &str,
+    ) -> bool {
+        let nonfinalized_blocks: u32 = chain_head_number
+            .saturating_sub(finalized_number)
+            .unique_saturated_into();
+        nonfinalized_blocks >= self.0
```
DamianStraszak, line 52: Can we log something when returning `true`? Ideally in `info` level.

bin/node/src/aleph_cli.rs
DamianStraszak: For the record, I still think it should be larger, but let's make it 20. What I think is crucial though is to make changes to our monitoring/alerts, so that there is an alert when we start backing off (currently it would not be even visible I guess?). Could you add an appropriate task in jira?

DamianStraszak (approved):

bin/node/src/aleph_cli.rs
timorl: Very good point, I've been thinking about this but did not get to creating the task until now. Added as 1643, and this should definitely be high priority (likely relatively simple though).

bin/node/src/service.rs
timorl: Good idea, added as `warn`, this shouldn't be happening during normal operation anyway.

mike1729 (approved):

PR #770, opened 2022-11-25
A0-1584: Unify MockMultiaddress

timorl:
# Description

Now we only have one `MockMultiaddress`, which should make changes to `Multiaddress` be less cursed.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/network/manager/session.rs
+++ b/finality-aleph/src/network/manager/session.rs
@@ -278,7 +278,7 @@ mod tests {
             mock::{crypto_basics, MockNetworkIdentity},
             NetworkIdentity,
         },
-        testing::mocks::validator_network::{random_identity, MockMultiaddress},
+        testing::mocks::validator_network::{random_multiaddress, MockMultiaddress},
         NodeIndex, SessionId,
     };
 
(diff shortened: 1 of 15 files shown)


finality-aleph/src/network/manager/session.rs
```diff
@@ -379,8 +380,8 @@ mod tests {
     async fn fails_to_create_with_non_unique_peer_id() {
         let mut crypto_basics = crypto_basics(NUM_NODES).await;
         let addresses = vec![
-            MockMultiaddress::random_with_id(MockPeerId::random()),
-            MockMultiaddress::random_with_id(MockPeerId::random()),
+            random_identity().0[0].clone(),
```
lesniak43, line 383: maybe create a function `random_address` that does exactly that?

lesniak43 (approved):
please consider running `cargo clippy --all-targets --all-features --no-deps` and maybe fixing some recent warnings

finality-aleph/src/network/manager/session.rs
timorl: Done, also fixed some of the clippy things – the ones I could reasonably consider parts of this PR.

PR #771, opened 2022-11-28
Authorities are staking finalization stall

maciejzelaszczyk:
# Description

This increases the number of nodes run from `docker/docker-compose.yml` to `6`. Handling of `6` nodes added to CI scripts.

## Type of change


timorl (approved):

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -454,8 +454,8 @@ jobs:
           node-count: 6
           reserved-seats: 3
           non-reserved-seats: 3
-          follow-up-finalization-check: true
-        timeout-minutes: 20
+          follow-up-finalization-check: false
```
timorl, line 457: With this check turned off this PR makes no sense. :/

maciejzelaszczyk: Agreed, I am just checking if turning it off will make this pass (99% it will). Further investigation ongoing.

Marcin-Radecki (approved):

e2e-tests/src/validators.rs
```diff
@@ -138,9 +138,20 @@ pub async fn prepare_validators<S: SignedConnectionApi + AuthorRpc>(
...
     }
 
-    for controller in accounts.controller_raw_keys.iter() {
+    let address_tokens = node.split(':').collect::<Vec<_>>();
+    let ws = address_tokens[0];
+    let ip_address = address_tokens[1];
+    let first_port = address_tokens[2].parse::<usize>()?;
+
+    // Assumes the same ip address and consecutive ports for nodes, e.g. ws://127.0.0.1:9943,
+    // ws://127.0.0.1:9944, etc.
```
timorl, line 147: This part seems out of place in this function, shouldn't we pass in a set of connections instead, or at least a set of addresses?

e2e-tests/src/validators.rs
```diff
@@ -138,9 +138,20 @@ pub async fn prepare_validators<S: SignedConnectionApi + AuthorRpc>(
         }));
     }
 
-    for controller in accounts.controller_raw_keys.iter() {
+    let address_tokens = node.split(':').collect::<Vec<_>>();
+    let ws = address_tokens[0];
+    let ip_address = address_tokens[1];
+    let first_port = address_tokens[2].parse::<usize>()?;
```
timorl, line 144: Eh, this should technically be `u16`, but I see it might be annoying to deal with in the loop below – if you can make this an `u16` without much trouble then please do, otherwise just resolve this comment.

e2e-tests/src/validators.rs
```diff
@@ -138,9 +138,20 @@ pub async fn prepare_validators<S: SignedConnectionApi + AuthorRpc>(
         }));
     }
 
-    for controller in accounts.controller_raw_keys.iter() {
+    let address_tokens = node.split(':').collect::<Vec<_>>();
+    let ws = address_tokens[0];
+    let ip_address = address_tokens[1];
```
timorl, line 143: Oh gods, I just noticed this totally isn't an IP address, but something sinister. Dx Maybe `let address_prefix = format!("{}:{}", address_tokens[0], address_tokens[1])` to avoid confusion? You only modify the port anyway.

timorl (approved):

e2e-tests/src/validators.rs
maciejzelaszczyk: Sure, I have extracted this to a separate function and passed the connections in.

PR #772, opened 2022-11-28
Clippy args change for aleph-node

maciejzelaszczyk:
# Description

Changes the restrictiveness of `clippy` checks.

## Type of change


.github/workflows/unit_tests.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           RUSTC_WORKSPACE_WRAPPER: sccache
         with:
           command: clippy
-          args: -- --no-deps -D warnings
+          args: --all-targets --all-features -- --no-deps -D warnings -A clippy::type_complexity
```
timorl, line 61: Could we also have `type_complexity` checks? I know AlephBFT has this, but imo that's a mistake and it would be nice to fix it there as well.

timorl (approved):

.github/workflows/unit_tests.yml
lesniak43: +1

lesniak43 (approved):

.github/workflows/unit_tests.yml
maciejzelaszczyk: Jira issue added :)

maciejzelaszczyk: I have reworked this PR to include the complexity checks. The setup should now prevent us from submitting overly complex types.

finality-aleph/src/network/clique/protocols/v0/mod.rs
```diff
@@ -263,17 +253,18 @@ mod tests {
 
     #[tokio::test]
     async fn parent_service_dead() {
-        let (
-            _id_incoming,
-            _pen_incoming,
-            _id_outgoing,
-            _pen_outgoing,
+        let MockPrelims {
+            id_incoming: _id_incoming,
```
timorl, line 257: The whole point with the `..` is that you can avoid writing all the unused variables.

timorl (approved):
One nit, although it requires many lines of changes. ;p

finality-aleph/src/network/clique/protocols/v0/mod.rs
maciejzelaszczyk: I left them because of the `_`/`_var` distinction, where `_` desugars to `_` and this results in a bunch of failures in the test module.

timorl: You should  only leave the ones that result in failures – I'm slightly surprized any do, but only the channels are even possible, so probably leave them?

maciejzelaszczyk: OK, so I have now removed all the unused non-channel variables and it works as expected. Removing a random channel variable still results in errors, so unless we want to check all of them in sequence, this seems like a good place to stop :D

PR #776, opened 2022-12-01
Port aleph-node 8.2 testnet to mainnet branch

Marcin-Radecki:
Following commands has been done:

```
git checkout release-8
git checkout -b port-8-X-Testnet-to-Mainnet
git merge release-8-mainnet # to pick up revert SCs change, resolved conflicts only in locks and toml files 
nvim bin/node/Cargo.toml # bump version to 0.8.2+mainnet
git commit -am "..."
git push -u origin HEAD
```


bin/runtime/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aleph-runtime"
-version = "0.8.0+mainnet"
+version = "0.8.1"
```
timorl, line 3: Is that  truly correct? This shouldn't be changing the runtime afaik, so this seems suspicious to me.

timorl (approved):
Just one comment, although might be important. <_<

bin/runtime/Cargo.toml
Marcin-Radecki: You're right, reverted

mike1729 (approved):

PR #778, opened 2022-12-02
Added pending_rewards call to nomination pool api to our runtime

Marcin-Radecki:
This change is required as part of adding nomination pools to aleph-runtime. 
See https://github.com/paritytech/polkadot/pull/5924/files for more reference.

### DIFF
--- a/bin/runtime/Cargo.toml
+++ b/bin/runtime/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aleph-runtime"
-version = "0.9.0"
+version = "0.8.4"
 authors = ["Cardinal Cryptography"]
 edition = "2021"
 homepage = "https://alephzero.org"
--- a/bin/runtime/Cargo.toml
+++ b/bin/runtime/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aleph-runtime"
-version = "0.8.3"
+version = "0.9.0"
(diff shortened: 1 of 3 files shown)


bin/runtime/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aleph-runtime"
-version = "0.8.3"
+version = "0.9.0"
```
kostekIV, line 3: should it be 8.4? or do we want it to be in net 'big' release?

DamianStraszak (approved):
Except for the version discussion, LGTM

bin/runtime/Cargo.toml
Marcin-Radecki: Okay so let me explain what I had in mind: in semver in general, we bump major version, not patch version. Patch version is only bumped when there's a hotfix to existing feature. This is at least how I understand semver. 

For sake of this PR I'll revert it to 0.8.4 as it's the current convention

kostekIV (approved):

bin/runtime/Cargo.toml
deuszx: It's `major.minor.patch` so here we would be bumping `minor` (not `major`) but it's probably a typo. And yes, I agree with @Marcin-Radecki .

PR #779, opened 2022-12-02
A0-1503: Take e2e config from env

obrok:
# Description

Takes all parts of the e2e config from the environment instead of using clap. When using `cargo test` to run the tests, this seems more natural. 

## Type of change

### DIFF
--- a/e2e-tests/src/config.rs
+++ b/e2e-tests/src/config.rs
@@ -1,49 +1,71 @@
-use std::env;
+use std::{env, str::FromStr};
 
 use aleph_client::{RootConnection, SignedConnection};
-use clap::{Args, Parser};
 use once_cell::sync::Lazy;
 use primitives::SessionIndex;
 
 use crate::accounts::{get_sudo_key, get_validators_keys, get_validators_seeds, NodeKeys};
 
 static GLOBAL_CONFIG: Lazy<Config> = Lazy::new(|| {
-    let unparsed = env::var("E2E_CONFIG").unwrap_or("".to_string());
-    let unparsed = format!("e2e {}", unparsed);
-    Config::parse_from(unparsed.split_whitespace())
+    let node = get_env("NODE_URL").unwrap_or_else(|| "ws://127.0.0.1:9943".to_string());
+    let validator_count = get_env("VALIDATOR_COUNT").unwrap_or_else(|| 5);
+    let validators_seeds = env::var("VALIDATORS_SEEDS")
+        .ok()
+        .map(|s| s.split(',').map(|s| s.to_string()).collect());
+    let sudo_seed = get_env("SUDO_SEED").unwrap_or_else(|| "//Alice".to_string());
+    let reserved_seats = get_env("RESERVED_SEATS");
+    let non_reserved_seats = get_env("NON_RESERVED_SEATS");
+    let upgrade_to_version = get_env("UPGRADE_VERSION");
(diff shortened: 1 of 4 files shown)


Marcin-Radecki (approved):

e2e-tests/src/config.rs
```diff
@@ -1,49 +1,71 @@
...
+    let validator_count = get_env("VALIDATOR_COUNT").unwrap_or_else(|| 5);
+    let validators_seeds = env::var("VALIDATORS_SEEDS")
+        .ok()
+        .map(|s| s.split(',').map(|s| s.to_string()).collect());
+    let sudo_seed = get_env("SUDO_SEED").unwrap_or_else(|| "//Alice".to_string());
+    let reserved_seats = get_env("RESERVED_SEATS");
+    let non_reserved_seats = get_env("NON_RESERVED_SEATS");
+    let upgrade_to_version = get_env("UPGRADE_VERSION");
+    let upgrade_session = get_env("UPGRADE_SESSION");
+    let upgrade_finalization_wait_sessions = get_env("UPGRADE_FINALIZATION_WAIT_SESSIONS");
```
kostekIV, line 20: looks good but we lost 'usage' message from clap - add smth that print help/usage in case of bad/missing arguments

kostekIV (approved):

e2e-tests/src/config.rs
obrok: Right, I already talked a bit with @Marcin-Radecki about this problem. I think this is a downside, but a minor one - the way the tests currently work, we have to verify each arg (option) in every test anyway, because almost all of them are optional. So I think the tests themselves can print out errors if a particular setting is not present.

kostekIV: I am ok with resolving this for now. Altough ideally there should be a usage fn here, can be done in another pr

obrok: To be clear: I'm not planning to add it in another PR or otherwise - I don't think it's a useful addition in this form of the code. If anything, we could add a note about the environment vars to the README for `e2e-tests`, and direct people to look in this file for an up-to-date list of available options. WDYT?

kostekIV: well, readme is the minimal thing we should have here. Lets remember that this will end up being used by people writing new e2e tests that were not familiar with the code before. Usage fn would be ideal in the long run and does not require that much work:
``` rust
if smth_failed {
    print_usage()
}
```

obrok: Disagree about the complexity, you just hid all the complexity under `smth_failed` and `print_usage`.

obrok: Example:

Someone sets an env variable called `URL` (we're expecting `NODE`). In clap you can detect this, but in this case, it doesn't even look like a failure. We just fall back to the default.

obrok: Added a README.

kostekIV: It is simple, we dont want to check all of the env variables anyway since it is not possible to know which are for us.
I had smth like this in mind (added second step to inform user about config): 
1. bad value provided or missing in non default case -> fail and print usage
2. log/print parsed Config
This should be enough and IMHO is simple

kostekIV: Resolving as readme is added but this should be added in the future imo

PR #780, opened 2022-12-02
Outsource hashing

pmikolajczyk41:


contracts/shielder/Cargo.toml
```diff
@@ -31,6 +31,7 @@ ark-bls12-381 = { version = "^0.3.0", default-features = false, features = ["sca
 openbrush = { git = "https://github.com/Supercolony-net/openbrush-contracts", rev = "v2.3.0", default-features = false, features = ["psp22", "ownable"] }
 
 snarcos-extension = { path = "../../bin/runtime/src/chain_extension/snarcos-extension", default-features = false, features = ["ink-as-dependency"] }
+house-snark = { git = "https://github.com/Cardinal-Cryptography/HouseSnark", branch = "reusability", package = "house-snark", default-features = false, features = ["shielder"] }
```
deuszx, line 34: Can we extract `shielder` contract from the main `aleph-node` repo? Seems a bit backwards that the main repo depends on some small external contract.

Also a bit backwards is that `house-snark` has a feature named like this contract that depends on it.

PR #782, opened 2022-12-02
Added pending_rewards API call to Testnet 8.X branch

Marcin-Radecki:
This is a slightly modified port of https://github.com/Cardinal-Cryptography/aleph-node/pull/778 to Testnet 8.X brach

Runtime `spec_version` is 44 which might sound controversial, but we need to use the first free number that is not used. This ie because spec_version needs to identify uniquely aleph-runtime, just in case of wrongly updated runtime on Tesnet or Mainnet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -283,7 +283,7 @@ dependencies = [
 
 [[package]]
 name = "aleph-runtime"
-version = "0.8.1"
+version = "0.8.2"
 dependencies = [
  "frame-executive",
  "frame-support",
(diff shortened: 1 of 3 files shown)


kostekIV (approved):
controversial approve

DamianStraszak (approved):
I'm not sure about 44 -- this then changes our policy about bumping by one in our PRs, i.e., we need to jump over 44 in this case.

I guess it's fine, but could you please spread awareness somehow (writing on dev-core or something)?

PR #788, opened 2022-12-06
A0-1664 Stop exposing sync connected events from substrate network

lesniak43:
https://cardinal-cryptography.atlassian.net/browse/A0-1664

### DIFF
--- a/finality-aleph/src/network/mock.rs
+++ b/finality-aleph/src/network/mock.rs
@@ -175,7 +175,7 @@ pub struct MockNetwork {
 }
 
 #[derive(Debug, Copy, Clone)]
-pub struct MockSenderError();
+pub struct MockSenderError;
 
 impl fmt::Display for MockSenderError {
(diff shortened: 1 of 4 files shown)


finality-aleph/src/network/mock.rs
```diff
@@ -182,17 +175,11 @@ pub struct MockNetwork {
 }
 
 #[derive(Debug, Copy, Clone)]
-pub enum MockSenderError {
-    SomeError,
-}
+pub struct MockSenderError();
```
timorl, line 178: I believe these parentheses are unnecessary.

timorl (approved):
Looks good as long as there is a plan to reacquire these tests.

finality-aleph/src/network/mock.rs
lesniak43: indeed

maciejnems (approved):
Rebased on top of main, so tests are no longer removed.

PR #790, opened 2022-12-06
A0-1592: make connection manager operate of validator network directly

maciejnems:
# Description
* Moved Validator Network to ConnectionManagerIO
* Simplified tests in Network Service. They are no longer dependent on VersionedAuthentication
 
## Type of change

- Refactor/SImplification of code

# Checklist:

- I have not added any tests, but they are not necessary


fixxxedpoint (approved):
lgtm

PR #791, opened 2022-12-06
A0-1667 Add timeouts to the dialer

lesniak43:
# Description

https://cardinal-cryptography.atlassian.net/browse/A0-1667

Currently the `TcpDialer` depends on OS timeouts, which are probably too long. We should wrap the `connect` call with a `timeout`.

## Type of change

### DIFF
--- a/finality-aleph/src/validator_network/outgoing.rs
+++ b/finality-aleph/src/validator_network/outgoing.rs
@@ -38,7 +38,8 @@ impl<PK: PublicKey, A: Data, ND: Dialer<A>> Display for OutgoingError<PK, A, ND>
     }
 }
 
-const DIAL_TIMEOUT: Duration = Duration::from_secs(43);
+/// Arbitrarily chosen timeout, should be more than enough.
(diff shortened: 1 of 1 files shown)


finality-aleph/src/validator_network/outgoing.rs
```diff
@@ -32,10 +33,13 @@ impl<PK: PublicKey, A: Data, ND: Dialer<A>> Display for OutgoingError<PK, A, ND>
                 "communication with {} failed, protocol error: {}",
                 addr, e
             ),
+            TimedOut => write!(f, "dial timeout",),
         }
     }
 }
 
+const DIAL_TIMEOUT: Duration = Duration::from_secs(43);
```
timorl, line 41: You should probably comment on the choice of the value, especially if it's not obvious.

And yes, I'm also sad that this likely means it will have to be changed. :<

timorl (approved):

maciejnems (approved):

PR #792, opened 2022-12-07
fix clashing args in purge-chain cmd

kostekIV:
base-path in PurgeBackup was clashing with base-path in the PurgeChain cmd causing shenanigans

### DIFF
--- a/bin/node/src/commands.rs
+++ b/bin/node/src/commands.rs
@@ -303,7 +303,14 @@ pub struct PurgeChainCmd {
 impl PurgeChainCmd {
     pub fn run(&self, database_config: DatabaseSource) -> Result<(), Error> {
         self.purge_chain.run(database_config)?;
-        self.purge_backup.run(self.purge_chain.yes)
+        self.purge_backup.run(
+            self.purge_chain.yes,
+            self.purge_chain
+                .shared_params
+                .base_path()
+                .expect("need base-path to be provided")
+                .expect("need base-path to be provided"),
+        )
     }
 }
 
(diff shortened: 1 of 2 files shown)


Marcin-Radecki (approved):

bin/node/src/commands.rs
```diff
@@ -319,16 +326,14 @@ impl CliConfiguration for PurgeChainCmd {
 
 #[derive(Debug, Parser)]
 pub struct PurgeBackupCmd {
-    #[clap(flatten)]
-    pub node_params: NodeParams,
+    /// Directory under which AlephBFT backup is stored
+    #[arg(long, default_value = DEFAULT_BACKUP_FOLDER)]
```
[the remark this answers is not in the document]
kostekIV: They were not used by this command only partially (base-path) and the rest of args are optional.

bin/node/src/commands.rs
```diff
@@ -303,7 +303,14 @@ pub struct PurgeChainCmd {
...
     pub fn run(&self, database_config: DatabaseSource) -> Result<(), Error> {
         self.purge_chain.run(database_config)?;
-        self.purge_backup.run(self.purge_chain.yes)
+        self.purge_backup.run(
+            self.purge_chain.yes,
+            self.purge_chain
+                .shared_params
+                .base_path()
+                .expect("need base-path to be provided")
+                .expect("need base-path to be provided"),
```
maciejnems, line 312: Right now this results in a really confusing error saying wrong things:
```
Thread 'main' panicked at 'need base-path to be provided', bin/node/src/commands.rs:314

This is a bug. Please report it at:

	docs.alephzero.org

```
Maybe instead of expects you could return an error here with message about not providing `--base-path` argument, which is mandatory?

kostekIV: done

maciejnems: Also now purge-backup could be before purge-chain, so that purge-chain does not happen without base-path

kostekIV: well, this seems like some change that I am not sure about - this was like that before. No logic changes were done to purge_chain and purge_backup so why would we need to change the order?

maciejnems: Before the update not passing `base_path` would result in command failing on cli::parse. So purge_chain was not called. If you do not want to change the order, you can just create `base_path` before purge_chain call, so it does not happen.

kostekIV: ok, fair. Done

maciejnems (approved):
LGTM

PR #793, opened 2022-12-07
[A0-1673] Crate relations

fbielejec:
# Description

Crate with all existing relations


relations/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+ark-r1cs-std = { version = "^0.3.0", default-features = false }
+ark-relations = { version = "^0.3.0", default-features = false }
+ark-serialize = { version = "^0.3.0", default-features = false }
+ark-snark = { version = "^0.3.0", default-features = false }
+ark-std = { version = "^0.3.0", default-features = false }
+blake2 = "0.9.2"
+
+[features]
+default = ["std"]
+std = []
```
pmikolajczyk41, line 28: is there any reason for that feature? eventually, this crate will be used in some contracts, so no-std for sure will be needed, but for other usages (like CLI) I'm not so sure that we will need any std functionality (maybe some (de)serializing, but we don't have it anyway)

relations/src/environment.rs
```diff
@@ -0,0 +1,139 @@
...
+pub type Marlin = ark_marlin::Marlin<CircuitField, MarlinPolynomialCommitment, Blake2s>;
+
+/// Serialized keys.
+pub struct RawKeys {
+    pub pk: Vec<u8>,
+    pub vk: Vec<u8>,
+}
+
+/// Common API for every proving system.
+pub trait ProvingSystem {
```
pmikolajczyk41, line 28: I'm wondering whether this crate should also allow to verify proofs... wdyt?

relations/src/linear.rs
```diff
@@ -0,0 +1,54 @@
...
+impl LinearEquationRelation {
+    pub fn new(a: u32, x: u32, b: u32, y: u32) -> Self {
+        Self { a, x, b, y }
+    }
+}
+
+impl<Field: PrimeField> ConstraintSynthesizer<Field> for LinearEquationRelation {
+    fn generate_constraints(self, cs: ConstraintSystemRef<Field>) -> Result<(), SynthesisError> {
+        // Watch out for overflows!!!
+        let x = UInt32::new_witness(ark_relations::ns!(cs, "x"), || Ok(&self.x))?;
```
pmikolajczyk41, line 36: please add todo for migrating from `UInt32` to `FpVar` (it will also make `impl GetPublicInput<CircuitField> for LinearEquationRelation` easier to implement)

pmikolajczyk41: the same for `xor` as well

pmikolajczyk41 (approved):

h4nsu (approved):

relations/src/environment.rs
fbielejec: very good idea, pushed it

relations/Cargo.toml
fbielejec: maybe SC test environments... but we're not using those. Removed

relations/src/linear.rs
fbielejec: :ok_hand:

PR #797, opened 2022-12-08
Add a condition for S3 steps to run only when AWS secrets are present

mikogs:
Implemented as suggested on the following GitHub Community discussion:
https://github.com/orgs/community/discussions/26726#discussioncomment-3253116


mikogs:
@Kosiare These files are required only in `deploy-mainnet` and `deploy-testnet` which fail when forked anyway (secrets are used to push to ECR etc.).

Kosiare (approved):

PR #803, opened 2022-12-12
A0-1663: remove unnecessary session start retries in connection manager

maciejnems:
# Description
`to_retry` was kept before, to retry session start in case we did not know our own identity. Now we know it from the start and the only way session start can fail is if we changed session handler type. In this case we do not want to retry session start.

### DIFF
--- a/finality-aleph/src/network/manager/service.rs
+++ b/finality-aleph/src/network/manager/service.rs
@@ -56,7 +56,7 @@ struct PreValidatorSession {
 }
 
 #[derive(Clone)]
-/// Stores all data needed for starting non validator session
+/// Stores all data needed for starting non-validator session
 struct PreNonvalidatorSession {
     session_id: SessionId,
     verifier: AuthorityVerifier,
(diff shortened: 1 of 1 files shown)


finality-aleph/src/network/manager/service.rs
```diff
@@ -307,21 +287,16 @@ where
         pre_session: PreValidatorSession,
         result_for_user: Option<oneshot::Sender<mpsc::UnboundedReceiver<D>>>,
     ) -> Result<ServiceActions<M, NI::AddressingInformation>, SessionHandlerError> {
-        match self.update_validator_session(pre_session.clone()).await {
-            Ok((actions, data_from_network)) => {
+        self.update_validator_session(pre_session.clone())
```
timorl, line 290: You no longer need to `clone` here.

Btw, we really should get clippy back in the pipeline. xd

finality-aleph/src/network/manager/service.rs
```diff
@@ -368,13 +343,7 @@ where
...
         pre_session: PreNonvalidatorSession,
     ) -> Result<(), SessionHandlerError> {
-        self.update_nonvalidator_session(pre_session.clone())
-            .await
-            .map_err(|e| {
-                self.to_retry
-                    .push((PreSession::Nonvalidator(pre_session), None));
-                e
-            })
+        self.update_nonvalidator_session(pre_session.clone()).await
```
timorl, line 346: Nor here.

timorl (approved):
Two nits, looks good.

PR #805, opened 2022-12-13
Dex whitelist

fbielejec:
# Description

- Add a query
- whitelist all possible pairs exchange contract deployment

### DIFF
--- a/.github/scripts/run_smartnet.sh
+++ b/.github/scripts/run_smartnet.sh
@@ -2,7 +2,9 @@
 
 set -e
 
-export NODE_IMAGE=public.ecr.aws/p6e8q1z1/aleph-node:latest
+export NODE_IMAGE=573243519133.dkr.ecr.us-east-1.amazonaws.com/feature-env-aleph-node:fe-benjamin_c643069
+# export NODE_IMAGE=public.ecr.aws/p6e8q1z1/aleph-node:latest
+
 # key derived from "//0"
 export NODE_ID=5D34dL5prEUaGNQtPPZ3yN5Y6BnkfXunKXXz6fo7ZJbLwRRH
 export ALICE=5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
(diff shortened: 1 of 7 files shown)


h4nsu (approved):

contracts/simple_dex/lib.rs
```diff
@@ -322,6 +322,11 @@ mod simple_dex {
             Ok(())
         }
 
+        #[ink(message)]
+        pub fn can_swap_pair(&self, from: AccountId, to: AccountId) -> bool {
```
deuszx, line 326: Please add documentation on the public endpoint. It gets compiled into `metadata.json` file and may be useful for frontend wallets like Contracts UI.

e2e-tests/src/test/button_game/mod.rs
```diff
@@ -159,6 +160,12 @@ pub fn simple_dex(config: &Config) -> Result<()> {
 
     dex.swap(account_conn, token1, balance_after, token3, mega(90))?;
     assert_recv_id(&mut events, "Swapped");
+
+    // can't swap a pair not on the whitelist
+
+    dex.remove_swap_pair(authority_conn, token3.into(), token1.into())?;
+    assert_recv_id(&mut events, "SwapPairRemoved");
```
deuszx, line 167: Why had the test below been failing before you added these two lines? i.e. it [the test] asserts that there was no trade made and I'm assuming that it used to pass so why did it?

fbielejec: cause before pair was not on a whitelist for swapping

.github/scripts/run_smartnet.sh
```diff
@@ -2,7 +2,8 @@
 
 set -e
 
-export NODE_IMAGE=public.ecr.aws/p6e8q1z1/aleph-node:latest
+export NODE_IMAGE=573243519133.dkr.ecr.us-east-1.amazonaws.com/feature-env-aleph-node:fe-benjamin_c643069
```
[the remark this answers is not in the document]
fbielejec: i don't know, maybe @mikogs does

contracts/simple_dex/lib.rs
fbielejec: done

obrok:
Duh, I already forgot that we have the separate marketplace thing _that I wrote myself_ :/

PR #808, opened 2022-12-13
Benchmark configuration cleanup

pmikolajczyk41:
- use `AllPalletsWithSystem::whitelisted_storage_keys`
 - drop `extern crate`

### DIFF
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -791,12 +791,8 @@ pub type Executive = frame_executive::Executive<
 >;
 
 #[cfg(feature = "runtime-benchmarks")]
-#[macro_use]
-extern crate frame_benchmarking;
-
-#[cfg(feature = "runtime-benchmarks")]
 mod benches {
-    define_benchmarks!([pallet_snarcos, Snarcos]);
+    frame_benchmarking::define_benchmarks!([pallet_snarcos, Snarcos]);
 }
 
 impl_runtime_apis! {
@@ -1033,7 +1029,7 @@ impl_runtime_apis! {
             Vec<frame_benchmarking::BenchmarkList>,
             Vec<frame_support::traits::StorageInfo>,
         ) {
-            use frame_benchmarking::{Benchmarking, BenchmarkList};
+            use frame_benchmarking::{Benchmarking, BenchmarkList, cb_list_benchmarks, list_benchmark};
             use frame_support::traits::StorageInfoTrait;
 
             let mut list = Vec::<BenchmarkList>::new();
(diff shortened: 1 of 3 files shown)


bin/runtime/src/lib.rs
```diff
@@ -1047,22 +1043,11 @@ impl_runtime_apis! {
...
-                // Event Count
-                hex_literal::hex!("26aa394eea5630e07c48ae0c9558cef70a98fdbe9ce6c55837576c60c7af3850").to_vec().into(),
-                // System Events
-                hex_literal::hex!("26aa394eea5630e07c48ae0c9558cef780d41e5e16056765bc8461851072c9d7").to_vec().into(),
-            ];
+            use frame_benchmarking::{
+                Benchmarking, BenchmarkBatch, TrackedStorageKey, cb_add_benchmarks, add_benchmark};
+            use frame_support::traits::WhitelistedStorageKeys;
+
+            let whitelist: Vec<TrackedStorageKey> = AllPalletsWithSystem::whitelisted_storage_keys();
```
deuszx, line 1050: Can you explain why we're doing that? I'm looking at the notion document about benchmarking right now and it doesn't contain the explanation either. So maybe put it there (in the docs). Also, while you're at it, can you update said doc to have code snippets more aliged with what we will have here after this PR? Thank you.

bin/runtime/src/lib.rs
```diff
@@ -1047,22 +1043,11 @@ impl_runtime_apis! {
...
-                hex_literal::hex!("c2261276cc9d1f8598ea4b6a74b15c2f57c875e4cff74148e4628f264b974c80").to_vec().into(),
-                // Execution Phase
-                hex_literal::hex!("26aa394eea5630e07c48ae0c9558cef7ff553b5a9862a516939d82b3d3d8661a").to_vec().into(),
-                // Event Count
-                hex_literal::hex!("26aa394eea5630e07c48ae0c9558cef70a98fdbe9ce6c55837576c60c7af3850").to_vec().into(),
-                // System Events
-                hex_literal::hex!("26aa394eea5630e07c48ae0c9558cef780d41e5e16056765bc8461851072c9d7").to_vec().into(),
-            ];
+            use frame_benchmarking::{
+                Benchmarking, BenchmarkBatch, TrackedStorageKey, cb_add_benchmarks, add_benchmark};
```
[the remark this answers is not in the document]
pmikolajczyk41: unfortunately, they are required due to the macro expansion in `benches` module... I couldn't get it complied without it

bin/runtime/src/lib.rs
pmikolajczyk41: I've updated these snippets already :) but I will add also a word about whitelisting

deuszx (approved):

PR #811, opened 2022-12-13
A0-1576: Move clique network into network

timorl:
# Description

Almost done with this series of refactors, 3/4 and the last one should be tiny.

## Type of change

Refactor

# Checklist:


finality-aleph/src/network/manager/service.rs
```diff
@@ -585,7 +585,7 @@ impl<GE: Display> Display for Error<GE> {
         match self {
             CommandsChannel => write!(f, "commands channel unexpectedly closed"),
             MessageChannel => write!(f, "message channel unexpectedly closed"),
-            ValidatorNetwork => write!(f, "validator network unexpectedly done"),
+            CliqueNetwork => write!(f, "validator network unexpectedly done"),
```
maciejnems, line 588:
```suggestion
            CliqueNetwork => write!(f, "clique network unexpectedly done"),
```

finality-aleph/src/network/manager/service.rs
```diff
@@ -558,14 +558,14 @@ pub struct IO<
...
-    VN: ValidatorNetwork<A::PeerId, A, DataInSession<D>>,
+    CN: CliqueNetwork<A::PeerId, A, DataInSession<D>>,
     GN: GossipNetwork<VersionedAuthentication<M, A>>,
 > where
     A::PeerId: PublicKey,
 {
     commands_from_user: mpsc::UnboundedReceiver<SessionCommand<D>>,
     messages_from_user: mpsc::UnboundedReceiver<(D, SessionId, Recipient)>,
-    validator_network: VN,
+    validator_network: CN,
```
maciejnems, line 568: Have you left it named `validator_network` here and in other places on purpose?

maciejnems (commented):
There are still some comments saying `validator network` and variables/fields named `validator_network...`. I'm not sure we want to keep them, or completely get rid of it?

finality-aleph/src/network/manager/service.rs
timorl: Yes, in here it is actually the validator network, because it is for the validators specifically, not just an abstract clique network.

finality-aleph/src/network/manager/service.rs
timorl: Actually I did the opposite, for the same reason as I kept the variable names before. So now the error is called `ValidatorNetwork` again.

finality-aleph/src/network/manager/service.rs
maciejnems: Got it

finality-aleph/src/network/clique/service.rs
```diff
@@ -301,8 +303,8 @@ where
                 },
                 // periodically reporting what we are trying to do
                 _ = status_ticker.tick() => {
-                    info!(target: "validator-network", "Validator Network status: {}", self.manager.status_report());
-                    debug!(target: "validator-network", "Validator Network legacy status: {}", self.legacy_manager.status_report());
+                    info!(target: LOG_TARGET, "Validator Network status: {}", self.manager.status_report());
```
maciejnems, line 306: So from what I understand, inside clique network, we should know nothing about validators. So shouldn't this and the following status say:
```
Clique Network status: ...
```
Also in this file comments at lines 77 and 98 probably should say clique instead of validator.

timorl: Yup, good points and hopefully fixed all the remaining instances now, thanks!

maciejnems: Just out curiosity, gossip network statuses will also be changed, right? Because right now they say `Network status...` which might be confusing a little. I'm worried I missed that in last review :/

maciejnems (approved):
LGTM

finality-aleph/src/network/clique/service.rs
timorl: This one says `"Clique Network status..."` so it shouldn't be _too_ confusing. Besides they are under different logging targets.

finality-aleph/src/network/clique/service.rs
```diff
@@ -133,7 +135,7 @@ where
         let dialer = self.dialer.clone();
         let next_to_interface = self.next_to_interface.clone();
         self.spawn_handle
-            .spawn("aleph/validator_network_outgoing", None, async move {
+            .spawn("aleph/clique_network_outgoing", None, async move {
```
Yithis, line 138: Since `LOG_TARGET` inspired me, that maybe almost every string should be const at the begining of the file? Only exception I can think of are error/expect messages, but probably there are more

finality-aleph/src/network/clique/service.rs
```diff
@@ -154,7 +156,7 @@ where
         let secret_key = self.secret_key.clone();
         let next_to_interface = self.next_to_interface.clone();
         self.spawn_handle
-            .spawn("aleph/validator_network_incoming", None, async move {
+            .spawn("aleph/clique_network_incoming", None, async move {
```
Yithis, line 159: Same as above, especially that it is the same string

finality-aleph/src/testing/clique_network.rs
```diff
@@ -0,0 +1,310 @@
...
+    large_message_interval: Option<usize>,
+    corrupted_message_interval: Option<usize>,
+    status_report_interval: Duration,
+) {
+    // create spawn_handle, we need to keep the task_manager
+    let task_manager =
+        TaskManager::new(Handle::current(), None).expect("should create TaskManager");
+    let spawn_handle = task_manager.spawn_handle();
+    // create peer identities
+    info!(target: LOG_TARGET, "generating keys...");
```
Yithis, line 111: I think that tests can have a little different target for clarity

scripts/run_nodes.sh
```diff
@@ -101,7 +101,7 @@ run_node() {
     --validator-port ${validator_port} \
     -laleph-party=debug \
     -laleph-network=debug \
-    -lvalidator-network=debug \
+    -lclique-network=debug \
```
Yithis, line 104: Would be great to have it automatized by all `LOG_TARGET` s

finality-aleph/src/network/clique/service.rs
timorl: If these are used only once (and these are, I think you missed the "outgoing"/"incoming" distinction) it's not as important. I think we should do the `LOG_TARGET` thing with log targets in general though, but this is definitely not the PR for that.

scripts/run_nodes.sh
timorl: That would require quite a bit of `sed`/`awk` magic, I think it's a good idea, but probably not for this PR.

Maybe we could get an ADR for logging targets and then executing it would also include magicking up the appropriate script here?

finality-aleph/src/testing/clique_network.rs
timorl: Good point, made a separate `const` for them and also for `tcp` (also makes sense to have that be separate) and made the one in `clique` non-public.

timorl:
I got an off-band accept for this PR from @Yithis (who is having some technical problems), so merging.

PR #813, opened 2022-12-14
A0-1459: Replace triggering workflow from indexer repo with kubectl cmds

mikogs:
Calling workflow in the `indexer` repository has been replaced with `kubectl` commands. 

We explicitly restart indexer databases when deploying contracts (if addresses.json has changed then they are going to be wipe out). We do not do that when a new version of indexer is released (in the `indexer` repo). Hence, it is more meaningful to have it here (same as it is for `the-button`). 

Once this is merged, `kubectl` commands will be removed from `indexer`'s workflow.

### DIFF
--- a/.github/workflows/contracts-e2e-tests-and-deploy.yaml
+++ b/.github/workflows/contracts-e2e-tests-and-deploy.yaml
@@ -303,23 +303,55 @@ jobs:
       - name: Cleanup cache
         uses: ./.github/actions/post-cache
 
-      - name: Trigger Indexer deployment workflow
-        run: |
-          curl -X POST 'https://api.github.com/repos/Cardinal-Cryptography/indexer/actions/workflows/build-and-deploy-to-devnet.yml/dispatches' \
-          -H "Accept: application/vnd.github+json" \
-          -H 'Authorization: Bearer ${{ secrets.CI_GH_TOKEN }}' \
-          -d '{ "ref":"master" }'
+      - name: GIT | Checkout aleph-apps repo
+        uses: actions/checkout@master
+        with:
+          repository: Cardinal-Cryptography/aleph-apps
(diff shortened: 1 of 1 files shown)


Kosiare (approved):

.github/workflows/contracts-e2e-tests-and-deploy.yaml
```diff
@@ -303,23 +303,57 @@ jobs:
...
         uses: azure/setup-kubectl@v2.0
         with:
           version: 'v1.23.6'
 
-      - name: Restart the-button deployment
+      - name: INDEXER | Destroy archive and squid apps
         shell: bash
         run: |
           aws eks --region eu-central-1 update-kubeconfig --name alephzero-devnet-eu-central-1-eks
+          kubectl delete -n indexer-squid --ignore-not-found=true deploy squid-api
```
fbielejec, line 328: Am I correct in thinking that these steps will be run on every push to the branch? Could we add a condition that the changes should touch the `contracts/` directory to trigger the deployment?

There was also a condition that checks the hash of the addresses.json file with the last deployed one (stored in S3) - is that gone now?

mikogs: Both checking hash of the addresses.json file and downloading addresses.json from S3 bucket are now done by init containers that execute before database or app starts. We don't want these to be part of the deployment workflows in the indexer or the-button repositories. The steps here that re-deploy archive and squid are basically steps that were earlier in the indexer repository's workflow.

As for the triggering this workflow. There was a `paths` setting set to `contracts` but it was removed on yapee's request. 
If we really need to make it run only when `contracts/**` have changed we could check adding an action with few conditions, like: https://github.com/marketplace/actions/has-changed-path or https://github.com/dorny/paths-filter ...

fbielejec (approved):

PR #814, opened 2022-12-15
A0-1538: granular benches for pallet `snarcos`

pmikolajczyk41:
We charge less for extrinsics that don't reach to verification procedure.


pallets/snarcos/src/lib.rs
```diff
@@ -254,29 +254,53 @@ pub mod pallet {
...
+                + public_input.len().saturating_sub(data_length_limit);
             ensure!(
-                proof.len() <= T::MaximumDataLength::get() as usize
-                    && public_input.len() <= T::MaximumDataLength::get() as usize,
-                (Error::<T>::DataTooLong, None)
+                data_length_excess == 0,
+                (
+                    Error::<T>::DataTooLong,
+                    Some(T::WeightInfo::verify_data_too_long(
+                        data_length_excess as u32
```
deuszx, line 265: Have we discussed this ever before and agreed that we will charge the exact amount of gas that is equal to the exceess data length?

pallets/snarcos/src/lib.rs
```diff
@@ -254,29 +254,53 @@ pub mod pallet {
...
+                )
             })?;
 
             let public_input: Vec<S::CircuitField> =
                 CanonicalDeserialize::deserialize(&*public_input).map_err(|e| {
                     log::error!("Deserializing public input failed: {:?}", e);
-                    (Error::<T>::DeserializingPublicInputFailed, None)
+                    (
+                        Error::<T>::DeserializingPublicInputFailed,
+                        Some(T::WeightInfo::verify_data_decoding_fails(
```
deuszx: In `weights.rs`, this function (`verify_data_decoding_fails`) returns constant cost, regardless of the input.

deuszx (approved):

bin/runtime/src/lib.rs
```diff
@@ -327,14 +327,16 @@ impl pallet_aleph::Config for Runtime {
 }
 
 parameter_types! {
-    // We allow 10kB keys. This is in 100% blind guess.
+    // We allow 10kB keys, proofs and public inputs. This is in 100% blind guess.
     pub const MaximumVerificationKeyLength: u32 = 10_000;
+    pub const MaximumDataLength: u32 = 10_000;
```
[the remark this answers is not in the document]
pmikolajczyk41: while this could make sense in general, this whole file has _a lot_ of such parameters, therefore for consistency I would leave it as it is now, especially that these two parameters are used (only) right below declaration

pallets/snarcos/src/benchmarking/suite.rs
```diff
@@ -115,9 +115,46 @@ benchmarks! {
 
     // Partial `verify` execution
 
+    verify_data_too_long {
+        // Excess. Unfortunately, anything like
+        // `let e in (T::MaximumDataLength::get() + 1) .. (T::MaximumDataLength::get() * 1_000)`
+        // doesn't compile.
+        let e in 1 .. T::MaximumDataLength::get() * 1_000;
```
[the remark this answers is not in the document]
pmikolajczyk41: it is handled by the `benchmarks` macro and therefore we use the expected syntax... basically, in macros you can use your custom syntax

pallets/snarcos/src/benchmarking/suite.rs
```diff
@@ -115,9 +115,46 @@ benchmarks! {
...
+        let proof = vec![255u8; (T::MaximumDataLength::get() + e) as usize];
```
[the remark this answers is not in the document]
pmikolajczyk41: we are checking how expensive it is to reject the extrinsic early

pallets/snarcos/src/lib.rs
pmikolajczyk41: We will charge `T::WeightInfo::verify_data_too_long(data_length_excess as u32)`, where `T::WeightInfo::verify_data_too_long(x)` tells us how much weight is consumed for handling extrinsic with arguments that has `x` bytes over limit.

pallets/snarcos/src/lib.rs
pmikolajczyk41: nevertheless, the semantics requires passing data length; in particular, I suppose that on another machine, this parameter might be discovered to have some impact

deuszx: If the costs are constant, this function could be parameter-less but it's probably not in our control?

pallets/snarcos/src/lib.rs
deuszx: Why not charge with `proof.len()` and `public_input.len()` though? Again, cost is constant so it probably doesn't matter.

pallets/snarcos/src/benchmarking/suite.rs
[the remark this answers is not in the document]
deuszx: I know. Still, it could have used something that is closer to valid Rust instead.

pallets/snarcos/src/lib.rs
deuszx: OK, I think I see the reason now. If `x` was valid (i.e. not exceeding the limit) then the cost of this operation would be different so there's no real reason to use _that_ value here instead.

pallets/snarcos/src/lib.rs
pmikolajczyk41: out of control

bin/runtime/src/lib.rs
```diff
@@ -327,14 +327,16 @@ impl pallet_aleph::Config for Runtime {
 }
 
 parameter_types! {
-    // We allow 10kB keys. This is in 100% blind guess.
+    // We allow 10kB keys, proofs and public inputs. This is in 100% blind guess.
```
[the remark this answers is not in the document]
pmikolajczyk41: +1

PR #815, opened 2022-12-16
A0-1576: Improve network manager interface

timorl:
# Description

Final, 4/4, part of this refactor.

## Type of change

Refactor.

# Checklist:


kostekIV (commented):
I am not sure what was added and what was jsut moved. Could the moving and changes be in separate prs

finality-aleph/src/party/manager/mod.rs
```diff
@@ -404,20 +405,20 @@ where
     async fn early_start_validator_session(
         &self,
         session: SessionId,
+        node_id: NodeIndex,
```
kostekIV, line 408: im not sure whether this should be provided externally. Previously we were sure whether node_id is a correct one. Now this might introduce some subtle bugs with idxes

timorl: That's true, but there is another method here that already uses this approach (`spawn_authority_task_for_session`). Furthermore, the only use of this method went out of it's way to avoid learning about `node_id` which was pretty weird.

The Extremely Correct way of doing this would be to remove the `node_idx` method from the trait and have the methods do this internally. But this would also require merging the early start method with the nonvalidator method to an extent and in general would be a bigger (although not enormous) refactor.

I would suggest changing it here for consistency/simplicity and adding a task to refactor this properly, wdyt?

kostekIV: I'm okay with that :)

kostekIV (approved):

maciejnems (approved):
Is a little bit hard to distinguish what was moved and what was changed but LGTM

finality-aleph/src/party/manager/mod.rs
timorl: Added as A0-1769.

PR #816, opened 2022-12-19
A0-1610: Contract support redux

obrok:
# Description

Rewrites contract interaction tooling (without events - coming in next PR) to work with subxt and new substrate.

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected), although this already doesn't work after the upgrade to subxt.

# Checklist:

* [x] I have added tests
* [x] I have made corresponding changes to the existing documentation
* [x] I have bumped aleph-client version if relevant


aleph-client/src/contract/convertible_value.rs
```diff
@@ -0,0 +1,120 @@
...
+                        bail!("Unexpected number of elements in Err variant: {:?}", &value);
+                    }
+                }
+                _ => (),
+            }
+        }
+
+        bail!("Expected {:?} to be an Ok(_) or Err(_) tuple.", value);
+    }
+}
```
deuszx, line 120: Can you add the impl for `Option<T>`?

aleph-client/src/contract/mod.rs
```diff
@@ -90,93 +79,109 @@ impl ContractInstance {
...
+    ) -> Result<()> {
+        self.contract_exec_value(conn, message, args, 0).await
+    }
+
+    /// Executes a 0-argument contract call sending the given amount of value with it.
+    pub async fn contract_exec_value0(
+        &self,
+        conn: &SignedConnection,
+        message: &str,
+        value: u128,
```
deuszx, line 141: Is this in pico AZERO or AZERO? 

Also, nit: can we rename it to `amount` or `pazero_amount`?

Longer-term: How about we introduce the `AZERO` struct with proper constructors and accessors.

aleph-client/src/contract/mod.rs
```diff
@@ -90,93 +79,109 @@ impl ContractInstance {
...
+        .map(|_| ())
     }
 
-    fn encode(&self, message: &str, args: &[&str]) -> Result<Vec<u8>> {
+    fn encode<S: AsRef<str> + Debug>(&self, message: &str, args: &[S]) -> Result<Vec<u8>> {
         self.transcoder.encode(message, args)
     }
+
+    fn decode(&self, message: &str, data: Vec<u8>) -> Result<Value> {
+        println!("Decoding: {:?} from {:?}", data, message);
```
deuszx, line 176: Leftover?

aleph-client/src/contract/convertible_value.rs
obrok: Sure, why not.

aleph-client/src/contract/mod.rs
obrok: > Longer-term: How about we introduce the AZERO struct with proper constructors and accessors.
 
All for this, it's very inconvenient to express values right now.

> Also, nit: can we rename it to amount or pazero_amount?

I don't like it. `Pico` and especially `p` isn't better for me personally, and we are just using these raw `u128` everywhere, so it wouldn't even be consistent.

deuszx: OK, so please add as part of the comment the unit.

e2e-tests/src/test/adder.rs
```diff
@@ -0,0 +1,71 @@
...
+    let config = setup_test();
+
+    let (conn, _authority, account) = basic_test_context(config).await?;
+    let contract = AdderInstance::new(
+        &config.test_case_params.adder,
+        &config.test_case_params.adder_metadata,
+    )?;
+
+    let before = contract.get(&conn).await?;
+    contract.add(&sign(&conn, &account), 10).await?;
```
deuszx, line 24: nit: extract `10` to a value and use instead

e2e-tests/src/test/adder.rs
```diff
@@ -0,0 +1,71 @@
...
+    fn from(instance: &'a AdderInstance) -> Self {
+        &instance.contract
+    }
+}
+
+impl<'a> From<&'a AdderInstance> for AccountId {
+    fn from(instance: &'a AdderInstance) -> Self {
+        instance.contract.address().clone()
+    }
+}
```
deuszx, line 46: Is it correct that we could implement this for `ContractInstance`:
```rust
impl<'a> From<&'a ContractInstance> for AccountId {
    ...
}
```
and not repeat it for every `ContractInstance`-wrapper

obrok: I think then you'd need to do `into().into()` at the usage site, which is so-so... I think it would need to be something like `AsRef` or `Deref` to make it seamless. In any case - I would consider those a refactor for later.

e2e-tests/src/test/helpers.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+impl From<KeyPairWrapper> for AccountId {
+    fn from(keypair: KeyPairWrapper) -> Self {
+        (&keypair).into()
+    }
+}
+
+/// Creates a copy of the `connection` signed by `signer`
+pub fn sign(conn: &Connection, signer: &KeyPairWrapper) -> SignedConnection {
```
deuszx, line 43: Why not make it a method on the instance of `KeyPairWrapper` that you control? 
```rust
impl KeyPairWrapper {
    pub fn signed_connection(&self, conn: &Connection) -> SignedConnection { ... }
}
```

I find 
```rust
key_pair.signed_connection(&conn)
```
to be a better dev experience than
```rust
use helpers::sign;

sign(signer, conn)
```

aleph-client/src/contract/mod.rs
obrok: > OK, so please add as part of the comment the unit.

Same as with the name - I don't see the point. I'm not even sure about the unit myself (other than `1` here being the smallest value expressible as a `Balance`). It's not that I wouldn't like this to be expressed better  it's that I think having some mentions of units in this file, but not in others would make it worse. For example, currently grepping `aleph-client` by things like `pico`, `unit`, `decimal`, gives nothing.

I'd propose raising this as a separate ticket - "Unit Overhaul".

deuszx (approved):
Code seems correct.

I'll leave the CI setup to one of our devops experts though. I'm fine with how current changes look like but maybe there's some convention/plan that I don't know of that they would prefer sticking to.

fbielejec (approved):

PR #820, opened 2022-12-20
bump versions

mike1729:
Bump legacy and current versions of aleph-bft.

### DIFF
--- a/finality-aleph/src/abft/common.rs
+++ b/finality-aleph/src/abft/common.rs
@@ -34,19 +34,6 @@ pub fn unit_creation_delay_fn(unit_creation_delay: UnitCreationDelay) -> DelaySc
     })
 }
 
-pub fn coord_request_delay_fn() -> DelaySchedule {
-    Arc::new(|t| match t {
-        0 => Duration::from_millis(0),
-        1 => Duration::from_millis(50),
-        2 => Duration::from_millis(1000),
-        _ => Duration::from_millis(3000 * (t as u64 - 2)),
-    })
-}
-
-pub fn coord_request_recipients_fn() -> RecipientCountSchedule {
-    Arc::new(|t| if t <= 2 { 3 } else { 1 })
-}
-
 pub struct DelayConfig {
     pub tick_interval: Duration,
     pub requests_interval: Duration,
--- a/finality-aleph/src/abft/common.rs
+++ b/finality-aleph/src/abft/common.rs
@@ -25,15 +25,26 @@ fn exponential_slowdown(
 }
 
 pub type DelaySchedule = Arc<dyn Fn(usize) -> Duration + Sync + Send + 'static>;
+pub type RecipientCountSchedule = Arc<dyn Fn(usize) -> usize + Sync + Send + 'static>;
 
 pub fn unit_creation_delay_fn(unit_creation_delay: UnitCreationDelay) -> DelaySchedule {
-    Arc::new(move |t| {
-        if t == 0 {
-            Duration::from_millis(2000)
-        } else {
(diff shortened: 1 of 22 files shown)


finality-aleph/src/abft/current.rs
```diff
@@ -76,11 +79,16 @@ pub fn create_aleph_config(
     unit_creation_delay: UnitCreationDelay,
 ) -> Config {
     let delay_config = DelayConfig {
-        tick_interval: Duration::from_millis(100),
+        tick_interval: Duration::from_millis(10),
```
kostekIV, line 82: why 100 -> 10?

finality-aleph/src/abft/legacy.rs
```diff
@@ -71,11 +75,16 @@ pub fn create_aleph_config(
     unit_creation_delay: UnitCreationDelay,
 ) -> Config {
     let delay_config = DelayConfig {
-        tick_interval: Duration::from_millis(100),
+        tick_interval: Duration::from_millis(10),
```
kostekIV, line 78: same question

finality-aleph/src/abft/common.rs
```diff
@@ -25,23 +25,39 @@ fn exponential_slowdown(
...
+    Arc::new(|t| match t {
+        0 => Duration::from_millis(0),
+        1 => Duration::from_millis(50),
+        2 => Duration::from_millis(1000),
+        _ => Duration::from_millis(3000 * (t as u64 - 2)),
     })
 }
 
+pub fn coord_request_recipients_fn() -> RecipientCountSchedule {
+    Arc::new(|t| if t <= 2 { 3 } else { 1 })
```
timorl, line 47: It would be nice to have some explanation for these functions in comments, as they are written it's very unclear what they are doing. In particular this one seems weird intuitively, since we are first trying to ask more nodes, and if that fails we start asking fewer? Although I might be misunderstanding what `t` is.

timorl (changes requested):
I think oyu missed bumping the version numbers, and that is immensely important.

Other than that I endorse @kostekIV 's comments, and in general it would be nice to have stuff a bit more commented here and there.

finality-aleph/src/abft/current.rs
mike1729: This and all changes in config are consequences of changes in aleph-bft

finality-aleph/src/abft/common.rs
mike1729: All those functions are taken from aleph-bft. The q is then do we want to make them public there or duplicate here?

finality-aleph/src/abft/legacy.rs
mike1729: Changed to default.

finality-aleph/src/abft/common.rs
mike1729: Changed to default.

kostekIV (approved):

finality-aleph/src/abft/current.rs
```diff
@@ -24,8 +16,10 @@ use crate::{
     CurrentNetworkData, Hasher, Keychain, NodeIndex, SessionId, SignatureSet, UnitCreationDelay,
 };
 
+use super::common::unit_creation_delay_fn;
```
timorl: We don't generally use `super` outside of tests, preferably just import it using `crate`. Although tbh I'm not sure why, but still, consistency.

timorl (approved):

finality-aleph/src/abft/current.rs
mike1729: done.

PR #822, opened 2022-12-21
A0-1761: Enhance gossip network interface

timorl:
# Description

Make the gossip network return the peer id of the source of received data and add two, currently unused, methods to the interface – they will be used by block sync.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/network/gossip/mod.rs
+++ b/finality-aleph/src/network/gossip/mod.rs
@@ -1,5 +1,6 @@
 //! A P2P-based gossip network, for now only for sending broadcasts.
 use std::{
+    collections::HashSet,
     fmt::{Debug, Display},
     hash::Hash,
 };
@@ -15,9 +16,25 @@ mod service;
 pub use service::Service;
 
 #[async_trait::async_trait]
-/// Interface for the gossip network, currently only supports broadcasting and receiving data.
+/// Interface for the gossip network. This represents a P2P network and a lot of the properties of
+/// this interface result from that. In particular we might know the ID of a given peer, but not be
+/// connected to them directly.
 pub trait Network<D: Data>: Send + 'static {
     type Error: Display + Send;
+    type PeerId: Clone + Debug + Eq + Hash + Send + 'static;
+
+    /// Attempt to send data to a peer. Might silently fail if we are not connected to them.
+    fn send(&mut self, data: D, peer_id: Self::PeerId) -> Result<(), Self::Error>;
+
+    /// Send data to a random peer, preferably from a list. It should send the data to a randomly
+    /// chosen peer from the provided list, but if it cannot (e.g. because it's not connected) it
+    /// will send to a random available peer. No guarantees any peer gets it even if no errors are
+    /// returned, retry appropriately.
+    fn send_to_random(
+        &mut self,
+        data: D,
+        peer_ids: HashSet<Self::PeerId>,
+    ) -> Result<(), Self::Error>;
 
     /// Broadcast data to all directly connected peers. Network-wide broadcasts have to be
     /// implemented on top of this abstraction. Note that there might be no currently connected
@@ -25,8 +42,8 @@ pub trait Network<D: Data>: Send + 'static {
     /// returned, retry appropriately.
     fn broadcast(&mut self, data: D) -> Result<(), Self::Error>;
 
-    /// Receive some data from the network.
-    async fn next(&mut self) -> Result<D, Self::Error>;
+    /// Receive some data from the network, including information about who sent it.
+    async fn next(&mut self) -> Result<(D, Self::PeerId), Self::Error>;
 }
 
 /// The Authentication protocol is used for validator discovery.
@@ -51,7 +68,7 @@ pub trait NetworkSender: Send + Sync + 'static {
 pub enum Event<P> {
     StreamOpened(P, Protocol),
     StreamClosed(P, Protocol),
-    Messages(Vec<(Protocol, Bytes)>),
+    Messages(P, Vec<(Protocol, Bytes)>),
 }
 
 #[async_trait::async_trait]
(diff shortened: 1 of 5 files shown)


finality-aleph/src/network/gossip/mod.rs
```diff
@@ -15,18 +16,34 @@ mod service;
...
-/// Interface for the gossip network, currently only supports broadcasting and receiving data.
+/// Interface for the gossip network. This represents a P2P network and a lot of the properties of
+/// this interface result from that. In particular we might know the ID of a given peer, but not be
+/// connected to them directly.
 pub trait Network<D: Data>: Send + 'static {
     type Error: Display + Send;
+    type PeerId: Clone + Debug + Eq + Hash + Send + 'static;
+
+    /// Attempt to send data to a peer. Might silently fail if we are not connected to them.
+    fn send(&mut self, data: D, peer_id: Self::PeerId) -> Result<(), Self::Error>;
```
maciejnems, line 27: nit: just a small preference I have is naming this `send_to`. Feel free to ignore this comment.

finality-aleph/src/network/gossip/service.rs
```diff
@@ -26,16 +34,16 @@ use crate::{
...
+    messages_for_user: mpsc::UnboundedSender<(D, N::PeerId)>,
     authentication_connected_peers: HashSet<N::PeerId>,
     authentication_peer_senders: HashMap<N::PeerId, TracingUnboundedSender<D>>,
     spawn_handle: SpawnTaskHandle,
 }
 
-struct ServiceInterface<D: Data> {
-    messages_from_service: mpsc::UnboundedReceiver<D>,
-    messages_for_service: mpsc::UnboundedSender<D>,
+struct ServiceInterface<D: Data, P: Clone + Debug + Eq + Hash + Send + 'static> {
```
maciejnems, line 44: Right now `ServiceInterface` just sends `Command` without any `Protocol`. And after that everything is just sent over `Authentication` protocol.

Maybe it would be nice to add a field to this struct that states which protocol to use. Then we can extend `Command` enum to also send `Protocol`. Something like:
```
Command::Send(Data, PeerId, Protocol)
```
This way we do not have to change `Network` trait to contain protocol in each sending call (broadcast, send, send_to_random). This way connection manager would not have to know about  `Protocol` enum. It would just have a network interface that sends through Authentication protocol. And from what I understand in the future we want to have several protocols (like Authentication and BlockSync).

finality-aleph/src/network/gossip/mod.rs
```diff
@@ -15,18 +16,34 @@ mod service;
 pub use service::Service;
 
 #[async_trait::async_trait]
-/// Interface for the gossip network, currently only supports broadcasting and receiving data.
+/// Interface for the gossip network. This represents a P2P network and a lot of the properties of
+/// this interface result from that. In particular we might know the ID of a given peer, but not be
+/// connected to them directly.
 pub trait Network<D: Data>: Send + 'static {
     type Error: Display + Send;
+    type PeerId: Clone + Debug + Eq + Hash + Send + 'static;
```
maciejnems, line 24: Maybe we could just reuse `PeerId` trait here from `finality-aleph/src/network/mod.rs`?

finality-aleph/src/network/gossip/service.rs
```diff
@@ -330,12 +396,25 @@ mod tests {
     #[async_trait::async_trait]
     impl Network<MockData> for TestData {
         type Error = Error;
+        type PeerId = MockPublicKey;
+
+        fn send(&mut self, data: MockData, peer_id: Self::PeerId) -> Result<(), Self::Error> {
```
maciejnems, line 401: We should probably add tests for `send` and `send_to_random`. For the latter it would be nice to have one test where peer_ids has intersection of only one peer and another where intersection is zero peers.

finality-aleph/src/network/gossip/mod.rs
```diff
@@ -15,18 +16,34 @@ mod service;
...
     /// Broadcast data to all directly connected peers. Network-wide broadcasts have to be
     /// implemented on top of this abstraction. Note that there might be no currently connected
     /// peers, so there are no guarantees any single call sends anything even if no errors are
     /// returned, retry appropriately.
     fn broadcast(&mut self, data: D) -> Result<(), Self::Error>;
 
-    /// Receive some data from the network.
-    async fn next(&mut self) -> Result<D, Self::Error>;
+    /// Receive some data from the network, including information about who sent it.
+    async fn next(&mut self) -> Result<(D, Self::PeerId), Self::Error>;
```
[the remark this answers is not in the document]
timorl: Responding to requests is one thing, another is noting information about which peers have some information – e.g. if we get a couple best block announcements with headers, we want to request the block related to the header from one of the peers which sent us the announcements.

This call does not guarantee that the returned peer is actually connected though – they might have disconnected in the meantime.

finality-aleph/src/network/gossip/service.rs
timorl: Added tests, good point.

finality-aleph/src/network/gossip/mod.rs
timorl: Eh, there are some differences between these traits, I would prefer not to unify them.

finality-aleph/src/network/gossip/service.rs
timorl: Yeah, good point, improved.

finality-aleph/src/network/gossip/mod.rs
timorl: Renamed then.

finality-aleph/src/network/gossip/service.rs
```diff
@@ -514,16 +593,14 @@ mod tests {
...
+                random_peer_id(),
+                vec![(PROTOCOL, message.clone().encode().into())],
+            ))
             .expect("Should handle");
 
-        assert_eq!(
-            test_data.next().await.expect("Should receive message"),
-            message,
-        );
+        let (received_message, _) = test_data.next().await.expect("Should receive message");
```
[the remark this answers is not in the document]
timorl: True, added.

Marcin-Radecki (approved):

maciejnems (approved):

PR #824, opened 2022-12-22
A0-1609 Removed unwrap()'s in aleph-client

Marcin-Radecki:
# Description

Motivation: get rid of bare `unwrap()` in `aleph-client`. One can just transform them to `expect(msg)`, but we can do better. We can benefit from leveraging `subxt::Error`, ie passing `Result<T, subxt::Error> to the client of `aleph-client`. This can be simple done as `.map_err(|e: subxt::Error| e.into())`, where the return type is `anyhow::Result` that is used already in `aleph-client`. Not all public API of `aleph-client` was modified in order not to bloat this PR. 

Deliberately only minor version of `aleph-client` was bumped to not bump major version twice. This is fine since we don't publish `aleph-client` to crates.io.

## Type of change

### DIFF
--- a/aleph-client/src/connections.rs
+++ b/aleph-client/src/connections.rs
@@ -3,7 +3,6 @@ use std::{thread::sleep, time::Duration};
 use anyhow::anyhow;
 use codec::Decode;
 use log::info;
-use primitives::BlockNumber;
 use subxt::{
     ext::sp_core::Bytes,
     metadata::DecodeWithMetadata,
@@ -113,18 +112,6 @@ impl Connection {
         let bytes: Bytes = self.client.rpc().request(&func_name, params).await?;
 
         Ok(R::decode(&mut bytes.as_ref())?)
-    }
-
-    pub async fn get_block_number_inner(
-        &self,
-        block: Option<BlockHash>,
-    ) -> anyhow::Result<Option<BlockNumber>> {
-        self.client
-            .rpc()
-            .header(block)
-            .await
-            .map(|maybe_header| maybe_header.map(|header| header.number))
-            .map_err(|e| e.into())
     }
 }
 
--- a/aleph-client/src/connections.rs
+++ b/aleph-client/src/connections.rs
@@ -3,6 +3,7 @@ use std::{thread::sleep, time::Duration};
 use anyhow::anyhow;
 use codec::Decode;
 use log::info;
+use primitives::BlockNumber;
 use subxt::{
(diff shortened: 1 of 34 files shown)


aleph-client/src/connections.rs
```diff
@@ -113,6 +114,18 @@ impl Connection {
 
         Ok(R::decode(&mut bytes.as_ref())?)
     }
+
+    pub async fn get_block_number_inner(
```
pmikolajczyk41, line 118: the name suggests that it should be a private method (?)

aleph-client/src/utility.rs
```diff
@@ -8,64 +8,63 @@ use crate::{
...
-            .header(Some(block))
+            .finalized_head()
             .await
-            .unwrap()
-            .map(|h| h.number)
+            .map_err(|e| e.into())
+    }
+
+    async fn get_block_number(&self, block: BlockHash) -> anyhow::Result<Option<BlockNumber>> {
+        self.get_block_number_inner(Some(block)).await
```
pmikolajczyk41, line 60: if you convert `get_block_number_inner` into a private method, I would move it to this module

e2e-tests/src/test/finalization.rs
```diff
@@ -10,12 +11,15 @@ pub async fn finalization() -> anyhow::Result<()> {
...
-    let finalized = connection.connection.get_finalized_block_hash().await;
+    let finalized = connection.connection.get_finalized_block_hash().await?;
     let finalized_number = connection
         .connection
         .get_block_number(finalized)
-        .await
-        .unwrap();
+        .await?
+        .ok_or(anyhow!(
+            "Failed to retrieve block number for hash {:?}",
```
pmikolajczyk41, line 20: nit: you can inline `finalized` into the literal, like `anyhow!("Failed to retrieve block number for hash {finalized:?}")`

pmikolajczyk41 (approved):
cool!

e2e-tests/src/test/finalization.rs
Marcin-Radecki: done

aleph-client/src/utility.rs
Marcin-Radecki: done

aleph-client/src/connections.rs
Marcin-Radecki: done

PR #828, opened 2022-12-23
Fixing broken compatibility

timorl:
# Description

Compatibility didn't work, because apparently no one checked?

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/party/manager/mod.rs
+++ b/finality-aleph/src/party/manager/mod.rs
@@ -348,11 +348,13 @@ where
                 info!(target: "aleph-party", "Running session with legacy-only AlephBFT version.");
                 self.legacy_subtasks(params)
             }
-            Ok(version) if version == CURRENT_VERSION => {
+            // The `as`es here should be removed, but this would require a pallet migration and I
+            // am lazy.
+            Ok(version) if version == CURRENT_VERSION as u32 => {
                 info!(target: "aleph-party", "Running session with AlephBFT version {}, which is current.", version);
                 self.current_subtasks(params)
             }
-            Ok(version) if version == LEGACY_VERSION => {
+            Ok(version) if version == LEGACY_VERSION as u32 => {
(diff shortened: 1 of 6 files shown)


finality-aleph/src/party/manager/mod.rs
```diff
@@ -348,11 +348,13 @@ where
                 info!(target: "aleph-party", "Running session with legacy-only AlephBFT version.");
                 self.legacy_subtasks(params)
             }
-            Ok(version) if version == CURRENT_VERSION => {
+            // The `as`es here should be removed, but this would require a pallet migration and I
```
maciejnems, line 351: <3

maciejnems (approved):
ups

mike1729 (approved):
D'oh!

scripts/run_nodes.sh
```diff
@@ -52,7 +52,7 @@ validator_ids_string="${validator_ids[*]}"
 validator_ids_string="${validator_ids_string//${IFS:0:1}/,}"
 
 echo "Bootstrapping chain for nodes 0..$((N_VALIDATORS - 1))"
-./target/release/aleph-node bootstrap-chain --base-path "$BASE_PATH" --account-ids "$validator_ids_string" --chain-type local > "$BASE_PATH/chainspec.json"
+./target/release/aleph-node bootstrap-chain --raw --base-path "$BASE_PATH" --account-ids "$validator_ids_string" --chain-type local > "$BASE_PATH/chainspec.json"
```
Marcin-Radecki, line 55: Why we added `--raw` here? This script is used mostly for development purposes to my knowledge, and having non-raw chainspec makes debugging easier when it comes to e.g. adding new pallet

timorl: But not having a raw chainspec also produces cryptic errors when you are trying to run it on two different versions.

Perhaps we could produce both?

PR #829, opened 2022-12-23
A0-1491: Change protocol naming convention and add block sync protocol

timorl:
# Description

This makes us follow the protocol naming convention outlined in https://github.com/paritytech/substrate/issues/7746. It also adds the block sync protocol, although currently unused – I wanted to make sure the abstraction will work with that.

## Type of change

# Checklist:

### DIFF
--- a/bin/node/src/service.rs
+++ b/bin/node/src/service.rs
@@ -234,6 +234,13 @@ fn setup(
             protocol_naming.clone(),
             Protocol::Authentication,
         ));
+    config
+        .network
+        .extra_sets
+        .push(finality_aleph::peers_set_config(
+            protocol_naming.clone(),
+            Protocol::BlockSync,
+        ));
 
     let (network, system_rpc_tx, tx_handler_controller, network_starter) =
         sc_service::build_network(sc_service::BuildNetworkParams {
--- a/bin/node/src/service.rs
+++ b/bin/node/src/service.rs
@@ -9,11 +9,11 @@ use aleph_primitives::AlephSessionApi;
 use aleph_runtime::{self, opaque::Block, RuntimeApi, MAX_BLOCK_SIZE};
 use finality_aleph::{
     run_nonvalidator_node, run_validator_node, AlephBlockImport, AlephConfig,
-    JustificationNotification, Metrics, MillisecsPerBlock, Protocol, SessionPeriod,
+    JustificationNotification, Metrics, MillisecsPerBlock, Protocol, ProtocolNaming, SessionPeriod,
 };
 use futures::channel::mpsc;
 use log::warn;
-use sc_client_api::{Backend, HeaderBackend};
+use sc_client_api::{Backend, BlockBackend, HeaderBackend};
 use sc_consensus_aura::{ImportQueueParams, SlotProportion, StartAuraParams};
 use sc_consensus_slots::BackoffAuthoringBlocksStrategy;
 use sc_network::NetworkService;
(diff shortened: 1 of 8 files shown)


bin/node/src/service.rs
```diff
@@ -212,14 +212,28 @@ fn setup(
...
     ),
     ServiceError,
 > {
+    let genesis_hash = client
+        .block_hash(0)
+        .ok()
+        .flatten()
+        .expect("we should have a hash");
+    let chain_prefix = match config.chain_spec.fork_id() {
+        Some(fork_id) => format!("/{}/{}", genesis_hash, fork_id),
```
lesniak43, line 226: what's the idea with `fork_id`?

lesniak43 (approved):

bin/node/src/service.rs
timorl: This is for when/if we ever hard fork. We don't expect this to happen and afaik `fork_id` currently returns `None`, so all good, but if it ever happens we definitely don't want the networks mixing.

lesniak43: then I'd say we should either not consider this case in this PR at all, or include the non-fork name in the fallback list

timorl: The latter completely misses the point, anything included in the fallback list _will_ be connected, and we don't want that.

As for the former, we could, but I am trying to stay completely in sync with the substrate naming convention, as I think in this case it's smart. The main reason is that if there ever is a fork it might be hard to remember that this is one of the things that need to be changed for stuff to work (or worse, someone hacks around the problem).

lesniak43: so why do you want to have fallback names now, but not in case of a hard fork? both situations require a transition period during which everyone updates their nodes, right?

timorl: Oh, nono, a hard fork would start completely disconnected from the previous network. This is _extremely_ unlikely to ever happen, but in case it does there will be no transition period like this.

finality-aleph/src/network/gossip/service.rs
```diff
@@ -262,6 +276,12 @@ impl<N: RawNetwork, D: Data> Service<N, D> {
                         self.authentication_peer_senders.insert(peer.clone(), tx);
                         rx
                     }
+                    Protocol::BlockSync => {
+                        let (tx, rx) = tracing_unbounded("mpsc_notification_stream_block_sync");
+                        self.block_sync_connected_peers.insert(peer.clone());
```
maciejnems, line 281: nit: could be added to status report

bin/node/src/service.rs
```diff
@@ -212,14 +212,28 @@ fn setup(
...
+        None => format!("/{}", genesis_hash),
+    };
+    let protocol_naming = ProtocolNaming::new(chain_prefix);
     config
         .network
         .extra_sets
-        .push(finality_aleph::peers_set_config(Protocol::Authentication));
+        .push(finality_aleph::peers_set_config(
```
maciejnems, line 233: So for now we do not add block sync config yet?

maciejnems (approved):

bin/node/src/service.rs
timorl: Uhh, whoopsie? Added.

Also, the two changes together let me catch a bug, so yeah, thanks. <_<""

finality-aleph/src/network/gossip/service.rs
timorl: Good point, added.

PR #833, opened 2022-12-28
A0-1796: add justification broadcast ticker

maciejnems:
# Description

Adding broadcast ticker that will be used when deciding whether we should broadcast justification, or wait.

## Type of change

- New feature (not yet in use)

# Checklist:

- I have added tests


finality-aleph/src/sync/mod.rs
```diff
@@ -0,0 +1,3 @@
+// TODO: remove when ticker is used
+#[allow(dead_code)]
```
timorl, line 2: I think you can slap this above the `mod sync` in `lib.rs` instead.

finality-aleph/src/sync/ticker.rs
```diff
@@ -0,0 +1,143 @@
...
+    current_periodic_timeout: Duration,
+    max_timeout: Duration,
+    min_timeout: Duration,
+}
+
+impl BroadcastTicker {
+    pub fn new(max_timeout: Duration, min_timeout: Duration) -> Self {
+        Self {
+            last_broadcast: Instant::now(),
+            current_periodic_timeout: max_timeout,
```
timorl, line 15: I think the `periodic_` is unnecessary here.

finality-aleph/src/sync/ticker.rs
```diff
@@ -0,0 +1,143 @@
+use tokio::time::{sleep, Duration, Instant};
+
+/// This struct is used for determining when we should broadcast justification so that it does not happen too often.
```
timorl, line 3: This will not only be about broadcasting justifications, so it would be nice if you could make all the comments more generic – just that we don't want to this to trigger more often than once every `min_timeout`, no less often than once every `max_timeout`, and that there is a special request for an asap tick. Maybe you could even avoid mentioning that this is for broadcasts at all, it technically is just a rate-limited on-demand ticker.

Anyway, this is just an extensive nit, apply however much or little of it as you want (although I would at least generalize away from justifications, otherwise it will have to be changed later).

finality-aleph/src/sync/ticker.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub fn new(max_timeout: Duration, min_timeout: Duration) -> Self {
+        Self {
+            last_broadcast: Instant::now(),
+            current_periodic_timeout: max_timeout,
+            max_timeout,
+            min_timeout,
+        }
+    }
+
+    /// Returns whether we should broadcast right now if we just imported a justification.
```
timorl, line 21: Just to be clearer about the lengthy comment above – e.g. this could be called `try_tick` and it would return whether the tick was successful.

The trigger to me writing these comments was the imo too detailed explanation of batches and justifications, generically, we just don't want the ticks to happen too often, and it's our private matter as to why. :P The comment about (potentially) broadcasting two justifications per batch likely should end up somewhere where this will be used rather than in this documentation.

finality-aleph/src/sync/ticker.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// If min_timeout elapsed since last broadcast, returns true, sets last broadcast to now and will
+    /// return true again if called after `self.min_timeout`.
+    /// If not, returns false and sets periodic broadcast timeout to `self.min_timeout`.
+    /// This is to prevent from sending every justification when importing a batch of them. This way,
+    /// when receiving batch of justifications we will broadcast the first justification and the highest known
+    /// after `self.min_timeout` using periodic broadcast.
+    pub fn try_broadcast(&mut self) -> bool {
+        let now = Instant::now();
+        if now.saturating_duration_since(self.last_broadcast) >= self.min_timeout {
+            self.last_broadcast = now;
```
timorl, line 31: Huh, bug – you should set `timeout` to `max_timeout` here. Maybe even better, have a private method `tick` that sets the last tick and timeout.

timorl: Oh, also make a regression test for that maybe – in this version as far as I can tell after a unsuccessful `try` followed by a successful one the next `wait` only takes `min_timeout` to fire.

finality-aleph/src/sync/ticker.rs
```diff
@@ -0,0 +1,143 @@
...
+            true
+        } else {
+            self.current_periodic_timeout = self.min_timeout;
+            false
+        }
+    }
+
+    /// Sleeps until next periodic broadcast should happen.
+    /// In case enough time elapsed, sets last broadcast to now and periodic timeout to `self.max_timeout`.
+    pub async fn wait_for_periodic_broadcast(&mut self) {
```
timorl, line 41: I think just `wait` should be enough as a name.

finality-aleph/src/sync/ticker.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    #[tokio::test]
+    async fn try_broadcast_after_wait_for_periodic_broadcast() {
+        let max_timeout = Duration::from_millis(700);
+        let min_timeout = Duration::from_millis(100);
+        let mut ticker = BroadcastTicker::new(max_timeout, min_timeout);
+
+        assert_eq!(
+            timeout(
+                max_timeout + min_timeout,
```
timorl, line 132: Why not `2 * max_timeout` like everywhere else?

finality-aleph/src/sync/ticker.rs
```diff
@@ -0,0 +1,143 @@
...
+    use super::BroadcastTicker;
+
+    #[tokio::test]
+    async fn try_broadcast() {
+        let max_timeout = Duration::from_millis(700);
+        let min_timeout = Duration::from_millis(100);
+        let mut ticker = BroadcastTicker::new(max_timeout, min_timeout);
+
+        assert!(!ticker.try_broadcast());
+        sleep(min_timeout).await;
```
timorl, line 62: Don't you want some wiggle room here? Otherwise I would be worried this will fail spuriously sometimes. < _<"

finality-aleph/src/sync/ticker.rs
```diff
@@ -0,0 +1,143 @@
...
+
+#[cfg(test)]
+mod tests {
+    use tokio::time::{sleep, timeout, Duration};
+
+    use super::BroadcastTicker;
+
+    #[tokio::test]
+    async fn try_broadcast() {
+        let max_timeout = Duration::from_millis(700);
```
timorl, line 57: I thought you can have `Duration` constants, but if not then perhaps have a struct/function for these that preferably also details the times one should wait to be sure everything is ok (e.g. `shorter_than_max_but_above_min = 2*min`), but that might be an overkill just for tests.

timorl (approved):
Mostly naming nits, but I found one minor bug. :3

finality-aleph/src/sync/ticker.rs
maciejnems: I thought it was supposed to work like that :P but yeah, there is no reason to broadcast soon after this is successful, so we can set here to `max_timeout`.

finality-aleph/src/sync/ticker.rs
maciejnems: This should not fail, as the sleep lasts at least `min_timeout`. The part I would be worried about are whenever there are `timeout` functions with close time limit. From what I tested, they seem to work, but I can add some wiggle room to them.

finality-aleph/src/sync/ticker.rs
maciejnems: No need to wait so much, but I think I will create two constants. One a little bit over `MIN_TIMEOUT` and one a little bit over `MAX_TIMEOUT`

finality-aleph/src/sync/ticker.rs
maciejnems: I will add duration constants

finality-aleph/src/sync/ticker.rs
maciejnems: Changed after renaming to `Ticker`

finality-aleph/src/sync/ticker.rs
maciejnems: Changed after renaming to `Ticker`

finality-aleph/src/sync/ticker.rs
```diff
@@ -0,0 +1,115 @@
+use tokio::time::{sleep, Duration, Instant};
+
+/// This struct is used for rate limiting as an on-demand ticker. It can be used for ticking
+/// at least once `max_timeout` but not more than once every `min_timeout`.
+pub struct Ticker {
```
kostekIV, line 5: I wonder about the usage of this. The tick rate depends on how often the calls are made and is not guaranteed to tick once every `max_timetout`. My suggestion is to either change docs to describe this or slightly refactor this struct:
1. fn run that runs ticker and sends a notification every max_timeout
2. fn try_tick that tries to tick immediately and sends a notification
It would probably be required to split Ticker into Ticker (the interface with which the caller interacts) and TickerWorker doing all of the work

finality-aleph/src/sync/ticker.rs
```diff
@@ -0,0 +1,115 @@
...
+            current_timeout: max_timeout,
+            max_timeout,
+            min_timeout,
+        }
+    }
+
+    /// Returns whether at least `min_timeout` time elapsed since last tick.
+    /// If `min_timeout` elapsed since last tick, returns true, sets last tick to now,
+    /// current timout to `max_timeout` and will Return true again if called after `min_timeout`.
+    /// If not, returns false and sets current timeout to `min_timeout`.
```
timorl, line 25: This explanation is a bit too technical, it shouldn't be mentioning current timeouts, but rather what will happen. So I would suggest something like "...returns true and records a tick. If not, calls to `wait` will return when `min_timeout` elapses until the next tick." Similarly the comment in `wait` would be "... When enough time elapsed returns and records a tick."

You are also missing a comment by `new`.

timorl (approved):

finality-aleph/src/sync/ticker.rs
timorl: This proposal doesn't actually improve things, you just move the problem from awaiting on the method to awaiting on the channel. Worse, it breaks the main guarantee we are interested in, that no ticks happen sooner than `min_timeout`, or at least I don't see how you can uphold it with a channel.

Not to mention it's massively more complicated. xd

kostekIV: It makes it more complicated here but easier to use for the user of the struct. Anyway, I'm fine here with just docs reflecting what I wrote in the first paragraph :)

maciejnems: I think that approach with ticker worker is unnecessary complicated and current usage for user is pretty simple. You just add `wait` to your main select loop and use `try_tick` whenever you would like to tick sooner in another branch of select. So in this case what I wrote in docs is guaranteed (apart from some stupid cases with heavy synchronous tasks). If you want to I can add this to the doc.

finality-aleph/src/sync/ticker.rs
```diff
@@ -0,0 +1,115 @@
...
+            true
+        } else {
+            self.current_timeout = self.min_timeout;
+            false
+        }
+    }
+
+    /// Sleeps until next tick should happen. In case enough time elapsed,
+    /// sets last tick to now and current timeout to `max_timeout`.
+    pub async fn wait(&mut self) {
```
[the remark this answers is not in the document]
maciejnems: I would leave it named as it is now

finality-aleph/src/sync/ticker.rs
kostekIV: Yes, but it is not guaranteed by the ticker itself

finality-aleph/src/sync/ticker.rs
```diff
@@ -0,0 +1,115 @@
...
+/// at least once `max_timeout` but not more than once every `min_timeout`.
+pub struct Ticker {
+    last_tick: Instant,
+    current_timeout: Duration,
+    max_timeout: Duration,
+    min_timeout: Duration,
+}
+
+impl Ticker {
+    pub fn new(max_timeout: Duration, min_timeout: Duration) -> Self {
```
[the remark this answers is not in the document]
maciejnems: So you would like me to add an assert here to panic in case of opposite?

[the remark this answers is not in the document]
maciejnems: I don't like asserts in production code, but I guess it should be harmless here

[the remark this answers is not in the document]
timorl: Alternatively you could `max_timeout=min_timeout` in this case rather than panic. Neither choice should matter inpractice, so eh.

finality-aleph/src/sync/ticker.rs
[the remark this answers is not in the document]
kostekIV: but we wait and tick here, its not clear from name that tick happens :(

finality-aleph/src/sync/ticker.rs
```diff
@@ -0,0 +1,122 @@
...
+            current_timeout: max_timeout,
+            max_timeout,
+            min_timeout,
+        }
+    }
+
+    /// Returns whether at least `min_timeout` time elapsed since last tick.
+    /// If `min_timeout` elapsed since last tick, returns true and records a tick.
+    /// If not, returns false and calls to `wait` will return when `min_timeout`
+    /// elapses until the next tick.
```
kostekIV, line 32: annoying nit: 
```suggestion
    /// Returns whether at least `min_timeout` time elapsed since the last tick.
    /// If `min_timeout` elapsed since the last tick, returns true and records a tick.
    /// If not, returns false and calls to `wait` will return when `min_timeout`
    /// elapses until the next tick.
```

finality-aleph/src/sync/ticker.rs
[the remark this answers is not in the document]
maciejnems: okaaay, I will change it

kostekIV (approved):

PR #834, opened 2022-12-28
A0-1613 Improving `Connection`s

krzysztofziobro:
# Description

Changing interfaces of `Connection` etc. : 
- Introduce `AsConnection` trait for casting to `Connection` and `AsSigned` for casting to `SignedConnection`
- Introduce `ConnectionApi` trait to enable using added methods not only for `Connection`, but also for `SignedConnection` and `RootConnection` + similar `SignedConnectionApi`
- Construct `Connection` from `&str`
- Hide fields

## Type of change

- Refactor

# Checklist:
- [x] I have bumped aleph-client version


obrok:
Why the custom `AsConnection` instead of `From<_>`?

krzysztofziobro:
> Why the custom `AsConnection` instead of `From<_>`?

Didn't want to wrap the `Client` + I think it's more transparent

aleph-client/src/connections.rs
```diff
@@ -14,19 +14,62 @@ use subxt::{
...
+            (0, Err(e)) => panic!("{:?}", e),
+            _ => {
+                sleep(Duration::from_secs(RETRY_WAIT_SECS));
+                retries -= 1;
+            }
+        }
+    }
+}
+
+pub trait AsConnection: Sync {
```
kostekIV, line 40: you could also add AsSignedConnection and AsRootConnection similar to this

kostekIV (commented):
Great work, couple comments

aleph-client/src/connections.rs
```diff
@@ -14,19 +14,62 @@ use subxt::{
...
+    >(
+        &self,
+        addrs: &StaticStorageAddress<T, Yes, Defaultable, Iterable>,
+        at: Option<BlockHash>,
+    ) -> Option<T::Target>;
+
+    async fn rpc_call<R: Decode>(&self, func_name: String, params: RpcParams) -> anyhow::Result<R>;
 }
 
 pub struct SignedConnection {
```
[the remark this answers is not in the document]
krzysztofziobro: I'm afraid that our `KeyPair` is not clonable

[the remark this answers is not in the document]
kostekIV: Yep, thats why we need to implement clone not just derive it. In e2e tests there are some examples of how to manually 'clone' keypair

krzysztofziobro:
I'm doing a pretty substantial changes right now, so you might want to wait a bit with reviewing

aleph-client/src/connections.rs
```diff
@@ -58,29 +109,60 @@ impl SudoCall for RootConnection {
...
+    }
+}
+
+impl AsSigned for SignedConnection {
+    fn as_signed(&self) -> &SignedConnection {
+        self
+    }
+}
+
+impl AsSigned for &SignedConnection {
```
fixxxedpoint, line 145: Is this necessary? Same for other impls.

aleph-client/src/connections.rs
```diff
@@ -12,21 +12,72 @@ use subxt::{
...
 }
 
+#[derive(Clone)]
 pub struct RootConnection {
-    pub connection: Connection,
-    pub root: KeyPair,
+    connection: SignedConnection,
+}
+
+pub trait AsConnection: Sync {
```
fixxxedpoint, line 34: I think this trait can be just internal to this crate? `Connection` except new, has no public API. You will need to remove it from other traits as well. Actually what you need here is something like `AsSubxtClient`?

aleph-client/src/pallets/author.rs
```diff
@@ -1,17 +1,16 @@
...
+use crate::{aleph_runtime::SessionKeys, ConnectionApi};
 
 #[async_trait::async_trait]
 pub trait AuthorRpc {
     async fn author_rotate_keys(&self) -> anyhow::Result<SessionKeys>;
 }
 
 #[async_trait::async_trait]
-impl AuthorRpc for Connection {
+impl<C: ConnectionApi> AuthorRpc for C {
```
fixxxedpoint, line 11: `AsConnection` should be enough here?

fixxxedpoint (changes requested):
Nicely done! Left few comments.

aleph-client/src/connections.rs
```diff
@@ -12,21 +12,72 @@ use subxt::{
...
+    fn as_signed(&self) -> &SignedConnection;
+}
+
+#[async_trait::async_trait]
+pub trait ConnectionApi: AsConnection {
+    async fn get_storage_entry<T: DecodeWithMetadata + Sync, Defaultable: Sync, Iterable: Sync>(
+        &self,
+        addrs: &StaticStorageAddress<T, Yes, Defaultable, Iterable>,
+        at: Option<BlockHash>,
+    ) -> T::Target;
```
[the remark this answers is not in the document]
kostekIV: this one is used for entries that we think are present (or have default values) and is left for user to decide whether they want this or a variant which returns an Option

aleph-client/src/connections.rs
```diff
@@ -174,7 +305,13 @@ impl RootConnection {
     ) -> anyhow::Result<Self> {
         let root_address = api::storage().sudo().key();
 
-        let root = match connection.client.storage().fetch(&root_address, None).await {
+        let root = match connection
+            .as_connection()
```
kostekIV, line 309: not needed

aleph-client/src/pallets/treasury.rs
```diff
@@ -91,25 +91,24 @@ impl TreasuryUserApi for SignedConnection {
 
 #[async_trait::async_trait]
 impl TreasurySudoApi for RootConnection {
-    async fn approve(&self, proposal_id: u32, status: TxStatus) -> anyhow::Result<BlockHash> {
+    async fn sudo_approve(&self, proposal_id: u32, status: TxStatus) -> anyhow::Result<BlockHash> {
```
kostekIV, line 94: its kinda redundant to call it sudo as its already in the name of the trait imho

kostekIV (approved):

aleph-client/src/pallets/treasury.rs
krzysztofziobro: right, but then `approve` becomes ambiguous if we import both traits

kostekIV: You can still qualify method in such case

krzysztofziobro: I thought it will be more readable this way, but I can change it

aleph-client/src/connections.rs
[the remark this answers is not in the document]
krzysztofziobro: Then we would need to specify `Sync` in every trait implementation for `ConnectionApi` - alternatively we can make `ConnectionApi` `Sync`

krzysztofziobro: + I'm not sure why we should hide potentially useful trait

aleph-client/src/pallets/author.rs
krzysztofziobro: Right, just wanted to be consistent

krzysztofziobro: Can change it if you think it's important

fixxxedpoint (approved):
Great work!

PR #835, opened 2022-12-29
A0-1396 Poseidon hash function

fbielejec:
# Description

An instance of a (parameterized) Poseidon hash function to be used with the relations crate. 

## Type of change

# Checklist:

- I have added tests


relations/Cargo.toml
```diff
@@ -22,22 +22,24 @@ ark-serialize = { version = "^0.3.0", default-features = false }
...
-    "ark-ff/std",
-    "ark-gm17/std",
-    "ark-groth16/std",
-    "ark-marlin/std",
-    "ark-poly/std",
-    "ark-poly-commit/std",
-    "ark-r1cs-std/std",
-    "ark-relations/std",
-    "ark-serialize/std",
-    "blake2/std",
```
pmikolajczyk41, line 42: do you have tab set to '  ' in toml formatting? this is kinda different to the whole codebase :P

poseidon/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+type CircuitField = ark_bls12_381::Fr;
+type FpVar = ark_r1cs_std::fields::fp::FpVar<CircuitField>;
+
+pub mod hash {
+    use ark_bls12_381::Fr;
+    use poseidon_permutation::Instance;
+
+    use crate::parameters::RATE_1_PARAMETERS;
+    /// hashes one field value, outputs a fixed length field value
+    pub fn one_to_one_hash(domain_separator: &Fr, value: Fr) -> Fr {
```
pmikolajczyk41, line 12: what is `domain_separator`?

pmikolajczyk41 (approved):
lgtm, but maybe before following up with 2to1, I suggest creating SC with this preimage, just to measure time needed for this computation

poseidon/src/parameters.rs
```diff
@@ -0,0 +1,36 @@
...
+
+// Poseidon parameters generated for the Fr (Fp256) finite field
+pub mod fr_parameters {
+    include!(concat!(env!("OUT_DIR"), "/parameters.rs"));
+}
+
+// Parameters for the 1:1 hash instance of Poseidon
+pub static RATE_1_PARAMETERS: Lazy<PoseidonParameters<Fr>> = Lazy::new(fr_parameters::rate_1);
+
+// taken from Penumbra (https://github.com/penumbra-zone/poseidon377/blob/a2d8c7a3288e2e877ac88a4d8fd3cc4ff2b52c04/poseidon377/src/r1cs.rs#L12)
```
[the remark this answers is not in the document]
fbielejec: - yes they do - we are generating parameters on  BLS-12-381
- ark and penumbra paramgen are Apache-2.0 compat, I'd say we are good?

relations/Cargo.toml
fbielejec: 2 spaces for tab. Fixed it

poseidon/src/lib.rs
fbielejec: It's arbitrary and used for encoding use cases (so nothing security critical). Basically because you can have several different "poseidons" (variable length, fixed length, various hashing rates etc) you will use this to encode them. 

There is a section explaining it [here](https://spec.filecoin.io/#section-algorithms.crypto.poseidon.domain-separation) and in the original Poseidon paper.

poseidon/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+edition = "2021"
+license = "Apache 2.0"
+authors = ["Cardinal Cryptography"]
+description = "An instantiation of the Poseidon SNARK-friendly hash function."
+
+[dependencies]
+ark-bls12-381 = { version = "^0.3.0" }
+ark-ff = "0.3"
+ark-r1cs-std = {version = "0.3" }
+ark-relations = { version = "0.3" }
```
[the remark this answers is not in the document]
fbielejec: done

poseidon/src/lib.rs
```diff
@@ -0,0 +1,43 @@
+mod parameters;
+
```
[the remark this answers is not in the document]
fbielejec: good point, I made it wasm-y now

relations/Cargo.toml
fbielejec: I hate tabs btw. wasteful

pmikolajczyk41:
regarding SC costs, you might find https://github.com/Cardinal-Cryptography/aleph-node/pull/852 helpful

poseidon/src/lib.rs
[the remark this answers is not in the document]
pmikolajczyk41: why `poseidon = { path = "../poseidon" }` is without `default-features = false`?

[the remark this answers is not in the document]
fbielejec: parameter generation was std - initially I assumed we would link to the generated rust file with parameters statically, but actually there is no need, and they can be generated in a non-std environment as well. Changed it now.

poseidon/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+    use crate::parameters::RATE_1_PARAMETERS;
+    /// hashes one field value, outputs a fixed length field value
+    pub fn one_to_one_hash(domain_separator: &Fr, value: Fr) -> Fr {
+        let parameters = RATE_1_PARAMETERS.clone();
+        let mut state = Instance::new(&parameters);
+        state.n_to_1_fixed_hash(vec![*domain_separator, value])
+    }
+}
+
+pub mod r1cs {
```
mike1729, line 21: Is this name standard for this level of abstraction? I'd go into either `circuit` or `constrained` directions but don't know standard nomenclature here yet.

poseidon/src/parameters.rs
```diff
@@ -0,0 +1,37 @@
+use ark_bls12_381::Fr;
+use ark_sponge::poseidon::PoseidonParameters as ArkSpongePoseidonParameters;
+use once_cell::sync::Lazy;
+use poseidon_paramgen::{Alpha, PoseidonParameters};
+
+// Poseidon parameters generated for the Fr (Fp256) finite field
+pub mod fr_parameters {
+    use ark_std::vec;
+    include!(concat!(env!("OUT_DIR"), "/parameters.rs"));
```
mike1729, line 9: This looks suspicious. Maybe add some dir `parameters` or `resources` and assume parameters are there.

poseidon/src/parameters.rs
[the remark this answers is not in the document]
mike1729: We're good and we don't even need this comment to be compliant.

poseidon/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+type FpVar = ark_r1cs_std::fields::fp::FpVar<CircuitField>;
+
+pub mod hash {
+    use ark_bls12_381::Fr;
+    use ark_std::vec;
+    use poseidon_permutation::Instance;
+
+    use crate::parameters::RATE_1_PARAMETERS;
+    /// hashes one field value, outputs a fixed length field value
+    pub fn one_to_one_hash(domain_separator: &Fr, value: Fr) -> Fr {
```
mike1729, line 14: I believe the `domain_separator` is needed mostly on the theoretic side, so I'd just fixed one and remove the parameter here and in the module below.

relations/src/preimage.rs
```diff
@@ -0,0 +1,141 @@
...
+    use ark_relations::r1cs::{ConstraintSynthesizer, ConstraintSystem};
+    use poseidon::hash;
+
+    use super::{PreimageRelation, DOMAIN_SEPARATOR};
+    use crate::{CircuitField, GetPublicInput};
+
+    #[test]
+    fn preimage_constraints_correctness() {
+        let preimage = CircuitField::from(17u64);
+        let preimage_hash = hash::one_to_one_hash(&DOMAIN_SEPARATOR, preimage);
```
mike1729, line 108: For a mathematician, it seems enormously disturbing to not name this var image :P

relations/src/preimage.rs
```diff
@@ -0,0 +1,141 @@
...
+
+impl<S: WithPublicInput> GetPublicInput<CircuitField> for PreimageRelation<S> {
+    fn public_input(&self) -> Vec<CircuitField> {
+        vec![self
+            .hash
+            .expect("Circuit should have public input assigned")]
+    }
+}
+
+#[cfg(test)]
```
mike1729, line 94: I'd add two more tests for negative cases to make sure that `is_satisfied` and `verify` are not implemented as `{ true }`

mike1729 (changes requested):
A couple comments but nothing big. Great work to make it work!

poseidon/build.rs
```diff
@@ -0,0 +1,42 @@
...
+use ark_std::vec;
+use poseidon_paramgen::poseidon_build;
+
+fn main() {
+    let security_level = match env::var("SECURITY_LEVEL") {
+        Ok(level) => match level.as_str() {
+            "80" => 80,
+            "128" => 128,
+            "256" => 256,
+            _ => panic!("Unsupported security level"),
```
deuszx, line 19:
```suggestion
            _ => panic!("Unsupported security level. Supported levels: 80, 128, 256"),
```

poseidon/src/parameters.rs
fbielejec: Here's the file generated by build.rs in my env, using just the defaults:
2023-01-11-1673447495_screenshot_1920x1080

relations/src/preimage.rs
fbielejec: iguess

poseidon/src/lib.rs
fbielejec: went with circuit, indeed perhaps we will go away from expressing them in r1cs

poseidon/src/lib.rs
fbielejec: done

relations/src/preimage.rs
fbielejec: if it did, that would be very unfortunate :)

I created a test where constraints are unsatisfied - imo there is no need to check whether prover works in that case, since they all check the circuit even before creating the proof (see [here](https://github.com/arkworks-rs/groth16/blob/master/src/prover.rs#L253))

poseidon/build.rs
```diff
@@ -0,0 +1,32 @@
...
+    // t = arity + 1, so t=2 is a 1:1 hash, t=3 is a 2:1 hash etc
+    // see https://spec.filecoin.io/#section-algorithms.crypto.poseidon.filecoins-poseidon-instances for similar specification used by Filecoin
+    let t_values = vec![2];
+
+    // Fr => Fp256
+    let parameters =
+        poseidon_build::compile::<Fr>(security_level, t_values, FrParameters::MODULUS, true);
+
+    let output_directory =
+        PathBuf::from(env::var("OUT_DIR").expect("OUT_DIR environmental variable must be set"))
```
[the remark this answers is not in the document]
fbielejec: > I know that we won't be running this build frequently but I'd turned `OUT_DIR` into a cli arg as well as `security_level`.

AFAIK this build will run every time a project that uses the dependency is build, so actually very frequently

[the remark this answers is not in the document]
mike1729: For given hyperparameters the output is deterministic, right? We could check in build if we have generated parameters for given inputs. Recalculating sth that is deterministic seems unnecessary, however I don't have a strong opinion here, it's still early in development we can start with either approach. 
 
Rather sidenote: from the docs you shared:

> --out-dir directory
    Copy final artifacts to this directory.

    This option is unstable and available only on the [nightly channel](https://doc.rust-lang.org/book/appendix-07-nightly-rust.html) and requires the -Z unstable-options flag to enable. See https://github.com/rust-lang/cargo/issues/6790 for more information.

poseidon/src/parameters.rs
mike1729: Let's keep the discussion in one thread, resolving this one.

poseidon/build.rs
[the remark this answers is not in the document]
fbielejec: I have not measured it but it's not long, as in a completely subjective fashion it does not make the compilation "feel long"). I assumed that what matters is the runtime - hence the parameters are instantiated lazily, once at runtime. 

I guess they will stabilize the feature soon - and the node is on nightly atm anyway :wink:

mike1729 (approved):

PR #836, opened 2022-12-29
A0-1795: add task queue

maciejnems:
# Description

Adding task queue that will be used in sync protocol

## Type of change

- New feature (not yet used)

# Checklist:

- I have added tests


finality-aleph/src/sync/task_queue.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// Awaits for the first and most overdue task and returns it. Returns `None` if there are no tasks.
+    pub async fn pop(&mut self) -> Option<T> {
+        let scheduled_task = self.queue.peek_mut()?;
+
+        sleep(
+            scheduled_task
+                .scheduled_time
+                .saturating_duration_since(Instant::now()),
+        )
+        .await;
```
timorl, line 75: It might be worth it to avoid the sleep if the duration is zero, I'm not sure the implementation necessarily handles this efficiently. Although if it does it's probably better to leave as-is... <_<"

timorl (approved):
A couple point of confusion from me, but that might be more my fault.

finality-aleph/src/sync/task_queue.rs
```diff
@@ -0,0 +1,103 @@
...
+impl<T: Eq> PartialOrd for ScheduledTask<T> {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl<T: Eq> Ord for ScheduledTask<T> {
+    /// Compare tasks so that earlier times come first in a max-heap.
+    fn cmp(&self, other: &Self) -> Ordering {
+        other.scheduled_time.cmp(&self.scheduled_time)
```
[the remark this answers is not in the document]
maciejnems: I was also curious about that, but it seems, this BinaryHeap can store duplicates of equal objects.

finality-aleph/src/sync/task_queue.rs
```diff
@@ -0,0 +1,103 @@
...
+        };
+        self.queue.push(ScheduledTask {
+            task,
+            scheduled_time,
+        });
+    }
+
+    /// Awaits for the first and most overdue task and returns it. Returns `None` if there are no tasks.
+    pub async fn pop(&mut self) -> Option<T> {
+        let scheduled_task = self.queue.peek_mut()?;
```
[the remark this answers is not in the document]
maciejnems: So that we can later use pop on it. We can also just use `peek` here and `queue.pop` at the end. Which one do you prefer?

deuszx:
Have you considered changing the order and `sleep` before pushing onto the queue? You wouldn't need to compare things and simple `VecDeq` should suffice since items should be inserted into the queue in the order from earliest to latest (to pop).

maciejnems:
> 



> Have you considered changing the order and `sleep` before pushing onto the queue? You wouldn't need to compare things (inside the binary tree) and simple `VecDeq` should suffice since items should be inserted into the queue in the order from earliest to latest (to pop).
> 
> Although, that would require making the producer `async` which isn't what you would want probably.

I don't think making push async is a good idea. It would make usage of this struct a lot harder. Also in this solution it is not possible to add something that should be popped before top of queue.

finality-aleph/src/sync/task_queue.rs
```diff
@@ -0,0 +1,108 @@
...
+    collections::BinaryHeap,
+    fmt::{Debug, Formatter},
+};
+
+use log::warn;
+use tokio::time::{sleep, Duration, Instant};
+
+use crate::sync::LOG_TARGET;
+
+#[derive(Clone, Eq, PartialEq)]
```
fixxxedpoint, line 12: Does it require `PartialEq...` only because of that `derive` clause. I don't see any reason for `T` to implement those traits. Removing it would allow to drop those constraints on `TaskQueue` as well.

finality-aleph/src/sync/task_queue.rs
```diff
@@ -0,0 +1,108 @@
...
+
+impl<T: Eq> Ord for ScheduledTask<T> {
+    /// Compare tasks so that earlier times come first in a max-heap.
+    fn cmp(&self, other: &Self) -> Ordering {
+        other.scheduled_time.cmp(&self.scheduled_time)
+    }
+}
+
+#[derive(Clone, Default)]
+pub struct TaskQueue<T: Eq + PartialEq> {
```
fixxxedpoint, line 32: Code itself doesn't require `T` to be anyhow related with some abstract `task`. Perhaps add some requirement for it or name it like `ScheduledQueue`. Or, perhaps that `ScheduledTask` should be implementing that `waiting` logic directly and then that whole queue is unecessary and can be just switched with any ordered queue. Consider also that `pop` can actually `block` whole queue until head task is ready.

fixxxedpoint (changes requested):

finality-aleph/src/sync/task_queue.rs
maciejnems: `ScheduledTask` needs to implement `Ord` and because of that it also needs to implement all `PartialOrd`, `Eq` and `PartialEq`. I guess we could manually implement `PartialEq` and `Eq` only for instant. Right now `Eq` together with `Ord` is not consistent.

finality-aleph/src/sync/task_queue.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    use super::TaskQueue;
+
+    #[tokio::test]
+    async fn test_scheduling() {
+        let mut q = TaskQueue::new();
+        q.schedule_in(2, Duration::from_millis(50));
+        q.schedule_in(1, Duration::from_millis(20));
+
+        assert!(timeout(Duration::from_millis(5), q.pop()).await.is_err());
```
fixxxedpoint, line 97: I would avoid such checks. It might fail just based on tokio's internal scheduling/timing. Similarly, adding `q.schedule_in(1, Duration::from_millis(20));` doesn't guarantee it will be queued before `2`. If possible, using `Instant` instead of `Duration` would solve second issue.

finality-aleph/src/sync/task_queue.rs
```diff
@@ -0,0 +1,108 @@
...
+        assert_eq!(
+            timeout(Duration::from_millis(20), q.pop()).await,
+            Ok(Some(1))
+        );
+        assert!(timeout(Duration::from_millis(10), q.pop()).await.is_err());
+        assert_eq!(
+            timeout(Duration::from_millis(20), q.pop()).await,
```
fixxxedpoint, line 104: Theoretically, only `~35 millis` might have passed since you added `2`. Just use `50` here.

fixxxedpoint (approved):
left few comments

finality-aleph/src/sync/task_queue.rs
```diff
@@ -0,0 +1,108 @@
...
+    /// Creates an empty queue.
+    pub fn new() -> Self {
+        Self {
+            queue: BinaryHeap::new(),
+        }
+    }
+
+    /// Schedules `task` for after `delay`.
+    pub fn schedule_in(&mut self, task: T, delay: Duration) {
+        let scheduled_time = match Instant::now().checked_add(delay) {
```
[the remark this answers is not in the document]
maciejnems: We wanted to schedule in the future. Either way Instant needs to be calculated, here or when scheduling. The case when instant is not possible should not happen, so if we drop this task it does not matter. This usage is easier.

finality-aleph/src/sync/task_queue.rs
maciejnems: Unless we have a mock for time I also won't like those tests. That said I think windows I left are enough so that this test is not flaky. Changing it to `Instant` just for the sake of tests does not sound good to me.

finality-aleph/src/sync/task_queue.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+impl<T> Ord for ScheduledTask<T> {
+    /// Compare tasks so that earlier times come first in a max-heap.
+    fn cmp(&self, other: &Self) -> Ordering {
+        other.scheduled_time.cmp(&self.scheduled_time)
+    }
+}
+
+#[derive(Clone, Default)]
```
deuszx, line 39: I wouldn't derive `Clone` for it as it can mask uncorrect usage - I'm pretty sure we don't want to clone the queue? We do need to make it safe to use across different threads though.

PR #837, opened 2022-12-29
A0-1770: Add basic interfaces used and exposed by sync

timorl:
# Description

The basic API for the sync. Only implemented for some trivial substrate types, but I checked and it is all doable.

## Type of change

# Checklist:


finality-aleph/src/sync/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    /// The identifier of this block.
+    fn id(&self) -> Self::Identifier;
+
+    /// The identifier of this block's parent.
+    fn parent_id(&self) -> Option<Self::Identifier>;
+}
+
+/// The justification of a block, including a header.
+pub trait Justification: Clone {
```
kostekIV, line 32: suggestion: Add method returning raw_justification here and in Verifier just pass Justification instead of header + raw_justification

finality-aleph/src/sync/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+    fn verify(&self, header: J::Header, raw_justification: RJ) -> Result<J, Self::Error>;
+}
+
+/// A facility for finalizing blocks using justifications.
+pub trait Finalizer<J: Justification> {
+    type Error: Display;
+
+    /// Finalize a block using this justification. Since the justification contains the header, we
+    /// don't need to additionally specify the block.
+    fn finalize(&self, justification: J) -> Result<(), Self::Error>;
```
kostekIV, line 54: suggestion, use anyhow + optionally thiserror

finality-aleph/src/sync/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+
+/// A stream of notifications about the chain state in the database changing.
+#[async_trait::async_trait]
+pub trait ChainStateNotifier<BI: BlockIdentifier> {
+    /// Returns a chain state notification when it is available.
+    async fn next(&self) -> ChainStateNotification<BI>;
+}
+
+/// The state of a block in the database.
+pub enum BlockState<J: Justification> {
```
kostekIV, line 73: suggestion: Add variant `IsOnFork(J::Header)` (or some better name)

kostekIV (approved):
lgtm

finality-aleph/src/sync/mod.rs
timorl: What would this mean?

finality-aleph/src/sync/mod.rs
timorl: Hm, my idea was to have this work as a constructor, but your comment made me realize that this is incoherent with sending justifications with headers over the network...

Changed it to maybe be better now – `Justification` has an associated type for an unverified version, can be changed into that and verifier should be the only way of performing the change the other way. Amusingly this should give us some type safety even if we set `Unverified = Self` in the implementation, so that's nice.

finality-aleph/src/sync/substrate.rs
```diff
@@ -0,0 +1,46 @@
...
+        H: Hasher,
+    {
+        self.hash.hash(state);
+        self.number.hash(state);
+    }
+}
+
+impl<H: SubstrateHeader> BlockIdentifier for BlockId<H> {
+    fn number(&self) -> u32 {
+        self.number.unique_saturated_into()
```
maciejnems, line 25: If you specified `H` to `SubstrateHeader<Number = u32>` you would not have to use saturated into here. Also there exists `BlockNumber` in aleph-primitives, but is inconsistent with `BlockNumber` in bin (both are set separately to u32), so one of them could be used here instead of `u32`. If you decide to do that, it can be also used in BlockIdentifier trait.

finality-aleph/src/sync/mod.rs
timorl: I would rather not add these, if we want them either we should get an ADR or have people who care more add them personally. :P

maciejnems (approved):

finality-aleph/src/sync/substrate.rs
timorl: Oh, nice, fixed the first part, although the definitions became a bit noisy. I used the `BlockNumber` from `primitives`, but I haven't fixed the inconsistency...

finality-aleph/src/sync/mod.rs
kostekIV: xD

finality-aleph/src/sync/mod.rs
kostekIV: Block is from fork that is not the best/finalized chain

kostekIV: but totally optional

maciejnems: I guess it could only apply to blocks that are not in finalized chain, as the ones not from best chain could still be in the best chain in the future.

This might be useful to know that a specific block will never be finalized. But is there a case where it could be useful to us?

timorl: We will be internally figuring out which blocks are hopeless forks anyway, and we might want to have our own best-block logic. In general we want to treat the DB more like a bag of blocks and what it thinks about their structure beyond "there is a finalized blockchain" is mostly irrelevant.

PR #838, opened 2022-12-29
A0-1610: Restore contract event support

obrok:
# Description

Restores contract event subscriptions that were removed during the subxt migration.

## Type of change

- New feature (non-breaking change which adds functionality [although it was previously broken by a breaking change])

# Checklist:

* [x] I have added tests
* [x] I have made corresponding changes to the existing documentation
* [x] I have bumped aleph-client version if relevant


aleph-client/src/contract/event.rs
```diff
@@ -0,0 +1,130 @@
...
+
+use crate::{contract::ContractInstance, AccountId, Connection};
+
+/// Represents a single event emitted by a contract.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct ContractEvent {
+    /// The address of the contract that emitted the event.
+    pub contract: AccountId,
+    /// The name of the event.
+    pub ident: Option<String>,
```
fbielejec, line 59: can we use full names for fields? Like _identifier_ or _event_name_

fbielejec (approved):
looks good, nice work

aleph-client/src/contract/event.rs
obrok: Maybe just `name` if you don't like `ident`?

aleph-client/src/contract/event.rs
```diff
@@ -0,0 +1,130 @@
...
+
+use std::collections::HashMap;
+
+use anyhow::{bail, Result};
+use contract_transcode::Value;
+use futures::{channel::mpsc::UnboundedSender, StreamExt};
+
+use crate::{contract::ContractInstance, AccountId, Connection};
+
+/// Represents a single event emitted by a contract.
```
deuszx, line 53:
```suggestion
/// Represents a single event emitted by the contract.
```

aleph-client/src/contract/event.rs
```diff
@@ -0,0 +1,130 @@
...
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct ContractEvent {
+    /// The address of the contract that emitted the event.
+    pub contract: AccountId,
+    /// The name of the event.
+    pub ident: Option<String>,
+    /// Data contained in the event.
```
deuszx, line 60: Maybe just "event data"?

aleph-client/src/contract/event.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    while let Some(block) = block_subscription.next().await {
+        if sender.is_closed() {
+            break;
+        }
+
+        let block = block?;
+
+        for event in block.events().await?.iter() {
+            let event = event?;
```
deuszx, line 85: Isn't this going to exit from the `listen_contract_events` (similarly the `event.as_event` below)? If so then we should add that to the comment.

aleph-client/src/contract/event.rs
```diff
@@ -0,0 +1,130 @@
...
+                }
+            }
+        }
+    }
+
+    Ok(())
+}
+
+/// The contract transcoder assumes there is an extra byte (that it discards) indicating the size of the data. However,
+/// data arriving through the subscription as used in this file don't have this extra byte. This function adds it.
```
deuszx, line 109:
```suggestion
/// data arriving through the subscription as used in this file doesn't have this extra byte. This function adds it.
```

aleph-client/src/contract/event.rs
```diff
@@ -0,0 +1,130 @@
...
+fn zero_prefixed(data: &[u8]) -> Vec<u8> {
+    let mut result = vec![0];
+    result.extend_from_slice(data);
+    result
+}
+
+fn build_event(address: AccountId, event_data: Result<Value>) -> Result<ContractEvent> {
```
deuszx, line 116: I would unwrap the `event_data` out of `Result` before calling this function. `build_event` shouldn't have to clean up the mess that someone else made.

e2e-tests/src/test/adder.rs
```diff
@@ -13,14 +18,31 @@ pub async fn adder() -> Result<()> {
...
+        <Result<(), anyhow::Error>>::Ok(())
+    };
+    let join = tokio::spawn(listen());
 
     let increment = 10;
     let before = contract.get(&conn).await?;
+
     contract.add(&account.sign(&conn), increment).await?;
+
+    let event = rx.next().await.context("No event received")??;
```
deuszx, line 41: Will `next()` resolve immediately? i.e. it will either return or not, without waiting. So we're assuming here that by the time we get to this line, the `contract.add(...)` transaction has already been processed by the chain?

e2e-tests/src/test/adder.rs
```diff
@@ -13,14 +18,31 @@ pub async fn adder() -> Result<()> {
...
         &config.test_case_params.adder_metadata,
-    )?;
+    )?);
+
+    let listen_conn = conn.clone();
+    let listen_contract = contract.clone();
+    let (tx, mut rx) = unbounded();
+    let listen = || async move {
+        listen_contract_events(&listen_conn, &[listen_contract.as_ref().into()], tx).await?;
+        <Result<(), anyhow::Error>>::Ok(())
```
deuszx, line 32: Do we need it? `listen_contract_events` already returns `Result<(), anyhow::Error>` so we should be OK if we drop the `;`

deuszx (commented):
I don't see anything inherently wrong with this but I left a couple questions/suggestions.

aleph-client/src/contract/event.rs
obrok: Disagree

aleph-client/src/contract/event.rs
obrok: I think both are correct, data can be plural or singular. I guess `don't` is more formal than `doesn't`.

aleph-client/src/contract/event.rs
obrok: 1. I don't think the cleanliness of the interface matters, because it's a private function.
2. It would add extra mangling at the call site. Cannot just move the `?` to the call site, because that's not what's supposed to happen - we're supposed to send a `Result`, so it would have to be some `if`, making the call site more complicated.

So overall, I think this is better.

aleph-client/src/contract/event.rs
deuszx: Right, didn't know that :)

e2e-tests/src/test/adder.rs
obrok: I'm not sure what the question is... `next` returns a future, it will not resolve immediately. Even if technically the event already occurred, there is no guarantee on how long it's going to take to fetch that event into the client. There is indeed a way to see events "immediately" after a contract call - a call result can return all the events in the block that the call ended up in. However, I intentionally didn't go for that method, because I think it's less flexible - it only allows you to observe events caused by your own calls. If anything, I'd say that can be a secondary feature, but this listening is something that must be there.

e2e-tests/src/test/adder.rs
obrok: An `|| async move` has problems with its own identity and so the compiler won't be able to deduce the type in that case. See https://rust-lang.github.io/async-book/07_workarounds/02_err_in_async_blocks.html

aleph-client/src/contract/event.rs
deuszx: Just b/c it's a private function doesn't mean it shouldn't have a clean interface. I understand your second point though -- still, I would do that on the callsite 
```rust
sender.unbounded_send(event.map(|event_data| build_event(address, event_data)))
```
doesn't seem very complicated.

aleph-client/src/contract/event.rs
obrok: If you're suggesting that a field named `data` in a struct named `Event` be commented `event data`, then I strongly disagree. This does not add any information and doesn't even put it in a more "human readable" form. I realize that the comment as-is doesn't add any information either, but at least it's a nominal sentence.

e2e-tests/src/test/adder.rs
deuszx: Bluh, I didn't know that. thanks.

e2e-tests/src/test/adder.rs
deuszx: > There is indeed a way to see events "immediately" after a contract call - a call result can return all the events in the block that the call ended up in. However, I intentionally didn't go for that method, because I think it's less flexible - it only allows you to _observe events caused by **your own calls**_.

I think that's exactly what we're interested in here, aren't we? Otherwise we might get here an event that was created by someone else's call that happened right after we sent the txn but wasn't result of that txn.

aleph-client/src/contract/event.rs
obrok: But this is a doc for the _data structure_. The structure _could_ be used without the `listen` function. The structure represents _some_ event by _some_ contract.

deuszx: Well, it's _the_ contract that lives under `contract` address.

Anyway, this discussion is already taking far longer than it should have so I'm resolving my comment.

e2e-tests/src/test/adder.rs
obrok: > I think that's exactly what we're interested in here, aren't we?

Depends on what you mean by "interested". If by "interested" you mean for the purposes of these tests, then yes. But that's not the (whole) point of the exercise. I would say the more important goal is to provide tooling for people to interact with the contracts, and subscribing to events without doing calls is the more flexible approach for that.

aleph-client/src/contract/event.rs
deuszx: IMO the version with `event.map` is more readable b/c it clearly "throws away" the error (signaling it's not of an interest to us in any of the contexts). So, I guess - 👎 .

e2e-tests/src/test/adder.rs
deuszx: OK, I'm convinced. Thanks.

PR #840, opened 2022-12-30
WIP: A0-1820: Add justifications and validator implementations

timorl:
# Description

Implements the validator and justifications. Makes the code a bit of a mess, mostly because some of this stuff was already there and still needs to be used externally while we still have the legacy justification sync.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/sync/substrate/justification.rs
+++ b/finality-aleph/src/sync/substrate/justification.rs
@@ -1,0 +1,171 @@
+use std::fmt::{Display, Error as FmtError, Formatter};
+
+use aleph_primitives::{BlockNumber, SessionAuthorityData};
+use codec::Encode;
+use log::warn;
+use sp_runtime::{
+    traits::{Block, Header as SubstrateHeader},
+    RuntimeAppPublic,
+};
+
+use crate::{
+    crypto::AuthorityVerifier,
+    justification::{AlephJustification, Verifier as LegacyVerifier},
+    session::session_id_from_num,
+    session_map::ReadOnlySessionMap,
+    sync::{Justification as JustificationT, Verifier},
+    AuthorityId, SessionPeriod,
+};
+
+/// A justification, including the related header.
+#[derive(Clone)]
+pub struct Justification<H: SubstrateHeader<Number = BlockNumber>> {
+    header: H,
+    raw_justification: AlephJustification,
+}
+
+impl<H: SubstrateHeader<Number = BlockNumber>> JustificationT for Justification<H> {
+    type Header = H;
+    type Unverified = Self;
+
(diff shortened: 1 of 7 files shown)


finality-aleph/src/sync/substrate/justification.rs
```diff
@@ -0,0 +1,166 @@
...
+            .get(session_id)
+            .map(|authority_data| authority_data.into())
+            .ok_or(VerificationError::UnknownSession)
+    }
+}
+
+impl<H: SubstrateHeader<Number = BlockNumber>> Verifier<Justification<H>> for FullVerifier {
+    type Error = VerificationError;
+
+    fn verify(&self, justification: Justification<H>) -> Result<Justification<H>, Self::Error> {
```
maciejnems, line 160: nit: this verify seems to be pretty heavy. On each call it locks on SessionMap, clones authorities and creates a new verifier. Still I'm not sure if there is a better way, to create a full verifier.

I guess we would have to somehow cache already existing verifiers and remove them from cache once they are not needed, which is already done in SessionMap :// maybe @DamianStraszak has an idea for that.

maciejnems (approved):
From what I understand outside of session map, the only write lock we have is for subscribing for authority list, so there is no possibility of new deadlocks to occur after changing it to sync code.

finality-aleph/src/sync/substrate/justification.rs
timorl: I agree this is annoying... I'll look into how hard it would be to rewrite ROSM some more to avoid this.

finality-aleph/src/sync/substrate/justification.rs
```diff
@@ -0,0 +1,166 @@
...
+pub struct SessionVerifier {
+    authority_verifier: AuthorityVerifier,
+    emergency_signer: Option<AuthorityId>,
+}
+
+impl From<SessionAuthorityData> for SessionVerifier {
+    fn from(authority_data: SessionAuthorityData) -> Self {
+        SessionVerifier {
+            authority_verifier: AuthorityVerifier::new(authority_data.authorities().to_vec()),
+            emergency_signer: authority_data.emergency_finalizer().clone(),
```
maciejnems, line 50: Also this clone should probably not happen, as we move `SessionAuthorityData`. That said I'm not sure how to get authorities and signer out of `SessionAuthorityData` ://

finality-aleph/src/sync/substrate/justification.rs
timorl: Ok, it seems I'll rewrite much more here after all, I'll open a separate PR for some smaller part of this and then write one for the verifier (although it will take longer). This one will likely get closed.

PR #842, opened 2023-01-03
A0-1619 New GH check for runtime metadata

Marcin-Radecki:
# Description

This PR introduces new check which runs `subxt codegen` in a container, just in case someone forgot to commit changes to that file.

## Type of change

# Checklist:


aleph-client/rustfmt.toml
```diff
@@ -0,0 +1,7 @@
+edition = "2021"
+use_field_init_shorthand = true
+reorder_modules = true
+
+imports_granularity = "Crate"
+group_imports = "StdExternalCrate"
+reorder_imports = true
```
deuszx, line 7: Does running it change anything? If so, can we add it and run on the whole project in a separate PR?

Marcin-Radecki: No, this is a bit of a quirk I did here. That is the same rustfmt toml as this one https://github.com/Cardinal-Cryptography/aleph-node/blob/main/rustfmt.toml , I've copied it since I wanted to build context to be in `aleph-client` directory. In that case docker prevents using `COPY` command from the parent directory, when the original `rustmft.toml` file originates. A symbolic link does not work either, so it must be a hard copy. I know it's a duplication.

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -182,6 +182,25 @@ jobs:
...
+      - name: Run one node in the background
+        uses: ./.github/actions/run-e2e-test
+        with:
+          node-count: 1
+          min-validator-count: 1
+
+      - name: check if runtime metadata matches
+        run: |
+          cd aleph-client/
+          docker build --tag subxt:latest -f docker/Dockerfile .
```
deuszx, line 202: We should use more descriptive tag -- this one makes it sound like the docker image _is_ subxt. Maybe `subxt-client-integration`?

deuszx: Instead of building it every time we should use pre-built container.

Marcin-Radecki: Both done

krzysztofziobro:
If I understand correctly, the last version of `aleph_zero.rs` is held by the image on ecr, but I do not see anything that would push it  automatically on commit to main. Are we leaving implementing such a mechanism for a separate PR?

deuszx:
> If I understand correctly, the last version of `aleph_zero.rs` is held by the image on ecr, but I do not see anything that would push it automatically on commit to main. Are we leaving implementing such a mechanism for a separate PR?

We will be failing PRs on merge to `main` if this returns `aleph_client.rs` different from the one in the branch.

aleph-client/src/aleph_zero.rs
```diff
@@ -19533,9 +19532,9 @@ pub mod api {
         let runtime_metadata_hash = client.metadata().metadata_hash(&PALLETS);
         if runtime_metadata_hash
             != [
-                155u8, 236u8, 207u8, 138u8, 149u8, 114u8, 96u8, 168u8, 143u8, 247u8, 148u8, 106u8,
-                148u8, 203u8, 48u8, 129u8, 14u8, 48u8, 155u8, 234u8, 78u8, 212u8, 73u8, 65u8,
-                212u8, 201u8, 174u8, 194u8, 72u8, 70u8, 240u8, 233u8,
+                10u8, 121u8, 157u8, 11u8, 147u8, 107u8, 235u8, 73u8, 90u8, 254u8, 82u8, 183u8,
+                112u8, 64u8, 213u8, 99u8, 23u8, 17u8, 10u8, 91u8, 124u8, 231u8, 209u8, 172u8, 59u8,
+                160u8, 15u8, 142u8, 149u8, 200u8, 95u8, 164u8,
```
deuszx, line 19537: Why is this PR passing CI if it's changing the `aleph_zero.rs`?

aleph-client/docker/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN rustup component add rustfmt --toolchain nightly
+
+COPY docker/docker_entrypoint.sh /subxt/docker_entrypoint.sh
+COPY src/aleph_zero.rs /subxt/aleph_zero_current.rs
+COPY rustfmt.toml /subxt/rustfmt.toml
+
+RUN chmod +x /subxt/docker_entrypoint.sh
+RUN rustc --version
+
+ENTRYPOINT ["./docker_entrypoint.sh"]
```
deuszx, line 15: I don't think the "main" Dockerfile in the `aleph-client` repository should be this one. Why not put it under `.github/scripts` directory?

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -182,6 +182,25 @@ jobs:
           if-no-files-found: error
           retention-days: 7
 
+  run-aleph-client-subxt-codegen-check:
```
deuszx, line 185: When is it going to run? On nightly or every PR/commit?

krzysztofziobro:
> > If I understand correctly, the last version of `aleph_zero.rs` is held by the image on ecr, but I do not see anything that would push it automatically on commit to main. Are we leaving implementing such a mechanism for a separate PR?
> 
> We will be failing PRs on merge to `main` if this returns `aleph_client.rs` different from the one in the branch.

Yes, but we need to update image on ecr after runtime changes on main, or we will keep failing even if `aleph_zero.rs` on branch is up-to-date with `aleph_zero.rs` on main, right?

deuszx:
You're onto something @krzysztofziobro . I'm wondering now whether we need docker here at all. Why not:
1. Run 1-node network
2. run `subxt codegen`
3. `git status --porcelain` (or sth different)

We should achieve what we're after - i.e. fail the PR if we didn't update `aleph_zero.rs`

aleph-client/src/aleph_zero.rs
Marcin-Radecki: It turned out sb forgot already to update `aleph_zero.rs` on `main`, hence I needed to fix it there.

aleph-client/docker/Dockerfile
Marcin-Radecki: I'll rename it to `Dockerfile.subxt-integration`

.github/workflows/e2e-tests-main-devnet.yml
Marcin-Radecki: Every push to PR, and every merge to `main`
Latter is not necessarily required, our current pipelines works like this. Ideally we should separate build from e2e testing phase, meaning that
* every commit to PR should built test binary only and runs e2e tests (now it build both test and production binary)
* every merge PR to main should just build production binary and not run e2e tests

Marcin-Radecki:
> You're onto something @krzysztofziobro . I'm wondering now whether we need docker here at all. Why not:
> 
>     1. Run 1-node network
> 
>     2. run `subxt codegen`
> 
>     3. `git status --porcelain` (or sth different)
> 
> 
> We should achieve what we're after - i.e. fail the PR if we didn't update `aleph_zero.rs`

Yes it does not make sense what I did there. I'll fix that we always compare to `aleph_zero.rs` from `main` , and not bake it in the dockerfile

Marcin-Radecki:
> > You're onto something @krzysztofziobro . I'm wondering now whether we need docker here at all. Why not:
> > ```
> > 1. Run 1-node network
> > 
> > 2. run `subxt codegen`
> > 
> > 3. `git status --porcelain` (or sth different)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We should achieve what we're after - i.e. fail the PR if we didn't update `aleph_zero.rs`
> 
> Yes it does not make sense what I did there. I'll fix that we always compare to `aleph_zero.rs` from `main` , and not bake it in the dockerfile

Actually it made sense till we changed approach to use pre-built container. Anyway, I'll try to improve situation here

Marcin-Radecki:
I've did rework of the Dockerfile, see the Readme.md added for more details

.github/workflows/e2e-tests-main-devnet.yml
deuszx: This is a pretty have action, it involves:
* building the node binary (since it requires `run-e2e-test`)
* running a network
* pulling a docker image and running it

so it should not be run on every push to a PR - ideally it would be run only once before merging.

Marcin-Radecki: This is not the scope of this PR I think.

deuszx: Can we move it to run on nightly then?

Marcin-Radecki: It's not heavy action, it just reuses current pipelines. The check itself runs 4 minutes and it runs parallel to e2e tests that runs 15 minutes, so it does not extend overall pipeline time. 

When e2e are moved to nightly, we can move this check as welll.

deuszx (approved):

PR #844, opened 2023-01-04
A0-1823: add substrate specific chain status

maciejnems:
# Description
Add substrate specific chain status implementation

## Type of change

- New feature (not yet used)


timorl (approved):

finality-aleph/src/sync/substrate/chain_status.rs
```diff
@@ -0,0 +1,147 @@
...
+                    "no justification for finalized block with hash {:?}",
+                    hash
+                )
+            }
+        }
+    }
+}
+
+/// Substrate implementation of ChainStatus trait
+struct SubstrateChainStatus<B, BE>
```
pmikolajczyk41, line 46: why not `pub` or `pub(super)`?

finality-aleph/src/sync/substrate/chain_status.rs
```diff
@@ -0,0 +1,147 @@
...
+where
+    BE: Backend<B>,
+    B: BlockT,
+    B::Header: SubstrateHeader<Number = BlockNumber>,
+{
+    fn header(&self, hash: B::Hash) -> Option<B::Header> {
+        let id = SubstrateBlockId::<B>::Hash(hash);
+        self.client
+            .header(id)
+            .expect("substrate client must respond")
```
pmikolajczyk41, line 66: idk what is the policy, but maybe it would be better to be defensive here and convert `Err` to `None`? or extend `Error` enum and return `Result` here

finality-aleph/src/sync/substrate/chain_status.rs
```diff
@@ -0,0 +1,147 @@
...
+    MissingHash(B::Hash),
+    MissingJustification(B::Hash),
+}
+
+impl<B: BlockT> Display for Error<B> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), FmtError> {
+        use Error::*;
+        match self {
+            MissingHash(hash) => {
+                write!(f, "no block for hash {:?}", hash)
```
pmikolajczyk41, line 32: as far as i can see, this variant is returned only when there exists a block with this hash, but it is not available yet in our backend - maybe a subtle rephrasing will be helpful

pmikolajczyk41 (approved):
cool, very clean and understandable even for people outside context :)

I've left a few minor comments - please resolve at your will

finality-aleph/src/sync/substrate/chain_status.rs
maciejnems: Yes, you are right. How about this message "Inconsistent state, no block for existing hash {:?}"? I could make something similar for justification.

finality-aleph/src/sync/substrate/chain_status.rs
maciejnems: Yep, we thought about it. The thing is we already use client in many places and use `expect` on the result. To be fair, if client is not working something really bad is going on an we probably can't fix it.

That said, from what I see, substrate does some error handling in this case, so I can add those here.

finality-aleph/src/sync/substrate/chain_status.rs
maciejnems: Should be pub, just forgot about it, as it is not used anywhere right now.

finality-aleph/src/sync/substrate/chain_status.rs
```diff
@@ -0,0 +1,147 @@
...
+            MissingJustification(hash) => {
+                write!(
+                    f,
+                    "no justification for finalized block with hash {:?}",
+                    hash
+                )
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
maciejnems: I guess we are still waiting on some decision to bring it to aleph node. I guess we don't have it in dependencies. We've been doing error handling this way for some time, so I would just stick to it for now.

finality-aleph/src/sync/substrate/chain_status.rs
pmikolajczyk41: something like `data availability problem: no block for hash {hash:?}`? should be a bit more precise

PR #845, opened 2023-01-05
A0-1611: Full multisig support in `aleph-client`

pmikolajczyk41:
This PR complements basic pallet API. Moreover, it introduces a new, contextual API, that is useful in places, where the whole aggregation is being done in one place (in the code).

**Review**: There are some (reverted) changes in commit history to `cliain` that allow to test the new API with just `cargo run multisig`.

### DIFF
--- a/aleph-client/src/pallets/multisig.rs
+++ b/aleph-client/src/pallets/multisig.rs
@@ -1,17 +1,32 @@
 use primitives::{Balance, BlockNumber};
 
 use crate::{
-    api, api::runtime_types, sp_weights::weight_v2::Weight, AccountId, BlockHash,
-    SignedConnectionApi, TxStatus,
+    aleph_runtime::RuntimeCall, api, api::runtime_types, sp_weights::weight_v2::Weight, AccountId,
+    BlockHash, SignedConnectionApi, TxStatus,
 };
 
 pub type CallHash = [u8; 32];
-pub type Call = Vec<u8>;
-pub type Timepoint = api::runtime_types::pallet_multisig::Timepoint<BlockNumber>;
+pub type Call = RuntimeCall;
+pub type Timepoint = runtime_types::pallet_multisig::Timepoint<BlockNumber>;
 pub type Multisig = runtime_types::pallet_multisig::Multisig<BlockNumber, Balance, AccountId>;
 
 #[async_trait::async_trait]
 pub trait MultisigUserApi {
+    async fn as_multi_threshold_1(
+        &self,
+        other_signatories: Vec<AccountId>,
+        call: Call,
+        status: TxStatus,
+    ) -> anyhow::Result<BlockHash>;
+    async fn as_multi(
+        &self,
+        threshold: u16,
+        other_signatories: Vec<AccountId>,
+        timepoint: Option<Timepoint>,
+        max_weight: Weight,
+        call: Call,
+        status: TxStatus,
+    ) -> anyhow::Result<BlockHash>;
     async fn approve_as_multi(
         &self,
         threshold: u16,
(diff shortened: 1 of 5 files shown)


aleph-client/src/pallets/multisig.rs
```diff
@@ -71,3 +129,388 @@ impl<S: SignedConnectionApi> MultisigUserApi for S {
...
+    /// Create new party from `signatories` and `threshold`.
+    ///
+    /// `signatories` can contain duplicates and doesn't have to be sorted. However, there must be
+    /// at least 2 unique accounts. There is also a virtual upper bound - `MaxSignatories` constant.
+    /// It isn't checked here (since it requires client), however, using too big party will fail
+    /// when performing any chain interaction.
+    ///
+    /// `threshold` must be between 2 and number of unique accounts in `signatories`. For threshold
+    /// 1, use special method `MultisigUserApi::as_multi_threshold_1`.
+    pub fn new(signatories: &[AccountId], threshold: MultisigThreshold) -> AnyResult<Self> {
```
kostekIV, line 150: in every other places `anyhow::Result` is used. Use it for consistency for now, we can add task to refactor it.

aleph-client/src/pallets/multisig.rs
```diff
@@ -71,3 +129,388 @@ impl<S: SignedConnectionApi> MultisigUserApi for S {
...
+    call_hash: CallHash,
+
+    /// The set of accounts, that already approved the call (via this context object), including the
+    /// author.
+    ///
+    /// `approvers.len() < party.threshold` always holds.
+    approvers: HashSet<AccountId>,
+}
+
+impl Context {
```
kostekIV, line 213: no new method?

aleph-client/src/pallets/multisig.rs
```diff
@@ -71,3 +129,388 @@ impl<S: SignedConnectionApi> MultisigUserApi for S {
...
+    call.using_encoded(blake2_256)
+}
+
+/// Ensure that the signer of `conn` is present in `party.signatories`. If so, return all other
+/// signatories.
+fn ensure_signer_in_party<S: SignedConnectionApi>(
+    conn: &S,
+    party: &MultisigParty,
+) -> anyhow::Result<Vec<AccountId>> {
+    let signer_account = account_from_keypair(conn.signer().signer());
```
kostekIV, line 508: isnt there something like conn.signer().account_id()

aleph-client/src/pallets/multisig.rs
```diff
@@ -1,20 +1,45 @@
...
-    SignedConnectionApi, TxStatus,
+    account_from_keypair, aleph_runtime::RuntimeCall, api, api::runtime_types,
+    sp_weights::weight_v2::Weight, AccountId, BlockHash, ConnectionApi, SignedConnectionApi,
+    TxStatus,
 };
 
 pub type CallHash = [u8; 32];
-pub type Call = Vec<u8>;
-pub type Timepoint = api::runtime_types::pallet_multisig::Timepoint<BlockNumber>;
+pub type Call = RuntimeCall;
```
Marcin-Radecki, line 16: those type aliases should be as well commented I think, otherwise there will be warns

aleph-client/src/pallets/multisig.rs
```diff
@@ -71,3 +129,388 @@ impl<S: SignedConnectionApi> MultisigUserApi for S {
...
+        self.max_weight = max_weight;
+    }
+
+    /// Set `call` only if `self.call_hash` is matching.
+    fn set_call(&mut self, call: &Call) -> anyhow::Result<()> {
+        ensure!(
+            self.call_hash == compute_call_hash(call),
+            "Call doesn't match to the registered hash"
+        );
+        self.call = Some(call.clone());
```
kostekIV, line 228: nit: dont borrow call, let the user decide whether the call should be cloned or can be moved

aleph-client/src/pallets/multisig.rs
```diff
@@ -71,3 +129,388 @@ impl<S: SignedConnectionApi> MultisigUserApi for S {
...
+        Ok(())
+    }
+
+    /// Register another approval. Consume `self` if the threshold has been met and `call` is
+    /// already known.
+    fn add_approval(mut self, approver: AccountId) -> Option<Self> {
```
kostekIV, line 234: there is no way to determine who approved, who was author etc after this consumes self which might be useful information loss.
suggestion: `fn add_approval(...) -> Either<Self, ConsumedContext>` where `ConsumedContext` only allows to read fields or smt like this

Marcin-Radecki (approved):
Great and quick job!

aleph-client/src/pallets/multisig.rs
pmikolajczyk41: sorry, I must have used it in one place by mistake and then copied it few times; changed to `anyhow::Result`

aleph-client/src/pallets/multisig.rs
pmikolajczyk41: I added a mod-private constructor, so that it can be created only by initiating signature aggregation

aleph-client/src/pallets/multisig.rs
pmikolajczyk41: perfect! I haven't noticed it before

aleph-client/src/pallets/multisig.rs
pmikolajczyk41: done

aleph-client/src/pallets/multisig.rs
pmikolajczyk41: done

kostekIV (approved):

PR #850, opened 2023-01-09
A0-1540: Verification error

pmikolajczyk41:
This PR introduces error enum for the verification process in pallet `snarcos`. It covers possible errors from `verify(..)` methods in `Groth16`, `GM17` and `Marlin` implementations from `arkworks`, i.e. 
 - `SynthesisError` (for the former two) - here we have only two variants: `MalformedVerifyingKey` and `UnexpectedError`. The first one is just extraction of its counterpart in `SynthesisError`. The second one covers all other variants - they are dedicated for constraint synthesis or proof generation phases.
 - `ark_marlin::Error<ark_poly_commit::Error>` (for the latter one) - here we unpack the `ark_marlin::Error` enum to either `AHPError` or `PolynomialCommitmentError`. In case `IndexTooLarge` happened, we fall back to `UnexpectedError` (this should never happen during verification process)

All the new variants are flat, i.e. they do not contain inner errors. This is because arkworks' errors don't implement basic traits and thus cannot be used in the runtime.

### DIFF
--- a/pallets/snarcos/src/systems.rs
+++ b/pallets/snarcos/src/systems.rs
@@ -1,11 +1,63 @@
 use ark_poly::polynomial::univariate::DensePolynomial;
 use ark_poly_commit::marlin_pc::MarlinKZG10;
+use ark_relations::r1cs::SynthesisError;
 use ark_serialize::CanonicalDeserialize;
 use ark_snark::SNARK;
 use ark_std::rand::{prelude::StdRng, SeedableRng};
 use blake2::Blake2s;
 use codec::{Decode, Encode};
+use frame_support::log::{error, info};
 use scale_info::TypeInfo;
+
+/// Possible errors from the verification process.
+#[derive(Copy, Clone, Eq, PartialEq, Debug, Decode, Encode, TypeInfo)]
+pub enum VerificationError {
+    /// The verifying key was malformed.
+    ///
+    /// May occur only for some non-universal system.
+    MalformedVerifyingKey,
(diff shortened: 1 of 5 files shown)


pallets/snarcos/src/systems.rs
```diff
@@ -1,12 +1,67 @@
...
+            }
+        }
+    }
+}
+
+impl From<ark_marlin::Error<ark_poly_commit::Error>> for VerificationError {
+    fn from(err: ark_marlin::Error<ark_poly_commit::Error>) -> Self {
+        match err {
+            ark_marlin::Error::AHPError(err) => {
+                info!("Encountered AHP error: {err:?}");
```
maciejzelaszczyk, line 50: Is there a reason why this is logged as `info` rather than an error? Same for the next case.

maciejzelaszczyk (approved):
This is very clean indeed! Nifty use of `Into` for error propagation.

pallets/snarcos/src/systems.rs
pmikolajczyk41: yes - in that case, we will still push the 'error status' back to the caller, which can be then handled. so in general, the logs shouldn't indicate that there's a program-level error

we are using info channel just to not miss the inner message (`AHPError`) - we cannot pass it to the caller due to unsatisfied traits, so it is the last place when the user can see it

maciejzelaszczyk: OK, gotcha

PR #851, opened 2023-01-09
Use parent hash as randomness in Marlin verification

pmikolajczyk41:
Universal verification systems use randomness for the verification process. We drop the constant seed in favor of parent hash.

### DIFF
--- a/pallets/snarcos/src/lib.rs
+++ b/pallets/snarcos/src/lib.rs
@@ -303,8 +303,14 @@ pub mod pallet {
                     )
                 })?;
 
-            let valid_proof = S::verify(&verification_key, &public_input, &proof)
-                .map_err(|err| (Error::<T>::VerificationFailed(err), None))?;
+            let parent_hash = <frame_system::Pallet<T>>::parent_hash();
+            let valid_proof = S::verify(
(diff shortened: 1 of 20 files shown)


DamianStraszak (approved):
Looks good, but I have no idea what is randomness is for really :)

pallets/snarcos/src/lib.rs
```diff
@@ -303,8 +303,14 @@ pub mod pallet {
...
                 })?;
 
-            let valid_proof = S::verify(&verification_key, &public_input, &proof)
-                .map_err(|err| (Error::<T>::VerificationFailed(err), None))?;
+            let parent_hash = <frame_system::Pallet<T>>::parent_hash();
+            let valid_proof = S::verify(
+                &verification_key,
+                &public_input,
+                &proof,
+                parent_hash.as_ref(),
```
deuszx, line 311: It's not _really_ random though as it can be influenced to some extent. Probably _good enough_ though.

pmikolajczyk41: usually proof generation lasts longer than block-time, so yeah, good enough :P

PR #852, opened 2023-01-09
A0-1766: Sketch out the pricing script

obrok:
# Description

I went with this weird approach of calling `cargo contract` from a python script, because:

1. The `py-substrate-client` does not seem to currently work with ink4 contracts on newest aleph-node, I suspect it's because of the new level of wrapping in `Result`, but didn't want to devote too much time to investigating that.
2. I thought (and @h4nsu was ecstatic to confirm) that a python script would be preferable to a bash script in terms of maintenance and helper libraries available.

An unfortunate downside is that the script is dependent on the actual directories with the contracts to be called, because that's how `cargo contract call` works. In particular, this means that we currently don't have PSP22, etc. contracts ready to use with this script, because the `benjamin` contracts are still on ink3. My plan:

1. Merge this to `main`
2. Merge `main` to `benjamin`
3. Migrate (at least) `wazero` and (one of the) PSP22 tokens to ink4
4. Extend this script to include pricing for operations on those

Example output:

| Operation                                      | Fee                        |
|------------------------------------------------|----------------------------|
| Instantiate contract with single storage value | 0.001929 AZERO ($0.001708) |
| Call contract with single storage update       | 0.001325 AZERO ($0.001173) |

## Type of change

# Checklist:

- [x] I have created new documentation


mike1729 (approved):
Great work, turned out to not be that hard as I initially fought. The q is now how to make it more general so that it's automatic or semi automatic, i.e. a user provides names of methods she wants to measure. Such UX should be good enough.

scripts/pricing/run.py
```diff
@@ -0,0 +1,96 @@
...
+
+def call(directory, contract, message, *args):
+    args = [x for a in args for x in ['--args', a]]
+    res = subprocess.run(['cargo', 'contract', 'call', '--contract', contract,
+                          '--message', message] + args + COMMON_ARGS, cwd=directory, check=True, stdout=PIPE)
+    return json.loads(res.stdout.decode('utf-8'))
+
+
+def event_field(event, field):
+    for f in [f for f in event['fields'] if f['name'] == field]:
```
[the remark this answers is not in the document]
obrok: Python noob here, sorry, picking idioms randomly.

obrok:
@deuszx 
> If you're calling into cargo contract anyway. why not write the whole thing in Rust and use any of the below:

I think this is much easier to work with. There's a bunch of overhead in rust related to types, compilation, etc., which I don't think gives us anything for a script like this. Note that this script constitutes about 50% of my lifetime exposure to python, and I still feel more comfortable writing it in python than rust :D

deuszx:
I find it educational and, in the long term, more practical to "eat your own dog food". I understand that using Python may be simpler short-term but having to use our own tools will lead to polishing them and making everyone's life better.

obrok:
> I find it educational and, in the long term, more practical to "eat your own dog food". I understand that using Python may be simpler short-term but having to use our own tools will lead to polishing them and making everyone's life better.

I asked, people seemed open to the idea of doing it in python, that's what I did, because it seemed best. You're welcome to create an ADR banning all non-rust languages :)

More broadly, I think we should provide non-rust tooling for our chain. In fact, I think js/python/ruby/etc. tooling would be of greater value for external integrators than aleph-client as these langs are more popular (although maybe not in the crypto space so YMMV).

Thirdly, remember that the second alternative I was considering was a bash script. If I submitted a PR with a bash script that does this, you wouldn't bat an eyelash, so just think of this as a glorified bash script.

deuszx:
Resolved off github -- my comments above were simply opinions and weren't meant to block the merge of this PR.

h4nsu (approved):
Some minor suggestions, feel free to ignore&resolve

scripts/pricing/run.py
```diff
@@ -0,0 +1,98 @@
...
+
+def find_fee(events, by_whom):
+    fee_event = next(filter(lambda e: e['name'] == 'TransactionFeePaid' and account_id(
+        event_field(e, 'who')) == by_whom, events), None)
+    return uint(event_field(fee_event, 'actual_fee'))
+
+
+with urllib.request.urlopen('https://api.coingecko.com/api/v3/simple/price?ids=aleph-zero&vs_currencies=usd') as response:
+    data = json.load(response)
+    aleph_usd = data['aleph-zero']['usd']
```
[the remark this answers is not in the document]
obrok: Just thought it was a nice sanity check - when discussing the results with @pmikolajczyk41 we (or at least I) were a bit confused what a result like 14274920402 means - is it a lot or a little?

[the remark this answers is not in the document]
h4nsu: 14274920402 means not much, but 0.014 AZERO is good enough for me.
What I don't like about this conversion is that it makes the results of the script very volatile - in practice every run could return a different value in USD, and results from a week ago are not valid today. But I'm probably overthinking  ;)

scripts/pricing/run.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def format_fee(fee):
+    return "%f AZERO ($%f)" % (fee / AZERO, fee / AZERO * aleph_usd)
+
+
+deploy_result = deploy(args.adder_dir)
+
+adder_address = deploy_result['contract']
```
deuszx, line 87: This approach will return invalid address if there are contracts instantiated in the constructor https://github.com/paritytech/cargo-contract/issues/777

obrok: Well, given it's a bug in cargo contract, I say ignore until they fix it, then update cargo contract, no?

deuszx: Or you can inspect the `.events[]` field in the JSON and use the last element where `.pallet == "Contracts"` and `.name = "Instantiated"`.

obrok: I'll keep that in mind if I run into this problem.

PR #860, opened 2023-01-13
`aleph-client`: Get tx events

pmikolajczyk41:
# Description

This PR introduces two things:
 - now all methods like `send_tx` return `TxInfo`, which currently is a simple struct containing `TxHash` and `BlockHash`
 - there's a new method `get_tx_events` that returns `ExtrinsicEvents` struct

There's also an example of usage in e2e test (fee).

Supporting contract events (i.e. merging new utility with decoding) will be done in another, non-breaking 

Reasoning for `ExtrinsicEvents`:
 - it is a `subxt`'s struct that already exposes handy API for filtering/extracting events; so it is ready to use so we don't have to provide our own, similar implementations
 - `ExtrinsicEvents` has `pub(crate)` constructor, which is used in two places in `subxt`: `TxInBlock::fetch_events` and `Extrinsic::events`. The former one is kinda out of external usage, so we have to create `Block`, `BlockBody` and `Extrinsic` intermediate objects. Anyway, works quite well and is still fairly concise

## Type of change

# Checklist:

### DIFF
--- a/aleph-client/src/connections.rs
+++ b/aleph-client/src/connections.rs
@@ -109,7 +109,7 @@ pub trait ConnectionApi: Sync {
 }
 
 /// Data regarding submitted transaction.
-#[derive(Copy, Clone, Eq, PartialEq, Debug, Default, Deserialize, Serialize)]
+#[derive(Copy, Clone, Eq, PartialEq, Debug, Deserialize, Serialize)]
 pub struct TxInfo {
     pub block_hash: BlockHash,
     pub tx_hash: TxHash,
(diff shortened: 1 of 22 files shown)


Marcin-Radecki (approved):

kostekIV (approved):

aleph-client/src/connections.rs
```diff
@@ -286,15 +305,25 @@ impl<S: AsSigned + Sync> SignedConnectionApi for S {
...
-            TxStatus::Submitted => return Ok(BlockHash::from_low_u64_be(0)),
+        let info: TxInfo = match status {
+            TxStatus::InBlock => progress
+                .wait_for_in_block()
+                .await?
+                .wait_for_success()
+                .await?
+                .into(),
+            TxStatus::Finalized => progress.wait_for_finalized_success().await?.into(),
+            // In case of Submitted block hash do not mean anything
```
kostekIV, line 316:
```suggestion
            // In case of Submitted block hash does not mean anything
```

PR #858, opened 2023-01-12
A0-1852 remove send-runtime-hook and fork-off from Deploy to Devnet w…

Marcin-Radecki:
…orkflow

# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

## Type of change

Please delete options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

- I have added tests
- I have made neccessary updates to the Infrastructure
- I have made corresponding changes to the existing documentation
- I have created new documentation
- I have bumped `spec_version` and `transaction_version`
- I have bumped aleph-client version if relevant

PR #862, opened 2023-01-13
Fix invalid variable used for sha_short

mikogs:
Fix variable used for SHA_SHORT. The whole step was copied over from a different workflow and the `steps.get_branch.outputs.sha_short` was just empty.


Kosiare (commented):
I couldn't comment those lines that haven't changed in this PR therefore I have to write it manually:

1. Line `43-44`:
Let's use `${{ env.AWS_REGION }}` here as well.

2. Line `49`
Why the tag of `TESTNET_IMAGE` is set to `${{ steps.vars.outputs.branch }}`? Shouldn't be the tag/release? 
Workflow is triggered on pre-release, therefore `echo "##[set-output name=branch;]$(echo ${GITHUB_REF##*/})"` will equal pre-release name or am I wrong?

Marcin-Radecki (approved):
One optional contribution:
* Remove `set-output` deprecated command as e.g. in https://github.com/Cardinal-Cryptography/aleph-node/pull/858/files

CezaryKierzyk (approved):

PR #863, opened 2023-01-13
Poseidon benchmarks

fbielejec:
# Description

Benchmarks for generating satisfyability proofs for a simple circuit using the Poseidon hash function
Quick run-of-the mill (busy laptop running intel i7):

```
preimage circuit size: 214
time:   [166.46 ms 167.74 ms 169.21 ms]
Found 13 outliers among 100 measurements (13.00%)
  6 (6.00%) high mild
  7 (7.00%) high severe
```

## Type of change

# Checklist:

- I have added tests


relations/src/preimage/relation.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+impl<S: WithPublicInput> GetPublicInput<CircuitField> for PreimageRelation<S> {
+    fn public_input(&self) -> Vec<CircuitField> {
+        vec![self
+            .hash
+            .expect("Circuit should have public input assigned")]
+    }
+}
```
deuszx, line 100: Can you rewrite it and use the macro instead?

pmikolajczyk41: that's my boy <3

relations/src/preimage/relation.rs
```diff
@@ -0,0 +1,100 @@
+use ark_ff::BigInteger256;
+// This relation showcases how to use Poseidon in r1cs circuits
```
pmikolajczyk41, line 2: nit: let's move it either to very top of this module or to `mod.rs`

relations/src/preimage/tests.rs
```diff
@@ -0,0 +1,99 @@
+use ark_bls12_381::Bls12_381;
+use ark_crypto_primitives::SNARK;
+use ark_ec::bls12::Bls12;
+#[cfg(test)]
+use ark_ff::BigInteger256;
+use ark_ff::Fp256;
+use ark_groth16::{Groth16, Proof, VerifyingKey};
+#[cfg(test)]
```
pmikolajczyk41, line 8: this and line 4. are not needed, are they?

relations/src/preimage/tests.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    (vk, input, proof)
+}
+
+#[test]
+pub fn frontend_to_backend_conversion() {
+    let frontend_preimage = 7u64;
+    let backend_preimage: CircuitField = CircuitField::from(frontend_preimage);
+
+    println!("preimage input in the field {:?}", backend_preimage);
```
pmikolajczyk41, line 83: maybe let's remove those `println`s

relations/src/preimage/tests.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    let expected_backend_hash: CircuitField = hash::one_to_one_hash([backend_preimage]);
+
+    println!("hash value in the field {:?}", expected_backend_hash);
+
+    let bint = BigInteger256::new([
```
pmikolajczyk41, line 89: I suppose that these values were generated by `hash::one_to_one_hash` and therefore I doubt whether this test actually shows anything...

relations/src/preimage/tests.rs
```diff
@@ -0,0 +1,99 @@
...
+    let cs = ConstraintSystem::new_ref();
+    circuit.generate_constraints(cs.clone()).unwrap();
+
+    let is_satisfied = cs.is_satisfied().unwrap();
+
+    assert!(!is_satisfied);
+}
+
+#[test]
+pub fn preimage_proving_and_verifying() {
```
pmikolajczyk41, line 51: doesn't have to be `pub`

I know that this is a low nit, but currently this module exports other functions and it is a bit confusing which things are actually used outside

relations/src/preimage/tests.rs
pmikolajczyk41: okay, I see the reason (`preimage_proving`); I would suggest moving it to `mod.rs` so that's exposed to both `benches` and `tests`; then you can put the whole `tests` module under `cfg(test)`

pmikolajczyk41 (approved):
Generally, looks good, only some minor comments left - resolve/ignore at your will

relations/src/preimage/relation.rs
fbielejec: done

relations/src/preimage/tests.rs
fbielejec: it's a safety net, it will catch a regression if we start playing with the hash function, then you can think whether it is legit that the output changed (e.g. maybe the ff was changed)

relations/src/preimage/tests.rs
fbielejec: ok

relations/src/preimage/tests.rs
fbielejec: done

relations/src/preimage/relation.rs
fbielejec: not sure I can - macro sems to work only for inline modules, and the module here is in it's own file

pmikolajczyk41: you can use an artificial module - it will be inlined anyway

deuszx: We could inline if it's needed. I think we should write all our relations using macro now, unless there's a good reason to not do it. We're committing to a single representation, API, struct names etc. It will make things less confusing and if we improve the macro's outputs - everything will be improved automatically.

fbielejec: i don't think tooling should manhandle anyone to write or arrange code in any particular manner, because there might be outside factors that influence that decision. Like here is a good example - modules separate code for tests, benchmarks and a relation.

pmikolajczyk41: sure, but you don't have to sacrifice anything here - you can just use the macro similarly to other relations

fbielejec: iguess

relations/src/preimage/relation.rs
```diff
@@ -0,0 +1,61 @@
...
+    #[relation_object_definition]
+    struct PreimageRelation {
+        /// private witness
+        #[private_input]
+        pub preimage: CircuitField,
+        /// public input
+        #[public_input(
+            frontend_type = "FrontendHash",
+            parse_with = "convert_hash"
+            // serialize_with = "flatten_sequence"
```
deuszx, line 26: Leftover.

relations/src/preimage/relation.rs
```diff
@@ -0,0 +1,61 @@
...
+    //     _phantom: PhantomData<S>,
+    // }
+
+    // impl<S: WithPublicInput> GetPublicInput<CircuitField> for PreimageRelation<S> {
+    //     fn public_input(&self) -> Vec<CircuitField> {
+    //         vec![self
+    //             .hash
+    //             .expect("Circuit should have public input assigned")]
+    //     }
+    // }
```
deuszx, line 60: Leftovers.

relations/src/preimage/relation.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// Preimage relation : H(preimage)=hash
+    /// where:
+    /// - hash : public input
+    /// - preimage : private witness
+    #[relation_object_definition]
+    struct PreimageRelation {
+        /// private witness
+        #[private_input]
+        pub preimage: CircuitField,
+        /// public input
```
deuszx, line 22: We don't need these comments anymore.

deuszx (approved):

PR #864, opened 2023-01-13
A0-1820: implement substrate specific verifier for sync protocol

maciejnems:
# Description

This Pr introduces caching mechanism for `SessionVerifier`'s. It allows for verifying justifications from all sessions without cloning authority set from `SessionMap` on each verify call. This approach was designed so that there is no two level caching mechanism. `VerifierCache` has a separate cache from `SessionMap`. It is simpler than `SessionMap` and updates cache on each get call, not in separate thread.

## Type of change

- New feature (not yet used)
- Refactor (`JustificationVerifier` renamed to `SessionVerifier` and moved to sync folder)

# Checklist:

- I have added tests

This work was based on https://github.com/Cardinal-Cryptography/aleph-node/pull/840


finality-aleph/src/sync/substrate/verification/cache.rs
```diff
@@ -0,0 +1,340 @@
...
+                f,
+                "justification from too old session {:?}. Should be at least {:?}",
+                session, lower_bound
+            ),
+            SessionInFuture(session, upper_bound) => write!(
+                f,
+                "justification from session {:?} without known authorities. Should be at most {:?}",
+                session, upper_bound
+            ),
+            UnknownSession => write!(f, "justification from unknown session"),
```
kostekIV, line 37:
```suggestion
            UnknownSession => write!(f, "justification from an unknown session"),
```

finality-aleph/src/sync/substrate/verification/cache.rs
```diff
@@ -0,0 +1,340 @@
...
+where
+    AP: AuthorityProvider<BlockNumber>,
+    FS: FinalizationStatus,
+{
+    sessions: HashMap<SessionId, SessionVerifier>,
+    session_period: SessionPeriod,
+    finalization_status: FS,
+    authority_provider: AP,
+    cache_size: SessionId,
+    /// Lowest curretnly available session.
```
kostekIV, line 56: nit: documenting private fields feels weird :P

finality-aleph/src/sync/substrate/verification/cache.rs
```diff
@@ -0,0 +1,340 @@
...
+            ));
+        }
+
+        if !self.sessions.contains_key(&session_id) {
+            self.download(session_id);
+        }
+
+        self.sessions
+            .get(&session_id)
+            .ok_or(CacheError::UnknownSession)
```
kostekIV, line 142: this error is strange. The error here really means that we don't know the authorities for this session which is known as we provide the block number to the function itself.

finality-aleph/src/sync/substrate/verification/cache.rs
```diff
@@ -0,0 +1,340 @@
...
+pub struct VerifierCache<AP, FS>
+where
+    AP: AuthorityProvider<BlockNumber>,
+    FS: FinalizationStatus,
+{
+    sessions: HashMap<SessionId, SessionVerifier>,
+    session_period: SessionPeriod,
+    finalization_status: FS,
+    authority_provider: AP,
+    cache_size: SessionId,
```
kostekIV, line 55: teribble nit: this should be just usize or (`type SessionCount = usize`). We use this to count sessions kept in the cache

finality-aleph/src/sync/substrate/verification/cache.rs
kostekIV, line 56:
```suggestion
    /// Lowest currently available session.
```

finality-aleph/src/sync/substrate/verification/cache.rs
```diff
@@ -0,0 +1,340 @@
...
+}
+
+impl<AP, FS> VerifierCache<AP, FS>
+where
+    AP: AuthorityProvider<BlockNumber>,
+    FS: FinalizationStatus,
+{
+    /// Download authorities for session and store in cache. It needs to be first session,
+    /// or first block from previous session needs to be finalized.
+    /// Otherwise nothing is downloaded.
```
kostekIV, line 89: sorry :( 
```suggestion
    /// Download authorities for the session and store them in the cache. It needs to be the first session,
    /// or the first block from the previous session needs to be finalized.
    /// Otherwise nothing is downloaded.
```

finality-aleph/src/sync/substrate/verification/cache.rs
```diff
@@ -0,0 +1,340 @@
...
+                self.authority_provider.next_authority_data(prev_first)
+            }
+        };
+
+        if let Some(session) = maybe_session {
+            self.sessions.insert(session_id, session.into());
+        };
+    }
+
+    /// Prune all sessions with number smaller than `session_id`
```
kostekIV, line 104:
```suggestion
    /// Prune all sessions with a number smaller than `session_id`
```

finality-aleph/src/sync/substrate/verification/cache.rs
```diff
@@ -0,0 +1,340 @@
...
+/// If the session is too new or ancient it will fail to return a SessionVerifier.
+/// Highest session verifier this cache returns is for the session after the current finalization session.
+/// Lowest session verifier this cache returns is for `top_returned_session` - `cache_size`.
+pub struct VerifierCache<AP, FS>
+where
+    AP: AuthorityProvider<BlockNumber>,
+    FS: FinalizationStatus,
+{
+    sessions: HashMap<SessionId, SessionVerifier>,
+    session_period: SessionPeriod,
```
kostekIV, line 52: instead of thi use SessionInfo. It will simplify the code in the impl block.

finality-aleph/src/sync/substrate/verification/cache.rs
```diff
@@ -0,0 +1,340 @@
...
+    }
+}
+
+impl<AP, FS> VerifierCache<AP, FS>
+where
+    AP: AuthorityProvider<BlockNumber>,
+    FS: FinalizationStatus,
+{
+    /// Download authorities for session and store in cache. It needs to be first session,
+    /// or first block from previous session needs to be finalized.
```
kostekIV, line 88: for me it means that the function itself checks whether the first block of the previous session is finalized. Change wording to assume or check this here

finality-aleph/src/sync/substrate/verification/cache.rs
```diff
@@ -0,0 +1,340 @@
...
+    /// Otherwise nothing is downloaded.
+    fn download(&mut self, session_id: SessionId) {
+        let maybe_session = match session_id {
```
kostekIV, line 91: nit rename to maybe_authority_data

finality-aleph/src/sync/substrate/verification/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+    },
+};
+
+mod cache;
+mod verifier;
+
+pub use verifier::SessionVerifier;
+
+/// Supplies finalized number. Will be unified together with other traits we used in A0-1839.
+pub trait FinalizationStatus {
```
kostekIV, line 32: if the naming will be different after unification ignore. I don't get the name of this trait with Status. Its rather smth like FInalizationInfo or smt like this.

finality-aleph/src/sync/substrate/verification/verifier.rs
```diff
@@ -0,0 +1,90 @@
...
+            CommitteeMultisignature(multisignature) => {
+                match self.authority_verifier.is_complete(&bytes, multisignature) {
+                    true => Ok(()),
+                    false => Err(BadMultisignature),
+                }
+            }
+            EmergencySignature(signature) => match self
+                .emergency_signer
+                .as_ref()
+                .ok_or(NoEmergencySigner)?
```
kostekIV, line 68: nice

kostekIV (changes requested):

finality-aleph/src/sync/substrate/verification/cache.rs
maciejnems: Well, to be honest this error should not happen, unless something with `AuthorityProvider` is broken. I can rename it to `UnknownAuthorities` if you think that is better.

finality-aleph/src/sync/substrate/verification/cache.rs
maciejnems: Yeah, I know. I left this comment here as I had no better/more descriptive name for this field.

finality-aleph/src/sync/substrate/verification/mod.rs
maciejnems: I intended it to be more aligned with `ChainStatus` struct. `FinalizationInfo` is also fine by me.

finality-aleph/src/sync/substrate/verification/cache.rs
maciejnems: Yeah, maybe it should. We should probably somehow systematize all such types, but right now almost nothing in `finality-aleph`, or `bin/node` is dependent on `aleph_primitives` types such as `BlockNumber`, `SessionIndex`, `SessionCount`. They are currently set separately in two different places. There is a plan to unify it in [A0-1839](https://cardinal-cryptography.atlassian.net/browse/A0-1839). For now I would leave it as it is. If `SessionId` can hold a session number, it can also hold a `SessionCount`.

finality-aleph/src/sync/substrate/verification/cache.rs
maciejnems: Good idea. I just wonder how exactly this should be done. I would prefer not to have `SessionInfo` as a  parameter of this struct.  Maybe just using implementation parameterized by `BlockNumber` would be enough.

kostekIV: why not though?

timorl (approved):
Nice, just address the comments from @kostekIV and GTG

finality-aleph/src/sync/substrate/verification/cache.rs
kostekIV: Still, I would prefer to not mix an id with a count. Using usize here would be fine

finality-aleph/src/sync/substrate/verification/cache.rs
maciejnems: I do not want cache to be dependent on `B: Block` as it is unnecessary. So I would have to either change `SessionInfo` trait to be dependent on `N: AtLeast32BitUnsigned`, or use some struct that has the same methods that is dependent on `BlockNumber`. Also `SessionInfo` always has and will have the same implementation, so I'm not even sure it should be a trait.

finality-aleph/src/sync/substrate/verification/cache.rs
maciejnems: It is a little bit tricky, to use entry here. It takes a mutable ref and then only if it is not occupied we would want to download. So download can't take the ref of `VerifierCache`... I would leave it as it is. Right now we know, that this error is because of `AuthorityProvider`

finality-aleph/src/sync/substrate/verification/cache.rs
maciejnems: Using usize would be inconvenient for all arithmetics here. Session id is already treated and used as session index, so seems better for arithmetics imo.

finality-aleph/src/sync/substrate/verification/cache.rs
timorl: You can make `download` into a function taking (mut?)ref to `AuthorityProvider` rather than a method, and returning an `Option` (then mapped into a result with the auth provider error). The `Entry` will only take a mutref to the cache itself, so the borrow checker shouldn't complain.

maciejnems: Yep, I feel like this complicates the code a little bit, but will add that. If in future (after refactor) authority provider would do all this logic all would be simpler.

finality-aleph/src/sync/substrate/verification/cache.rs
maciejnems: Changed the function a little bit and changed description

finality-aleph/src/sync/substrate/verification/cache.rs
kostekIV: Its just misleading that the sessionid is used to count something. Using usize should not be that big of a deal imho

finality-aleph/src/sync/substrate/verification/cache.rs
kostekIV: ok, feel free to resolve this. Although if you would like to do that I am leaving this open ;)

finality-aleph/src/sync/substrate/verification/cache.rs
maciejnems: It is misleading and to do it right we should rename it to `SessionIndex`. We already use it to count something in several places. We use partial ord from `u32`. `SessionId`'s are generated by a sequence. Substrate assumes they are a sequence and are `u32`.

So I would go with renaming it. And also changing it to `aleph_primitives::SessionIndex` / deriving traits such ass `Add`, `Sub` and others. All this should be done in a future refactor.

Alternatively creating a new type `SessionIndex = aleph_primitives::SessionIndex`, using it here and changing `pub struct SessionId(SessionIndex);`. If we want it to still be and id, we can leave it as an id in all other places, but use `SessionIndex` wherever there is functionality depending on it being and Index (so here, sesssion_map.rs, session.rs, party.rs). (I could do it in this PR)

The thing I don't want to do is use `usize` or `u32` here, like in `session_map.rs`.

finality-aleph/src/sync/substrate/verification/cache.rs
maciejnems: I will create a subtask for using `SessionInfo` here and refactoring it.

finality-aleph/src/sync/substrate/verification/cache.rs
kostekIV: w8, in most places you use `cache_size.0` anyway and other place is in fn `prune` which is called like that also
```rust
self.prune(SessionId(
                session_id.0.saturating_sub(self.cache_size.0) + 1,
            ));
```
, so I dont understand the problem here. Secondly, this variable indicates the size of the cache and has nothing to do with a session - so no shenanigans with SessionIndex and derivatives of it would make sense. Using usize should be natural here. My initial comment with using the type alias `SessionCount = usize` was to just indicate that we are counting sessions here.

maciejnems: Ah, yeah, you are right. I can use:
```
self.lower_bound.0.saturating_add(self.cache_size.saturated_into())
```
I will add it to this PR.

Still, changing `SessionId` to `SessionIndex = aleph_primitives::SessionIndex/u32` would remove need of using `session_id.0`. This would simplify a lot of stuff. Pattern of wrapping basic types does not really make sense if you can just get the inside of this type.

kostekIV (approved):

PR #867, opened 2023-01-16
Migrate extension to ink!4. Remove contract example

pmikolajczyk41:
# Description

Three changes to the SC side:
 - migrate extension to the new ink!
 - remove the contract showing usage of extension, because of 3 reasons:
   - Shielder already presents how to perform interaction with the extension
   - there are build-checks for this contract, so we don't have certainty in CI that it indeed works
   - chain extension API might soon be changed due to 727 ventures grant; there is now less code to migrate later
 - hide `preimage` (and thus `poseidon`) behind a feature flag in `relations`

## Type of change


relations/src/lib.rs
```diff
@@ -18,6 +19,7 @@ pub use environment::{
 };
 pub use linear::LinearEquationRelation;
 pub use merkle_tree::{MerkleTreeRelation, Root};
+#[cfg(any(feature = "preimage", feature = "preimage-std"))]
```
deuszx, line 22: I understand that it's so that the project compiles but in which situations you imagine we would enable these features?

Also, if we only ever use them in the combination like this - _any of the two_ - can we replace it with a single feature?

deuszx (approved):

relations/src/lib.rs
pmikolajczyk41: > in which situations you imagine we would enable these features

once we substitute 'tangling' with 'poseidon'

> replace it with a single feature

if 'poseidon' is compiling perfectly as we intend, there'll be no feature; otherwise, we might need it in both `std` and `no-std` context

poseidon/Cargo.toml
```diff
@@ -8,20 +8,21 @@ description = "An instantiation of the Poseidon SNARK-friendly hash function."
...
 ark-relations = { version = "^0.3.0", default-features = false }
 ark-sponge = { git = "https://github.com/penumbra-zone/sponge", branch = "r1cs", default-features = false, features = ["r1cs"] }
-once_cell = "1.8"
-poseidon-paramgen = { version = "0.1", default-features = false }
-poseidon-permutation = { git = "https://github.com/penumbra-zone/poseidon377", default-features = false }
 ark-std = { version = "^0.3.0", default-features = false }
+ark-ff = { version = "^0.3.0", default-features = false }
+once_cell = { version = "1.17", default-features = false, features = ["critical-section"] }
+
+poseidon-paramgen = { git = "https://github.com/pmikolajczyk41/poseidon377", branch = "no-std", default-features = false }
```
fbielejec, line 18: this is merged now, correct?

poseidon/Cargo.toml
```diff
@@ -8,20 +8,21 @@ description = "An instantiation of the Poseidon SNARK-friendly hash function."
...
-ark-ff = { version = "^0.3.0", default-features = false }
 ark-r1cs-std = {version = "^0.3.0" , default-features = false }
 ark-relations = { version = "^0.3.0", default-features = false }
 ark-sponge = { git = "https://github.com/penumbra-zone/sponge", branch = "r1cs", default-features = false, features = ["r1cs"] }
-once_cell = "1.8"
-poseidon-paramgen = { version = "0.1", default-features = false }
-poseidon-permutation = { git = "https://github.com/penumbra-zone/poseidon377", default-features = false }
 ark-std = { version = "^0.3.0", default-features = false }
+ark-ff = { version = "^0.3.0", default-features = false }
+once_cell = { version = "1.17", default-features = false, features = ["critical-section"] }
```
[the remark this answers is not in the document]
pmikolajczyk41: turns on `sync::Lazy`, idk how on earth it works in poseidon377 without this feature...

fbielejec (commented):
not sure if it's worth touching poseidon in this PR (other than hiding it behind a feature)

Cargo.toml
```diff
@@ -17,7 +17,6 @@ exclude = [
     "benches/payout-stakers",
     "bin/cliain",
     "bin/runtime/src/chain_extension/snarcos-extension",
-    "contracts/snarcos-contract",
```
fbielejec, line 20: I agree this contract is maybe not the best as an example, but could still be somehow usefull - e.g. compiling it in CI with relations crate to catch any std creeping in? I looked into various ways of checking whether a given crate compiles in a no_std environment and unfortunately it is not that straightforward, e.g. there is no way to capture it in a starightforward unit test :/

poseidon/Cargo.toml
pmikolajczyk41: right, corrected

Cargo.toml
pmikolajczyk41: I would revive it once we have stable building procedures for contracts (hopefully in a public docker image); also @deuszx had kinda strong opinion on that contract (to remove it)

poseidon/Cargo.toml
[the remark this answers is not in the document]
fbielejec: are you sure, we are using the sync module directly: https://docs.rs/once_cell/latest/once_cell/sync/index.html 

This features seems to use some other crate (called `critical_section`) that is 
https://github.com/matklad/once_cell/blob/master/Cargo.toml#L52

that seems to be for use with some exotic, embedded device architectures: https://docs.rs/critical-section/1.1.1/critical_section/

fbielejec (approved):

poseidon/Cargo.toml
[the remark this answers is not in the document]
pmikolajczyk41: I couldn't get it compiled without this feature :man_shrugging: 

reverted poseidon anyway

Cargo.toml
deuszx: To be honest, I didn't look at this from the perspective that @fbielejec is bringing up - i.e. the litmus test. It's true that the zk-apps will be more difficult to keep aligned with the repo here (unless we add some kind of nightly tests to it) but, at the same time, a lot of the interactions between contract <-> chain extension is happening via pattern matching on the usigned ints so it's hard to tell how many errors would keeping the `snarcos-contract` give us.

PR #870, opened 2023-01-17
A0-1845: Authorities api

kostekIV:
# Description
Now, pallet-aleph is entirely responsible for providing authority sets to the finality aleph gadget. Part of the work needed for disuniting Aura and AlephBFT committee.

# Checklist:


timorl (approved):

DamianStraszak (commented):
Not sure where this is going yet, but looks good so far :)

Waiting for the migration though :)

pallets/aleph/src/lib.rs
```diff
@@ -134,18 +138,32 @@ pub mod pallet {
         StorageValue<_, VersionChange, OptionQuery>;
 
     impl<T: Config> Pallet<T> {
-        pub(crate) fn initialize_authorities(authorities: &[T::AuthorityId]) {
+        pub(crate) fn initialize_authorities(
+            authorities: &[T::AuthorityId],
+            next_authorities: &[T::AuthorityId],
+        ) {
             if !authorities.is_empty() {
                 assert!(
```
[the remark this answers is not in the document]
kostekIV: done (not in new code because of the default value tho)

pallets/aleph/src/lib.rs
```diff
@@ -105,10 +106,21 @@ pub mod pallet {
         DEFAULT_FINALITY_VERSION
     }
 
+    /// Default finality version. Relevant for sessions before the first version change occurs.
```
DamianStraszak, line 109: Comment also got copied.

primitives/src/lib.rs
```diff
@@ -164,7 +164,7 @@ impl<AccountId> Default for EraValidators<AccountId> {
 
 #[derive(Encode, Decode, PartialEq, Eq, Debug)]
 pub enum ApiError {
-    DecodeKey,
+    MissingAuthoritiesForNextSession,
```
DamianStraszak, line 167: Not entirely sure but this change might be problematic.

Because of change in this type, the runtime Api changes, and the new host knows only the new runtime Api. So after an update, the host will use the types for the new Api to call the old on-chain Api. This might sound confusing, and rightfully so, so I'm happy to explain on slack or during a call. 

What's bad is that this will compile fine. Even during runtime this might be fine, because we will be lucky and things will deserialize correctly. But all the time we will be deserializing to a different type than the one that the data was serialized from.

pallets/aleph/src/lib.rs
kostekIV: fixed

DamianStraszak (approved):

PR #875, opened 2023-01-18
`aleph-client`: Cast `ConvertibleValue` to a sequence

pmikolajczyk41:
# Description

This PR extends the range of types to which we can cast `ConvertibleValue`. Useful directly in shielderish CLI tools.

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/aleph-client/src/contract/convertible_value.rs
+++ b/aleph-client/src/contract/convertible_value.rs
@@ -45,7 +45,7 @@ macro_rules! try_from_flat_value {
         impl TryFrom<ConvertibleValue> for $ty {
             type Error = anyhow::Error;
 
-            fn try_from(value: ConvertibleValue) -> Result<$ty, Self::Error> {
+            fn try_from(value: ConvertibleValue) -> anyhow::Result<$ty> {
                 match value.0 {
                     Value::$variant(value) => Ok(value.try_into()?),
                     _ => anyhow::bail!("Expected {:?} to be {}", value, $desc),
@@ -69,7 +69,7 @@ try_from_flat_value!(i128, Int, "signed integer");
 impl TryFrom<ConvertibleValue> for AccountId {
     type Error = anyhow::Error;
 
-    fn try_from(value: ConvertibleValue) -> Result<AccountId, Self::Error> {
+    fn try_from(value: ConvertibleValue) -> Result<AccountId> {
         match value.0 {
             Value::Literal(value) => {
                 AccountId::from_str(&value).map_err(|_| anyhow!("Invalid account id"))
@@ -85,7 +85,7 @@ where
 {
     type Error = anyhow::Error;
 
-    fn try_from(value: ConvertibleValue) -> Result<Result<T>, Self::Error> {
+    fn try_from(value: ConvertibleValue) -> Result<Result<T>> {
         if let Value::Tuple(tuple) = &value.0 {
             match tuple.ident() {
                 Some(x) if x == "Ok" => {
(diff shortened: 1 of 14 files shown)


aleph-client/src/contract/convertible_value.rs
```diff
@@ -40,38 +40,31 @@ impl Deref for ConvertibleValue {
...
+            fn try_from(value: ConvertibleValue) -> Result<$ty, Self::Error> {
+                match value.0 {
+                    Value::$variant(value) => Ok(value.try_into()?),
+                    _ => anyhow::bail!("Expected {:?} to be {}", value, $desc),
+                }
+            }
         }
-    }
+    };
 }
```
obrok, line 56: Reaction: Nice, I like self-contained, private macros like this.

aleph-client/src/contract/convertible_value.rs
```diff
@@ -193,3 +186,124 @@ where
         }
     }
 }
+
+impl<Elem: TryFrom<ConvertibleValue, Error = anyhow::Error>> TryFrom<ConvertibleValue>
+    for Vec<Elem>
+{
+    type Error = anyhow::Error;
+
+    fn try_from(value: ConvertibleValue) -> std::result::Result<Self, Self::Error> {
```
obrok, line 195: Question:

```suggestion
    fn try_from(value: ConvertibleValue) -> Result<Self> {
```

should work here, no?

aleph-client/src/contract/convertible_value.rs
```diff
@@ -193,3 +186,124 @@ where
...
+        let seq = match value.0 {
+            Value::Seq(seq) => seq,
+            _ =>  bail!("Failed parsing `ConvertibleValue` to `Vec<T>`. Expected `Seq(_)` but instead got: {:?}", value),
+        };
+
+        let mut result = vec![];
+        for element in seq.elems() {
+            result.push(ConvertibleValue(element.clone()).try_into()?);
+        }
```
obrok, line 204: Suggestion: something along the lines of

```suggestion
        seq.elems().into_iter().map(|x| ConvertibleValue(x).try_into()).collect()
```
? Not sure if better.

obrok (approved):
Some minor stylistic points, feel free to ignore any and all of them.

aleph-client/src/contract/convertible_value.rs
```diff
@@ -193,3 +186,124 @@ where
...
+    use contract_transcode::Value::{Bool, Char, Int, Seq, UInt};
+
+    use crate::contract::ConvertibleValue;
+
+    #[test]
+    fn converts_boolean() {
+        let cast: bool = ConvertibleValue(Bool(true))
+            .try_into()
+            .expect("Should cast successfully");
+        assert_eq!(true, cast);
```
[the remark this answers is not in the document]
pmikolajczyk41: `assert2` would be nice, but then it should be applied across the whole crate and I'm not eager to do that in this PR :|

one-liners would be cool indeed, but there are two problems:
 - no implementation for `anyhow::Error == anyhow::Error`, so there must be unwrapping
 - the first line has type ascription, which is required while casting; otherwise I would have to use `assert_eq!(-41, <ConvertibleValue as TryInto<i32>>::try_into(ConvertibleValue(Int(-41))).unwrap());`, which is pretty ugly

aleph-client/src/contract/convertible_value.rs
pmikolajczyk41: meh, you would have to:
 - add `.clone()` to `x`
 - convert `Vec<Result<_>>` into `Result<Vec<_>>` through further `iter().collect()`
and it gets meh

generally, I find loop a better option when the mapper returns `Result` or `Option`

aleph-client/src/contract/convertible_value.rs
pmikolajczyk41: right - I simplified it in every place in the module

mike1729 (approved):

PR #877, opened 2023-01-19
`aleph-client`: Fetch contract events

pmikolajczyk41:
# Description

This PR introduces `get_contract_events` call which, similarly to `get_tx_events`, fetches all events that were emitted by a contract call and decodes them using metadata. This covers part of cases, when one had to use subscribing mechanism. 

One e2e test has been adapted to the new API.

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/aleph-client/src/contract/event.rs
+++ b/aleph-client/src/contract/event.rs
@@ -1,67 +1,13 @@
-//! Utilities for listening for contract events.
+//! This module provides utilities corresponding to the events emitted by a contract.
 //!
-//! To use the listening feature you will need to pass a connection, some contracts and an
-//! `UnboundedSender` to the [listen_contract_events] function. You most likely want to
-//! `tokio::spawn` the resulting future, so that it runs concurrently.
-//!
-//! ```no_run
-//! # use std::sync::Arc;
-//! # use std::sync::mpsc::channel;
-//! # use std::time::Duration;
-//! # use aleph_client::{AccountId, Connection, SignedConnection};
-//! # use aleph_client::contract::ContractInstance;
-//! # use aleph_client::contract::event::{listen_contract_events};
-//! # use anyhow::Result;
-//! use futures::{channel::mpsc::unbounded, StreamExt};
-//!
-//! # async fn example(conn: Connection, signed_conn: SignedConnection, address1: AccountId, address2: AccountId, path1: &str, path2: &str) -> Result<()> {
-//! // The `Arc` makes it possible to pass a reference to the contract to another thread
-//! let contract1 = Arc::new(ContractInstance::new(address1, path1)?);
-//! let contract2 = Arc::new(ContractInstance::new(address2, path2)?);
-//!
-//! let conn_copy = conn.clone();
-//! let contract1_copy = contract1.clone();
-//! let contract2_copy = contract2.clone();
-//!
-//! let (tx, mut rx) = unbounded();
-//! let listen = || async move {
-//!     listen_contract_events(&conn, &[contract1_copy.as_ref(), contract2_copy.as_ref()], tx).await?;
-//!     <Result<(), anyhow::Error>>::Ok(())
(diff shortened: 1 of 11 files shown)


e2e-tests/src/test/adder.rs
```diff
@@ -19,26 +18,22 @@ pub async fn adder() -> Result<()> {
...
     let before = contract.get(&conn).await?;
 
-    contract.add(&account.sign(&conn), increment).await?;
+    let tx_info = contract.add(&account.sign(&conn), increment).await?;
+    let events = get_contract_events(&conn, &contract.contract, tx_info).await?;
+    let event = match events.len() {
+        0 => return Err(anyhow!("No event received")),
+        1 => events[0].clone(),
+        _ => return Err(anyhow!("Too many events received")),
+    };
```
deuszx, line 35: Tip: You can extract element while pattern matching on the array
```rust
match &*events {
  &[event] => event,
 _ => return Err(...)
}
```

Use or not :)

e2e-tests/src/test/adder.rs
```diff
@@ -19,26 +18,22 @@ pub async fn adder() -> Result<()> {
...
-    )?);
-
-    let listen_conn = conn.clone();
-    let listen_contract = contract.clone();
-    let (tx, mut rx) = unbounded();
-    let listen = || async move {
-        listen_contract_events(&listen_conn, &[listen_contract.as_ref().into()], tx).await?;
-        <Result<(), anyhow::Error>>::Ok(())
-    };
-    let join = tokio::spawn(listen());
```
obrok, line 34: Request: Write a new test instead of changing this code. This is not code testing the adder contract - the adder contract is used for testing the client. As far as I can tell you're removing the tests for event listening and replacing them with the tests for this new mechanism.

aleph-client/src/contract/event.rs
```diff
@@ -41,14 +41,39 @@
...
+//! # async fn example(conn: Connection, signed_conn: SignedConnection, address: AccountId, path: &str) -> Result<()> {
+//! let contract = ContractInstance::new(address, path)?;
+//!
+//! let tx_info = contract.contract_exec0(&signed_conn, "some_method").await?;
+//!
+//! println!("Received events {:?}", get_contract_events(&conn, &contract, tx_info).await);
+//!
+//! #   Ok(())
+//! # }
+//! ```
```
obrok, line 64: Suggestion: not that there is these two ways of getting events, I propose:

1. Moving the docs for the two ways to the docstrings of the two functions.
2. Changing this docstring to focus on a comparison between the two methods.

obrok (changes requested):

e2e-tests/src/test/adder.rs
deuszx: The message could have been `"expected single event but got {:?}, events"` but it's your call :)

obrok: I agree with @deuszx for the record - less code = better

aleph-client/src/contract/event.rs
pmikolajczyk41: done

e2e-tests/src/test/adder.rs
pmikolajczyk41: I agree - I've made two testcases, but decided not to extract common setup checks, since it requires abstracting some action, that takes inner action and returns an event, which makes things quite illegible

however, if you have some suggestion, I'm eager to fix this

e2e-tests/src/test/adder.rs
pmikolajczyk41: simplified

obrok (approved):

e2e-tests/src/test/adder.rs
obrok: Looks good

PR #880, opened 2023-01-23
`aleph-client`: Custom gas limit

pmikolajczyk41:
# Description

1. We bump default gas limit from 10B to 250B (~25% of the block capacity). This is because all Liminal-related contracts require much more than 10B gas and therefore we cannot rely directly on aleph-client (which used to have this limit hardcoded)
2. We allow user to override this limit.

## Type of change

# Checklist:


aleph-client/src/contract/mod.rs
```diff
@@ -168,8 +177,8 @@ impl ContractInstance {
             self.address.clone(),
             value,
             Weight {
-                ref_time: Self::MAX_GAS,
-                proof_size: Self::MAX_GAS,
+                ref_time: self.max_gas_override.unwrap_or(DEFAULT_MAX_GAS),
+                proof_size: self.max_gas_override.unwrap_or(DEFAULT_MAX_GAS),
```
deuszx, line 181: These are almost never equivalent and usually much different. Maybe there should be a way to override them separately in the final version of the client.

deuszx: What does it mean _in our case_? When I dry run any contract (bulletin board, psp22, etc.) these two are always different and non-zero.

pmikolajczyk41: if i'm not mistaken, it is non-zero, because that's what pallets' dispatchables return, but it isn't incurred

below you have both weights non-zero, but proof_size is not included in the actual fee (there's the standard weight + some length fee)
image

obrok (approved):

aleph-client/src/contract/mod.rs
deuszx: The thing is you have to provide "enough" `proofSize` limit to get the txn executed - that's how it works with the JS lib (i.e. I can't get txn to execute if `proofSize == 0`). If we are not charging the `proofSize` cost then it's even worse IMO b/c it's confusing: we require people to estimate/set it properly but in the end we will not charge for that? Even though it is returned in the final result?

pmikolajczyk41: added possibility to override proof size limit

aleph-client/src/contract/mod.rs
```diff
@@ -59,23 +59,42 @@ use crate::{
...
+/// Default gas limit, which allows up to 25% of block execution time.
+pub const DEFAULT_MAX_GAS: u64 = 250_000_000_000u64;
+/// Default proof size limit, which allows up to 25% of block execution time.
+pub const DEFAULT_MAX_PROOF_SIZE: u64 = 250_000_000_000u64;
+
 /// Represents a contract instantiated on the chain.
 pub struct ContractInstance {
     address: AccountId,
     transcoder: ContractMessageTranscoder,
+    max_gas_override: Option<u64>,
```
krzysztofziobro, line 71: Does not really matter, but why not just `max_gas: u64`?

aleph-client/src/contract/mod.rs
```diff
@@ -59,23 +59,42 @@ use crate::{
     AccountId, Balance, ConnectionApi, SignedConnectionApi, TxStatus,
 };
 
+/// Default gas limit, which allows up to 25% of block execution time.
```
krzysztofziobro, line 62: Where can I check that this is indeed 25% ?

krzysztofziobro (approved):
Looks fine, just two questions

aleph-client/src/contract/mod.rs
pmikolajczyk41: so that you can easily back off to the default, by setting to None (it doesn't require knowing `DEFAULT_MAX_GAS`)

aleph-client/src/contract/mod.rs
pmikolajczyk41: our block has 10^12 weight capacity - it should be somewhere in `primitives`

krzysztofziobro: Hmm, from what I found `MAX_BLOCK_WEIGHT` is 400 * 10^9. Maybe I should look for another constant?

krzysztofziobro: From what I understand we have 40% of the block for production, so new default would be >60% of production time

krzysztofziobro: So from what I understand it is indeed 25% of total block time, but >60% of actual max gas limit

pmikolajczyk41: yes, that's right

pmikolajczyk41: I've changed the docstring to be more precise

PR #881, opened 2023-01-23
A0-1718 remove old authentications

krzysztofziobro:
# Description
Removing old authentications.

## Type of change


finality-aleph/src/network/session/compatibility.rs
```diff
@@ -19,129 +16,36 @@ type ByteCount = u16;
...
-            NewOnly(authentication) => vec![V2(authentication)],
-            LegacyOnly(legacy_authentication) => {
-                vec![V1(AuthenticationBroadcast(legacy_authentication))]
-            }
-            Both(authentication, legacy_authentication) => vec![
-                V2(authentication),
-                V1(AuthenticationBroadcast(legacy_authentication)),
-            ],
-        }
+        vec![V2(authentication)]
```
timorl, line 29: The `use` above is a bit of an overkill, just spell the whole thing out in one line. ;p

finality-aleph/src/network/session/compatibility.rs
```diff
@@ -19,129 +16,36 @@ type ByteCount = u16;
...
 
-/// One of the possible messages we could have gotten as part of discovery.
-/// Ignores whether the old authentication was a broadcast or not, we don't send the
-/// non-broadcasts anymore anyway, and treating them as broadcasts doesn't break anything.
-#[derive(Clone, Debug, PartialEq, Eq, Hash)]
-pub enum DiscoveryMessage<M: Data, A: AddressingInformation + TryFrom<Vec<M>> + Into<Vec<M>>> {
-    Authentication(Authentication<A>),
-    LegacyAuthentication(LegacyAuthentication<M>),
-}
+pub type DiscoveryMessage<A> = Authentication<A>;
```
timorl, line 33: Ooh, nice hack, love it.

finality-aleph/src/network/session/manager.rs
```diff
@@ -7,13 +7,14 @@ use std::{
 use futures::channel::mpsc;
 use log::{debug, info};
 
+use super::Authentication;
```
timorl, line 10: Ah, we generally avoid `super` (not sure whether that's actually good, but this is the current convention). In particular some stuff below is imported from `super`, but without using the keyword, so this is now inconsistent even within this one file.

timorl (approved):
Nice, thanks for the trimming work. <3

finality-aleph/src/network/session/compatibility.rs
krzysztofziobro: fixed

finality-aleph/src/network/session/manager.rs
krzysztofziobro: fixed

finality-aleph/src/network/session/compatibility.rs
krzysztofziobro: thx

maciejnems (approved):
Very Nice

PR #890, opened 2023-01-27
A0-1304: add state pruning compatibility

maciejnems:
# Description

This PR add compatibility for running Aleph Node with pruning turned on. For now it stays as an experimental feature and because of that whenever using it, appropriate warning will be logged.

To use pruning one needs to provide `--experimental-pruning` parameter. When providing it, `--state-pruning` and `--blocks-pruning` can be set to possible values:
* `archive` - Keep all blocks/states
* `archive-canonical` - Keep only finalized blocks/states
* `number` -  Keep the last `number` of finalized blocks/states

For now `archive-canonical` setting does not work for state pruning. 

Default values are
* `--state-pruning=archive`
* `--blocks-pruning=archive-canonical`

Using experimental pruning should result in warning similar to:
```
WARN main aleph_node: Experimental_pruning was turned on. Usage of this flag can lead to misbehaviour, which can be punished. State pruning: Constrained(Constraints { max_blocks: Some(1000), max_mem: None }); Blocks pruning: Some(1000);
```
Changing pruning parameters without setting experimental pruning should result in:
```
WARN main aleph_node: Pruning not supported. Switching to keeping all block bodies and states.    
```

## Changelog
- Fix experimental pruning warnings not showing up (logger is not yet initialized before `cli.create_runner(...)`)
- Change `runtime_api` calls in code to reference last finalized block (not genesis block, which can be discarded)
- `SessionMap` now supports old states being discarded during catch-up phase (only state required is of first block, of session in which `top_finalized` block is. This requirement can be narrowed down, and we can ask for state of last finalized block. Not needed yet, as we plan to use `2048` as recommended pruning setting)
- Add sanity check, so that we are not suprised in the future about pruning parameters values

# Checklist:

- I have added tests

### DIFF
--- a/bin/node/src/main.rs
+++ b/bin/node/src/main.rs
@@ -4,7 +4,7 @@ use aleph_node::{new_authority, new_full, new_partial, Cli, Subcommand};
 #[cfg(feature = "try-runtime")]
 use aleph_runtime::Block;
 use log::warn;
-use sc_cli::{clap::Parser, PruningParams, SubstrateCli};
+use sc_cli::{clap::Parser, CliConfiguration, PruningParams, SubstrateCli};
 use sc_network::config::Role;
 use sc_service::PartialComponents;
 
@@ -116,8 +116,8 @@ fn main() -> sc_cli::Result<()> {
             let runner = cli.create_runner(&cli.run)?;
             if cli.aleph.experimental_pruning() {
                 warn!("Experimental_pruning was turned on. Usage of this flag can lead to misbehaviour, which can be punished. State pruning: {:?}; Blocks pruning: {:?};", 
-                    cli.run.import_params.pruning_params.state_pruning(),
-                    cli.run.import_params.pruning_params.blocks_pruning(),
+                    cli.run.state_pruning()?.unwrap_or_default(),
+                    cli.run.blocks_pruning()?,
                 );
             } else if overwritten_pruning {
                 warn!("Pruning not supported. Switching to keeping all block bodies and states.");
--- a/bin/node/src/main.rs
+++ b/bin/node/src/main.rs
@@ -4,27 +4,29 @@ use aleph_node::{new_authority, new_full, new_partial, Cli, Subcommand};
 #[cfg(feature = "try-runtime")]
 use aleph_runtime::Block;
 use log::warn;
-use sc_cli::{clap::Parser, SubstrateCli};
+use sc_cli::{clap::Parser, PruningParams, SubstrateCli};
 use sc_network::config::Role;
 use sc_service::PartialComponents;
 
+const STATE_PRUNING: &str = "archive";
+const BLOCKS_PRUNING: &str = "archive-canonical";
+
+fn pruning_changed(params: &PruningParams) -> bool {
+    let state_pruning_changed =
+        params.state_pruning != Some(STATE_PRUNING.into()) && params.state_pruning.is_some();
+
+    let blocks_pruning_changed =
+        params.blocks_pruning != Some(BLOCKS_PRUNING.into()) && params.blocks_pruning.is_some();
+
+    state_pruning_changed || blocks_pruning_changed
+}
+
 fn main() -> sc_cli::Result<()> {
     let mut cli = Cli::parse();
+    let overwritten_pruning = pruning_changed(&cli.run.import_params.pruning_params);
     if !cli.aleph.experimental_pruning() {
-        if cli
-            .run
-            .import_params
-            .pruning_params
-            .blocks_pruning
-            .is_some()
-            || cli.run.import_params.pruning_params.state_pruning != Some("archive".into())
-        {
-            warn!("Pruning not supported. Switching to keeping all block bodies and states.");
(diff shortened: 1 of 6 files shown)


bin/node/src/main.rs
```diff
@@ -112,6 +114,15 @@ fn main() -> sc_cli::Result<()> {
             .into()),
         None => {
             let runner = cli.create_runner(&cli.run)?;
+            if cli.aleph.experimental_pruning() {
+                warn!("Experimental_pruning was turned on. Usage of this flag can lead to misbehaviour, which can be punished. State pruning: {:?}; Blocks pruning: {:?};",
```
timorl, line 118: I love the idea of this log, but the things that get printed are hard to read – could you print them out in a more friendly format?

finality-aleph/src/session_map.rs
```diff
@@ -281,76 +264,95 @@ where
...
-        }
+        self.handle_first_block_of_session(current_session).await;
 
-        let mut last_updated = current_session;
+        current_session
+    }
+
+    pub async fn run(mut self) {
+        let mut notifications = self.finality_notificator.notification_stream();
+        let mut last_updated = self.catch_up().await;
```
timorl, line 342: I think technically you can catch up to a session while the current session advances, and you'll never get a notification, potentially leaving you without one of the authorities. I guess this should only happen on startup, limiting impact in practice, and as far as I can tell it was also the case before these changes.

I don't think you have to do anything about this, leaving this comment so you can let me know if I'm wrong or in case you actually see a simple way of fixing this. Otherwise feel free to resolve.

timorl (approved):
Lovely! <3

finality-aleph/src/session_map.rs
kostekIV: Not really, the magic of unbounded channel :P We have open notifications channel before calling catch_up

timorl: Ah, if that's a guarantee of the notification stream then all's good.

bin/node/src/main.rs
maciejnems: I wanted to print to what value the arguments actually evaluate to. I can also print here the value of passed arguments.

timorl: I meant it would be nice if you printed something representing what they are evaluated to, but without all the `Ok(Some....` boilerplate, more human-readable.

maciejnems: Ok, done

timorl: Uhm, that is definitely better, but I still wouldn't call it "human-readable". :P If making it more expressive is hard then it can stay as-is, but if you can make it write out the contents in an even nicer form that would be appreciated.

maciejnems: But that would require digging inside substrate even more and in future might require update, if they change something :// anyway, only state-pruning with number is this ugly. Others are just enum names.

finality-aleph/src/sync/substrate/verification/cache.rs
```diff
@@ -38,7 +38,7 @@ impl Display for CacheError {
             UnknownAuthorities(session) => {
                 write!(
                     f,
-                    "authorities for session {:?} not present on chain even though they should be",
+                    "authorities for session {:?} not present on chain. Most likely state is already pruned",
```
kostekIV, line 41: The older message was better. We expected the authorities to be present. Semi-guessing that they are pruned even though the running might not be turned on could be misleading.

finality-aleph/src/session_map.rs
```diff
@@ -281,76 +264,95 @@ where
...
 
-    pub async fn run(mut self, period: SessionPeriod) {
-        let mut notifications = self.finality_notificator.notification_stream();
+        // lets catch up with previous sessions
+        for session in starting_session.0..current_session.0 {
+            let id = SessionId(session);
+            if let Some(authority_data) = self.authorities_for_session(id) {
+                self.session_map.update(id, authority_data).await;
+            } else {
+                debug!(target: "aleph-session-updater", "No authorities for session {:?} during catch-up. Most likely already pruned.", id.0)
```
kostekIV, line 318: Im pretty sure at this point we should already panic. If we don't know the authorities for some older sessions how can we make sure the current session authorities are correct

kostekIV: right, good point

kostekIV (approved):

finality-aleph/src/sync/substrate/verification/cache.rs
maciejnems: True, that said, chain part of the message is also confusing now. How about "authorities for session {:?} not known even though they should be"?

mike1729:
I know that the name `archive-cannonical` is implied by the substrate naming but can we rename it to `archive-finalized`?

maciejnems:
> I know that the name `archive-cannonical` is implied by the substrate naming but can we rename it to `archive-finalized`?

This would unfortunately require messing with SubstrateCli and overwriting it in our main. Not sure this is worth it ://

PR #892, opened 2023-01-27
A0-1897 validator-network authorization

fixxxedpoint:
# Description

Added authorization api for the validator-network. Now, we start sending/receiving from a node only after successful authorization.

## Type of change

# Checklist:

- I have added tests


lesniak43:
wow...

lesniak43:
The code looks solid and well thought out, but I have to admit that it's a little overwhelming... I think that the best approach to review it will be to describe how I'd try to tackle this issue, and then we may discuss the proposed approach - is that ok?

So. We have two threads, [S]ervice and [W]orker, where S spawned W. Currently there's a one-way communication channel between them, created in S:
`let (result_for_parent, mut worker_results) = mpsc::unbounded();`
which was supposed to be used just once, after the worker establishes the connection.

Now we need to be able to ask thread S from inside W about the peer that connected to us. Note that the answer may be given synchronously, so there's no need (I'm not saying there's no reason) to introduce additional abstractions at this stage. Also, since this is a one-time (per created worker) interaction, I'm perfectly fine with sending a `Oneshot` channel attached to the request, so that S may use it to respond - the alternative being keeping a dict of spawned worker threads inside S to be able to manage a two-way communication seems a little too complicated right now.

This approach would require to introduce an `Enum` for the second message type. If we prefer to keep the channels separate, then I guess this is also fine, and in thread S we'd have something like:
```
let (result_for_parent, mut worker_results) = mpsc::unbounded();
let (authorizator, mut authorization_handler) = mpsc::unbounded(); // names to be changed?
```
and then
```
tokio::select! {
    ...
    (pk, oneshot) = authorization_handler.next() => oneshot.send(self.manager.is_authorized(&pk)), // this is fully sync
    ...
```

The worker W would just simply create the `Oneshot` pair, use the `authorizator` to send the request, and `await` on the receiver. 

I'm not claiming that this approach is better in general, but it has a huge advantage of being probably the simplest solution, given the current design of the network. This PR has over 700 lines of code, so it seems more like an introduction to a large refactor, and I'm not sure if this was the goal here, although it might be a good idea to talk about this on the next planning?

I'd like to hear your opinion before reviewing this PR.

finality-aleph/src/network/clique/protocols/v0/mod.rs
```diff
@@ -98,6 +102,17 @@ pub async fn incoming<SK: SecretKey, D: Data, S: Splittable>(
         "Incoming handshake with {} finished successfully.", public_key
     );
 
+    let authorized = handle_authorization::<SK>(authorization_requests_sender, public_key.clone())
+        .await
+        .map_err(|_| ProtocolError::NotAuthorized)?;
+    if !authorized {
```
timorl, line 108: There is some double-logic happening here, the error above is called `NotAuthorized`, but is actually about authorization failing for other reasons, and then here the actual lack of authorization happens.

To be more in line with the rest of this function I would suggest having `handle_authorization` return `ProtocolError`, in practice either `NoParentConnection` or `NotAuthorized`. Then don't log it here, but just `?` it. The function wouldn't have to return a `bool` at all.

I think it would also be simpler if you didn't send a `bool` through the oneshot, just a unit, and use dropping to signal lack of interest to have less states to worry about, like the surrounding code does. However, if you think sending a `bool` provides useful information, then do it, but return `NoParentConnection` on a closed channel and `NotAuthorized` on a `false`.

finality-aleph/src/network/clique/protocols/v0/mod.rs
```diff
@@ -123,14 +138,45 @@ pub async fn incoming<SK: SecretKey, D: Data, S: Splittable>(
     }
 }
 
+pub async fn handle_authorization<SK: SecretKey>(
+    authorization_requests_sender: mpsc::UnboundedSender<(SK::PublicKey, oneshot::Sender<bool>)>,
+    public_key: SK::PublicKey,
+) -> Result<bool, ()> {
+    let (sender, receiver) = oneshot::channel();
+    if let Err(err) = authorization_requests_sender.unbounded_send((public_key.clone(), sender)) {
+        warn!(
```
timorl, line 147: Related to the previous point – if you return meaningful errors from here then you don't have to `log` inside this function, simplifying the code quite a bit. Although maybe that's just my preference for "log as high up the stack as possible"?

finality-aleph/src/network/clique/service.rs
```diff
@@ -233,12 +244,14 @@ where
     pub async fn run(mut self, mut exit: oneshot::Receiver<()>) {
         let mut status_ticker = time::interval(STATUS_REPORT_INTERVAL);
         let (result_for_parent, mut worker_results) = mpsc::unbounded();
+        let (authorization_requests_sender, mut authorization_requests) = mpsc::unbounded();
         use ServiceCommand::*;
         loop {
+            let listener = &mut self.listener;
```
timorl, line 250: Huh, is this necessary now?

finality-aleph/src/network/clique/protocols/mod.rs
```diff
@@ -69,6 +72,7 @@ impl<PK: PublicKey> Display for ProtocolError<PK> {
             CardiacArrest => write!(f, "heartbeat stopped"),
             NoParentConnection => write!(f, "cannot send result to service"),
             NoUserConnection => write!(f, "cannot send data to user"),
+            NotAuthorized => write!(f, "user not authorized"),
```
timorl, line 75: Probably "peer" rather than "user", we use the latter to refer to the local user of the network.

finality-aleph/src/network/clique/protocols/v0/mod.rs
```diff
@@ -123,14 +138,45 @@ pub async fn incoming<SK: SecretKey, D: Data, S: Splittable>(
     }
 }
 
+pub async fn handle_authorization<SK: SecretKey>(
```
timorl, line 141: Maybe just `authorize`? You are not handling an authorization here, just trying to get the connection authorized. But this is silly naming stuff, a complete nit, feel free to ignore. ^^"

timorl (approved):
I forgot I wasn't supposed to be reviewing this, and when I remembered I was almost done anyway. Dx

Have this approve and some comments then. <_<"

kostekIV (approved):
other than @timorl comments about errors and logs looks good to me

finality-aleph/src/network/clique/protocols/v0/mod.rs
fixxxedpoint: Changed.

finality-aleph/src/network/clique/protocols/v0/mod.rs
fixxxedpoint: changed

finality-aleph/src/network/clique/service.rs
fixxxedpoint: removed

finality-aleph/src/network/clique/protocols/mod.rs
fixxxedpoint: changed

finality-aleph/src/network/clique/protocols/v0/mod.rs
fixxxedpoint: I choose `check_authorization`.

PR #895, opened 2023-01-31
A0-1763: Remove v0 connections from clique

krzysztofziobro:
# Description

Removing unidirectional connections along with other "legacy" code

## Type of change


finality-aleph/src/network/clique/protocols/v1/mod.rs
```diff
@@ -28,6 +27,19 @@ enum Message<D: Data> {
     Heartbeat,
 }
 
+async fn check_authorization<SK: SecretKey>(
```
timorl, line 30: Huh, what? This is part of @fixxxedpoint 's task, right? Why do I see it in this diff?

timorl (approved):
-1k lines, lets go!

finality-aleph/src/network/clique/protocols/v1/mod.rs
fixxxedpoint: moved from `v0/mod.rs`

fixxxedpoint (approved):
lgtm

finality-aleph/src/network/clique/protocols/v1/mod.rs
timorl: Ah...

PR #896, opened 2023-01-31
A0-1910 Added migration for a Scheduler

Marcin-Radecki:
# Description

PR 3/... for fixing Testetn migrations. This time custom Scheduler migration to `StorageVersion` 3 with some asserts that the `Agendas` map is empty. There's no such original migration that can be used here.

[try-runtime.log](https://github.com/Cardinal-Cryptography/aleph-node/files/10546463/try-runtime.log)

```
2023-01-31 13:48:20.596  INFO main remote-ext: downloaded data for module Scheduler (count: 1 / prefix: 3db7a24cfdc9de785974746c14a99df9).    
...
2023-01-31 13:48:23.864  INFO                 main runtime::scheduler::migration: Migrated 0 agendas, bumped StorageVersion to V3    
```

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/bin/runtime/src/migrations/custom_scheduler_migration.rs
+++ b/bin/runtime/src/migrations/custom_scheduler_migration.rs
@@ -10,7 +10,7 @@ use sp_std::vec::Vec;
 
 use crate::Weight;
 
-const TARGET: &'static str = "runtime::scheduler::migration";
+const TARGET: &str = "runtime::scheduler::migration";
 
 /// Custom migrations the scheduler pallet from V0 to V3 that only bumps StorageVersion to 3
 pub struct MigrateToV3<T>(sp_std::marker::PhantomData<T>);
--- a/bin/runtime/src/migrations/custom_scheduler_migration.rs
+++ b/bin/runtime/src/migrations/custom_scheduler_migration.rs
@@ -1,0 +1,77 @@
+use codec::{Decode, Encode};
+use frame_support::{
+    log,
+    pallet_prelude::{Get, StorageVersion},
+    traits::OnRuntimeUpgrade,
+};
+use pallet_scheduler::{Agenda, Config, Pallet};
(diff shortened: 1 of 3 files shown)


timorl (approved):
Cool (almost) noop. I'm not particularly good at runtime, but LGTM.

bin/runtime/src/migrations/custom_scheduler_migration.rs
```diff
@@ -0,0 +1,77 @@
...
+            log::warn!(
+                target: TARGET,
+                "skipping v0 to v3 migration: executed on wrong storage version.\
+				Expected version 0, found {:?}",
+                version,
+            );
+            return T::DbWeight::get().reads(1);
+        }
+
+        let agendas = Agenda::<T>::iter_keys().count() as u32;
```
fixxxedpoint, line 31: nit: no idea whether this is somehow important. Line `38` suggests this call takes one read, but skipping through code suggests it can take up to `agendas` storage reads.

fixxxedpoint (approved):
LGTM. Left one question. No idea how to test it.

bin/runtime/src/migrations/custom_scheduler_migration.rs
Marcin-Radecki: Good catch! Since `Agenda` is `StorageMap`, `iter_keys` returns an iterator that is consumed by `count()`, which calls `next()` repeatedly until `None`. I adapted code so now correct weight is returned.

fixxxedpoint: It depends on how is the `Iterator` for `StorageMap` implemented. I just looked at `some` version and it looks like it `reads` one by one (no batching). Probably number of reads for that method is something like `agendas + 1` (+1 for reading `None` for next), so line should have +1 (but I don't know :)).

PR #897, opened 2023-01-31
nomination migrations

kostekIV:
[try-runtime.log](https://github.com/Cardinal-Cryptography/aleph-node/files/10556627/try-runtime.log)

tested with `RUST_LOG=runtime=trace,try-runtime::cli=trace,executor=trace ./target/release/aleph-node try-runtime --no-spec-check-panic  --chain /tmp/chainspec.json on-runtime-upgrade live -p NominationPools -p Elections -p Staking --uri wss://ws.test.azero.dev:443 > ~/Desktop/try-runtime.log 2>&1`


bin/runtime/src/migrations/custom_nomination_pools.rs
```diff
@@ -0,0 +1,139 @@
...
+            input.read(&mut buffer)?;
+            if let Ok(new) = OldRewardPool::<B>::decode(&mut buffer.clone().as_slice()) {
+                return Ok(EitherRewardPool::Old(new));
+            }
+
+            RewardPool::<T>::decode(&mut buffer.as_slice()).map(|old| EitherRewardPool::New(old))
+        }
+    }
+
+    fn dissolve_pool<T: Config>(id: PoolId) {
```
Marcin-Radecki, line 44: We don't want to do as original `dissolve_pool` does? ie  that part " All accounts are dusted and the leftover
 of the reward account is returned to the depositor."

Now I think unfortunately there's more work needed, as original `dissolve_pool` is called in the situation the pool depositor withdraw as last one points. So we can't just remove those PoolMembers, we need to try at least to transfer their tokens back.

Marcin-Radecki (changes requested):
Due to the fact we need to make an attempt to transfer back bonded poiints

bin/runtime/src/migrations/custom_nomination_pools.rs
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        #[cfg(feature = "try-runtime")]
+        fn post_upgrade(_: Vec<u8>) -> Result<(), &'static str> {
+            // new version must be set.
+            assert_eq!(Pallet::<T>::on_chain_storage_version(), 2);
+
+            // no reward or bonded pool has been skipped.
+            assert_eq!(
+                RewardPools::<T>::iter().count() as u32,
```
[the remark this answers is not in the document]
kostekIV: this checks whether the count in the CountedStorageMap is correct after migration

bin/runtime/src/migrations/custom_nomination_pools.rs
```diff
@@ -0,0 +1,139 @@
...
+                }
+
+                log::debug!(target: "runtime::nomination-pools", "deleting member {:?}", key.encode());
+                None
+            });
+
+            log::debug!(target: "runtime::nomination-pools", "deleted pools {:?}", old_ids);
+            StorageVersion::new(2).put::<Pallet<T>>();
+
+            T::DbWeight::get().reads(1)
```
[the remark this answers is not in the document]
kostekIV: right, forget about this part

bin/runtime/src/migrations/custom_nomination_pools.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    enum EitherRewardPool<T: Config, B> {
+        Old(OldRewardPool<B>),
+        New(RewardPool<T>),
+    }
+
+    impl<T: Config, B: Decode> Decode for EitherRewardPool<T, B> {
+        fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+            let len = input.remaining_len()?.unwrap_or_default();
```
DamianStraszak, line 33: This is not how `decode` should be implemented. The idea of codec is that you decode by reading the stream until you read enough bytes. The remaining bytes in the stream likely correspond to some other objects. I understand this is only to hack this migration through, but still...

This is only used in `translate`, correct?

What is `translate` using underneath? Is it using `decode_all`?

DamianStraszak: Also you should use `decode_all` everywhere, I guess.

bin/runtime/src/migrations/custom_nomination_pools.rs
```diff
@@ -0,0 +1,139 @@
...
+    #[derive(Decode)]
+    pub struct OldRewardPool<B> {
+        pub balance: B,
+        pub total_earnings: B,
+        pub points: U256,
+    }
+
+    enum EitherRewardPool<T: Config, B> {
+        Old(OldRewardPool<B>),
+        New(RewardPool<T>),
```
DamianStraszak, line 28: Is it correct that old is `16+16+32` bytes and new is `16+16+16` bytes?

DamianStraszak (changes requested):

bin/runtime/src/migrations/custom_nomination_pools.rs
kostekIV: that is correct

bin/runtime/src/migrations/custom_nomination_pools.rs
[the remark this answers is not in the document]
kostekIV: fixed

bin/runtime/src/migrations/custom_nomination_pools.rs
kostekIV: done (as discussed offline)

bin/runtime/src/migrations/custom_nomination_pools.rs
kostekIV: This is only used in the translate, changed it to use decode_all

bin/runtime/src/migrations/custom_nomination_pools.rs
```diff
@@ -0,0 +1,171 @@
...
+            RewardPool::<T>::decode_all(&mut buffer.as_slice())
+                .map(|old| EitherRewardPool::New(old))
+        }
+    }
+
+    fn dissolve_pool<T: Config>(id: PoolId) {
+        let bonded_account = Pallet::<T>::create_bonded_account(id);
+
+        let bonded_amount = T::StakingInterface::active_stake(&bonded_account).unwrap_or_default();
+        let r = T::StakingInterface::unbond(bonded_account.clone(), bonded_amount);
```
Marcin-Radecki, line 55: Please call `Chill` before unbonding

Marcin-Radecki (approved):

bin/runtime/src/migrations/custom_nomination_pools.rs
kostekIV: done

Marcin-Radecki (approved):

bin/runtime/src/migrations/custom_nomination_pools.rs
DamianStraszak: Heh. The only reason this works is because the old format uses more bytes than the new one, AND, you are trying in the order `Old`, `New`. If you tried `New`, `Old`, then it wouldn't work :P. I just wanted to make sure we all understand this, and that this is walking on thin ice :)

kostekIV: now with decode_all in the decode impl order should not matter, right?

DamianStraszak: Correct.

DamianStraszak (approved):

PR #898, opened 2023-01-31
Merge main + migrate to ink4

obrok:
# Description

Merge main, migrating to ink4 in the process.


.github/workflows/contracts-e2e-tests-and-deploy.yaml
```diff
@@ -92,22 +90,15 @@ jobs:
         env:
           AWS_REGION: us-east-1
 
-      - name: Run one-node smartnet
+      - name: Run nodes
         shell: bash
         run: |
-          .github/scripts/run_smartnet.sh &
+          scripts/run_nodes.sh
```
deuszx, line 96: I've found that `run_nodes.sh -v 1 -n 1 -b false` works as well but it runs minimal network of 2 nodes (with 1 validator).

aleph-client/docker/subxt-integration-entrypoint.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+# an ugly workaround for the fact we want to ignore rustdoc warnings in generated runtime file
+echo "#[doc(hidden)]" > aleph_zero.rs
```
deuszx, line 4: This file seems to be different now on `main` branch https://github.com/Cardinal-Cryptography/aleph-node/blob/main/aleph-client/docker/subxt-integration-entrypoint.sh

aleph-client/src/waiting.rs
```diff
@@ -1,56 +1,196 @@
...
+    /// * `status` - a [`BlockStatus`] of the era we wait for
+    async fn wait_for_era(&self, era: EraIndex, status: BlockStatus);
+
+    /// Wait for given session to happen.
+    /// * `session` - number of the session to wait for
+    /// * `status` - a [`BlockStatus`] of the session we wait for
+    async fn wait_for_session(&self, session: SessionIndex, status: BlockStatus);
+}
+
+/// nWaiting _from_ the current moment of time API
```
deuszx, line 77:
```suggestion
/// Waiting _from_ the current moment of time API
```

bin/cliain/src/contracts.rs
```diff
@@ -58,48 +33,47 @@ fn storage_deposit(storage_deposit_limit: Option<u128>) -> Option<Compact<u128>>
...
+    });
+
+    let _block_hash = signed_connection
+        .upload_code(
+            wasm,
+            storage_deposit(storage_deposit_limit),
+            TxStatus::InBlock,
+        )
+        .await?;
+    let code_stored_event = event_handler.await?;
```
deuszx, line 68: In a separate PR, I think we should use the new [`get_txn_events`](https://github.com/Cardinal-Cryptography/aleph-node/blob/main/aleph-client/src/contract/event.rs#L55) approach here that is more resilient to accidental consumption of someoneelse's event.

deuszx: This applies to almost all methods in this file most probably.

contracts/adder/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+
+scale = { package = "parity-scale-codec", version = "3", default-features = false, features = ["derive"] }
+scale-info = { version = "2.3", default-features = false, features = ["derive"], optional = true }
+
+[lib]
+name = "adder"
+path = "lib.rs"
+crate-type = [
+	# Used for normal contract Wasm blobs.
+	"cdylib",
```
deuszx, line 19: Looks like formatting here is a bit different.

contracts/scripts/deploy.sh
```diff
@@ -248,15 +248,15 @@ cargo contract build --release
...
 cd "$CONTRACTS_PATH"/access_control
 
 ACCESS_CONTROL_CODE_HASH=$(cargo contract upload --url "$NODE" --suri "$AUTHORITY_SEED")
 ACCESS_CONTROL_CODE_HASH=$(echo "$ACCESS_CONTROL_CODE_HASH" | grep hash | tail -1 | cut -c 14-)
-ACCESS_CONTROL=$(cargo contract instantiate --url "$NODE" --constructor new --suri "$AUTHORITY_SEED" --skip-confirm)
-ACCESS_CONTROL=$(echo "$ACCESS_CONTROL" | grep Contract | tail -1 | cut -c 14-)
-ACCESS_CONTROL_PUBKEY=$(docker run --rm --entrypoint "/bin/sh" "${NODE_IMAGE}" -c "aleph-node key inspect $ACCESS_CONTROL" | grep hex | cut -c 23- | cut -c 3-)
+ACCESS_CONTROL=$(cargo contract instantiate --url "$NODE" --constructor new --suri "$AUTHORITY_SEED" --skip-confirm --output-json)
+ACCESS_CONTROL=$(echo "$ACCESS_CONTROL" | jq -r '.contract')
+ACCESS_CONTROL_PUBKEY=$("$ROOT"/target/release/aleph-node key inspect $ACCESS_CONTROL | grep hex | cut -c 23- | cut -c 3-)
```
deuszx, line 259: Just FYI @fbielejec , I dropped the usage of docker here and use node binary directly.

contracts/scripts/test.sh
```diff
@@ -4,32 +4,21 @@ set -euo pipefail
...
+  SIMPLE_DEX=$(jq --raw-output ".simple_dex" < "$CONTRACTS_PATH"/addresses.json) \
+  WRAPPED_AZERO=$(jq --raw-output ".wrapped_azero" < "$CONTRACTS_PATH"/addresses.json) \
+  BUTTON_GAME_METADATA=$CONTRACTS_PATH/button/target/ink/metadata.json \
+  TICKET_TOKEN_METADATA=$CONTRACTS_PATH/ticket_token/target/ink/metadata.json \
+  REWARD_TOKEN_METADATA=$CONTRACTS_PATH/game_token/target/ink/metadata.json \
+  MARKETPLACE_METADATA=$CONTRACTS_PATH/marketplace/target/ink/metadata.json \
+  SIMPLE_DEX_METADATA=$CONTRACTS_PATH/simple_dex/target/ink/metadata.json \
+  WRAPPED_AZERO_METADATA=$CONTRACTS_PATH/wrapped_azero/target/ink/metadata.json \
+  RUST_LOG="aleph_e2e_client=info" \
+  cargo test button -- --test-threads 1 --nocapture
```
deuszx, line 22: I think we need to drop the `button` and go for the 
```suggestion
  cargo test -- --test-threads 1 --nocapture
```
otherwise we won't test `simple_dex` at the very least.

contracts/wrapped_azero/lib.rs
```diff
@@ -133,23 +128,21 @@ pub mod wrapped_azero {
...
-                    instance.metadata.name = Some("wAzero".into());
-                    instance.metadata.symbol = Some("wA0".into());
-                    instance.metadata.decimals = 12; // same as AZERO
-
-                    instance.access_control = AccountId::from(ACCESS_CONTROL_PUBKEY);
-                }),
-                Err(why) => panic!("Could not initialize the contract {:?}", why),
+            let access_control = AccountId::from(ACCESS_CONTROL_PUBKEY);
+            let access_control = AccessControlRef::from_account_id(access_control);
+            if access_control.has_role(caller, Role::Initializer(code_hash)) {
```
deuszx, line 133: [nit] I think it's already possible to return `Result<Self>` from the constructor as well so here, we could do:
```rust
self.check_role(caller, Role::Initializer(code_hash))?;
let mut metadata = metadata::Data::default();
metadata.name = Some("wAzero".into());
metadata.symbol = Some("wA0".into());
metadata.decimals = 12; // same as AZERO

Self {
    psp22: psp22::Data::default(),
    metadata,
    access_control,
}
```

contracts/adder/lib.rs
```diff
@@ -0,0 +1,65 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+//! This is a simple example contract for use with e2e tests of the aleph-client contract interaction.
+
+#[ink::contract]
+mod adder {
+    #[ink(storage)]
+    pub struct Adder {
+        name: Option<[u8; 20]>,
```
deuszx, line 9: [nit]: There's a `String` [type in ink](https://docs.rs/ink_prelude/latest/ink_prelude/string/struct.String.html). Your choice.

contracts/adder/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+            Ok(())
+        }
+
+        #[ink(message)]
+        pub fn get(&self) -> u32 {
+            self.value
+        }
+
+        #[ink(message)]
+        pub fn set_name(&mut self, name: Option<[u8; 20]>) {
```
deuszx, line 50: [nit]: I know it's an examplary contract but I would expect a `set` method to take non-optional values.

contracts/button/errors.rs
```diff
@@ -21,6 +25,10 @@ pub enum GameError {
     CantRetrieveOwnCodeHash,
     /// Overflow error
     Arithmethic,
+    /// Error from the marketplace contract
+    MarketplaceError(MarketplaceError),
+    /// Error while calling another contract
+    ContractCall(LangError),
```
deuszx, line 31: Just FYI, This is removed in the next release of ink (release candidate) as `LangError`s are not expected to be recoverable so the defaults will be to panic - https://github.com/paritytech/ink/pull/1512

deuszx (approved):
Only [blocking comment](https://github.com/Cardinal-Cryptography/aleph-node/pull/898#discussion_r1092539511) from me, otherwise LGTM. Awesome work.

deuszx (dismissed):
One [blocking comment](https://github.com/Cardinal-Cryptography/aleph-node/pull/898#discussion_r1092539511).

contracts/scripts/deploy.sh
fbielejec: but why :fearful:

deuszx: No dependency on the docker here anymore and you have to compile `aleph-node` anyway to run these tests and the docker image wasn't available for me locally anyway. 

Although, now that I think of it, you're calling `deploy.sh` in contexts where there might be no `aleph-node` binary available 🤔 I'll be fine with reverting this change and having local version of it for myself.

.github/workflows/contracts-e2e-tests-and-deploy.yaml
obrok: Well, this doesn't work on CI, so I'll have to figure something else out tomorrow... I'm not really working today, just listening to meetings :D

contracts/scripts/test.sh
obrok:
```
yapee@skadi:~/Projects/aleph-node/e2e-tests$ cargo test button -- --list 

test::button_game::back_to_the_future: test
test::button_game::button_game_reset: test
test::button_game::early_bird_special: test
test::button_game::marketplace: test
test::button_game::simple_dex: test
test::button_game::the_pressiah_cometh: test
test::button_game::wrapped_azero: test
```

I don't think so - the filter is applied to the whole test name, and all of these are in `button_game`, so they are all included.

contracts/button/errors.rs
obrok: Yeah, I know, we'll deal with that if/when we migrate to that release.

contracts/adder/lib.rs
obrok: Why? The type of this value is `Option<_>` so if you want to reset it back to the default of `None`, then you need `set` to take an `Option` as well.

contracts/adder/lib.rs
obrok: But that one won't have constant-sized storage, will it? Can it even be used as args/return values of contract methods because of that?

bin/cliain/src/contracts.rs
obrok: This is merged code, not interested in changing any of this.

contracts/adder/Cargo.toml
obrok: Merged code, same.

aleph-client/src/waiting.rs
obrok: Merged code, same.

aleph-client/src/waiting.rs
```diff
@@ -1,56 +1,196 @@
...
+                let event = event.expect("Failed to obtain event from the block!");
+                if let Ok(Some(ev)) = event.as_event::<T>() {
+                    if predicate(&ev) {
+                        return ev;
+                    }
+                }
+            }
         }
+
+        panic!("No more blocks");
```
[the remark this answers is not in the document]
obrok: I have no opinion on this, maybe follow up with the person who added this line?

aleph-client/docker/subxt-integration-entrypoint.sh
obrok: Well, this merges `main` at some particular point in time, so it might have gotten removed since.

contracts/scripts/test.sh
deuszx: Oh, right. I forgot there's `button_game` as part of the test path.

.github/workflows/contracts-e2e-tests-and-deploy.yaml
deuszx: The main difference being that `run_smartnet.sh` runs a docker network while `run_nodes.sh` plain binaries.

deuszx (approved):
Changes to the contract-related code (and required changes to client) look good.

.github/workflows/contracts-e2e-tests-and-deploy.yaml
obrok: This is moot now, I rolled the contract tests into the main `e2e` workflow.

fbielejec (approved):

PR #899, opened 2023-02-01
A0-1910 Added custom transaction payment migration

Marcin-Radecki:
# Description

Last PR for fixing Testniet Migration. This time simple bump of custom `StorageVersion` to `V2` from `None`. 

[try-runtime.log](https://github.com/Cardinal-Cryptography/aleph-node/files/10556683/try-runtime.log)

```
2023-02-01 10:36:07.075  INFO main remote-ext: downloaded data for module TransactionPayment (count: 2 / prefix: 3f1467a096bcd71a5b6a0c8155e20810).    
....
2023-02-01 10:37:04.424  INFO                 main runtime::transaction_payment::migration: NextFeeMultiplier remains the same. Bumped StorageVersion to V2    
```

## Type of change


bin/runtime/src/migrations/custom_transaction_payment_migration.rs
```diff
@@ -0,0 +1,84 @@
...
+    V2,
+}
+
+/// Custom migrations the transaction payment pallet from V1 to V2 that only bumps StorageVersion to 2
+pub struct BumpTransactionVersionToV2<T>(sp_std::marker::PhantomData<T>);
+
+impl<T: Config> OnRuntimeUpgrade for BumpTransactionVersionToV2<T> {
+    fn on_runtime_upgrade() -> Weight {
+        let version = StorageVersion::get();
+        if version.is_some() && version.unwrap() == Releases::V2 {
```
deuszx, line 35: nit: this
```rust
let version = StorageVersion::get();
if version == Some(Releases::V2) { ... }
```
should work too

bin/runtime/src/migrations/custom_transaction_payment_migration.rs
```diff
@@ -0,0 +1,84 @@
...
+            log::warn!(
+                target: TARGET,
+                "skipping V1Ancient to V2 migration: executed on wrong storage version.\
+				Expected version V1Ancient, found {:?}",
```
deuszx, line 39: We can be confident here (since we just checked it in the `if`) that the `version` is `Releases::V2`.

bin/runtime/src/migrations/custom_transaction_payment_migration.rs
```diff
@@ -0,0 +1,84 @@
...
+        StorageVersion::put(Releases::V2);
+        T::DbWeight::get().reads_writes(1, 1)
+    }
+
+    #[cfg(feature = "try-runtime")]
+    fn pre_upgrade() -> Result<Vec<u8>, &'static str> {
+        assert_eq!(
+            StorageVersion::get(),
+            None,
+            "Can only upgrade from version V1Ancient"
```
deuszx, line 54: Pardon my ignorance - but if `StorageVersion::get()` is `Some(Releases::V2)` already, won't this `assert_eq!` panic?

deuszx: Ah, this is feature-gated so wouldn't happen on production chain. Still, is this some kind of "standard" when it comes to writing these things?

bin/runtime/src/migrations/custom_transaction_payment_migration.rs
Marcin-Radecki: good point changed

maciejnems (approved):

bin/runtime/src/migrations/custom_transaction_payment_migration.rs
Marcin-Radecki: Right, done

bin/runtime/src/migrations/custom_transaction_payment_migration.rs
Marcin-Radecki: Yes, the flow is as follows:
1. `pre_upgrade` and `post_upgrade` are for developer and revievers to test changes using Testnet data.
2. During real Testnet chain update, those hooks are not called, onlt `on_runtime_upgrade`.

1, and 2. can happen in a different time, even few weeks apart, and on different chains, so there are not really coupled. Therefore we need to check in 2. their own checks just to be sure

deuszx (approved):

PR #900, opened 2023-02-01
A0-1916: Proper tangling and path verification

pmikolajczyk41:
# Description

This PR fixes two things:
 - tangling
 - merkle path checking

# Story

The root problem was with in the tangling implementation. Two things here.
1. Tangling was really vulnerable to `0` in its input - once there were sufficiently many zero entries, the result was zero as well.
1. We were working on `ByteVar` (`Uint8<CircuitField>`) type, which is completely improper for arithmetic in a circuit. Although it doesn't expose calculation API (like `+` or `*`), it is not idiot-proof. We can still retrieve its underlying `u8` value with `.value()` method. **However, any operation on these values, doesn't have any impact on the circuit itself**. Unfortunately, this problem was hidden by the former point and thus we couldn't learn that something is no yes.

Similar bug was also present in the Merkle path checking, but again, it wasn't visible since, everything worked well with zeroed notes. In particular, `leaf_index` was perfectly broken.

# Fix for tangling

Firstly, we give up working on bytes. Although `arkworks` exposes API for converting `ByteVar`/`Vec<ByteVar>` to `Vec<CircuitField>` and back (with `ToBytesGadget::to_bytes`), **it is not homomorphic** with the conversion `CircuitField::to_bytes` (via underlying `BigInteger256`). And actually we need such homomorphism, since we are tangling both outside and inside circuit.

Secondly, in critical moments, we replace `0` inputs with an index-dependent constant value (see implementation).

# Fix for Merkle path checking

Initially we had just a leaf index and depending on its bits we restored path shape. However, **this logic must be baked into a circuit**. For this, `Boolean` type has `select` method, which together with the `CondSelectGadget` trait enables conditional execution. In order to make things a bit cleaner, we introduced `PathShapeVar` (that represents leaf index explicitly as circuit bits).

### DIFF
--- a/relations/src/shielder/circuit_utils.rs
+++ b/relations/src/shielder/circuit_utils.rs
@@ -1,6 +1,7 @@
 use core::borrow::Borrow;
 #[cfg(feature = "std")]
 use std::fmt::{Display, Formatter};
+use std::ops::Index;
 
 use ark_r1cs_std::{
     alloc::{AllocVar, AllocationMode},
@@ -34,9 +35,13 @@ impl PathShapeVar {
     pub(super) fn len(&self) -> usize {
         self.shape.len()
     }
+}
 
-    pub(super) fn at(&self, i: usize) -> &Boolean<CircuitField> {
-        &self.shape[i]
+impl Index<usize> for PathShapeVar {
+    type Output = Boolean<CircuitField>;
+
+    fn index(&self, index: usize) -> &Self::Output {
+        &self.shape[index]
     }
 }
 
--- a/relations/src/shielder/circuit_utils.rs
+++ b/relations/src/shielder/circuit_utils.rs
@@ -7,10 +7,7 @@ use ark_r1cs_std::{
     boolean::Boolean,
     R1CSVar,
 };
-use ark_relations::{
-    ns,
-    r1cs::{ConstraintSystemRef, Namespace, SynthesisError, SynthesisError::AssignmentMissing},
-};
+use ark_relations::r1cs::{Namespace, SynthesisError};
 
 use crate::CircuitField;
 
--- a/relations/src/shielder/circuit_utils.rs
+++ b/relations/src/shielder/circuit_utils.rs
@@ -1,0 +1,72 @@
+use core::borrow::Borrow;
+#[cfg(feature = "std")]
+use std::fmt::{Display, Formatter};
+
+use ark_r1cs_std::{
+    alloc::{AllocVar, AllocationMode},
+    boolean::Boolean,
+    R1CSVar,
+};
+use ark_relations::{
+    ns,
+    r1cs::{ConstraintSystemRef, Namespace, SynthesisError, SynthesisError::AssignmentMissing},
+};
+
+use crate::CircuitField;
+
+#[derive(Clone, Debug)]
+pub(super) struct PathShapeVar {
+    shape: Vec<Boolean<CircuitField>>,
+}
+
+#[cfg(feature = "std")]
+impl Display for PathShapeVar {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{:?}",
+            self.shape
+                .iter()
+                .map(|b| b.value().map(|boo| if boo { "left" } else { "right" }))
+                .collect::<Vec<_>>()
+        )
+    }
+}
+
+impl PathShapeVar {
+    pub(super) fn len(&self) -> usize {
+        self.shape.len()
+    }
+
+    pub(super) fn at(&self, i: usize) -> &Boolean<CircuitField> {
+        &self.shape[i]
+    }
+}
+
+impl AllocVar<(u8, Result<u64, SynthesisError>), CircuitField> for PathShapeVar {
+    fn new_variable<T: Borrow<(u8, Result<u64, SynthesisError>)>>(
+        cs: impl Into<Namespace<CircuitField>>,
+        f: impl FnOnce() -> Result<T, SynthesisError>,
+        mode: AllocationMode,
+    ) -> Result<Self, SynthesisError> {
+        let ns = cs.into();
+        let cs = ns.cs();
+
+        let mut shape = vec![];
+
+        let (path_length, maybe_leaf_index) = *f()?.borrow();
+
+        for i in 0..path_length {
+            shape.push(Boolean::new_variable(
+                cs.clone(),
+                || {
+                    let current_index = maybe_leaf_index? / (1 << i);
+                    Ok(current_index & 1 != 1 || current_index == 1)
+                },
+                mode,
+            )?);
+        }
+
+        Ok(Self { shape })
+    }
+}
--- a/relations/src/shielder/mod.rs
+++ b/relations/src/shielder/mod.rs
@@ -40,7 +40,7 @@ pub use withdraw::{
 use crate::environment::{CircuitField, FpVar};
 
 fn convert_hash(front: [u64; 4]) -> CircuitField {
-    CircuitField::from(BigInteger256::new(front))
+    CircuitField::new(BigInteger256::new(front))
 }
 
 fn convert_vec(front: Vec<[u64; 4]>) -> Vec<CircuitField> {
--- a/relations/src/shielder/mod.rs
+++ b/relations/src/shielder/mod.rs
@@ -12,7 +12,7 @@ pub mod types;
 mod withdraw;
 
 use ark_ff::{BigInteger256, PrimeField, Zero};
-use ark_r1cs_std::{alloc::AllocVar, eq::EqGadget};
+use ark_r1cs_std::{alloc::AllocVar, eq::EqGadget, R1CSVar};
 use ark_relations::{
     ns,
     r1cs::{ConstraintSystemRef, SynthesisError, SynthesisError::UnconstrainedVariable},
@@ -84,6 +84,11 @@ fn check_merkle_proof(
 
         current_note = tangle_in_field(&next_level)?;
         leaf_index /= 2;
+    }
+
+    if !cs.is_in_setup_mode() {
+        println!("merkle {}", merkle_root.value().unwrap());
+        println!("curren {}", current_note.value().unwrap());
     }
 
     merkle_root.enforce_equal(&current_note)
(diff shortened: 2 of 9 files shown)


relations/src/shielder/tangle.rs
```diff
@@ -1,67 +1,81 @@
...
+) -> Result<FpVar, SynthesisError> {
+    let fallback = FpVar::constant(fallback.into());
+    fp.is_zero()?.select(&fallback, fp)
 }
 
 /// Recursive and index-bounded implementation of the first step of the `tangle` procedure.
-fn _tangle_in_field(bytes: &mut [ByteVar], low: usize, high: usize) -> Result<(), SynthesisError> {
-    // Bottom level case: computing suffix sums. We have to do some loop-index boilerplate, because
-    // Rust doesn't support decreasing range iteration.
+fn _tangle_in_circuit(elems: &mut [FpVar], low: usize, high: usize) -> Result<(), SynthesisError> {
```
mike1729, line 67: I assume that the underscore at the beginning comes from python convention but in rust this naming is reserved for sth else.  Maybe sth like `tangle_in_circuit_step` or  `tangle_in_circuit_rec`?

relations/src/shielder/tangle.rs
```diff
@@ -1,67 +1,81 @@
...
-                u8::overflowing_add(
-                    bytes[i].value().unwrap_or_default(),
-                    bytes[i + 1].value().unwrap_or_default(),
-                )
-                .0,
-            );
+            let previous = dezeroize_in_circuit(&elems[i + 1], (2 * low + 1) as u64)?;
+            let current = dezeroize_in_circuit(&elems[i], (2 * high + 1) as u64)?;
+
+            elems[i] = (previous.inverse()? + current.inverse()?)
```
mike1729, line 76: Use `1/a + 1/b = (a+b)/ab` as this reduces 2 costly inverses to 1.  Also, if you `dezeroize_in_circuit` here if `a+b` then you don't need to `dezeroize_in_circuit` later as this is a prime field ;)

mike1729 (changes requested):

relations/src/shielder/note.rs
```diff
@@ -49,24 +42,18 @@ pub fn compute_note(
...
-
-    note_from_bytes(tangle::<4>(bytes).as_slice())
+    tangle(&[
+        CircuitField::from(token_id as u64),
+        CircuitField::from(token_amount),
+        CircuitField::from(trapdoor),
+        CircuitField::from(nullifier),
+    ])
+    .0
+     .0
```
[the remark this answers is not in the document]
pmikolajczyk41: default formatter :man_shrugging:

relations/src/shielder/note.rs
```diff
@@ -49,24 +42,18 @@ pub fn compute_note(
...
 }
 
 pub fn compute_parent_hash(left: FrontendNote, right: FrontendNote) -> FrontendNote {
-    let bytes = [
-        BigInteger256::new(left).to_bytes_le(),
-        BigInteger256::new(right).to_bytes_le(),
-    ]
-    .concat();
-    note_from_bytes(tangle::<2>(bytes).as_slice())
+    tangle(&[convert_hash(left), convert_hash(right)]).0 .0
```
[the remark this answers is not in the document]
pmikolajczyk41: again, default formatter :shrug:

relations/src/shielder/tangle.rs
pmikolajczyk41: changed to `do_...`

relations/src/shielder/tangle.rs
pmikolajczyk41: fair point, but I chose to remove inverting `previous` - now it is just suffix sums of inverses (with some const scaling); when I started reworking towards `(a+b)/ab` the code looked too much complicated for its quality xD

relations/src/shielder/tangle.rs
```diff
@@ -73,8 +77,8 @@ fn _tangle_in_circuit(elems: &mut [FpVar], low: usize, high: usize) -> Result<()
             let previous = dezeroize_in_circuit(&elems[i + 1], (2 * low + 1) as u64)?;
             let current = dezeroize_in_circuit(&elems[i], (2 * high + 1) as u64)?;
 
-            elems[i] = (previous.inverse()? + current.inverse()?)
-                * FpVar::constant(CircuitField::from(i as u64));
+            elems[i] =
+                (previous + current.inverse()?) * FpVar::constant(CircuitField::from(i as u64));
```
mike1729, line 81: Please update the module documentation.

mike1729 (approved):

relations/src/shielder/tangle.rs
pmikolajczyk41: done

relations/src/shielder/circuit_utils.rs
```diff
@@ -0,0 +1,69 @@
...
+impl AllocVar<(u8, Result<u64, SynthesisError>), CircuitField> for PathShapeVar {
+    fn new_variable<T: Borrow<(u8, Result<u64, SynthesisError>)>>(
+        cs: impl Into<Namespace<CircuitField>>,
+        f: impl FnOnce() -> Result<T, SynthesisError>,
+        mode: AllocationMode,
+    ) -> Result<Self, SynthesisError> {
+        let ns = cs.into();
+        let cs = ns.cs();
+
+        let mut shape = vec![];
```
deuszx, line 52: Would it make perf* different if we pre-allocated this vector upfront? We know its length and size of the elements is constant.

pmikolajczyk41: I don't think so, especially that usually path length will be < 20, so this should be effective anyway

deuszx: I thought that pushing to a vector re-allocates and re-writes but that's not the case. It's the `unsafe` code that just inc the length and moves one pointer. https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#1831 👍

relations/src/shielder/tangle.rs
```diff
@@ -1,67 +1,85 @@
...
-                .cloned()
-                .reduce(|x, y| x.xor(&y).unwrap())
-                .unwrap()
-        })
-        .collect())
+pub(super) fn tangle_in_circuit(input: &[FpVar]) -> Result<FpVar, SynthesisError> {
+    let mut input_expanded = input
+        .iter()
+        .cycle()
+        .take(EXPAND_TO)
```
deuszx, line 49: So this method works only on 128 elements? Anything above them will get thrown away?

pmikolajczyk41: yes, but if you are trying to make some tangling on >128 elements in your circuit, than it might be infeasible and should definitely be changed

relations/src/shielder/tangle.rs
```diff
@@ -123,15 +150,64 @@ fn _tangle(bytes: &mut [u8], low: usize, high: usize) {
...
+    };
+
+    #[test]
+    fn tangling_is_homomorphic() {
+        let input = vec![
+            CircuitField::from(0u64),
+            CircuitField::from(100u64),
+            CircuitField::from(17u64),
+            CircuitField::from(19u64),
+        ];
```
deuszx, line 185: Out of curiosity - is there anything specific in this example that makes you confident that tangling is homomorphic?

relations/src/shielder/mod.rs
```diff
@@ -55,53 +56,33 @@ fn convert_account(front: [u8; 32]) -> CircuitField {
...
-
-    let mut current_hash_bytes = leaf_bytes;
-    let mut hash_bytes = vec![current_hash_bytes.clone()];
+    let mut current_note = leaf;
+    let zero_note = CircuitField::zero();
 
     for i in 0..max_path_len {
         let sibling = FpVar::new_witness(ns!(cs, "merkle path node"), || {
-            Ok(path.get(i as usize).unwrap_or(&zero))
+            Ok(path.get(i as usize).unwrap_or(&zero_note))
```
deuszx, line 78: Since we use `i as usize` everywhere, you could cast in the `for` loop above 
```rust
for i in 0..max_path_len as usize
```

relations/src/shielder/circuit_utils.rs
```diff
@@ -0,0 +1,69 @@
...
+        )
+    }
+}
+
+impl PathShapeVar {
+    pub(super) fn len(&self) -> usize {
+        self.shape.len()
+    }
+
+    pub(super) fn at(&self, i: usize) -> &Boolean<CircuitField> {
```
deuszx, line 38: nit: one could implement [`Index`](https://doc.rust-lang.org/std/ops/trait.Index.html) and get the `[m]` access to `PathShapeVar` elements.

relations/src/shielder/circuit_utils.rs
```diff
@@ -0,0 +1,69 @@
...
+
+        let mut shape = vec![];
+
+        let (path_length, maybe_leaf_index) = *f()?.borrow();
+
+        for i in 0..path_length {
+            shape.push(Boolean::new_variable(
+                cs.clone(),
+                || {
+                    let current_index = maybe_leaf_index? / (1 << i);
```
deuszx, line 60: Why does the `maybe_leaf_index` have to be unwrapped inside the `f` of `Boolean_new_variable` and cannot be unwrapped immediately in line 54?

deuszx (commented):
Pretty amazing.

relations/src/shielder/circuit_utils.rs
pmikolajczyk41: yes, that's the tricky part :) both for generating keys (`RelationWithoutInput`) and for generating proof (`RelationWithFullInput`) we know `max_path_length`; however, only during proof generation we actually need `leaf_index` **and only during proof generation this closure (for `new_variable`) will be actually triggered**

relations/src/shielder/mod.rs
pmikolajczyk41: done

relations/src/shielder/circuit_utils.rs
pmikolajczyk41: done

relations/src/shielder/circuit_utils.rs
deuszx: I see - so the circuit is the same for generating keys and proof but only in the second case the `leaf_index` is known and used.

pmikolajczyk41: exactly

deuszx (approved):

PR #902, opened 2023-02-03
A0-1797: Main logic for justification sync

timorl:
# Description

Part one of the justification sync service.

## Type of change

# Checklist:

- I have added tests


finality-aleph/src/sync/handler.rs
```diff
@@ -0,0 +1,486 @@
...
+        forest::{Error as ForestError, Forest, JustificationAddResult},
+        BlockIdFor, BlockIdentifier, ChainStatus, Finalizer, Header, Justification, PeerId,
+        Verifier,
+    },
+};
+
+/// How many justifications we will send at most in response to a broadcast.
+const MAX_SMALL_JUSTIFICATION_BATCH: u32 = 10;
+// Silly workaround to make range matching actually work...
+const MAX_SMALL_JUSTIFICATION_BATCH_PLUS_ONE: u32 = MAX_SMALL_JUSTIFICATION_BATCH + 1;
```
lesniak43, line 14: really?...

lesniak43 (approved):
Nice.

finality-aleph/src/sync/handler.rs
timorl: Oh believe me, I'm very open to improvement suggestions. Dx

lesniak43: I mean, maybe don't make this a const, but rather create a local variable right before the match?

lesniak43: oh, it does not work...

lesniak43: ok, I'm not sure if this qualifies as "improvement", but I'd suggest:
1. nested `match`, outer branch matches `Some(n)`
2. checked `n -= 1`
3. use non inclusive ranges: `0`, `1..MAX`, and `MAX..`

lesniak43: I mean it's ugly, but we don't have to worry that someone forgets to update the other const in the future, and since that particular number was chosen pseudorandomly, then it might be necessary to tune it later...

lesniak43: hmm, then again, the compiler will complain

lesniak43: it's fine, closing...

finality-aleph/src/sync/handler.rs
```diff
@@ -0,0 +1,486 @@
...
+        let mut result = Vec::with_capacity(MAX_SMALL_JUSTIFICATION_BATCH as usize);
+        let mut number = id.number() + 1;
+        while result.len() < MAX_SMALL_JUSTIFICATION_BATCH as usize {
+            match self
+                .chain_status
+                .finalized_at(number)
+                .map_err(Error::ChainStatus)?
+            {
+                Some(justification) => result.push(justification.into_unverified()),
+                None => break,
```
maciejnems, line 149: Is there a reason not to send last block of this session here if it exists? Seems to me like if we did that here, we could just make one function `justification_batch_for` which would take batch size as an argument.

maciejnems (commented):
Great PR. I'm just not sure about some things

finality-aleph/src/sync/handler.rs
```diff
@@ -0,0 +1,486 @@
...
+            .id()
+            .number()
+            .checked_sub(remote_top.header().id().number())
+        {
+            // If we are just one behind then normal broadcasts should remedy the situation.
+            Some(0..=1) => Ok(SyncActions::noop()),
+            Some(2..=MAX_SMALL_JUSTIFICATION_BATCH) => Ok(SyncActions::response(
+                self.small_justification_batch_from(remote_top.header().id())?,
+            )),
+            Some(MAX_SMALL_JUSTIFICATION_BATCH_PLUS_ONE..) => Ok(SyncActions::response(
```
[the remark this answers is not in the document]
lesniak43: hmm, seems like a good idea, don't remember why we decided to send only the top justification (maybe this was just to simplify the algorithm, not considering efficiency)

finality-aleph/src/sync/handler.rs
```diff
@@ -0,0 +1,486 @@
...
+        Ok(NetworkData::Justifications(result, self.state()?))
+    }
+
+    fn top_understandable_for(
+        &self,
+        id: BlockIdFor<J>,
+    ) -> Result<NetworkData<J>, Error<J, CS, V, F>> {
+        use Error::*;
+        let block_session = session_id_from_block_num(id.number(), self.period);
+        let understandable_session = SessionId(block_session.0 + 1);
```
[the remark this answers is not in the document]
timorl: The idea is that after receiving this justification the peer will send an explicit request for many justifications and actually receive everything in a response. The justification itself is just as high as possible to ensure they will be convinced that they should do that.

finality-aleph/src/sync/handler.rs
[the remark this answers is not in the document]
timorl: We don't want to send a lot of justifications without an explicit request, sine they might be of nontrivial size.

finality-aleph/src/sync/handler.rs
timorl: Hmm, for short fresh batches this means that we are above the top finalized block, as they have justifications for all the blocks. So this would never send the last block in the sense you are implying above. It is also a much simpler break condition than in the other method.

After looking at this for a while I agree that the methods could be unified into one (the more complicated one), with the behaviour remaining identical in non-bugged situations. I am not sure this is that helpful though – it seems to me that for readability having the usual situation be represented by a much simpler method, and only have weird edge cases (needed only for ancient sessions) in the more rarely used one is better. But I'm not very convinced either way – what do you two think? cc @lesniak43

finality-aleph/src/sync/handler.rs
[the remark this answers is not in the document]
maciejnems: But we do that in case of a small justification batch. Why not do the same here with a small number of justifications?

finality-aleph/src/sync/handler.rs
```diff
@@ -0,0 +1,486 @@
...
+        Ok(Handler {
+            chain_status,
+            verifier,
+            finalizer,
+            forest,
+            period,
+        })
+    }
+
+    fn large_justification_batch_from(
```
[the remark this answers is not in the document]
timorl: The reasoning behind the naming is that `top_understandable_for`(someone who has the perspective of this block) and `*justification_batch_(starting_)from`. I'm not sure what `for` would stand for in your proposal, so if we wanted to unify this I would rather go for changing to `top_understandable_from`(the perspective of this block). WDYT?

Technically we could, but all the uses get a value anyway (from the `id` method of `Header`), so this wouldn't really help with anything.

finality-aleph/src/sync/handler.rs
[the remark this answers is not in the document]
lesniak43: Yeah, but I understand that @maciejnems suggests sending the same set of justifications as in the `2..=MAX_SMALL_JUSTIFICATION_BATCH` case, plus one additional - this makes sense since now the peer expects at most `MAX_SMALL_JUSTIFICATION_BATCH` justifications in response to the state broadcast, so `+1` is not that much of a difference?
On the other hand, it's 2 messages faster, but may complicate the code, and the usual case is a node that's very far behind, so this will probably not make a difference? Hmm, the question is how often a node will lag slightly more than 10 justifications behind.

finality-aleph/src/sync/handler.rs
maciejnems: yes, in typical situation it would not send the justification, and if we used the logic of a more complicated method, it would just ended the loop, as last block of session would be over top finalized. 

I just feel, like there could be some strange state we did not expect and in this case our node would not send the justifications. That said, it would probably quickly fix itself, as soon as gap between two nodes is big enough.

So I guess we could leave two different functions here.

finality-aleph/src/sync/handler.rs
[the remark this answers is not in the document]
timorl: Ah, I see the argument. IMO we should assume here that the difference is much larger than `MAX_SMALL_JUSTIFICATION_BATCH`, so we don't get much out of sending these justifications, while the code gets much more complex. If it turns out that we often are _just_ above this threshold then it would be much better to just increase the threshold.

finality-aleph/src/sync/handler.rs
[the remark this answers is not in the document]
maciejnems: Ok, `from` and taking value sounded confusing to me as it looked like a constructor. If it stands for "starting from" it makes sense to me. Feel free to keep it as it is.

As of getting the value, now you have to pass/clone it, and after that it clones inside again. So potentially 2 clones instead of one. This is something that I thought of lately when looking at sync code, that in a lot of places we take value, where instead we could just take reference.

maciejnems (approved):
LGTM

PR #904, opened 2023-02-07
staking

kostekIV:
Simply set StorageVersion for pallet staking. Since we copied the 
release struct the storage returns None not the default from the staking pallet.

`RUST_LOG=runtime=trace,try-runtime::cli=trace,executor=trace ./target/release/aleph-node try-runtime --no-spec-name-check  --chain /tmp/chainspec.json on-runtime-upgrade live -p NominationPools -p Elections -p Staking -p Aleph --uri wss://ws.azero.dev:443 > ~/Desktop/try-runtime.log 2>&1
`
[try-runtime.log](https://github.com/Cardinal-Cryptography/aleph-node/files/10673715/try-runtime.log)

### DIFF
--- a/bin/runtime/src/migrations/staking.rs
+++ b/bin/runtime/src/migrations/staking.rs
@@ -54,7 +54,7 @@ impl<T: Config> OnRuntimeUpgrade for BumpStorageVersionFromV7ToV10<T> {
                     "💸 Migration being executed on the wrong storage \
                     version, expected Releases::V10_0_0 or None"
                 );
-                return T::DbWeight::get().reads(1);
+                T::DbWeight::get().reads(1)
             }
         }
     }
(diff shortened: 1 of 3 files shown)


bin/runtime/src/migrations/staking.rs
```diff
@@ -0,0 +1,82 @@
...
+                );
+                return T::DbWeight::get().reads(1);
+            }
+        }
+    }
+
+    #[cfg(feature = "try-runtime")]
+    fn pre_upgrade() -> Result<(), &'static str> {
+        frame_support::ensure!(
+            StorageVersion::get() == Some(Releases::V7_0_0) || StorageVersion::get() == None,
```
DamianStraszak, line 65: For consistency this should be also `V10_0_0` I guess.

DamianStraszak (approved):

bin/runtime/src/migrations/staking.rs
kostekIV: done

DamianStraszak (approved):

PR #905, opened 2023-02-07
Scheduler migration

kostekIV:
[try-runtime.log](https://github.com/Cardinal-Cryptography/aleph-node/files/10673905/try-runtime.log)

### DIFF
--- a/bin/runtime/src/migrations/mod.rs
+++ b/bin/runtime/src/migrations/mod.rs
@@ -1,4 +1,3 @@
-
 mod scheduler;
 mod staking;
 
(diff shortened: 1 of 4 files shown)


bin/runtime/src/migrations/mod.rs
```diff
@@ -0,0 +1,5 @@
+mod scheduler;
+mod staking;
+
+pub use scheduler::MigrateToV3 as SchedulerMigrateToV3;
+pub use staking::BumpStorageVersionFromV7ToV10;
```
DamianStraszak, line 5: I guess you can try to make the naming consistent. The approach for `Scheduler` is good.

DamianStraszak (approved):

bin/runtime/src/migrations/mod.rs
kostekIV: changed it in this pr

PR #906, opened 2023-02-07
History migration

kostekIV:
[try-runtime.log](https://github.com/Cardinal-Cryptography/aleph-node/files/10674180/try-runtime.log)

### DIFF
--- a/bin/runtime/src/migrations/history.rs
+++ b/bin/runtime/src/migrations/history.rs
@@ -15,7 +15,7 @@ impl<T: Config> OnRuntimeUpgrade for MigrateToV1<T> {
     #[cfg(feature = "try-runtime")]
     fn pre_upgrade() -> Result<(), &'static str> {
         frame_support::ensure!(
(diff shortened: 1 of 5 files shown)


bin/runtime/src/migrations/history.rs
```diff
@@ -0,0 +1,35 @@
...
+    fn on_runtime_upgrade() -> Weight {
+        StorageVersion::new(1).put::<Pallet<T>>();
+        log::info!(target: "runtime::historical", "migrated to V1");
+        T::DbWeight::get().reads_writes(0, 1)
+    }
+
+    #[cfg(feature = "try-runtime")]
+    fn pre_upgrade() -> Result<(), &'static str> {
+        frame_support::ensure!(
+            StorageVersion::get::<Pallet<T>>() == 0 || StorageVersion::get::<Pallet<T>>() == 0,
```
DamianStraszak, line 18: should probably be `1` in the second option

DamianStraszak (approved):

PR #910, opened 2023-02-07
Use no_std variants in relations crate.

deuszx:
When using `relations` as dependency of a `no_std` project (measuring performance of SNARKs in the browser), the following compilation errors were emitted (among others):
```sh
error[E0433]: failed to resolve: use of undeclared crate or module `std`
  --> /home/mateusz/aleph/aleph-node/relations/src/shielder/tangle.rs:29:5
   |
29 | use std::ops::Add;
   |     ^^^ use of undeclared crate or module `std`

error[E0433]: failed to resolve: use of undeclared crate or module `std`
 --> /home/mateusz/aleph/aleph-node/relations/src/shielder/circuit_utils.rs:4:5
  |
4 | use std::ops::Index;
  |     ^^^ use of undeclared crate or module `std`

```
switching to proper no std variants fixes the problem.

### DIFF
--- a/relations/src/shielder/circuit_utils.rs
+++ b/relations/src/shielder/circuit_utils.rs
@@ -1,7 +1,6 @@
-use core::borrow::Borrow;
+use core::{borrow::Borrow, ops::Index};
 #[cfg(feature = "std")]
 use std::fmt::{Display, Formatter};
-use std::ops::Index;
 
 use ark_r1cs_std::{
     alloc::{AllocVar, AllocationMode},
@@ -9,6 +8,7 @@ use ark_r1cs_std::{
     R1CSVar,
 };
 use ark_relations::r1cs::{Namespace, SynthesisError};
+use ark_std::vec::Vec;
 
 use crate::CircuitField;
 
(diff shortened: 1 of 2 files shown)


relations/src/shielder/circuit_utils.rs
```diff
@@ -54,7 +54,7 @@ impl AllocVar<(u8, Result<u64, SynthesisError>), CircuitField> for PathShapeVar
         let ns = cs.into();
         let cs = ns.cs();
 
-        let mut shape = vec![];
+        let mut shape = Vec::new();
```
pmikolajczyk41, line 57: fyi: `vec` macro is available in `core` or `ark_ff` or `ark_std` crates as well

pmikolajczyk41 (approved):
I strongly suggest adding `cargo nono` to `snarkeling` pipelines. See: https://github.com/penumbra-zone/poseidon377/pull/37/files#diff-73e17259d77e5fbef83b2bdbbe4dc40a912f807472287f7f45b77e0cbf78792dR58

relations/src/shielder/circuit_utils.rs
deuszx: This one is from `ark_std` though (or at least that's what I expect).

PR #911, opened 2023-02-07
Use cargo nono

pmikolajczyk41:
Ensure that some crates are `no-std`-compatible.

`poseidon` will be checked as well, as soon as it becomes `no-std` friendly (soon).

### DIFF
--- a/.github/workflows/check-excluded-packages.yml
+++ b/.github/workflows/check-excluded-packages.yml
@@ -84,7 +84,7 @@ jobs:
       - name: Install nono
         run: cargo install cargo-nono
 
-      - name: Assert that packages are 'no-std'-compatible
+      - name: Assert that packages are compatible with no-std
         run: |
           packages=("relations" "bin/runtime/src/chain_extension/snarcos-extension")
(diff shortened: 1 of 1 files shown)


.github/workflows/check-excluded-packages.yml
```diff
@@ -81,5 +81,19 @@ jobs:
             popd
           done
 
+      - name: Install nono
+        run: cargo install cargo-nono
+
+      - name: Assert that packages are compatible with no-std
+        run: |
+          packages=("relations" "bin/runtime/src/chain_extension/snarcos-extension")
```
deuszx, line 89: Does this extension have to be compatible with no-std or just wasm32-unknown-unknown?

deuszx (approved):

.github/workflows/check-excluded-packages.yml
pmikolajczyk41: it is used by contracts, which are compiled with both `std` and `no-std`, so I suppose that it has to be

mike1729 (approved):

PR #914, opened 2023-02-08
A0-1923: add e2e tests for pruning

maciejnems:
# Description

Test for nodes with pruning turned on.

## Type of change

Two new tests are added.
* catch up with state pruning 90 (short session is on)
* multiple restarts with state pruning 2048 (no short session)

Removed archive argument from tests (it does not need to be passed in normal usage).
Catch up tests waits for 90 finalized blocks, to check if pruning works.
Multiple restarts test waits for 256 finalized blocks, as default pruning setting for substrate is 256.

# Checklist:

- I have added tests

### DIFF
--- a/.github/workflows/e2e-tests-main-devnet.yml
+++ b/.github/workflows/e2e-tests-main-devnet.yml
@@ -62,7 +62,7 @@ jobs:
       - name: GIT | Checkout source code
         uses: actions/checkout@v2
 
-      - name: Install Rust toolchain
+      - name: Install Rust toolchainrun_e2e_testrun_e2e_test
         uses: actions-rs/toolchain@v1
 
       - name: Install Protoc
@@ -796,6 +796,9 @@ jobs:
     name: Test catching up
     runs-on: ubuntu-20.04
     needs: build-new-node
+    strategy:
+      matrix:
+        pruning: ['', '--state-pruning 90']
     steps:
       - name: Checkout source code
         uses: actions/checkout@v2
@@ -815,12 +818,15 @@ jobs:
         env:
           # Relative to local-tests/ directory
           ALEPH_NODE_BINARY: aleph-test-node/aleph-node
-        run: ./.github/scripts/test_catch_up.sh
+        run: ./.github/scripts/test_catch_up.sh ${{ matrix.pruning }}
 
   test-multiple-restarts:
     name: Test multiple restarts
     runs-on: ubuntu-20.04
     needs: build-new-node
+    strategy:
+      matrix:
+        pruning: ['', '--state-pruning 2048']
     steps:
       - name: Checkout source code
         uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


local-tests/test_catch_up.py
```diff
@@ -46,10 +58,10 @@ def printt(s): print(ctime() + ' | ' + s)
 
 chain.start('aleph', nodes=[4, 5])
 
-printt('Waiting for finalization')
-chain.wait_for_finalization(0)
 printt('Waiting for authorities')
 chain.wait_for_authorities()
+printt('Waiting for finalization')
+chain.wait_for_finalization(90)
```
h4nsu, line 64: Why 90?

local-tests/test_multiple_restarts.py
```diff
@@ -22,28 +29,34 @@ def printt(s): print(ctime() + ' | ' + s)
...
 
 printt('Starting the chain')
 chain.start('aleph')
 
-printt('Waiting for finalization')
-chain.wait_for_finalization(0)
 printt('Waiting for authorities')
 chain.wait_for_authorities()
+printt('Waiting for finalization')
+chain.wait_for_finalization(256)
```
h4nsu, line 59: Same here: why 256?

local-tests/test_multiple_restarts.py
```diff
@@ -73,5 +86,5 @@ def printt(s): print(ctime() + ' | ' + s)
 
     # Check if the murdered node started catching up with reasonable nr of blocks.
     if diff <= delta:
-        printt(f'Too small catch up for validators: {validator_diff}')
```
h4nsu, line 76: whooops ^^'

h4nsu (approved):
Looks good modulo mysterious numbers used as `wait_for_finalization()` arguments.

.github/workflows/e2e-tests-main-devnet.yml
```diff
@@ -62,7 +62,7 @@ jobs:
       - name: GIT | Checkout source code
         uses: actions/checkout@v2
 
-      - name: Install Rust toolchain
+      - name: Install Rust toolchainrun_e2e_testrun_e2e_test
```
Kosiare, line 65: Accidentally something got pasted here?

Kosiare (approved):
LGTM, besides that one strange step name.

.github/workflows/e2e-tests-main-devnet.yml
maciejnems: Yep, exactly

local-tests/test_catch_up.py
maciejnems: Because we want pruning to start. Changed that to be more clear

local-tests/test_multiple_restarts.py
maciejnems: Because pruning starts in substrate at 256. We want to check it does not prune anything

PR #915, opened 2023-02-08
Fork off: child support

krzysztofziobro:
# Description

Adding support for child tries to fork-off. This should make some basic contracts usable on the new chain (anything that depends on block numbers is still broken).

## Type of change

- Bug fix (?)
- New feature (?)

# Checklist:


deuszx:
I have to admit, without knowledge about internal design of the storage it's difficult to review for the correctness/completeness. Can we add e2e tests for this new feature?

krzysztofziobro:
> I have to admit, without knowledge about internal design of the storage it's difficult to review for the correctness/completeness. Can we add e2e tests for this new feature?

Right, I may add some tests for contract interaction (or something else that uses child tries, but I am not aware off any other useful pallets that use them) on the "forked-off" chain, but there are no e2e tests for fork-off as for now, so the setup will probably take a while.

DamianStraszak (changes requested):

fork-off/src/jsonrpc_client.rs
```diff
@@ -28,6 +28,22 @@ pub trait Rpc {
         start_key: Option<StorageKey>,
         at: Option<BlockHash>,
     ) -> Result<Vec<StorageKey>, Error>;
+
+    #[rpc(name = "childstate_getKeys")]
```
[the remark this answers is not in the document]
krzysztofziobro: changed

fork-off/src/main.rs
```diff
@@ -56,9 +56,12 @@ async fn main() -> anyhow::Result<()> {
...
 
-    let initial_state: Storage =
-        serde_json::from_value(initial_spec["genesis"]["raw"]["top"].take())
-            .expect("Deserialization of state from given chainspec file failed");
+    // Initialize with state from chainspec + empty child storage
+    let initial_state = Storage {
+        top: serde_json::from_value(initial_spec["genesis"]["raw"]["top"].take())
+            .expect("Deserialization of state from given chainspec file failed"),
+        child_storage: ChildStorage::new(),
+    };
```
deuszx, line 64: We can create a constructor for `Storage` that accepts the `initial_spec` only:
```rust
impl Storage {
  pub fn new(initial_spec) -> Result<Self, anyhow::Error> {
    ...
  }
}
```

krzysztofziobro: done

fork-off/src/fetching.rs
```diff
@@ -37,10 +37,22 @@ impl StateFetcher {
...
             let mut output_guard = output.lock();
-            output_guard.insert(key, value);
-            if output_guard.len() % LOG_PROGRESS_FREQUENCY == 0 {
-                info!("Fetched {} values", output_guard.len());
+            output_guard.top.insert(key.clone(), value);
+            if let Ok(child_storage_map) = child_storage_map_res {
+                info!("Fetched child trie with {} keys", child_storage_map.len());
+                output_guard
+                    .child_storage
+                    .insert(key.without_child_storage_prefix(), child_storage_map);
```
deuszx, line 51: We can omit these if `child_storage_map` is empty.

fork-off/src/fetching.rs
```diff
@@ -37,10 +37,22 @@ impl StateFetcher {
                 .await
                 .unwrap();
 
+            let child_storage_map_res = self
+                .client
+                .get_child_storage_for_key(key.clone(), &block)
+                .await;
```
deuszx, line 43: Above if `self.client.get_storage` returns error (`RpcError`) we call `unwrap()` and short-circuit the loop.

fork-off/src/jsonrpc_client.rs
```diff
@@ -71,6 +89,46 @@ impl Client {
...
+                    empty_prefix.clone(),
+                    CHUNK_SIZE,
+                    start_key,
+                    Some(at.clone()),
+                )
+                .await?;
+
+            let values = self
+                .client
+                .get_child_storage_entries(child_key.clone(), keys.clone(), Some(at.clone()))
```
deuszx, line 116: I'm wondering - the `get_child_storage_entries` might be called with up to `CHUNK_SIZE` keys, can it happen that the amount of data returned here would be too much and the query would fail?

fork-off/src/fetching.rs
krzysztofziobro: Here success of the rpc means that this is indeed a key which is used to store a child trie, so we probably want to insert that anyway.

fork-off/src/fetching.rs
krzysztofziobro: Error here means that we passed an incorrect storage key: since it is a correct key in general, we know that there is just no child trie here (now I realize that we actually do not know that - error can probably have a different cause, will check that)

fork-off/src/fetching.rs
deuszx: I see, so we would insert empy storage which is an information as well.

fork-off/src/fetching.rs
deuszx: If there is no child trie there we should be able to recognize that differently. Maybe change the return type to `Result<Option<_>, RpcError>` and handle it on the caller's site adequately.

fork-off/src/jsonrpc_client.rs
krzysztofziobro: Hmm... it's still hard bound by `CHUNK_SIZE` * (what can fit in a block), but yeah, I'm not sure if that's good enough, will change it just to be sure

fork-off/src/fetching.rs
krzysztofziobro: Sounds right

fork-off/src/jsonrpc_client.rs
deuszx: When you say _what can fit in a block_ do you mean block's maximum size (in bytes) or its maximum gas? The reason I'm asking is that, in theory, you can have a very small transaction that will write lots of keys (child storage entries?) - just write new keys in a loop until you run out of gas.

Of course, this is not very practical example but there are other cases where a simple transaction can write a lot of keys.

On the other hand, maybe it's not worthwhile to spend too much time on this since it's a testing tool that won't affect any running system and we can fix as we go and encounter this problem.

**tl;dr;**: I think I'm fine leaving this _as-is_ and deal with it if we ever have to.

fork-off/src/fetching.rs
deuszx: If you do that, we can `unwrap()` here as well (or just `.await?`) since the only errors that would be returned are RPC errors (connection etc.) since the _child storage does not exist_ (valid case) would be represented as `None` and be handled properly below.

fork-off/src/jsonrpc_client.rs
krzysztofziobro: I mean size, it's not possible to write more data under a key than what fits in a block (right?)
Not sure how a lots of keys can hurt us here - we have 1:1 for (keys in child trie) : (child trie storage entries)

krzysztofziobro: Right, right, for some reason I didn't consider the case in which transaction is small but runtime/some contract generates large "value" to store it under some key

deuszx: But again, I don't think we _have to_ handle that case here right away. I think it will be most instructive to simply run this on already-existing networks and see whether it works as expected and handle the problems as they go.

fork-off/src/jsonrpc_client.rs
```diff
@@ -71,6 +89,65 @@ impl Client {
...
+            Err(RpcError::JsonRpcError(err)) => {
+                // Ugly :(
+                if err.message == "Client error: Invalid child storage key" {
+                    Ok(None)
+                } else {
+                    Err(RpcError::JsonRpcError(err))
+                }
+            }
+            _ => res.map(Some),
+        }
```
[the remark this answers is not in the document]
krzysztofziobro: changed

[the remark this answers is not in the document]
deuszx: Thanks

DamianStraszak (approved):

fork-off/src/fetching.rs
krzysztofziobro: I think it's done

deuszx: Yes, it is. If @DamianStraszak approved I guess it means that works :)

deuszx (approved):

PR #926, opened 2023-02-14
A0-1903: update to 0.9.34

maciejnems:
# Description

Update to substrate 0.9.34

## Type of change

* Needed to remove fee calculation test, as fill_block method is no longer in frame system. It is instead in a `pallet_root_testing` (https://github.com/paritytech/substrate/commit/3e71d606b7d1e91d9c1701c0a443530eefca1a39), which should not be used in production code. Will create a task for bringing this test back


bin/runtime/src/lib.rs
```diff
@@ -104,7 +106,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("aleph-node"),
     impl_name: create_runtime_str!("aleph-node"),
     authoring_version: 1,
-    spec_version: 46,
+    spec_version: 47,
```
Marcin-Radecki, line 109: On Testnet we have now 51, please bump to at least 52. 

In the future, we'll use a smarter way to version runtimes, e.g. the one Polkadot uses, but for now let's stick with assumption that spec version on mainnet must be different than testnet, to avoid mistakes during deployments

e2e-tests/src/test/fee.rs
```diff
@@ -1,103 +1,106 @@
...
 
-async fn fill_blocks(target_ratio: u8, blocks: u32, connection: &RootConnection) {
-    for _ in 0..blocks {
-        connection
-            .fill_block(target_ratio, TxStatus::InBlock)
-            .await
-            .unwrap_or_else(|err| panic!("Error while sending a fill_block transation: {:?}", err));
-    }
-}
+// use aleph_client::{RootConnection, pallets::system::SystemSudoApi};
```
Marcin-Radecki, line 10: Since there's a lot of commented code, can we just add a comment why we're doing so

Marcin-Radecki (approved):
Look good, please run this migration https://github.com/paritytech/substrate/commit/2704ab3d348f18f9db03e87a725e4807b91660d8#diff-d3a8ee05060d4cc1bb98df686418ec5aeba54698bd793a46c9095dde576eaa77 , most likely list of accounts needs to be empty

e2e-tests/src/test/fee.rs
maciejnems: To be honest, I can also just remove it. Git history will keep it

maciejnems: Btw, I added a comment about it, probably in a wrong place though

bin/runtime/src/lib.rs
```diff
@@ -742,6 +746,13 @@ construct_runtime!(
     }
 );
 
+pub struct EmptyList;
+impl Get<Vec<AccountId>> for EmptyList {
+    fn get() -> Vec<AccountId> {
+        vec![]
+    }
+}
```
deuszx, line 754: 😎

deuszx:
> Will create a task for bringing this test back

Did that happen?

deuszx (approved):

kostekIV:
> > Will create a task for bringing this test back
> 
> Did that happen?

didn't find it. [Here](https://cardinal-cryptography.atlassian.net/browse/A0-2016) newl created issue

PR #927, opened 2023-02-14
Migrating contracts on testnet

DamianStraszak:
# Description

Custom contracts migrations that fix the problem with missed migrations.

## Type of change

# Checklist:

Logs: [node0.log](https://github.com/Cardinal-Cryptography/aleph-node/files/10741080/node0.log)

### DIFF
--- a/bin/runtime/src/migrations/custom_contracts_migration.rs
+++ b/bin/runtime/src/migrations/custom_contracts_migration.rs
@@ -22,8 +22,8 @@ type BalanceOf<T> =
 const TARGET: &str = "runtime::custom_contract_migration";
 
 /// Performs all necessary migrations based on `StorageVersion`.
-pub struct Migration<T: Config>(PhantomData<T>);
-impl<T: Config> OnRuntimeUpgrade for Migration<T> {
+pub struct CustomMigrateToV8<T: Config>(PhantomData<T>);
+impl<T: Config> OnRuntimeUpgrade for CustomMigrateToV8<T> {
     fn on_runtime_upgrade() -> Weight {
         let version = StorageVersion::get::<Pallet<T>>();
         let mut weight = Weight::zero();
--- a/bin/runtime/src/migrations/custom_contracts_migration.rs
+++ b/bin/runtime/src/migrations/custom_contracts_migration.rs
@@ -143,7 +143,7 @@ mod v8 {
         }
     }
 
-	// Returns the used weight and whether this migration is the final one.
+    // Returns the used weight and whether this migration is the final one.
     pub fn migrate<T: Config>() -> (Weight, bool) {
         log::info!(
             target: TARGET,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -236,7 +236,7 @@ dependencies = [
 
 [[package]]
 name = "aleph-node"
-version = "0.9.0"
+version = "0.9.1"
 dependencies = [
  "aleph-runtime",
  "finality-aleph",
@@ -290,7 +290,7 @@ dependencies = [
 
 [[package]]
 name = "aleph-runtime"
-version = "0.9.0"
+version = "0.9.1"
 dependencies = [
  "frame-executive",
  "frame-support",
--- a/aleph-client/src/aleph_zero.rs
+++ b/aleph-client/src/aleph_zero.rs
@@ -1,5 +1,5 @@
-#[doc(hidden)]
 #[allow(dead_code, unused_imports, non_camel_case_types)]
+#[allow(clippy::all)]
 pub mod api {
     use super::api as root_mod;
     pub static PALLETS: [&str; 21usize] = [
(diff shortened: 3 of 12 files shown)


bin/runtime/src/migrations/custom_contracts_migration.rs
```diff
@@ -0,0 +1,306 @@
...
+            sp_io::storage::next_key(&previous_key).filter(|n| n.starts_with(&prefix))
+        {
+            previous_key = next;
+            let value = sp_io::storage::get(&previous_key)
+                .expect("The loop conditions checks for existence of the key; qed");
+            match decode_to_either_contract_info::<T>(&mut &value[..]) {
+                Ok(EitherContractInto::Old(old)) => {
+                    if total_items >= MAX_NUM_ITEMS_PROCESSED_PER_MIGRATION {
+                        num_old_contracts_skipped.saturating_accrue(1);
+                        continue;
```
kostekIV, line 174: we could just break from the while loop here and loose information about num_old_contracts_skipped and num_new_contracts but gain some time on skipping the decoding of the rest contracts.

bin/runtime/src/migrations/custom_contracts_migration.rs
```diff
@@ -0,0 +1,306 @@
...
+                    let new_contract_info = ContractInfo {
+                        trie_id: old.trie_id,
+                        code_hash: old.code_hash,
+                        storage_bytes,
+                        storage_items,
+                        storage_byte_deposit,
+                        storage_item_deposit,
+                        storage_base_deposit,
+                    };
+                    unhashed::put::<ContractInfo<T>>(&previous_key, &new_contract_info);
```
kostekIV, line 209: is that safe? old version had fewer bytes, could we corrupt db by overwriting smth in memory?

kostekIV (commented):
I think you use a different version of the subxt-cli. On this branch, it was '0.25.0'

bin/runtime/src/migrations/custom_contracts_migration.rs
DamianStraszak: `previous_key` is a key corresponding to a single entry in the `ContractInfoOf` map. unhashed::put::<ContractInfo<T>> takes the key and a value, serializes the value, and writes the bytes under this key. We do simply "translate" the old format to the new one, and indeed this code is just lower level details of `translate`.

bin/runtime/src/migrations/custom_contracts_migration.rs
```diff
@@ -0,0 +1,306 @@
...
+        trie_id: TrieId,
+        code_hash: CodeHash<T>,
+        storage_bytes: u32,
+        storage_items: u32,
+        storage_byte_deposit: BalanceOf<T>,
+        storage_item_deposit: BalanceOf<T>,
+        storage_base_deposit: BalanceOf<T>,
+    }
+
+    const OLD_INFO_ENCODING_LEN: usize = 33 + 32 + 16;
```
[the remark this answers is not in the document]
DamianStraszak: TrieId is a vec of `32` bytes always, in which case the encoding is `33` bytes. This is 100% consistent with on-chain data. Even if this was wrong for some reason we would just skip it and then see non-zero `num_contracts_erred`, which we don't see :) 

`translate_values` is too restrictive, as it forces you to know the type of the values, in our case there is no such type, also `translate_values` does not allow to skip anything, as we must do here.
What I do here, is pretty much roll out `translate_values` by pasting the implementation and tweaking it a little bit, to fit our needs.

bin/runtime/src/migrations/custom_contracts_migration.rs
```diff
@@ -0,0 +1,306 @@
...
+impl<T: Config> OnRuntimeUpgrade for CustomMigrateToV8<T> {
+    fn on_runtime_upgrade() -> Weight {
+        let version = StorageVersion::get::<Pallet<T>>();
+        let mut weight = Weight::zero();
+        log::info!(
+            target: TARGET,
+            "On-chain version of pallet contracts is {:?}",
+            version
+        );
+        if version < 8 {
```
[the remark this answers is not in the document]
DamianStraszak: First of all, we don't care what happens if someone runs this migration on some old version. It will break their chain most likely :P And yes, there are older version which had different `ContractInfo` type.

In our case, the storage is in version `7` (or actually somewhere in between 7 and 8 :P). However, there is no `StorageVersion` key on chain, and hence we will see the version as `0`.

Marcin-Radecki (approved):

bin/runtime/src/migrations/custom_contracts_migration.rs
DamianStraszak: Agreed, this could save a bit, but not a lot actually. In the best case scenario this saves around 4k reads, which is 3% of the typical number of 150k reads we make per migration. 

I like to see the full data logged in the summary, so that we can make sure everything is progressing as expected...

kostekIV (approved):

PR #930, opened 2023-02-15
A0-1973: Tune down contracts constants

pmikolajczyk41:
# Description

We follow Alex's advice and decrease two limits for `pallet_contracts`.

## Type of change

# Checklist:

### DIFF
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -678,9 +678,9 @@ impl pallet_contracts::Config for Runtime {
     type DeletionQueueDepth = DeletionQueueDepth;
     type DeletionWeightLimit = DeletionWeightLimit;
     type Schedule = Schedule;
-    type CallStack = [pallet_contracts::Frame<Self>; 31];
+    type CallStack = [pallet_contracts::Frame<Self>; 5];
     type AddressGenerator = pallet_contracts::DefaultAddressGenerator;
-    type MaxCodeLen = ConstU32<{ 128 * 1024 }>;
+    type MaxCodeLen = ConstU32<{ 123 * 1024 }>;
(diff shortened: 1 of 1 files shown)


h4nsu (approved):

bin/runtime/src/lib.rs
```diff
@@ -678,9 +678,9 @@ impl pallet_contracts::Config for Runtime {
     type DeletionQueueDepth = DeletionQueueDepth;
     type DeletionWeightLimit = DeletionWeightLimit;
     type Schedule = Schedule;
-    type CallStack = [pallet_contracts::Frame<Self>; 31];
+    type CallStack = [pallet_contracts::Frame<Self>; 5];
```
deuszx, line 681: Just FYI - this may cause currently deployed contracts to stop working.

pmikolajczyk41: yes, but if your contract needs 5 nested cross-contract calls, then probably you don't have an optimal architecture anyway

deuszx: Yes. But previously you didn't know that when dry-running so could have decided to deploy it. I'm not blocking, just saying we should remember to mention it in the release notes.

DamianStraszak: I'm kind of concerned why this is so small. Note that in EVM this is 256, so even 31 was already small, but now we are changing it to 5?! Would be great to know the core reason here...

DamianStraszak: 5 seems too small to me, each `delegate_call`-type upgradable contract already adds 2 frames. This is severely limiting the smart contract functionality

pmikolajczyk41:
replaced by #938

PR #932, opened 2023-02-16
Exchange tangle with poseidon. Remove once-cell

pmikolajczyk41:
1. We remove tangling and substitute it with calls to poseidon
2. We remove `once_cell` dep and usages from poseidon (both node's runtime and contracts were complaining about some low level things)
3. We introduce `circuit` feature to `poseidon` so that e.g. contracts or chain extension don't have to fetch irrelevant arkworks deps


mike1729 (approved):

poseidon/build.rs
```diff
@@ -12,6 +12,7 @@ use poseidon_paramgen::poseidon_build;
 fn main() {
     let security_level = match env::var("SECURITY_LEVEL") {
         Ok(level) => match level.as_str() {
+            "32" => 32,
```
deuszx, line 15: This is not mentioned in the PR's description. Is it left here intentionally?

deuszx (approved):
I think it would be fair to ask @fbielejec for a review as well. It's a bit of his child.

poseidon/build.rs
pmikolajczyk41: Leftover, but harmless and potentially helpful

PR #934, opened 2023-02-16
Expose poseidon via chain extension

pmikolajczyk41:
_This is **POC**_ and therefore some things are left to do (like tests, verbose errors etc)


bin/runtime/src/chain_extension/baby-liminal-extension/lib.rs
```diff
@@ -87,6 +87,15 @@ pub trait BabyLiminalExtension {
         input: Vec<u8>,
         system: ProvingSystem,
     );
+
+    #[ink(extension = 43, returns_result = false, handle_status = false)]
+    fn poseidon_one_to_one(input: [[u64; 4]; 1]) -> [u64; 4];
+
+    #[ink(extension = 44, returns_result = false, handle_status = false)]
+    fn poseidon_two_to_one(input: [[u64; 4]; 2]) -> [u64; 4];
```
fbielejec, line 95: why like this, why not one, two and four arguments respectively?

pallets/baby-liminal/Cargo.toml
```diff
@@ -74,6 +75,7 @@ runtime-benchmarks = [
     "frame-support/runtime-benchmarks",
     "frame-system/runtime-benchmarks",
     "sp-runtime/runtime-benchmarks",
+    "liminal-ark-poseidon",
```
fbielejec, line 78: /std missing?

bin/runtime/src/chain_extension/baby-liminal-extension/lib.rs
pmikolajczyk41: so that decoding on the chain side is easier :P

unfortunately, when you implement `ChainExtension` and have `env` object, you cannot read arguments sequentially (the pointer is not moved); that sucks ofc

poseidon/build.rs
```diff
@@ -18,7 +18,7 @@ fn main() {
             "256" => 256,
             _ => panic!("Unsupported security level. Supported levels: 80, 128, 256"),
         },
-        Err(_) => 128,
+        Err(_) => 32,
```
fbielejec, line 21: isn't it to low for production?

pmikolajczyk41: it is, but see the discussion on Slack - for 128, the contract won't fit into the limits

pmikolajczyk41: oh, now it won't be needed to store it in a contract - will revert; thanks!

pallets/baby-liminal/Cargo.toml
fbielejec: I meant the "std" feature above benchmarks feature,  I cannot comment on non-edited lines :/

pallets/baby-liminal/src/benchmarking/suite.rs
```diff
@@ -156,5 +156,19 @@ benchmarks! {
         )
     }
 
+    // Cryptography
+
+    poseidon_one_to_one {} : {
```
fbielejec, line 161: shouldn't the benchmarks be over a range of values? You could easily plug a range of integers there

pallets/baby-liminal/Cargo.toml
pmikolajczyk41: it's because it's optional and it is needed only for benchmarks; perfectly, `poseidon` should be benchmarked outside the pallet, but I don't know how to do it quickly :man_shrugging: we can think about it later, it's not the worst place for that imho

bin/runtime/src/chain_extension/mod.rs
```diff
@@ -202,4 +220,47 @@ impl BabyLiminalChainExtension {
         };
         Ok(RetVal::Converging(return_status))
     }
+
+    fn create_fr(x: SingleHashInput) -> Fr {
```
mike1729, line 224: nit: `into_field_element`

pallets/baby-liminal/src/benchmarking/suite.rs
pmikolajczyk41: yes, but hashing function should run roughly the same time for different outputs; but I can add an artificial input anyway

mike1729 (approved):

pallets/baby-liminal/src/benchmarking/suite.rs
pmikolajczyk41: done

bin/runtime/src/chain_extension/mod.rs
pmikolajczyk41: done

poseidon/build.rs
pmikolajczyk41: reverted

pallets/baby-liminal/Cargo.toml
fbielejec: no, no it's a perfect place

fbielejec (approved):

PR #938, opened 2023-02-21
Enforce heap pages

mike1729:
# Description

Increase total memory to 256MB, decrease call stack to 16 and enforce heap pages limit at cli.

## Type of change

# Checklist:

### DIFF
--- a/bin/node/src/main.rs
+++ b/bin/node/src/main.rs
@@ -6,10 +6,11 @@ use aleph_runtime::Block;
 use log::warn;
 use sc_cli::{clap::Parser, CliConfiguration, PruningParams, SubstrateCli};
 use sc_network::config::Role;
-use sc_service::PartialComponents;
+use sc_service::{Configuration, PartialComponents};
 
 const STATE_PRUNING: &str = "archive";
 const BLOCKS_PRUNING: &str = "archive-canonical";
+const HEAP_PAGES: u64 = 4094;
 
 fn pruning_changed(params: &PruningParams) -> bool {
     let state_pruning_changed =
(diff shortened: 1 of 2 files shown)


kostekIV (approved):

bin/node/src/main.rs
```diff
@@ -6,10 +6,11 @@ use aleph_runtime::Block;
 use log::warn;
 use sc_cli::{clap::Parser, CliConfiguration, PruningParams, SubstrateCli};
 use sc_network::config::Role;
-use sc_service::PartialComponents;
+use sc_service::{Configuration, PartialComponents};
 
 const STATE_PRUNING: &str = "archive";
 const BLOCKS_PRUNING: &str = "archive-canonical";
+const HEAP_PAGES: u64 = 4094;
```
deuszx, line 13: Why is it 4094 and not 4096?

mike1729: Due to a typo ;)

bin/runtime/src/lib.rs
```diff
@@ -681,7 +681,7 @@ impl pallet_contracts::Config for Runtime {
     type DeletionQueueDepth = DeletionQueueDepth;
     type DeletionWeightLimit = DeletionWeightLimit;
     type Schedule = Schedule;
-    type CallStack = [pallet_contracts::Frame<Self>; 31];
+    type CallStack = [pallet_contracts::Frame<Self>; 16];
```
pmikolajczyk41, line 684: Not 15? There's one more implicit frame for the initial contract call

mike1729: Our depth is 17 and call stack is 16 :)

deuszx: What @pmikolajczyk41 meant, I think, is that we want to allow for 16 in total, including the starting frame, and at least the `integrity_test` will add 1 to the `CallStack::size` return value https://github.com/paritytech/substrate/blob/8d3eace503c3956fbeda321686fc46faa62cebc6/frame/contracts/src/lib.rs#L392-L393

deuszx (approved):

PR #940, opened 2023-02-22
Check contstraints for memory parameters

mike1729:
Adds a simple test to check if parameters are right.


bin/runtime/src/lib.rs
```diff
@@ -1015,10 +1015,35 @@ impl_runtime_apis! {
 
 #[cfg(test)]
 mod tests {
-    use crate::VERSION;
+    use super::*;
+    use primitives::HEAP_PAGES;
+    use smallvec::Array;
```
deuszx, line 1020: Where do we use this exactly? IIRC, `smallvec` gives _some_ performance gains by pre-allocating space up to some small numbers of vec elements. I'm not sure it's worth adding new dependency just for it, especially in tests.

bin/runtime/src/lib.rs
```diff
@@ -1015,10 +1015,35 @@ impl_runtime_apis! {
...
+        let max_heap_size = <Runtime as pallet_contracts::Config>::Schedule::get()
+            .limits
+            .max_memory_size();
+        // Max call depth is CallStack::size() + 1
+        let max_call_depth = <Runtime as pallet_contracts::Config>::CallStack::size() as u32 + 1;
+        // Max code len
+        let max_code_len: u32 = <Runtime as pallet_contracts::Config>::MaxCodeLen::get();
+
+        let lhs = max_call_depth * (72 * max_code_len + max_heap_size + MAX_STACK_SIZE);
+        let rhs = MAX_RUNTIME_MEM * 3 / 4;
```
DamianStraszak, line 1045: Could you write a one-line comment where this 3/4 is coming from, and that this is not actually enforced in the runtime :P

DamianStraszak (approved):

bin/runtime/src/lib.rs
DamianStraszak: We can maybe move to `dev-dependencies`?

mike1729: @deuszx `CallStack` is of this type so I need to import it to call `CallStack::size`.
@DamianStraszak sure.

bin/runtime/src/lib.rs
deuszx: I was hesitating whether I should ask the same question or I will come across as 🙈 🙈 🙉

mike1729: Done.

PR #950, opened 2023-02-27
Update ink deps in baby liminal chain extension.

deuszx:
Bumps ink version to stable releas of ink4.

### DIFF
--- a/bin/runtime/src/chain_extension/baby-liminal-extension/lib.rs
+++ b/bin/runtime/src/chain_extension/baby-liminal-extension/lib.rs
@@ -4,7 +4,6 @@ use ink::env::Environment;
 use scale::{Decode, Encode};
 #[cfg(feature = "std")]
 use scale_info::TypeInfo;
-use sp_std::vec::Vec;
 
 /// Gathers all the possible errors that might occur while calling `pallet_baby_liminal::store_key` or
 /// `pallet_baby_liminal::verify`.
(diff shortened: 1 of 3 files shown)


bin/runtime/src/chain_extension/baby-liminal-extension/lib.rs
```diff
@@ -74,27 +73,27 @@ pub trait BabyLiminalExtension {
     ///
     /// The extension method ID matches the one declared in runtime:
     /// `BABY_LIMINAL_STORE_KEY_FUNC_ID`.
-    #[ink(extension = 41, returns_result = false)]
+    #[ink(extension = 41)]
```
deuszx, line 76: Attribute no longer existing in latest ink release.

fbielejec (approved):

pmikolajczyk41 (approved):

bin/runtime/src/chain_extension/baby-liminal-extension/lib.rs
pmikolajczyk41: and the current behaviour is like `returns_result = false`?

deuszx: Looks like it from the examples give in the docs https://use.ink/macros-attributes/chain-extension/#attributes

PR #956, opened 2023-03-01
Bake poseidon parameters in library

mike1729:
We generate fixed parameters and ship them with the library. If new are needed, there is a binary to recreate them.


poseidon/Cargo.toml
```diff
@@ -21,14 +21,18 @@ ark-sponge = { git = "https://github.com/penumbra-zone/sponge", branch = "r1cs",
 poseidon-parameters = { git = "https://github.com/penumbra-zone/poseidon377", rev = "50699746", default-features = false }
 poseidon-permutation = { git = "https://github.com/penumbra-zone/poseidon377", rev = "50699746", default-features = false }
 
-[build-dependencies]
-ark-bls12-381 = { version = "^0.3.0" }
-ark-ff = { version = "^0.3.0" }
+poseidon-paramgen = { git = "https://github.com/penumbra-zone/poseidon377", rev = "50699746" , optional = true, default-features = false  }
+
+[lib]
+name = "liminal_ark_poseidon"
```
pmikolajczyk41, line 27:
```suggestion
name = "liminal-ark-poseidon"
```

poseidon/Cargo.toml
```diff
@@ -21,14 +21,18 @@ ark-sponge = { git = "https://github.com/penumbra-zone/sponge", branch = "r1cs",
...
 
-poseidon-paramgen = { git = "https://github.com/penumbra-zone/poseidon377", rev = "50699746" }
+[[bin]]
+name = "liminal-ark-poseidon-paramgen"
+path = "src/generate_parameters.rs"
+required-features = ["std", "build-binary"]
 
 [features]
-default = ["std"]
+default = ["build-binary"]
```
pmikolajczyk41, line 35: why `build-binary` is default? and why `std` is not default anymore?

poseidon/src/parameters.rs
```diff
@@ -1,2 +1,81 @@
-use ark_ff::vec;
-include!(concat!(env!("OUT_DIR"), "/parameters.rs"));
+use ark_ff::PrimeField;
```
pmikolajczyk41, line 1: we may add a header with a warning about not changing this file

poseidon/Cargo.toml
```diff
@@ -38,9 +42,11 @@ std = [
     "ark-sponge/std",
     "poseidon-parameters/std",
     "poseidon-permutation/std",
+    "poseidon-paramgen/std",
```
pmikolajczyk41, line 45: meh, that's sad that we have to include it also for library in `std` :| but that's better than introducing `build-binary-std` feature I guess

pmikolajczyk41 (approved):

poseidon/src/generate_parameters.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
fbielejec, line 1: there is just no way this will every work with std right? There are io operations - therefore I see no point in pretending it will.

fbielejec:
```
  let security_level = match env::var("SECURITY_LEVEL") {
        Ok(level) => match level.as_str() {
            "32" => 32,
            "80" => 80,
            "128" => 128,
            "256" => 256,
        },
        Err(_) => 128,
    };
```
panic message is inconsistent with the accepted values. Maybe ditch 32 altogether anyway, it was a hack to get a faster dev build times?

poseidon/Cargo.toml
fbielejec: I'm not so sure you need it here - instead opt in for the default feature (which is the) when adding the paramgen `dependency` - it's always used in std context anyway

poseidon/Cargo.toml
```diff
@@ -21,14 +21,18 @@ ark-sponge = { git = "https://github.com/penumbra-zone/sponge", branch = "r1cs",
 poseidon-parameters = { git = "https://github.com/penumbra-zone/poseidon377", rev = "50699746", default-features = false }
 poseidon-permutation = { git = "https://github.com/penumbra-zone/poseidon377", rev = "50699746", default-features = false }
 
-[build-dependencies]
-ark-bls12-381 = { version = "^0.3.0" }
-ark-ff = { version = "^0.3.0" }
+poseidon-paramgen = { git = "https://github.com/penumbra-zone/poseidon377", rev = "50699746" , optional = true, default-features = false  }
```
fbielejec, line 24: default-features = true?

mike1729: then `cargo nono` compains.

poseidon/Cargo.toml
mike1729: It's default to not to have to use it in build and `std` is not to make `cargo nono` happy.

poseidon/src/parameters.rs
mike1729: Yeah, I'm about to do that.

poseidon/src/generate_parameters.rs
mike1729: Yeah, but we have to mark it for `cargo nono`

fbielejec (approved):

PR #961, opened 2023-03-03
Delegate attribute parsing to `darling`

pmikolajczyk41:
# Description

We give up parsing field attributes on our side (which was pretty sketchy and a bit fragile). For that we use `darling`. The only drawback is that now empty attributes must include `()` (`#[constant]` -> `#[constant()]`). This can be removed by the cost of digging back into metadata on macro side.

## Type of change

# Checklist:

### DIFF
--- a/relations/src/proc_macro/Cargo.toml
+++ b/relations/src/proc_macro/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "liminal-ark-relation-macro"
-version = "0.1.0"
+version = "0.1.1"
 edition = "2021"
 authors = ["Cardinal"]
 documentation = "https://docs.rs/?"
@@ -14,6 +14,7 @@ description = "Procedural macro for concise defining SNARK relations."
 proc-macro = true
 
 [dependencies]
+darling = "0.14.3"
 proc-macro2 = "1.0"
 syn = { version = "1.0", features = ["full"] }
 quote = "1.0"
--- a/relations/src/proc_macro/src/intermediate_representation.rs
+++ b/relations/src/proc_macro/src/intermediate_representation.rs
@@ -1,3 +1,4 @@
+use darling::FromMeta;
 use proc_macro2::{Ident, Span};
 use syn::{
     spanned::Spanned, Error as SynError, Field, Fields, FieldsNamed, Item, ItemFn, ItemMod,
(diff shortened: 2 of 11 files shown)


relations/src/proc_macro/src/intermediate_representation.rs
```diff
@@ -34,56 +29,92 @@ pub(super) struct IR {
...
+        Ok(RelationField { field, attrs })
     }
 }
 
+#[derive(Clone, Default, FromMeta)]
+pub(super) struct PublicFieldAttrs {
+    /// The value of the `frontend_type` modifier, if any.
+    pub frontend_type: Option<String>,
+    /// The value of the `parse_with` modifier, if any.
+    pub parse_with: Option<String>,
```
pmikolajczyk41, line 80: `darling` doesn't support flattening (https://github.com/TedDriggs/darling/issues/146)

relations/src/proc_macro/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "liminal-ark-relation-macro"
-version = "0.1.0"
+version = "0.1.1"
```
deuszx, line 3: I think we should bump `minor` version here instead, if we want to follow semver rules, since we're breaking the interface for all of the downstream consumers of the macro -- i.e. they have to add `()` to the attributes.

deuszx (approved):
I like that we can delete some code but why do you say it was sketchy?

maciejzelaszczyk (approved):
LGTM

relations/Cargo.toml
```diff
@@ -27,7 +27,8 @@ ark-serialize = { version = "^0.3.0", default-features = false }
 ark-snark = { version = "^0.3.0", default-features = false }
 ark-std = { version = "^0.3.0", default-features = false }
 blake2 = { version = "0.9.2", default-features = false }
-liminal-ark-relation-macro = { version = "0.1.0" }
+#liminal-ark-relation-macro = { version = "0.1.0" }
+liminal-ark-relation-macro = { path = "src/proc_macro" }
```
pmikolajczyk41, line 31: temporary ofc

maciejzelaszczyk (approved):
Nice. One change and a lot of little inconveniences removed.

relations/src/proc_macro/README.md
```diff
@@ -0,0 +1,113 @@
+# liminal-ark-relation-macro
+
+This crate provides `snark_relation` procedural macro for concise defining SNARK relations.
+Given minimal relation definition, this macro will generate all the required code for creating and casting partial relation objects, public input serialization and circuit generation.
+
+## General usage
+
+The `#[snark_relation]` attribute is intended for modules. Such module must define two items:
+  1. *relation object*: the collection of all constant, public and private relation data. The
+     struct should be defined with `#[relation_object_definition]` attribute.
```
deuszx, line 10:
```suggestion
     struct must be defined with `#[relation_object_definition]` attribute.
```
?
Since there's no other way it would work. _Should_ sounds as if it was optional.

deuszx (approved):

relations/src/proc_macro/README.md
pmikolajczyk41: done

PR #963, opened 2023-03-03
A0-1921: Add merge relation

maciejzelaszczyk:
# Description

Introduces the _merge_ relation for the pooling of two notes.

## Type of change

# Checklist:
- I have added tests

### DIFF
--- a/relations/src/shielder/merge.rs
+++ b/relations/src/shielder/merge.rs
@@ -1,11 +1,11 @@
 use liminal_ark_relation_macro::snark_relation;
 
 /// It expresses the facts that:
-///  - `first_old_note` is a prefix of the result of hashing together the `token_id`,
+///  - `first_old_note` is the result of hashing together the `token_id`,
 ///    `first_old_token_amount`, `first_old_trapdoor` and `first_old_nullifier`,
-///  - `second_old_note` is a prefix of the result of hashing together the `token_id`,
+///  - `second_old_note` is the result of hashing together the `token_id`,
 ///    `second_old_token_amount`, `second_old_trapdoor` and `second_old_nullifier`,
-///  - `new_note` is a prefix of the result of hashing together the `token_id`, `new_token_amount`,
+///  - `new_note` is the result of hashing together the `token_id`, `new_token_amount`,
 ///    `new_trapdoor` and `new_nullifier`,
 ///  - `new_token_amount = token_amount + old_token_amount`
 ///  - `first_merkle_path` is a valid Merkle proof for `first_old_note` being present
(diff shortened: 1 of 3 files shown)


relations/src/shielder/merge.rs
```diff
@@ -0,0 +1,319 @@
+use liminal_ark_relation_macro::snark_relation;
+
+/// It expresses the facts that:
+///  - `first_old_note` is a prefix of the result of hashing together the `token_id`,
```
pmikolajczyk41, line 4: "prefix of ..." is no more - this is just a hash

relations/src/shielder/merge.rs
```diff
@@ -0,0 +1,319 @@
...
+        let merkle_root = FpVar::new_input(ns!(cs, "merkle root"), || self.merkle_root())?;
+        let first_path_shape = PathShapeVar::new_witness(ns!(cs, "first path shape"), || {
+            Ok((*self.max_path_len(), self.first_leaf_index().cloned()))
+        })?;
+
+        check_merkle_proof(
+            merkle_root.clone(),
+            first_path_shape,
+            first_old_note,
+            self.first_merkle_path().cloned().unwrap_or_default(),
```
pmikolajczyk41, line 168: maybe let's use `unwrap()` - the path should never be empty, and if it is, we should panic

relations/src/shielder/merge.rs
```diff
@@ -0,0 +1,319 @@
...
+            second_leaf_index,
+            first_old_note,
+            second_old_note,
+            first_old_token_amount,
+            second_old_token_amount,
+            new_token_amount,
+        )
+    }
+
+    #[test]
```
pmikolajczyk41, line 287: we don't have negative testcases - maybe we should add some of them to ensure that every of the 4 checks is actually happening

pmikolajczyk41 (approved):
nice!

relations/src/shielder/merge.rs
maciejzelaszczyk: OK, good catch, changed.

mike1729 (approved):

relations/src/shielder/merge.rs
maciejzelaszczyk: This actually not true in our tests so far ;) Key generation uses an empty circuit for which the constraints are generated here https://github.com/Cardinal-Cryptography/aleph-node/blob/0af5afb4356829d56425e2dac16c7a7253141762/relations/src/shielder/deposit_and_merge.rs#L221. This is the likely reason why @mike1729 went for `unwrap_or_default`. The way I see it, we can leave the default version and add additional checks in the tests. I am open to other suggestions, though.

pmikolajczyk41 (approved):
good to go, thanks!

PR #965, opened 2023-03-03
adding key ownership and lockabel deposits

fbielejec:
# Description

Adding ownership info for verification keys. A configurable amount of native token is locked until the key is deleted. Only original owner is able to delete or overwrite a key. 
Deposits are tracked such that even if runtime changed the correct amount is unlocked.
In case of key overwrites we do not bother with deposit difference if it changed in the meantime - it will be returned at key delete anyway.

### DIFF
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -330,13 +330,16 @@ parameter_types! {
     // We allow 10kB keys, proofs and public inputs. This is a 100% blind guess.
     pub const MaximumVerificationKeyLength: u32 = 10_000;
     pub const MaximumDataLength: u32 = 10_000;
+    pub const VerificationKeyDepositAmount: u128 = 1_000_000_000;
 }
 
 impl pallet_baby_liminal::Config for Runtime {
+    type Currency = Balances;
     type RuntimeEvent = RuntimeEvent;
     type WeightInfo = pallet_baby_liminal::AlephWeight<Runtime>;
     type MaximumVerificationKeyLength = MaximumVerificationKeyLength;
     type MaximumDataLength = MaximumDataLength;
+    type VerificationKeyDepositAmount = VerificationKeyDepositAmount;
 }
 
 impl_opaque_keys! {
--- a/pallets/baby-liminal/src/lib.rs
+++ b/pallets/baby-liminal/src/lib.rs
@@ -7,8 +7,11 @@ mod systems;
 mod tests;
 mod weights;
 
-use frame_support::pallet_prelude::StorageVersion;
-use frame_system::ensure_root;
+use frame_support::{
+    pallet_prelude::StorageVersion,
+    traits::{Currency, ReservableCurrency},
+};
+use frame_system::ensure_signed;
 pub use pallet::*;
 pub use systems::{ProvingSystem, VerificationError};
 pub use weights::{AlephWeight, WeightInfo};
@@ -18,9 +21,12 @@ const STORAGE_VERSION: StorageVersion = StorageVersion::new(0);
 
 /// We store verification keys under short identifiers.
 pub type VerificationKeyIdentifier = [u8; 4];
+pub type VerificationKeyDeposit<T> =
+    <<T as Config>::Currency as Currency<<T as frame_system::Config>::AccountId>>::Balance;
 
 #[frame_support::pallet]
 pub mod pallet {
+
     use ark_serialize::CanonicalDeserialize;
     use frame_support::{
         dispatch::PostDispatchInfo, log, pallet_prelude::*, sp_runtime::DispatchErrorWithPostInfo,
@@ -35,6 +41,7 @@ pub mod pallet {
     pub trait Config: frame_system::Config {
         type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
         type WeightInfo: WeightInfo;
+        type Currency: Currency<Self::AccountId> + ReservableCurrency<Self::AccountId>;
 
         /// Limits how many bytes verification key can have.
         ///
@@ -46,6 +53,12 @@ pub mod pallet {
         /// Limits how many bytes proof or public input can have.
         #[pallet::constant]
         type MaximumDataLength: Get<u32>;
+
+        /// Deposit amount for storing a verification key
+        ///
+        /// Will get locked and returned upon deleting the key by the owner
+        #[pallet::constant]
+        type VerificationKeyDepositAmount: Get<VerificationKeyDeposit<Self>>;
     }
 
     #[pallet::error]
@@ -70,6 +83,13 @@ pub mod pallet {
         VerificationFailed(VerificationError),
         /// Proof has been found as incorrect.
         IncorrectProof,
+
+        /// Unsigned request
+        BadOrigin,
+        /// User has insufficient funds to lock the deposit for storing verification key        
+        CannotAffordDeposit,
+        /// Caller is not the owner of the key
+        NotOwner,
     }
 
     #[pallet::event]
@@ -94,12 +114,23 @@ pub mod pallet {
     pub struct Pallet<T>(_);
 
     #[pallet::storage]
+    #[pallet::getter(fn get_verification_key)]
     pub type VerificationKeys<T: Config> = StorageMap<
         _,
         Twox64Concat,
         VerificationKeyIdentifier,
         BoundedVec<u8, T::MaximumVerificationKeyLength>,
     >;
+
+    #[pallet::storage]
+    #[pallet::getter(fn get_verification_key_owner)]
+    pub type VerificationKeyOwners<T: Config> =
+        StorageMap<_, Twox64Concat, VerificationKeyIdentifier, T::AccountId>;
+
+    #[pallet::storage]
+    #[pallet::getter(fn get_verification_key_deposit)]
+    pub type VerificationKeyDeposits<T: Config> =
+        StorageMap<_, Twox64Concat, T::AccountId, VerificationKeyDeposit<T>>;
 
     #[pallet::call]
     impl<T: Config> Pallet<T> {
@@ -114,11 +145,11 @@ pub mod pallet {
         #[pallet::call_index(0)]
         #[pallet::weight(T::WeightInfo::store_key(key.len() as u32))]
         pub fn store_key(
-            _origin: OriginFor<T>,
+            origin: OriginFor<T>,
             identifier: VerificationKeyIdentifier,
             key: Vec<u8>,
         ) -> DispatchResult {
-            Self::bare_store_key(identifier, key).map_err(|e| e.into())
+            Self::bare_store_key(origin, identifier, key).map_err(|e| e.into())
         }
 
         /// Deletes a key stored under `identifier` in `VerificationKeys` map.
@@ -130,7 +161,15 @@ pub mod pallet {
             origin: OriginFor<T>,
             identifier: VerificationKeyIdentifier,
         ) -> DispatchResult {
-            ensure_root(origin)?;
+            let who = ensure_signed(origin).map_err(|_| Error::<T>::BadOrigin)?;
+            let owner = VerificationKeyOwners::<T>::get(identifier)
+                .ok_or(Error::<T>::UnknownVerificationKeyIdentifier)?;
+
+            ensure!(who == owner, Error::<T>::NotOwner);
+
+            let deposit = VerificationKeyDeposits::<T>::get(&owner).unwrap(); // cannot fail since the key has owner and owner must have made a deposit
+            T::Currency::unreserve(&owner, deposit);
+
             VerificationKeys::<T>::remove(identifier);
             Self::deposit_event(Event::VerificationKeyDeleted);
             Ok(())
@@ -139,7 +178,7 @@ pub mod pallet {
         /// Overwrites a key stored under `identifier` in `VerificationKeys` map with a new value `key`
         ///
         /// Fails if `key.len()` is greater than `MaximumVerificationKeyLength`.
-        /// Can only be called by a root account.
+        /// Can only be called by the original owner of the key.
         #[pallet::call_index(2)]
         #[pallet::weight(T::WeightInfo::overwrite_key(key.len() as u32))]
         pub fn overwrite_key(
@@ -147,7 +186,11 @@ pub mod pallet {
             identifier: VerificationKeyIdentifier,
             key: Vec<u8>,
         ) -> DispatchResult {
-            ensure_root(origin)?;
+            let who = ensure_signed(origin).map_err(|_| Error::<T>::BadOrigin)?;
+            let owner = VerificationKeyOwners::<T>::get(identifier)
+                .ok_or(Error::<T>::UnknownVerificationKeyIdentifier)?;
+
+            ensure!(who == owner, Error::<T>::NotOwner);
 
             ensure!(
                 key.len() <= T::MaximumVerificationKeyLength::get() as usize,
@@ -209,9 +252,12 @@ pub mod pallet {
         /// or other dispatchable-related overhead. Thus, it is more suited to call directly from
         /// runtime, like from a chain extension.
         pub fn bare_store_key(
+            origin: OriginFor<T>,
             identifier: VerificationKeyIdentifier,
             key: Vec<u8>,
         ) -> Result<(), Error<T>> {
+            let who = ensure_signed(origin).map_err(|_| Error::<T>::BadOrigin)?;
+
             ensure!(
                 key.len() <= T::MaximumVerificationKeyLength::get() as usize,
                 Error::<T>::VerificationKeyTooLong
@@ -222,10 +268,17 @@ pub mod pallet {
                 Error::<T>::IdentifierAlreadyInUse
             );
 
+            let deposit = T::VerificationKeyDepositAmount::get();
+            T::Currency::reserve(&who, deposit).map_err(|_| Error::<T>::CannotAffordDeposit)?;
+
             VerificationKeys::<T>::insert(
                 identifier,
                 BoundedVec::try_from(key).unwrap(), // must succeed since we've just check length
             );
+
+            // will never overwrite anything since we already check the VerificationKeys map
+            VerificationKeyOwners::<T>::insert(identifier, &who);
+            VerificationKeyDeposits::<T>::insert(&who, deposit);
 
             Self::deposit_event(Event::VerificationKeyStored);
             Ok(())


pallets/baby-liminal/src/lib.rs
```diff
@@ -222,11 +268,18 @@ pub mod pallet {
...
+            let deposit = T::VerificationKeyDepositAmount::get();
+            T::Currency::reserve(&who, deposit).map_err(|_| Error::<T>::CannotAffordDeposit)?;
+
             VerificationKeys::<T>::insert(
                 identifier,
                 BoundedVec::try_from(key).unwrap(), // must succeed since we've just check length
             );
 
+            // will never overwrite anything since we already check the VerificationKeys map
+            VerificationKeyOwners::<T>::insert(identifier, &who);
```
deuszx, line 280: Why do we require that `identifier`s are unique across whole chain? Can't we enrich it with `origin` (or maybe even better 
- an account_id of contract that called the method) and "namespace" it? Why is it that people can "claim" specific identifiers for themselves and prevent others from using it?

fbielejec: There is  2^32 unique identifier possibilities, one per every pair of humans on Earth. I just don't see a use-case in increasing it even further.

deuszx: That's a fair point. I was thinking about other people that would want to use similar technic to derive their keys: `[b'T', b'E', b'S', b'T']` or something.

But this raises another point -- how can people find which keys are taken? Should they use some source of random data for this? If they try to use a mnemonic they could try using something that is already registered. Maybe pallet should be the one to tell under what identifier their key was registered - i.e. turn the dependencies around.

pallets/baby-liminal/src/lib.rs
```diff
@@ -130,7 +161,15 @@ pub mod pallet {
...
             identifier: VerificationKeyIdentifier,
         ) -> DispatchResult {
-            ensure_root(origin)?;
+            let who = ensure_signed(origin).map_err(|_| Error::<T>::BadOrigin)?;
+            let owner = VerificationKeyOwners::<T>::get(identifier)
+                .ok_or(Error::<T>::UnknownVerificationKeyIdentifier)?;
+
+            ensure!(who == owner, Error::<T>::NotOwner);
+
+            let deposit = VerificationKeyDeposits::<T>::get(&owner).unwrap(); // cannot fail since the key has owner and owner must have made a deposit
```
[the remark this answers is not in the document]
fbielejec: yes, although it should probably be benchmarked like the two other ones

pallets/baby-liminal/src/lib.rs
```diff
@@ -46,6 +53,12 @@ pub mod pallet {
         /// Limits how many bytes proof or public input can have.
         #[pallet::constant]
         type MaximumDataLength: Get<u32>;
+
+        /// Deposit amount for storing a verification key
+        ///
+        /// Will get locked and returned upon deleting the key by the owner
+        #[pallet::constant]
+        type VerificationKeyDepositAmount: Get<VerificationKeyDeposit<Self>>;
```
pmikolajczyk41, line 61: I think that we should have something like `DepositPerByte` like pallets contracts and mulitsig have - since we allow for different proof systems, the key lengths may vary from each other. However, in order to avoid problems with such cost changing over time, we should also store what exactly _was_ deposit for every key independently and return that value after key removal.

pallets/baby-liminal/src/lib.rs
```diff
@@ -18,9 +21,12 @@ const STORAGE_VERSION: StorageVersion = StorageVersion::new(0);
 
 /// We store verification keys under short identifiers.
 pub type VerificationKeyIdentifier = [u8; 4];
```
pmikolajczyk41, line 23: up to you, but we may want to take this opportunity and increase length a bit (coming up with 4-letter shortcuts is not always comfy)

pallets/baby-liminal/src/lib.rs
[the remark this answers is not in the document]
pmikolajczyk41: yes, we should move this calculation to benchmark suite

pallets/baby-liminal/src/lib.rs
```diff
@@ -114,40 +151,53 @@ pub mod pallet {
...
             identifier: VerificationKeyIdentifier,
         ) -> DispatchResult {
-            ensure_root(origin)?;
+            let who = ensure_signed(origin).map_err(|_| Error::<T>::BadOrigin)?;
+            let owner = VerificationKeyOwners::<T>::get(identifier)
+                .ok_or(Error::<T>::UnknownVerificationKeyIdentifier)?;
+
+            ensure!(who == owner, Error::<T>::NotOwner);
+
+            let deposit = VerificationKeyDeposits::<T>::get(&owner).unwrap(); // cannot fail since the key has owner and owner must have made a deposit
```
pmikolajczyk41, line 176: if I understand correctly, an account may have up to one key registered - otherwise we will return them all deposits at once

pallets/baby-liminal/src/lib.rs
pmikolajczyk41: I have slight preference to @deuszx 's vision, especially that some people might copy Shielder and want to use the same key identifiers

pallets/baby-liminal/src/lib.rs
```diff
@@ -222,12 +275,19 @@ pub mod pallet {
...
+
             VerificationKeys::<T>::insert(
                 identifier,
                 BoundedVec::try_from(key).unwrap(), // must succeed since we've just check length
             );
 
-            Self::deposit_event(Event::VerificationKeyStored);
+            // will never overwrite anything since we already check the VerificationKeys map
+            VerificationKeyOwners::<T>::insert(identifier, &who);
+            VerificationKeyDeposits::<T>::insert(&who, deposit);
```
pmikolajczyk41, line 288: this will overwrite the previous deposit, won't it?

pmikolajczyk41 (changes requested):

pallets/baby-liminal/src/lib.rs
fbielejec: if we make a  (AccountId, [u;4]) tuple an identifier, that this same identifier would need to be passed from the contracts when verifying.  Let's be honest here, we will be the only ones ever registering those keys.

pallets/baby-liminal/src/lib.rs
fbielejec: done

pallets/baby-liminal/src/lib.rs
fbielejec: changed - there is a tuple stored for deposits now

pallets/baby-liminal/src/lib.rs
pmikolajczyk41: I'm fine with extending 4 bytes to more (8/16)

pallets/baby-liminal/src/lib.rs
fbielejec: it's under a tuple now

pallets/baby-liminal/src/lib.rs
fbielejec: :ok_hand:

pallets/baby-liminal/src/lib.rs
fbielejec: ok

pallets/baby-liminal/src/lib.rs
deuszx: It would be a non-issue if we made it a tupple with _caller's_ `AccountId` - every contract would have its own and we would be passing that automatically -- or rather, read the caller's address in the chain extension. That's IMO the best solution for this.

pallets/baby-liminal/src/lib.rs
```diff
@@ -114,40 +154,61 @@ pub mod pallet {
...
             origin: OriginFor<T>,
             identifier: VerificationKeyIdentifier,
             key: Vec<u8>,
         ) -> DispatchResult {
-            ensure_root(origin)?;
+            let who = ensure_signed(origin).map_err(|_| Error::<T>::BadOrigin)?;
+            let owner = VerificationKeyOwners::<T>::get(identifier);
+            let previous_key = VerificationKeys::<T>::get(identifier);
+
+            if let Some(owner) = owner {
```
deuszx, line 202: When would that be a `None`? If someone calles `ovewrite_key` instead of `store_key`?

fbielejec: yes

bin/runtime/src/lib.rs
```diff
@@ -334,13 +334,16 @@ parameter_types! {
     // We allow 10kB keys, proofs and public inputs. This is a 100% blind guess.
     pub const MaximumVerificationKeyLength: u32 = 10_000;
     pub const MaximumDataLength: u32 = 10_000;
+    pub const VerificationKeyDepositAmount: u128 = 1_000_000_000;
```
pmikolajczyk41, line 337:
```suggestion
    pub const VerificationKeyDepositPerByte: u128 = 1_000_000_000;
```

pmikolajczyk41: and we can use either `LEGACY_DEPOSIT_PER_BYTE`, `CONTRACT_DEPOSIT_PER_BYTE` or `MILLI_AZERO`

pallets/baby-liminal/src/lib.rs
fbielejec: I don't understand this comment - what would be a non-issue?

pallets/baby-liminal/src/lib.rs
```diff
@@ -114,40 +154,61 @@ pub mod pallet {
...
+            Self::bare_store_key(origin, identifier, key).map_err(|e| e.into())
         }
 
         /// Deletes a key stored under `identifier` in `VerificationKeys` map.
         ///
-        /// Can only be called by a root account.
+        /// Returns the deposit locked. Can only be called by the key owner.
         #[pallet::call_index(1)]
-        #[pallet::weight(T::DbWeight::get().writes(1))]
+        #[pallet::weight(T::DbWeight::get().reads(2) + T::DbWeight::get().writes(2))]
```
pmikolajczyk41, line 168: use benchmark results?

fbielejec: trying, but it's brooooken :)

fbielejec: ok, fixed

bin/runtime/src/lib.rs
fbielejec: done

pallets/baby-liminal/src/lib.rs
```diff
@@ -114,40 +154,61 @@ pub mod pallet {
...
+            let owner = VerificationKeyOwners::<T>::get(identifier);
+            let previous_key = VerificationKeys::<T>::get(identifier);
+
+            if let Some(owner) = owner {
+                ensure!(who == owner, Error::<T>::NotOwner)
+            };
+
+            if let Some(previous) = previous_key {
+                ensure!(
+                    previous.len() == key.len(),
```
mike1729, line 208: This check is only due to the guessed value of `MaximumVerificationKeyLength` or there is some other reason?

fbielejec: yes they very well could be - for different proof system for example.

I though about that (returning the difference in deposits) it seemed like adding additional computational overhead for very little gain, we can revisit this though. 

When would a key be overwritten:
- relation changed (unlikey)
-  proof system changed (very unlikey in a production system)
- ... ?

- Thoughts @pmikolajczyk41 ?

fbielejec: Implemented reimbursement / top up when overwriting. So you can overwrite with a key of any length

pallets/baby-liminal/src/lib.rs
```diff
@@ -71,9 +74,8 @@ pub mod pallet {
         UnknownVerificationKeyIdentifier,
         /// Provided verification key is longer than `MaximumVerificationKeyLength` limit.
         VerificationKeyTooLong,
-        /// When you override a key new key stored must be of the same size in bytes
-        NewKeyMustHaveSameLength,
-
+        // /// When you override a key new key stored must be of the same size in bytes
+        // NewKeyMustHaveSameLength,
```
deuszx, line 78: Leftover?

fbielejec: yup

pallets/baby-liminal/src/lib.rs
deuszx: Then why not, instead, remove the `overwrite_key` method and just tell people to use `insert_key` if they want to overwrite?

pallets/baby-liminal/src/lib.rs
```diff
@@ -197,31 +201,48 @@ pub mod pallet {
...
 
             VerificationKeys::<T>::try_mutate_exists(identifier, |value| -> DispatchResult {
                 // should never fail, since length is checked above
-                *value = Some(BoundedVec::try_from(key).unwrap());
+                *value = Some(BoundedVec::try_from(key.clone()).unwrap());
 
                 Ok(())
             })?;
 
+            // TODO : reimburse or lock additional funds
```
deuszx, line 219: Leftover.

pallets/baby-liminal/src/lib.rs
```diff
@@ -197,31 +201,48 @@ pub mod pallet {
...
+            let previous_deposit = VerificationKeyDeposits::<T>::get((&who, &identifier))
+                .unwrap_or_else(|| 0u32.into());
```
deuszx, line 221: Now, that's sketchy. I understand the idea behind this default is to catch a case where `overwrite_key` was called instead of `insert_key`, so there's no `previous_deposit`, but it feels dangerous. I'd rather prefer is `overwrite_key` was not trying to also cover `store_key` (see my previous comment for alternative).

fbielejec: feeling is a weak evidence - if you could give me an example of how this can be dangerous we can change that, but I just don't see it in the situation where the caller locks & unlocks own funds.

pallets/baby-liminal/src/lib.rs
fbielejec: removed

pallets/baby-liminal/src/lib.rs
deuszx: It's just a bad code that tries to do two different things b/c user might make a mistake. There's no `overwrite` in `HashMap` - `insert` does it if the `key` is already taken.

deuszx (changes requested):
At the very least, we should add tests for reimbursements.

pallets/baby-liminal/src/lib.rs
fbielejec: I still don't understand - are you proposing to always be automatically passing the contract callers `AccountId` as an argument to the pallet's functions that need a key identifier? That makes little sense to me, there isn't a 1:1 mapping there, key is registered once and used by many. It's as if every user of the Tornado had to register own key - and for every relation.

pallets/baby-liminal/src/lib.rs
fbielejec: Because you cannot store a key with the same identifier twice

pallets/baby-liminal/src/lib.rs
deuszx: No, it would be used later on by a contract (in our case - `shielder`) so it's registered and used from the same context every time. If someone else wants to deploy their own shielder they can do that and keys won't conflict as the same key will be registered and used from a different context - different contract. Users aren't using the key directly, it's a contract that does it when they post a txn.

pallets/baby-liminal/src/lib.rs
mike1729: Couldn't we just return in the first `if let` in line `203` if there are no `owner` of `identifier`?

pallets/baby-liminal/src/lib.rs
fbielejec: I am completely lost here. Keys are unique for relations, why are we discussing confounding them with accountIds ? E.g. why do I need to keep track of some account id to deposit or withdraw funds?

pallets/baby-liminal/src/lib.rs
fbielejec: Is UPSERT in SQL statements considered bad code then?

pallets/baby-liminal/src/lib.rs
deuszx: This is what I commented on at the very beginning of this thread. Look at how we're choosing the key identifiers: `[b'D', b'E', b'P', b'O', b'S', b'I', b'T']` so not only we're making it "first is the winner" when it comes to taking the keys but also we're making it impossible for users to "just clone" our shielder setup b/c they will have to change all the identifiers under which they're registering the keys.

pallets/baby-liminal/src/lib.rs
deuszx: You're not writing SQL here.

pallets/baby-liminal/src/lib.rs
mike1729: Even if we tell people, they can still use this call and because of that we need this check.

pallets/baby-liminal/src/lib.rs
fbielejec: Yes, bu surely that does not mean there's nothing to learn from 50+ years of RDBMS history and we have to be reinventing the wheel again either.

deuszx: I think the difference is that when working with dbs you're expecting other clients to update the state, so you can't tell whether the record already exists or not. Here, the design prohibits anyone else but the owner of the entry to modify it so it makes more sense to assume (s)he already knows whether it should be calling `store_*` or `overwrite_*`.

pallets/baby-liminal/src/lib.rs
deuszx: Not if we didn't have `overwrite_key` at all and do everything in `store_key`: detect if there already exists an entry for that key and if so overwrite it.

fbielejec: Generally speaking there is a couple good reasons to keep these separate - one is the api of the StorageMap, another is that it gets hard to benchmark two paths like that

deuszx: OK. That's a good point, I can understand that. But aren't you "violating" your own rule right now in `overwrite_key`? If it's user is calling `overwrite_*` instead of `store_*` then you are using the method for not something it was intended to -- it should also be hard to benchmark.

pallets/baby-liminal/src/lib.rs
fbielejec: no they don't, they can the same keys that were already registered. I'd argue this is simpler, easier less error-prone and generally design limiting state space size and duplication in blockchain setting is a good thing

mike1729 (approved):

pallets/baby-liminal/src/lib.rs
fbielejec: The answer is provenance and likeness - it's just easier to support insert for overwrite, it incurs no additional db operations and these dominate the weights.

pallets/baby-liminal/src/tests/suite.rs
```diff
@@ -51,44 +67,63 @@ fn does_not_overwrite_registered_key() {
...
+        let reserved_balance_before = reserved_balance(1);
+        let deposit = put_key();
+        let reserved_balance_after = reserved_balance(1);
+
+        assert_eq!(reserved_balance_after - reserved_balance_before, deposit);
+        assert_ok!(BabyLiminal::overwrite_key(owner(), IDENTIFIER, vk()));
+
+        assert_eq!(reserved_balance_after, reserved_balance(1));
+        assert_ok!(BabyLiminal::delete_key(owner(), IDENTIFIER));
+        assert_eq!(reserved_balance_before, reserved_balance(1));
```
deuszx, line 126: We're still missing the test for overwriting with a key of different length.

pallets/baby-liminal/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![cfg_attr(not(feature = "std"), no_std)]
+#![recursion_limit = "256"]
```
deuszx, line 2: Can you put as a comment explanation why we need this? I've never seen it anywhere.

pallets/baby-liminal/src/lib.rs
```diff
@@ -114,40 +155,56 @@ pub mod pallet {
...
             identifier: VerificationKeyIdentifier,
         ) -> DispatchResult {
-            ensure_root(origin)?;
+            let who = ensure_signed(origin).map_err(|_| Error::<T>::BadOrigin)?;
+            let owner = VerificationKeyOwners::<T>::get(identifier)
+                .ok_or(Error::<T>::UnknownVerificationKeyIdentifier)?;
+
+            ensure!(who == owner, Error::<T>::NotOwner);
+
+            let deposit = VerificationKeyDeposits::<T>::take((&owner, &identifier)).unwrap(); // cannot fail since the key has owner and owner must have made a deposit
```
deuszx, line 180: But if it _does_ fail, it will crash "just" the runtime or the whole node?

pallets/baby-liminal/src/lib.rs
```diff
@@ -114,40 +155,56 @@ pub mod pallet {
...
             Ok(())
         }
 
         /// Overwrites a key stored under `identifier` in `VerificationKeys` map with a new value `key`
         ///
         /// Fails if `key.len()` is greater than `MaximumVerificationKeyLength`.
-        /// Can only be called by a root account.
+        /// Can only be called by the original owner of the key.
+        /// It will require the caller to lock up additional funds (if the new key occupies more storage)
+        /// or reimburse the difference if it is shorter in it's byte-length.
```
deuszx, line 193:
```suggestion
        /// or reimburse the difference if it is shorter in its byte-length.
```

pallets/baby-liminal/src/tests/suite.rs
fbielejec: added

pallets/baby-liminal/src/lib.rs
fbielejec: because compiler asked. there's one in runtime

pallets/baby-liminal/src/lib.rs
fbielejec: runtime

pallets/baby-liminal/src/lib.rs
deuszx: Still pretty sure I'm correct: it will reimburse the difference if the new key is shorter - if _its_ byte-length is shorter.

pallets/baby-liminal/src/tests/suite.rs
```diff
@@ -51,44 +71,73 @@ fn does_not_overwrite_registered_key() {
...
+#[test]
+fn key_deposits() {
+    new_test_ext().execute_with(|| {
+        let reserved_balance_before = reserved_balance(1);
+        let deposit = put_key();
+        let reserved_balance_after = reserved_balance(1);
+
+        assert_eq!(reserved_balance_before + deposit, reserved_balance_after);
+
+        let key_size = 1000;
```
deuszx, line 127: Can we do `vk().len() * 2` instead? So that it's obvious we will be overwriting with a key that is longer. This also prevents from future mistakes when the `vk()` is replaced with something that is longer than 1000.

pallets/baby-liminal/src/tests/suite.rs
```diff
@@ -51,44 +71,73 @@ fn does_not_overwrite_registered_key() {
...
+        let per_byte_fee: u64 =
+            <TestRuntime as crate::Config>::VerificationKeyDepositPerByte::get();
+        let long_key = vec![0u8; key_size];
+
+        let free_balance_before = free_balance(1);
+        assert_ok!(BabyLiminal::overwrite_key(owner(), IDENTIFIER, long_key));
```
deuszx, line 133: We're missing a test for overwriting with a key shorter than `vk()`.

fbielejec: added

pallets/baby-liminal/src/tests/suite.rs
fbielejec: ok

pallets/baby-liminal/src/lib.rs
deuszx, line 180: Hurts my eyes that this line is so long due to inlining the comment.

fbielejec: I think you should make a pipeline for that

fbielejec: Because nothing & nobody else thinks this is a problem

deuszx: Ah... looks like we never put anything about line length into the rustfmt -- I could have a line long of 1000 chars and it would still pass the formatter 🙈

pallets/baby-liminal/src/benchmarking/suite.rs
```diff
@@ -35,7 +40,14 @@ benchmarks! {
     overwrite_key {
         let l in 1 .. T::MaximumVerificationKeyLength::get();
         let key = vec![0u8; l as usize];
-    } : _(RawOrigin::Root, IDENTIFIER, key)
+        let _ = insert_key::<T>(key.clone ());
+    } : _(caller::<T>(), IDENTIFIER, key)
```
deuszx, line 44: I think this benchmark is incomplete. This benchmark does the following:

**[setup phase]**
1. Generate a key `k` of length `l`
2. Write `k` to storage

**[actual benchmark]**
1. `overwrite_key(k)`

Since `k` was already stored in the setup phase, the call to `overwrite_key` we're benchmarking will exit early when comparing old and new keys will have the same length (cmp operator would return `Equal`). We're skipping one `T::Currency::(un)reserve` and one write to storage (replacing previous deposit with new one.

mike1729: I believe the benchmark runs `overwrite` for keys of increasing length so the path with reserve is covered but unreserved is indeed missing.

fbielejec: there is no diff in # read/writes whether it's reserve or unreserve.

deuszx: > I believe the benchmark runs `overwrite` for keys of increasing length so the path with reserve is covered but unreserved is indeed missing.

I would have thought that only `_(caller::<T>(), IDENTIFIER, key)` is benchmarked, not the setup phase, so either the setup is wrong and we keep overwriting there (`insert_key(k_len=1)`, `insert_key(key_len=2)`, ..., `insert_key(key_len=l)`) and benchmark just the last insert or the benchmarked method is run for every setup separately - we insert key with length 1 and try to overwrite with the same key etc.

deuszx: I can't believe that this:
```rust
Equal => {
                            // do nothing
                        }
```
is equal in terms of complexity to:
```rust
Greater => {
    // lock the difference deposit - prev
    let difference = deposit - previous_deposit;
    T::Currency::reserve(&who, difference)
        .map_err(|_| Error::<T>::CannotAffordDeposit)?;
    *maybe_previous_deposit = Some(deposit);
}
```
but it seems like I'm the only one interested in checking this so I'll approve it now.

deuszx (approved):

pmikolajczyk41 (approved):
blind approval to unblock merging

PR #968, opened 2023-03-06
[A0-2073] Update dependencies to stable ink4.

deuszx:
Updates `aleph_client` and `cliain` to use stable versions of `contract-*` crates used for interacting with the chain.

Updates contracts' in `/contracts` repository to use stable ink4 release (instead of beta).

### DIFF
--- a/aleph-client/Cargo.lock
+++ b/aleph-client/Cargo.lock
@@ -74,12 +74,10 @@ version = "2.13.0"
 dependencies = [
  "anyhow",
  "async-trait",
- "contract-metadata 2.0.1",
  "contract-transcode",
  "frame-support",
  "futures",
  "hex",
- "ink_metadata",
  "log",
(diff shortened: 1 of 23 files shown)


h4nsu (approved):
If we're doing the "super-hard-freeze" with ~, we could consider going with the most recent version (~4.0.1) rather than 4.0.0. Just a suggestion, probably doesn't make much difference

deuszx:
> If we're doing the "super-hard-freeze" with ~, we could consider going with the most recent version (~4.0.1) rather than 4.0.0. Just a suggestion, probably doesn't make much difference

`~4.0.0` should mean it's `>= 4.0.0 and < 4.1.0` if I'm reading correctly: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

deuszx:
Superseeded by #995

PR #972, opened 2023-03-07
Fail early with helpful error message.

deuszx:
If `aleph-node:latest` docker image can't be found locally we should fail early and inform user with actionable instructions. Otherwise docker will try to pull `aleph-node:latest` from the official docker hub and fail with confusing messages.

## Type of change

### DIFF
--- a/.github/scripts/run_consensus.sh
+++ b/.github/scripts/run_consensus.sh
@@ -17,6 +17,15 @@ MIN_VALIDATOR_COUNT=4
 
 export NODE_IMAGE
 
+if [[ $(docker images | grep ${NODE_IMAGE}) ]]; then
+  echo "${NODE_IMAGE} found locally"
+else
+  echo "${NODE_IMAGE} not found locally."
+  echo "Build image first with:"
+  echo "docker build -t aleph-node:latest -f docker/Dockerfile ."
+  exit 1;
+fi
+
 mkdir -p docker/data/
 
 function usage {


.github/scripts/run_consensus.sh
```diff
@@ -17,6 +17,15 @@ MIN_VALIDATOR_COUNT=4
 
 export NODE_IMAGE
 
+if [[ $(docker images | grep ${NODE_IMAGE}) ]]; then
+  echo "${NODE_IMAGE} found locally"
+else
+  echo "${NODE_IMAGE} not found locally."
+  echo "Build image first with:"
+  echo "docker build -t aleph-node:latest -f docker/Dockerfile ."
```
fixxxedpoint, line 25:
```suggestion
  echo "docker build -t ${NODE_IMAGE} -f docker/Dockerfile ."
```

fixxxedpoint (approved):

.github/scripts/run_consensus.sh
```diff
@@ -17,6 +17,15 @@ MIN_VALIDATOR_COUNT=4
 
 export NODE_IMAGE
 
+if [[ -z $(docker inspect ${NODE_IMAGE} | grep "No such object") ]]; then
```
deuszx, line 20: @fixxxedpoint , I had to change this condition as `grep aleph-node:latest` didn't match any - the output of `docker images` is:
```sh
~/aleph/aleph-node$ docker images
REPOSITORY                                  TAG       IMAGE ID       CREATED          SIZE
aleph-node                                  latest    006771953aa3   10 minutes ago   291MB
```
hence `grep aleph-node:latest` didn't catch it. Now we call `docker inspect` which respects name and tag.

Marcin-Radecki, line 20: Alternatively, we can just check output status of `docker inspect`, as it is more future-proof, ie output of the command might change etc 
```
if docker inspect "${NODE_IMAGE}" > /dev/null; then
```

Marcin-Radecki (approved):

.github/scripts/run_consensus.sh
deuszx: Like this:
```sh
if [[ $(docker inspect ${NODE_IMAGE} > /dev/null) ]]; then 
    echo "image found"; 
else 
    echo "image not found"; 
fi
```
?
because I'm getting `image not found` even though it is available.

PR #985, opened 2023-03-07
Merge yamllint changes

obrok:
Hopefully, this is the last big thing we need to merge in from main and we can merge into main afterwards.


.github/workflows/contracts-deploy.yml
```diff
@@ -50,11 +56,14 @@ jobs:
 
       - name: Install binaryen
         shell: bash
+        # yamllint disable rule:line-length
```
deuszx, line 59: Do we need the `binaryen` (and `cargo contract` below) if we're using dockerized `cargo contract` in `contracts/scripts/deploy.sh`?

obrok: Honestly, I tried to not touch this file as much as possible, I don't really know how to test the changes. Perhaps @fbielejec could check the deployments again after all this is merged.

fbielejec: I would definitely not touch any workflow files before it's all unified

fbielejec (approved):

PR #994, opened 2023-03-09
Dockerize contracts/scripts/test.sh

obrok:
It should now work with nodes started with `.github/scripts/run_consensus.sh` out of the box, as described in the comment in the file.

### DIFF
--- a/contracts/scripts/test.sh
+++ b/contracts/scripts/test.sh
@@ -9,6 +9,9 @@
 
 set -euox pipefail
 
+# Move to the root of the repo
+cd "$(dirname -- "$(readlink -f -- "${BASH_SOURCE[@]}")")/../.."
+
 export CONTRACTS_PATH
 CONTRACTS_PATH=$(pwd)/contracts
 
--- a/contracts/scripts/test.sh
+++ b/contracts/scripts/test.sh
@@ -1,15 +1,35 @@
 #!/bin/bash
 
-set -euo pipefail
+# To run this:
+#
+# .github/scripts/run_consensus.sh
+# source contracts/env/dev
+# contracts/scripts/deploy.sh
+# contracts/scripts/test.sh
 
-E2E_PATH=$(pwd)/e2e-tests
+set -euox pipefail
+
+export CONTRACTS_PATH
 CONTRACTS_PATH=$(pwd)/contracts
 
-# shellcheck source=contracts/scripts/test_env.sh
-. "$CONTRACTS_PATH/scripts/test_env.sh"
-
-pushd "$E2E_PATH"
-cargo test button -- --test-threads 1 --nocapture
-popd
-
-exit $?
+docker run \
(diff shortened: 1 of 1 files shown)


contracts/scripts/test.sh
```diff
@@ -1,15 +1,35 @@
...
+# source contracts/env/dev
+# contracts/scripts/deploy.sh
+# contracts/scripts/test.sh
 
-E2E_PATH=$(pwd)/e2e-tests
-CONTRACTS_PATH=$(pwd)/contracts
-
-# shellcheck source=contracts/scripts/test_env.sh
-. "$CONTRACTS_PATH/scripts/test_env.sh"
+set -euox pipefail
```
deuszx, line 10: Can we check somewhere at the beginning here that the `${PWD}` is root directory of the `aleph-node` project? Otherwise it will not work.

contracts/scripts/test.sh
```diff
@@ -1,15 +1,35 @@
...
+# .github/scripts/run_consensus.sh
+# source contracts/env/dev
+# contracts/scripts/deploy.sh
+# contracts/scripts/test.sh
 
-E2E_PATH=$(pwd)/e2e-tests
-CONTRACTS_PATH=$(pwd)/contracts
-
-# shellcheck source=contracts/scripts/test_env.sh
-. "$CONTRACTS_PATH/scripts/test_env.sh"
```
deuszx, line 9: We can remove the file now, right?

contracts/scripts/test.sh
```diff
@@ -1,15 +1,38 @@
...
+  -e SIMPLE_DEX_METADATA="/code/contracts/simple_dex/target/ink/simple_dex.json" \
+  -e WRAPPED_AZERO_METADATA="/code/contracts/wrapped_azero/target/ink/wrapped_azero.json" \
+  -e RUST_LOG="aleph_e2e_client=info" \
+  -v "$(pwd)":/code \
+  -v ~/.cargo/registry:/usr/local/cargo/registry \
+  -v ~/.cargo/git:/usr/local/cargo/git \
+  -w /code/e2e-tests \
+  --rm \
+  --network host \
+  rust:1.67-buster \
```
deuszx, line 37: Now that I think of it - maybe we should reuse the `ink-dev` we already have, instead of downloading yet another image. It [`ink-dev`] should already have all the necessary dependencies to run the test.

deuszx:
```sh
mateusz@DESKTOP-O514E9R:~$ docker images
REPOSITORY                           TAG           IMAGE ID       CREATED        SIZE
rust                                 1.67-buster   2b1b9476651a   7 days ago     1.36GB
public.ecr.aws/p6e8q1z1/ink-dev      1.0.0         5e90a54426fa   9 days ago     867MB
```

deuszx (approved):

PR #998, opened 2023-03-10
A0-1797: Block sync service

timorl:
# Description

The rest of the block sync logic.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/sync/mod.rs
+++ b/finality-aleph/src/sync/mod.rs
@@ -91,6 +91,7 @@ pub enum ChainStatusNotification<H: Header> {
 }
 
 /// A stream of notifications about the chain status in the database changing.
+/// We assume that this will return all the events, otherwise we will end up with a broken state.
 #[async_trait::async_trait]
 pub trait ChainStatusNotifier<H: Header> {
     type Error: Display;
--- a/finality-aleph/src/sync/mod.rs
+++ b/finality-aleph/src/sync/mod.rs
@@ -10,6 +10,7 @@ mod forest;
 mod handler;
 #[cfg(test)]
 mod mock;
+mod service;
 mod substrate;
 mod task_queue;
 mod ticker;
@@ -19,9 +20,9 @@ pub use substrate::SessionVerifier;
 const LOG_TARGET: &str = "aleph-block-sync";
 
 /// The identifier of a connected peer.
-pub trait PeerId: Clone + Hash + Eq {}
+pub trait PeerId: Debug + Clone + Hash + Eq {}
 
-impl<T: Clone + Hash + Eq> PeerId for T {}
+impl<T: Debug + Clone + Hash + Eq> PeerId for T {}
 
 /// The identifier of a block, the least amount of knowledge we can have about a block.
 pub trait BlockIdentifier: Clone + Hash + Debug + Eq + Codec + Send + Sync + 'static {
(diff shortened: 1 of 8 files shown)


finality-aleph/src/sync/service.rs
```diff
@@ -0,0 +1,274 @@
...
+    fn handle_justifications(&mut self, justifications: Vec<J::Unverified>, peer: N::PeerId) {
+        use SyncAction::*;
+        let mut previous_block_id = None;
+        for justification in justifications {
+            let maybe_block_id = match self
+                .handler
+                .handle_justification(justification, peer.clone())
+            {
+                Ok(action) => match action {
+                    Response(data) => {
```
lesniak43, line 157: when will this branch run?

finality-aleph/src/sync/service.rs
```diff
@@ -0,0 +1,274 @@
...
+        match self.handler.handle_state(state, peer.clone()) {
+            Ok(action) => self.perform_sync_action(action, peer),
+            Err(e) => warn!(
+                target: LOG_TARGET,
+                "Error handling sync state from {:?}: {}.", peer, e
+            ),
+        }
+    }
+
+    fn handle_justifications(&mut self, justifications: Vec<J::Unverified>, peer: N::PeerId) {
```
lesniak43, line 148: Please add some comments to this method, sth. like "this will create backup requests for every but last justification" and so on.
Do we assume that `justifications` is sorted by block number? If we do, then the first step should be to sort the Vec, as it comes from the network.

finality-aleph/src/sync/service.rs
```diff
@@ -0,0 +1,274 @@
...
+    fn backup_request(&mut self, block_id: BlockIdFor<J>) {
+        self.tasks.schedule_in(block_id, Duration::from_secs(5));
+    }
+
+    fn delayed_request(&mut self, block_id: BlockIdFor<J>) {
+        self.tasks.schedule_in(block_id, Duration::from_millis(500));
+    }
+
+    fn request(&mut self, block_id: BlockIdFor<J>) {
+        self.tasks.schedule_in(block_id, Duration::ZERO);
```
lesniak43, line 80: Do we check if tasks are duplicated?

lesniak43 (approved):
Seems ok, although we really need to add some integration tests ASAP.

finality-aleph/src/sync/mod.rs
```diff
@@ -82,20 +83,20 @@ pub trait Finalizer<J: Justification> {
...
+pub enum ChainStatusNotification<H: Header> {
     /// A block has been imported.
-    BlockImported(BI),
+    BlockImported(H),
     /// A block has been finalized.
-    BlockFinalized(BI),
+    BlockFinalized(H),
 }
 
 /// A stream of notifications about the chain status in the database changing.
```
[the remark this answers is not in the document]
timorl: Yup, at least within categories. It would be extremely strange if they didn't, underneath they have a channel with just one service producing them.

finality-aleph/src/sync/service.rs
timorl: I added a separate task for a rework of this – there is currently even a somewhat convoluted vulnerability here due to this assumption, but merging the PR with this bug will let us work on this in parallel with other issues.

finality-aleph/src/sync/mod.rs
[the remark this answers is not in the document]
lesniak43: can we at least add a comment that if at least one block import notification comes in out of order, then our code enters an unrecoverable state?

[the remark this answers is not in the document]
lesniak43: I'm pretty sure that at some point we've decided that we won't fully trust the notifier provided by substrate, due to the risk of missing notifications.

finality-aleph/src/sync/service.rs
timorl: Hm, right, reworked the interface a tad.

finality-aleph/src/sync/mod.rs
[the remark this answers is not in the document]
timorl: Yeah, but we later checked and they actually provide all the notifications if you don't do something weird (although maybe creating blocks will make this strange?). Anyway, I think if there are any problems they should be in assumptions that will come out during practical testing, and then we will fix this if it's broken.

finality-aleph/src/sync/service.rs
```diff
@@ -0,0 +1,266 @@
...
+            }
+            Uninterested => (),
+        }
+    }
+
+    fn handle_chain_event(&mut self, event: ChainStatusNotification<J::Header>) {
+        use ChainStatusNotification::*;
+        match event {
+            BlockImported(header) => {
+                if let Err(e) = self.handler.block_imported(header) {
```
lesniak43, line 234: This error is converted to a log and ignored - how are we expected to recover from it?

lesniak43 (approved):

finality-aleph/src/sync/service.rs
lesniak43: Ok, we're not, by design.

PR #999, opened 2023-03-10
A0-2090: add custom substrate test client with BlockNumber = u32

maciejnems:
# Description

Test client substrate provides has hardcoded `Header<Number = u64>`. This is incompatible with some interfaces we have (they require `Header<Number = u32>`). Because of that we our own custom `TestClient`, `TestClientBuilder`, and `TestClientBuilderExt`.  This aims to do that and replaces all usages of old TestClient with the new one.

It also aims to unify testing folder a little bit.


finality-aleph/src/testing/mocks/mod.rs
```diff
@@ -1,22 +1,28 @@
...
+pub use session_info::{SessionInfoProviderImpl, VerifierWrapper};
+use sp_runtime::traits::BlakeTwo256;
+use substrate_test_runtime::Extrinsic;
 
-pub(crate) type TBlock = substrate_test_runtime::Block;
-pub(crate) type THeader = substrate_test_runtime::Header;
-pub(crate) type THash = substrate_test_runtime::Hash;
-pub(crate) type TNumber = substrate_test_runtime::BlockNumber;
+type Hashing = BlakeTwo256;
+pub type TBlockNumber = BlockNumber;
```
timorl, line 16: Hm, since the whole purpose of this exercise is to make `BlockNumber` the only block number we use I think we should just use it for tests as well, directly and without this alias. What do you think?

timorl (approved):
One wide ranging comment for discussion, other than that LGTM.

finality-aleph/src/testing/mocks/mod.rs
maciejnems: Yeah, I thought about it. We can change all those occurrences to `BlockNumber`. My thinking was to leave it as it is in case it changes in the future but it will likely never happen.

timorl: Yeah, I would change it to `BlockNumber` then.

lesniak43 (approved):

PR #1000, opened 2023-03-10
A0-1876: refactor session info

maciejnems:
# Description

Previously session info was either taken from implementations of `SessionInfo` trait, or functions that were inside this trait. Now the functions are private, and session info is taken from `SessionInfo` struct. 

`SessionBoundries` struct and `AuthorityProvider` trait will be refactored in another PR to not be generic over `B: Block`

## Type of change

- Refactor

### DIFF
--- a/finality-aleph/src/nodes/mod.rs
+++ b/finality-aleph/src/nodes/mod.rs
@@ -18,7 +18,7 @@ use crate::{
     },
     mpsc,
     mpsc::UnboundedSender,
-    session::SessionInfo as SessionBoundInfo,
+    session::SessionBoundaryInfo,
     session_map::ReadOnlySessionMap,
     sync::SessionVerifier,
     BlockchainBackend, JustificationNotification, Metrics, MillisecsPerBlock, SessionPeriod,
@@ -45,14 +45,14 @@ struct JustificationParams<B: Block, H: ExHashT, C, BB> {
 
 struct SessionInfoProviderImpl {
     session_authorities: ReadOnlySessionMap,
-    session_info: SessionBoundInfo,
+    session_info: SessionBoundaryInfo,
 }
 
 impl SessionInfoProviderImpl {
     fn new(session_authorities: ReadOnlySessionMap, session_period: SessionPeriod) -> Self {
         Self {
             session_authorities,
-            session_info: SessionBoundInfo::new(session_period),
+            session_info: SessionBoundaryInfo::new(session_period),
         }
     }
 }
--- a/finality-aleph/src/nodes/mod.rs
+++ b/finality-aleph/src/nodes/mod.rs
@@ -3,6 +3,7 @@ mod validator_node;
 
 use std::{future::Future, sync::Arc};
 
+use aleph_primitives::BlockNumber;
 pub use nonvalidator_node::run_nonvalidator_node;
 use sc_client_api::Backend;
 use sc_network::NetworkService;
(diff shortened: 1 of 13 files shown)


finality-aleph/src/session.rs
```diff
@@ -29,25 +27,44 @@ impl<B: Block> SessionBoundaries<B> {
...
+    pub fn new(session_period: SessionPeriod) -> Self {
+        Self { session_period }
+    }
+}
+
+/// Struct for getting the session boundaries.
+impl SessionInfo {
+    /// Returns session id of the session that block belongs to.
+    pub fn session_id_from_block_num(&self, n: BlockNumber) -> SessionId {
+        session_id_from_block_num(n, self.session_period)
```
timorl, line 56: Eh, I'm pretty sure this would be better defined just internally, without a separate function. It seems the only reason to have the separate function (although not even for this specific case!) is `SessionBoundaries`, but that I would say should be returned by a method of `SessionInfo`, e.g. `boundaries_for_session(&self, id: SessionId) -> SessionBoundaries`. You don't need the separate functions then (I think?).

finality-aleph/src/nodes/mod.rs
```diff
@@ -15,9 +16,9 @@ use crate::{
     justification::{
         JustificationHandler, JustificationRequestSchedulerImpl, SessionInfo, SessionInfoProvider,
     },
-    last_block_of_session, mpsc,
+    mpsc,
     mpsc::UnboundedSender,
-    session_id_from_block_num,
+    session::SessionInfo as SessionBoundInfo,
```
timorl, line 21: This suggests that perhaps this struct should be called `SessionBoundaryInfo` or something, since `SessionInfo` is a bit too generic and results in confusion with different session information here.

finality-aleph/src/session_map.rs
```diff
@@ -14,19 +14,17 @@ use tokio::sync::{
...
-    first_block_of_session, session_id_from_block_num, ClientForAleph, SessionId, SessionPeriod,
-};
+use crate::{session::SessionInfo, ClientForAleph, SessionId, SessionPeriod};
 
 const PRUNING_THRESHOLD: u32 = 10;
 type SessionMap = HashMap<SessionId, SessionAuthorityData>;
 type SessionSubscribers = HashMap<SessionId, Vec<OneShotSender<SessionAuthorityData>>>;
 
-pub trait AuthorityProvider<B> {
+pub trait AuthorityProvider<N> {
```
timorl, line 23: Eh? What happened here?

lesniak43 (approved):

timorl (approved):
Some small comments, but in general great improvement.

finality-aleph/src/session.rs
lesniak43: +1

finality-aleph/src/session_map.rs
maciejnems: I changed it to `N`, as this was not generic over `Block`, but over number. Not related to the PR, but wanted to squeeze that in somewhere.

finality-aleph/src/nodes/mod.rs
maciejnems: Yeeeaaah, this could definitely be done. Will do.

finality-aleph/src/session.rs
maciejnems: Great idea, will do

timorl: Sure, sounds good. I only want it done, doesn't have to be in this PR. :P

finality-aleph/src/session_map.rs
timorl: Ah, could you then also change the `block` in the calls to `block_number` or something, to make this clearer?

PR #1011, opened 2023-03-14
A0-2102: Bump EndBug/add-and-commit GH action version from 5.1.0 to 9.1.1

mikogs:
This PR updates EndBug/add-and-commit action from version 5.1.0 to 9.1.1, as suggested by Dependabot on https://github.com/Cardinal-Cryptography/aleph-node/pull/980.

It's quite a bump so changelog has been checked and there are breaking changes in version 7.0.0: https://github.com/EndBug/add-and-commit/releases/tag/v7.0.0 
A major breaking change is the token input that has been removed.  That token will be now taken from the previously used `actions/checkout@v3` action.  Thus, if a repository `aleph-apps` is checked out in `aleph-apps` directory, our `add-and-commit` action can be used on that directory without passing the token.

New version has been tested on some temporary branches:  
* the following workflow has been created on a temporary branch: https://github.com/Cardinal-Cryptography/aleph-node/blob/test-new-addandcommit-action/.github/workflows/test-addandcommit-action.yml
* and result of its run can be found here: https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/4411153835/jobs/7729381015 
* along with the commit that it was meant to add here: https://github.com/Cardinal-Cryptography/aleph-apps/commit/62c3c51fb4ff4940edb9d636e63a89168005ffdb . 

Please check above to ensure it's all right.

### DIFF
--- a/.github/workflows/deploy-mainnet.yml
+++ b/.github/workflows/deploy-mainnet.yml
@@ -100,7 +100,7 @@ jobs:
             aleph-runtime-${{ steps.vars.outputs.sha_short }}.tar.gz
 
       - name: GIT | Checkout aleph-apps repo
-        uses: actions/checkout@master
+        uses: actions/checkout@v3
         with:
           ref: mainnet
           repository: Cardinal-Cryptography/aleph-apps
@@ -132,16 +132,13 @@ jobs:
           done
 
       - name: GIT | Commit changes to aleph-apps repository.
-        uses: EndBug/add-and-commit@v5.1.0
+        uses: EndBug/add-and-commit@v9.1.1
         with:
           author_name: AlephZero Automation
           author_email: alephzero@10clouds.com
           message: "Updating Mainnet docker image tag for release: ${{ steps.vars.outputs.branch }}"
           add: "*.yaml"
           cwd: "aleph-apps"
-          branch: mainnet
-        env:
-          GITHUB_TOKEN: ${{ secrets.CI_GH_TOKEN }}
 
   slack:
     name: Slack notification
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy-mainnet.yml
```diff
@@ -132,16 +132,13 @@ jobs:
...
       - name: GIT | Commit changes to aleph-apps repository.
-        uses: EndBug/add-and-commit@v5.1.0
+        uses: EndBug/add-and-commit@v9.1.1
         with:
           author_name: AlephZero Automation
           author_email: alephzero@10clouds.com
           message: "Updating Mainnet docker image tag for release: ${{ steps.vars.outputs.branch }}"
           add: "*.yaml"
           cwd: "aleph-apps"
-          branch: mainnet
```
timorl, line 142: Just to make sure – this is now not necessary, because the checkout step picks the branch and that's enough?

timorl (approved):

.github/workflows/deploy-mainnet.yml
Kosiare: I got the same question like @timorl

Kosiare (approved):

Marcin-Radecki (approved):

.github/workflows/deploy-mainnet.yml
mikogs: yes @timorl , that's correct.  If you check the temporary workflow that has been created to test this, branch was taken from `actions/checkout@v3`.  There is no `branch` input in the latest version of `EndBug/add-and-commit`.

PR #1017, opened 2023-03-14
Run checks on pallet changes

timorl:
# Description

Excluded packages should be checked on pallet changes as well.

## Type of change

# Checklist:


timorl:
This only confuses pipelines more, bad idea...

PR #1021, opened 2023-03-15
Removed paths from workflows

Marcin-Radecki:
# Description

This PR removed `paths` from workflows, so now all checks are triggered unconditionally. This has below motivations:
* we've had an issue that checks excluded packages broke due to one PR slipped change that updated excluded package build, as this workflow has bug which does not trigger it when `relations`  
* build and push cliain worfklow has bug which does not trigger on some paths, e.g. when `aleph-client` is changed

This PR also removes `matrix` from check excluded packages as now we run N times check excluded packages, we don't need that matrix anyway as single workflow run takes now 8 minutes. 

## Type of change

# Checklist:


.github/workflows/build-and-push-cliain.yml
```diff
@@ -6,12 +6,7 @@ on:
     branches:
       - 'main'
     paths:
```
obrok, line 8:
```suggestion
```

deuszx, line 8: Why not remove the `paths` entry entirely?

Marcin-Radecki: Right, will remove

obrok (approved):
I'm not a fan of `paths`, because I think it's too easy to make a bug (like not building when a dependency changes), so +1 from me. I prefer the `matrix` approach even when the builds are fast, but it does have some drawbacks too, so no objections from me there.

.github/workflows/build-and-push-cliain.yml
Marcin-Radecki: done

Marcin-Radecki: done

deuszx (approved):

PR #1023, opened 2023-03-15
fix clean script

fbielejec:
# Description

Fixes the clean script

### DIFF
--- a/contracts/scripts/clean.sh
+++ b/contracts/scripts/clean.sh
@@ -34,26 +34,11 @@ function get_address {
 
 function remove_contract_code {
   local code_hash=$(cat "$CONTRACTS_PATH"/addresses.json | jq --raw-output ".$1")
-
   if [ "None" == "$(docker run --network host -e RUST_LOG=info "$CLIAIN_IMAGE" --seed "$AUTHORITY_SEED" --node "$NODE" contract-owner-info --code-hash "$code_hash")" ]; then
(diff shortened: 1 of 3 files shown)


docker/smartnet-compose.yml
```diff
@@ -32,6 +32,7 @@ services:
       - 0.0.0.0:30333:30333
       - 0.0.0.0:9933:9933
       - 0.0.0.0:9943:9943
+      - 0.0.0.0:9944:9943
```
mike1729, line 35: Shouldn't it be 9944:9944?

mike1729 (approved):
One tiny thing, besides that g2g.

docker/smartnet-compose.yml
fbielejec: it's on purpose, it bind's the same port to two ports on host - some tools default to 9943, some to 9944 so this satisfies both

deuszx: This is surprising since a node binds to websocket to either 9943 OR 9944. Like here, https://github.com/Cardinal-Cryptography/aleph-node/blob/main/docker/docker-compose.base.yml#L24 -- node0 binds to 9943 while node1 to 9944 so this here is fixing something that should have been fixed on the "client" side - i.e. decide to which node the client wants to talk to. The underlying node (in this docker) can't bind to both at the same time.

PR #1028, opened 2023-03-16
refactor SessionBoundaries

maciejnems:
# Description

After this PR;
* `HashNum` only be created with `N = BlockNumber`
* `SessionBoundaries` no longer is generic over `B: Block`
* `SessionBoundaries` is created using `SessionBoundaryInfo` as suggested in https://github.com/Cardinal-Cryptography/aleph-node/pull/1000#discussion_r1133938346

Next steps for the refactor will be:
* `AuthorityProvider` not generic over `N` and `FinalityNotificator` not generic over `B` in result `SessionMapUpdater` does not need to be generic over `B`
* `RequestBlocks` not generic over `B`. Could be generic over `H: Hash`
* other changes, for example somehow unifying `HashNum` with `BlockIdentifier` in `sync`

## Type of change

- Refactor

### DIFF
--- a/finality-aleph/src/data_io/chain_info.rs
+++ b/finality-aleph/src/data_io/chain_info.rs
@@ -1,9 +1,10 @@
 use std::sync::Arc;
 
+use aleph_primitives::BlockNumber;
 use log::error;
 use lru::LruCache;
 use sc_client_api::HeaderBackend;
-use sp_runtime::traits::{Block as BlockT, Header as HeaderT, NumberFor, One};
+use sp_runtime::traits::{Block as BlockT, Header as HeaderT, NumberFor};
 
 use crate::{data_io::ChainInfoCacheConfig, BlockHashNum};
 
@@ -20,6 +21,7 @@ pub trait ChainInfoProvider<B: BlockT> {
 impl<C, B> ChainInfoProvider<B> for Arc<C>
 where
     B: BlockT,
+    B::Header: HeaderT<Number = BlockNumber>,
     C: HeaderBackend<B>,
 {
     fn is_block_imported(&mut self, block: &BlockHashNum<B>) -> bool {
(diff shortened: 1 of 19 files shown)


finality-aleph/src/data_io/chain_info.rs
```diff
@@ -197,7 +201,7 @@ where
         let mut curr_block = self.aux_finalized.clone();
         while curr_block.num > num {
             let parent_hash = self.chain_info_provider.get_parent_hash(&curr_block)?;
-            curr_block = (parent_hash, curr_block.num - NumberFor::<B>::one()).into();
+            curr_block = (parent_hash, curr_block.num - 1).into();
```
kostekIV, line 204: nit: I would still use BlockNumber::one() or B::Header::Number::One() but thats just me :P

kostekIV (approved):
I kinda dont like that we need to specify everywhere `B::Header: HeaderT<Number = BlockNumber>` where B: BlockT is specified but that's just me.

maciejnems:
> I kinda dont like that we need to specify everywhere `B::Header: HeaderT<Number = BlockNumber>` where B: BlockT is specified but that's just me.

Yeaaah I also do not like it :/ Not sure other solutions would be good though. We  could:
* Use unstable feature for associated type bounds
* Use our own trait: ```trait OurBlock : Block<Header = Self::H> {
    type H: Header<Number = BlockNumber>;
}```

Also if we added BlockIdentifier trait to more places in code, in some places we would not need to specify Header. Only in places where we directly operate on client. I think this is the direction we want to take in future.

finality-aleph/src/data_io/chain_info.rs
maciejnems: And what is the reason for that? Right now here we have `BlockNumber = u32`. And if its type changes, compiler will just shout an error here. Also this type should never change. At least it does not have to in the next 100 years.

PR #1029, opened 2023-03-16
A0-2098: refactor AuthorityProvider, FinalityNotificator and SessionMapUpdater

maciejnems:
# Description

Refactor `AuthorityProvider` and `FinalityNotificator` not to be generic over `B: Block`. In result also `SessiomMapUpdater` no longer is generic over `B: Block`

## Type of change

Refactor

### DIFF
--- a/finality-aleph/src/session_map.rs
+++ b/finality-aleph/src/session_map.rs
@@ -20,11 +20,11 @@ const PRUNING_THRESHOLD: u32 = 10;
 type SessionMap = HashMap<SessionId, SessionAuthorityData>;
 type SessionSubscribers = HashMap<SessionId, Vec<OneShotSender<SessionAuthorityData>>>;
 
-pub trait AuthorityProvider<N> {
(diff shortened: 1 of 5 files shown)


kostekIV (approved):

finality-aleph/src/session_map.rs
```diff
@@ -693,8 +651,8 @@ mod tests {
             );
         }
 
-        for block in blocks {
-            sender.unbounded_send(to_notification(block)).unwrap();
+        for n in PRUNING_THRESHOLD + 1..2 * PRUNING_THRESHOLD + 1 {
```
timorl, line 654: I'm a bit nervous about this lacking parentheses. <_<"

timorl (approved):
Nice <3

finality-aleph/src/session_map.rs
maciejnems: Kind of changed that

PR #1036, opened 2023-03-17
A0-2120: Make yaml-validate workflow fail on errors

mikogs:
It's better we actually make that YAML Validate workflow fail on error.  We have almost overlooked one syntax mistake in another PR.
Now github-workflows-validator always exits with code 0, at least for now (we were about to change it later once workflows got tidied up).  However, all errors are prefixed with `E` so we can check if these appear in the output and if so add `exit 1`.


mikogs:
`contracts-tests` workflow is broken so I'm not sure what to do with it.  We can force-squash-and-merge it and ignore the problem, or naturally fix it (tidying up that workflow will probably take some time as we need to build a new dev environments for contracts, one that would be similar to benjamin).
Anyway, feel free to modify and/or merge that PR whilst I'll be on holidays.

Marcin-Radecki (approved):

PR #1053, opened 2023-03-20
A0-2103 Interface for user Justifications

lesniak43:
# Description

https://cardinal-cryptography.atlassian.net/browse/A0-2103

Add interface for accepting additional justifications. These will be coming from AlephBFT or the parts of Substrate sync that we'll remove in next stages of the sync rewrite.

## Type of change

# Checklist:


finality-aleph/src/sync/service.rs
```diff
@@ -259,6 +275,18 @@ impl<
                 maybe_event = self.chain_events.next() => match maybe_event {
                     Ok(chain_event) => self.handle_chain_event(chain_event),
                     Err(e) => warn!(target: LOG_TARGET, "Error when receiving a chain event: {}.", e),
+                },
+                maybe_justification = self.justifications_from_user.next() => match maybe_justification {
+                    // if the Justification comes from user, then it means we already have the
+                    // corresponding block, and it is or will soon be imported independently,
```
kostekIV, line 281: I dont get that part, we have a block but we will soon import it?

kostekIV (approved):

finality-aleph/src/sync/service.rs
```diff
@@ -266,6 +282,19 @@ impl<
                     Ok(chain_event) => self.handle_chain_event(chain_event),
                     Err(e) => warn!(target: LOG_TARGET, "Error when receiving a chain event: {}.", e),
                 },
+                maybe_justification = self.justifications_from_user.next() => match maybe_justification {
+                    // If the Justification comes from user (i.e. AlephBFT), then it means we already
+                    // have received the corresponding block, and it is or soon will be imported,
+                    // independently, into the Substrate database.
+                    // Therefore, we do not create a task requesting the block from peers.
```
timorl, line 289: Could you instead clearly state what the assumption about the justification is (i.e. "we will end up having the related block in some way and no action from sync is necessary") and add it as a usage requirement to the trait documentation? This would make it more likely that any users will be aware that they should be following this assumption when using the interface.

timorl (approved):
Looks good, one comment about docs.

finality-aleph/src/sync/service.rs
lesniak43: sure, how about now?

PR #872, opened 2023-01-17
Exclude finality-aleph unit tests

mike1729:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

## Type of change

Please delete options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

- I have added tests
- I have made neccessary updates to the Infrastructure
- I have made corresponding changes to the existing documentation
- I have created new documentation
- I have bumped `spec_version` and `transaction_version`
- I have bumped aleph-client version if relevant

PR #1062, opened 2023-03-21
Added build check for liminal binary to PR pipeline

Marcin-Radecki:
# Description

1. Adds new job in PR pipeline that checks `aleph-node` build with liminal flag
2. Add a workflow that triggers after merge to main, that builds `aleph-node` with liminal feature and pushes to ECR to `aleph-node-liminal` repository

## Type of change

# Checklist:


.github/workflows/_build-liminal-node.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+
+      - name: Install WASM target
+        run: rustup target add wasm32-unknown-unknown
+
+      - name: Build test binary and runtime
+        run: |
+          cargo build --release -p aleph-node --features "liminal"
```
pmikolajczyk41, line 27: should we use `--profile production` instead? if so, then the path in the next step should be adjusted to `target/production/aleph-node`

.github/workflows/on-main-branch-commit-push-liminal-node-to-ecr.yml
```diff
@@ -0,0 +1,83 @@
...
+      - main
+
+concurrency:
+  group: ${{ github.ref }}-${{ github.workflow }}
+  cancel-in-progress: true
+
+jobs:
+  build-liminal-node:
+    name: Build liminal node
+    uses: ./.github/workflows/_build-test-node-and-runtime.yml
```
pmikolajczyk41, line 16: this will build a standard node, won't it?

.github/workflows/on-main-branch-commit-push-liminal-node-to-ecr.yml
```diff
@@ -0,0 +1,83 @@
...
+        uses: actions/checkout@v3
+
+      - name: Call action get-ref-properties
+        id: get-ref-properties
+        uses: ./.github/actions/get-ref-properties
+
+      - name: Download node production binary from artifacts
+        uses: actions/download-artifact@v2
+        with:
+          name: aleph-node-liminal
```
pmikolajczyk41, line 34: this is supposed to be build by the previous job, right?

.github/workflows/_build-liminal-node.yml
Marcin-Radecki: The problem is that it makes longer the whole build by 8 minutes, and consumes our multi core self hosted runner in the most inefficient way, as LTO linking that lasts this 8 minutes occupies only one CPU. So we need to make sure this is necessaryto use production profile here

.github/workflows/_check-build-liminal-node.yml
```diff
@@ -0,0 +1,23 @@
...
+
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+
+      - name: Install WASM target
+        run: rustup target add wasm32-unknown-unknown
+
+      - name: Build liminal binary and runtime
+        run: |
+          cargo check --release -p aleph-node --features "liminal"
```
deuszx, line 23: I know that I had told you that `cargo check` would suffice but there are some tests that are liminal-specific so we should run them _somewhere_ as well. /cc @pmikolajczyk41

Marcin-Radecki: No problem, should I change it to `cargo build` ? Actually even better, I can re-use the same workflow here

deuszx: No, not `cargo build`. I was thinking about tests. Let's wait for @pmikolajczyk41 as he's more familiar with what we have where.

.github/workflows/on-main-branch-commit-push-liminal-node-to-ecr.yml
```diff
@@ -0,0 +1,83 @@
...
+          path: target/release/
+
+      - name: Build production docker node image
+        id: build-image
+        run: |
+          chmod +x target/release/aleph-node
+          docker build --tag aleph-node:latest -f ./docker/Dockerfile .
```
deuszx, line 41: Later on you have `CURRENT_IMAGE: aleph-node-liminal` but here you're using a non-liminal tag. It should probably be `--tag aleph-node-liminal:latest`.

.github/workflows/on-main-branch-commit-push-liminal-node-to-ecr.yml
```diff
@@ -0,0 +1,83 @@
...
+
+      - name: Configure AWS Mainnet credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_MAINNET_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_MAINNET_SECRET_ACCESS_KEY }}
+          aws-region: eu-central-1
```
deuszx, line 48: This is something I didn't understand when I made my own, short-lived, PR -- why do we need AWS creds here? I didn't include it and it still pushed to ECR just fine.

.github/workflows/on-main-branch-commit-push-liminal-node-to-ecr.yml
```diff
@@ -0,0 +1,83 @@
...
+        id: login-public-ecr
+        uses: docker/login-action@v2
+        with:
+          registry: public.ecr.aws
+          username: ${{ secrets.AWS_MAINNET_ACCESS_KEY_ID }}
+          password: ${{ secrets.AWS_MAINNET_SECRET_ACCESS_KEY }}
+        env:
+          AWS_REGION: us-east-1
+
+      - name: Push aleph-node image to ECR
```
deuszx, line 60:
```suggestion
      - name: Push aleph-node-liminal image to ECR
```

.github/workflows/on-main-branch-commit-push-liminal-node-to-ecr.yml
```diff
@@ -0,0 +1,83 @@
...
+        env:
+          CURRENT_IMAGE: aleph-node-liminal
+          ECR_PUSH_IMAGE: public.ecr.aws/p6e8q1z1/aleph-node-liminal
+        run: |
+          docker tag ${{ env.CURRENT_IMAGE }} \
```
deuszx, line 65: Don't you need the `:latest` here as well?

deuszx: Or is it implicit?

.github/workflows/on-main-branch-commit-push-liminal-node-to-ecr.yml
```diff
@@ -0,0 +1,83 @@
...
+            ${{ env.ECR_PUSH_IMAGE }}:${{ steps.get-ref-properties.outputs.sha }}
+          docker push ${{ env.ECR_PUSH_IMAGE }}:${{ steps.get-ref-properties.outputs.sha }}
```
deuszx, line 67: @pmikolajczyk41 Do we want the `latest` tag as well and depend on that in shielder? Otherwise, even if we add e2e tests to the CI of the shielder, we won't catch any regressions since we will be depending on a fixed docker image.

.github/workflows/_build-liminal-node.yml
pmikolajczyk41: for standard node ofc

.github/workflows/_check-build-liminal-node.yml
pmikolajczyk41: we could run just runtime tests, like: https://github.com/Cardinal-Cryptography/aleph-node/pull/872/files (so that we don't waste time for tests that are already run in other places)

.github/workflows/_build-liminal-node.yml
Marcin-Radecki: for standard node, it is already there https://github.com/Cardinal-Cryptography/aleph-node/blob/main/.github/workflows/_build-production-node-and-runtime.yml#L34

.github/workflows/_check-build-liminal-node.yml
Marcin-Radecki: That will significantly affect the shape of this PR. Let's discuss this offline.

.github/workflows/on-main-branch-commit-push-liminal-node-to-ecr.yml
Marcin-Radecki: good catch

.github/workflows/on-main-branch-commit-push-liminal-node-to-ecr.yml
pmikolajczyk41: I would push both: ref and latest

.github/workflows/on-main-branch-commit-push-liminal-node-to-ecr.yml
Marcin-Radecki: Right, looks like it is not needed.

.github/workflows/on-main-branch-commit-push-liminal-node-to-ecr.yml
```diff
@@ -0,0 +1,80 @@
+---
+name: Builds and pushes aleph-node liminal to ECR on push to main branch
+
+on:
+  push:
+    branches:
+      - main
+      # temp for testing
+      - add-test-liminal-binary-to-workflows
```
deuszx, line 9: Did it work?

.github/workflows/_check-build-liminal-node.yml
deuszx: @pmikolajczyk41 can you ticket it please?

.github/workflows/on-main-branch-commit-push-liminal-node-to-ecr.yml
```diff
@@ -0,0 +1,80 @@
+---
+name: Builds and pushes aleph-node liminal to ECR on push to main branch
```
deuszx, line 2:
```suggestion
name: Builds and pushes aleph-node-liminal to ECR on push to main branch
```

.github/workflows/on-main-branch-commit-push-liminal-node-to-ecr.yml
```diff
@@ -0,0 +1,80 @@
...
+jobs:
+  build-liminal-node:
+    name: Build liminal node
+    uses: ./.github/workflows/_build-liminal-node.yml
+    secrets: inherit
+
+  push-liminal-node-image-to-ecr:
+    name: Push liminal node docker image to the ECR
+    needs: [build-liminal-node]
+    runs-on: ubuntu-20.04
```
deuszx, line 24: Are you not using `self-hosted` here on purpose?

.github/workflows/_check-build-liminal-node.yml
pmikolajczyk41: https://github.com/Cardinal-Cryptography/aleph-node/pull/1078 this should be enough

.github/workflows/on-main-branch-commit-push-liminal-node-to-ecr.yml
Marcin-Radecki: Yes, it's not needed, as it is in phase after merge and takes about 20 seconds anyway

.github/workflows/on-main-branch-commit-push-liminal-node-to-ecr.yml
Marcin-Radecki: After some more changes, but eventually yes: https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/4498596592

pmikolajczyk41 (approved):

PR #1065, opened 2023-03-21
A0-1798: Replace old justification sync

lesniak43:
# Description

Enables the new justification sync:
* starts the service,
* accepts data availability justification requests,
* deals with justifications from the aggregator,
* deals with justifications from block sync

Removes the old justification sync.

Summary of main changes per file or crate:

`bin/node`
* rpc - added `Translator` and sending justifications
* run sync

`finality-aleph`
* `import` - added `Translator`
* `justification` - removed lots of code, only the actual struct + compatibility left
* `lib.rs` - shrunk API
* `network` - `gossip/service.sr` enabling gossip network for the new sync
* `nodes` - removed lots of stuff (shrunk to a single file), run sync code
* `party/manager` - similar to `rpc`
* `sync/substrate/verification/verifier` - removed legacy stuff
* `testing` - gossip-related additions, removing lots of tests for removed code

## Type of change

# Checklist:

- I have added tests


finality-aleph/src/nodes.rs
```diff
@@ -13,21 +13,28 @@ use sp_runtime::traits::{Block, Header};
...
-    AlephConfig, BlockchainBackend,
+    sync::{
+        ChainStatus, Justification, JustificationTranslator, Service as SyncService,
+        SubstrateChainStatusNotifier, SubstrateFinalizationInfo, SubstrateJustification,
+        VerifierCache,
+    },
+    AlephConfig,
 };
 
+const VERIFIER_CACHE_SIZE: usize = 2;
```
kostekIV, line 36: What units are used here? Blocks, sessions or mbs? :P

finality-aleph/src/session_map.rs
```diff
@@ -210,10 +210,6 @@ impl SharedSessionMap {
 }
 
 impl ReadOnlySessionMap {
-    pub async fn get(&self, id: SessionId) -> Option<SessionAuthorityData> {
```
kostekIV, line 213: just rm pub instead of moving this to tests :P

kostekIV (approved):
couple of nits, looks good :)

bin/node/src/aleph_node_rpc.rs
```diff
@@ -106,17 +121,18 @@ where
                     "Provided justification cannot be converted into correct type".into(),
                 )
             })?);
+        let justification = self
```
fixxxedpoint, line 124: This comment doesn't belong to this review - you can resolve it. It is inconsistent (?): it can early exit if a justification can't be translated (that is good), but returns Ok, even when the provided justification is not valid (someone receiving it on that channel will decide about it?).

bin/node/src/service.rs
```diff
@@ -252,16 +259,18 @@ fn setup(
             warp_sync: None,
         })?;
 
+    let chain_status = SubstrateChainStatus::new(backend.clone())
```
fixxxedpoint, line 262: nit: Maybe you can share that `SubstrateChainStatus` between `setup` and `new_authority`.  Probably that is not an issue here, but it's not great to share same error messages here and there.

bin/node/src/service.rs
```diff
@@ -390,11 +399,12 @@ pub fn new_authority(
     if aleph_config.external_addresses().is_empty() {
         panic!("Cannot run a validator node without external addresses, stopping.");
     }
-    let blockchain_backend = BlockchainBackendImpl { backend };
+    let chain_status = SubstrateChainStatus::new(backend)
```
fixxxedpoint, line 402: nit: I don't like the idea that `SubstrateChainStatus` implements also `JustificationTranslator`. Relation looks more like: the only implementation of `JustificationTranslator` uses impl of `ChainStatus`. The `JustificationTranslator` trait also doesn't look very useful at the moment - only single implementation (shared with some other impl by SubstrateChainStatus), no mocks.

finality-aleph/src/network/gossip/service.rs
```diff
@@ -246,41 +270,60 @@ impl<N: RawNetwork, D: Data> Service<N, D> {
...
                     Protocol::Authentication => {
                         let (tx, rx) = tracing_unbounded(
                             "mpsc_notification_stream_authentication",
                             QUEUE_SIZE_WARNING,
                         );
                         self.authentication_connected_peers.insert(peer.clone());
                         self.authentication_peer_senders.insert(peer.clone(), tx);
-                        rx
+                        self.spawn_handle.spawn(
+                            "aleph/network/peer_sender",
```
fixxxedpoint, line 323: Maybe they should use different names since they are used for different protocols? Can they?

finality-aleph/src/network/gossip/service.rs
```diff
@@ -361,21 +403,30 @@ impl<N: RawNetwork, D: Data> Service<N, D> {
...
                     None => {
                         error!(target: "aleph-network", "Network event stream ended.");
                         return;
                     }
                 },
-                maybe_message = self.messages_from_user.next() => match maybe_message {
-                    Some(Command::Broadcast(message, protocol)) => self.broadcast(message, protocol),
-                    Some(Command::SendToRandom(message, peer_ids, protocol)) => self.send_to_random(message, peer_ids, protocol),
-                    Some(Command::Send(message, peer_id, protocol)) => self.send(message, peer_id, protocol),
+                maybe_message = self.messages_from_authentication_user.next() => match maybe_message {
```
fixxxedpoint, line 415: nit: divide it into two `match` statements instead of repeating `Some`.

finality-aleph/src/nodes.rs
```diff
@@ -113,17 +120,36 @@ where
...
+        block_sync_network,
+        chain_events,
+        chain_status.clone(),
+        verifier,
+        finalizer,
+        session_period,
+        justification_rx,
+    ) {
+        Ok(x) => x,
+        Err(e) => panic!("Failed to run Sync service: {}", e),
```
fixxxedpoint, line 150: nit: `Failed to initialize...`? Run is called later.

finality-aleph/src/nodes.rs
```diff
@@ -38,21 +45,21 @@ pub async fn new_pen(mnemonic: &str, keystore: Arc<dyn CryptoStore>) -> Authorit
...
+pub async fn run_validator_node<B, H, C, CS, BE, SC>(aleph_config: AlephConfig<B, H, C, SC, CS>)
 where
     B: Block,
     B::Header: Header<Number = BlockNumber>,
     H: ExHashT,
     C: crate::ClientForAleph<B, BE> + Send + Sync + 'static,
     C::Api: aleph_primitives::AlephSessionApi<B>,
     BE: Backend<B> + 'static,
-    BB: BlockchainBackend<B> + Send + 'static,
+    CS: ChainStatus<SubstrateJustification<B::Header>> + JustificationTranslator<B::Header>,
```
fixxxedpoint, line 56: nit: maybe don't merge them?

fixxxedpoint (approved):
lgtm. Left few nits.

finality-aleph/src/import.rs
```diff
@@ -77,82 +77,85 @@ where
...
         }
     }
 
     fn send_justification(
         &mut self,
-        block_id: IdentifierFor<B>,
-        justification: Justification,
-    ) -> Result<(), SendJustificationError<B>> {
-        debug!(target: "aleph-justification", "Importing justification for block {:?}", block_id.number());
+        hash: B::Hash,
```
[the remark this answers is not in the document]
timorl: Yes, but this conflicts with the translator abstraction. This will be fixed later, with the couple or PRs Maciek left us – we didn't want to merge them before, because they touch quite a lot of code that we are removing here.

There are currently actually a couple other ugly things we introduce in this PR that Maciek's refactors will/would have addressed. A bit of a bad timing, but likely the version with all this ugliness will never be run on any *net.

finality-aleph/src/import.rs
```diff
@@ -77,82 +77,85 @@ where
 /// A wrapper around a block import that also extracts any present jsutifications and send them to
 /// our components which will process them further and possibly finalize the block.
 #[derive(Clone)]
-pub struct AlephBlockImport<B, I>
```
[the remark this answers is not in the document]
timorl: Done.

finality-aleph/src/network/gossip/service.rs
lesniak43: chatgpt says: Using a nested match expression is also a valid approach, but it may lead to more verbose code and potentially make it harder to read and understand the logic. Additionally, in this case, using nested matches may not be necessary because the cases being handled are not particularly complex.

oh, and also @timorl said something similar ;) let's leave it that way, shall we?

bin/node/src/service.rs
```diff
@@ -252,16 +259,18 @@ fn setup(
...
         let pool = transaction_pool.clone();
-
         Box::new(move |deny_unsafe, _| {
-            let deps = crate::rpc::FullDeps::<Block, _, _> {
+            let deps: RpcFullDeps<Block, _, _, _> = crate::rpc::FullDeps {
                 client: client.clone(),
                 pool: pool.clone(),
                 deny_unsafe,
                 import_justification_tx: import_justification_tx.clone(),
+                justification_translator: chain_status.clone(),
```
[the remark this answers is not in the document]
timorl: Yup:
```
267 |         Box::new(move |deny_unsafe, _| {
    |                  ^^^^^^^^^^^^^^^^^^^^^ this closure implements `FnOnce`, not `Fn`
...
273 |                 justification_translator: chain_status,
    |                                           ------------ closure is `FnOnce` because it moves the variable `chain_status` out of its environment
...
286 |         rpc_builder,
    |         ----------- the requirement to implement `Fn` derives from here
```

Interesting case though.

bin/node/src/aleph_node_rpc.rs
timorl: Yup, I think the return here should only be interpreted as explaining whether the submission process worked, not the operation itself. This sounds fine to me, but perhaps I'm biased since I know how annoying it would be to implement returning the result of the whole operation. ;p

bin/node/src/service.rs
timorl: Hmmm, trueeee... This whole thing could use a refactor, but in this specific case we are making things strictly, and potentially noticeably, worse.

Moved the creation around a bit to make it only happen once; the price is `setup` now takes it as an argument.

bin/node/src/service.rs
timorl: Yeah, I'm not sure what to do about this. I originally wanted to implement this for everything that implements `ChainStatus`, but that obviously doesn't work for typical reasons. :/ If you have some specific better idea I would be grateful, but preferably not in this PR (unless it would make it smaller?). This also related to the last comment ("Maybe don't merge them?").

finality-aleph/src/network/gossip/service.rs
timorl: I'm pretty sure they can and good idea, added prefixes.

finality-aleph/src/nodes.rs
timorl: Good point, fixed.

finality-aleph/src/network/gossip/service.rs
timorl: I hate that ChatGPT agrees with me. D'x

fixxxedpoint: I didn't mean nested....

fixxxedpoint:
```suggestion
	maybe_message = self.messages_from_authentication_user.next() => {
		let message = match maybe_message {
			Some(msg) => msg,
			None => {
				error!(target: "aleph-network", "Authentication user message stream ended.");
				return;
			},
		};
		match message {
			Command::Broadcast(message) => self.broadcast_authentication(message),
			Command::SendToRandom(message, peer_ids) => self.send_to_random_authentication(message, peer_ids),
			Command::Send(message, peer_id) => self.send_authentication_data(message, peer_id),
		}
	}
```

fixxxedpoint: haha super ugly. Move those to separate functions, return `Result` and use `?` operator, etc.

finality-aleph/src/justification/requester.rs
```diff
@@ -0,0 +1,97 @@
...
+        Ok(self
+            .chain_status
+            .children(id)?
+            .into_iter()
+            .map(|header| header.id())
+            .collect())
+    }
+
+    fn perform_request(&self) -> Result<(), CS::Error> {
+        self.block_requester.clear_justification_requests();
```
fixxxedpoint, line 58: Is this call necessary? `extra_requests.rs` seems to handle cleaning well. If it's necessary, a comment here would be good idea.

fixxxedpoint (changes requested):
LGTM. Left some small request.

finality-aleph/src/justification/requester.rs
timorl: Oh, and it definitely won't happen more than once per minute.

fixxxedpoint (approved):
great work

PR #1076, opened 2023-03-23
A0-2174: Fix nightly pipeline crashing due to missing artifact

mikogs:


.github/workflows/nightly-pipeline.yaml
```diff
@@ -10,21 +10,26 @@ concurrency:
...
 
 jobs:
-  build-new-node:
-    name: Build node and runtime artifacts (PR version)
+  build-new-production-node:
+    name: Build production node and runtime artifacts (PR version)
     uses: ./.github/workflows/_build-production-node-and-runtime.yml
     secrets: inherit
 
+  build-new-test-node:
```
Marcin-Radecki, line 18: Not sure why we need that step. Nightly should run on production binary, not test one.

Marcin-Radecki:
Optional: is it possible to:
* remove `check_determinism` from nightly , as it is done for each PR anyway
* extract `build-test-docker`and  `build-test-client` as separate action or worfklow, similiar to existing `_build-test-node-and-e2e-client-image.yml`, but they should operate on production binary

Marcin-Radecki (approved):

PR #1079, opened 2023-03-23
A0-2156-stop-the-world-calls

fbielejec:
# Description

Adding calls that halt trading in case of a major emergency
Simplified wAZERO prior to mainnet deployment

## Type of change


contracts/marketplace/lib.rs
```diff
@@ -181,6 +221,18 @@ pub mod marketplace {
             Ok(())
         }
 
+        /// Update the length of the auction.
+        ///
+        /// Can only be called by an account with an Admin role
+        #[ink(message)]
+        pub fn set_auction_length(&mut self, new_auction_length: BlockNumber) -> Result<(), Error> {
```
pmikolajczyk41, line 228: should it be allowed only for a halted / ended contract? I'm guessing that changing auction length just before some user's transaction might introduce some unexpected and confusing behaviour

contracts/simple_dex/lib.rs
```diff
@@ -155,6 +164,10 @@ mod simple_dex {
             amount_token_in: Balance,
             min_amount_token_out: Balance,
         ) -> Result<(), DexError> {
+            if self.halted {
```
pmikolajczyk41, line 167: should we guard `deposit` and `withdrawal` as well? and some other write messages?

contracts/wrapped_azero/lib.rs
```diff
@@ -148,15 +155,55 @@ pub mod wrapped_azero {
...
+        pub fn halt(&mut self) -> Result<()> {
+            let caller = self.env().caller();
+            let this = self.env().account_id();
+            self.check_role(caller, Role::Admin(this))?;
+            self.halted = true;
+
+            // emit event
+            Self::emit_event(self.env(), Event::Halted(Halted {}));
+
+            Ok(())
```
pmikolajczyk41, line 178: it would be very cool to have a common interface/something for all haltable contracts (thank you captain obvious!)

.github/workflows/contracts-deploy.yml
```diff
@@ -46,7 +48,14 @@ jobs:
...
   build-and-deploy-contracts:
-    needs: [detect-contracts-changes]
+    # TODO: e2e test are flaky AF
+    # and they block the deployment
+    # which in turns stalls development
+    # as frontend developers rely heavily on the devnet environment.
+    # Therefore we made an executive decision to deploy to the devnet
+    # regardless of the contracts e2e tests status.
+    # Once fixed this dependency can be brough back
+    # needs: [detect-contracts-changes]
```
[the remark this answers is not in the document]
Marcin-Radecki: I think it's a compromise of approach of using paths and not using paths to detect if pull request should trigger. In this solution, this PR would trigger always in GH interface, yet it will be run actually only on those path changes. If we ever go back to this solution, this is the best option

.github/workflows/contracts-deploy.yml
```diff
@@ -55,7 +64,7 @@ jobs:
       (github.event_name == 'pull_request' &&
        github.event.action == 'labeled' &&
        github.event.label.name == '[AZERO] DEPLOY-CONTRACTS')
-    runs-on: ubuntu-20.04
+    runs-on: self-hosted
```
Marcin-Radecki, line 67: If we switch to self-hosted runner, we can install all those pre-requisites (jq, node etc) upfront so we save overall time. 
CC @mikogs

contracts/marketplace/lib.rs
```diff
@@ -114,12 +122,44 @@ pub mod marketplace {
...
+        pub fn resume(&mut self) -> Result<(), Error> {
+            self.ensure_role(self.admin())?;
+            self.halted = false;
+            Self::emit_event(self.env(), Event::Resumed(Resumed {}));
+            Ok(())
+        }
+
+        /// Is the contract in a halted state
+        #[ink(message)]
+        pub fn is_halted(&mut self) -> bool {
```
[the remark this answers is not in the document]
fbielejec: removed

contracts/marketplace/lib.rs
```diff
@@ -114,12 +122,44 @@ pub mod marketplace {
...
+        }
+
+        /// Resume after emergency halt
+        ///
+        /// Can only be called by the contract's Admin.
+        #[ink(message)]
+        pub fn resume(&mut self) -> Result<(), Error> {
+            self.ensure_role(self.admin())?;
+            self.halted = false;
+            Self::emit_event(self.env(), Event::Resumed(Resumed {}));
```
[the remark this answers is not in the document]
fbielejec: fixed

contracts/marketplace/lib.rs
```diff
@@ -114,12 +122,44 @@ pub mod marketplace {
...
+        /// Emergency halt of all operations
+        ///
+        /// Can only be called by the contract's Admin.
+        #[ink(message)]
+        pub fn halt(&mut self) -> Result<(), Error> {
+            self.ensure_role(self.admin())?;
+            self.halted = true;
+
+            // emit event
+            Self::emit_event(self.env(), Event::Halted(Halted {}));
```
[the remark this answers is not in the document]
fbielejec: fixed - it's a no op now

.github/workflows/contracts-deploy.yml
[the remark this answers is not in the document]
fbielejec: exactly :point_up:

.github/workflows/contracts-deploy.yml
```diff
@@ -110,7 +119,6 @@ jobs:
       - name: Run deploy.sh script
         shell: bash
         run: |
-          . "$HOME/.cargo/env"
```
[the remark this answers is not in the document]
fbielejec: it failed on the self hosted runner, but in any case this just exports path to cargo which is set in the docker container anyway

contracts/marketplace/lib.rs
```diff
@@ -114,12 +122,44 @@ pub mod marketplace {
                     total_proceeds: starting_price.saturating_div(sale_multiplier),
                     tickets_sold: 1,
                     access_control,
+                    halted: false,
```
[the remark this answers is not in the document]
fbielejec: honestly - meeeh, I don't know what I should check at this point that is not covered by e.g. tests or just by knowing what these contracts do

contracts/button/lib.rs
```diff
@@ -328,7 +328,7 @@ pub mod button_game {
             self.last_presser = None;
             self.last_press = now;
             self.total_rewards = 0;
-            self.round.checked_add(1).ok_or(GameError::Arithmethic)?;
+            self.round = self.round.checked_add(1).ok_or(GameError::Arithmethic)?;
```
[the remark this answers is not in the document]
fbielejec: sure

contracts/marketplace/lib.rs
fbielejec: out of abundance of caution we can add that, sure

.github/workflows/contracts-deploy.yml
fbielejec: - jq definitely
- whatever LTS version of node is the most current

contracts/marketplace/lib.rs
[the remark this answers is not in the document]
fbielejec: forgot to commit, same for the button - see now

contracts/simple_dex/lib.rs
```diff
@@ -120,6 +140,56 @@ mod simple_dex {
         pub access_control: AccessControlRef,
         // a set of pairs that are availiable for swapping between
         pub swap_pairs: Mapping<SwapPair, ()>,
+        pub halted: bool,
+    }
+
+    impl Haltable for SimpleDex {
```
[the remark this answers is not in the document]
fbielejec: take a look at this - a shot at a common implementation

.github/workflows/contracts-deploy.yml
mikogs: @Marcin-Radecki Both node16 and jq are there on the AWS runners, along with many others.  It should be just fine.

Marcin-Radecki: Right, thanks

contracts/wrapped_azero/lib.rs
```diff
@@ -129,31 +125,32 @@ pub mod wrapped_azero {
...
+                metadata,
             }
         }
 
+        /// Terminates the contract.
+        ///
+        /// No-op by default, can only be compiled with a flag in dev environments
+        #[ink(message)]
+        pub fn terminate(&mut self) -> Result<()> {
+            cfg_if::cfg_if! { if #[cfg( feature = "devnet" )] {
```
pmikolajczyk41, line 146: simple `#[cfg(feature = "devnet")]` doesn't work?

pmikolajczyk41 (approved):
I didn't resolve a few comments, but none of them is a blocker and I'm fine if you just resolve them

generally, great work :muscle:

contracts/wrapped_azero/lib.rs
fbielejec: noup. I might revisit it - but in another PR, this not part of the ticket

.github/workflows/_check-excluded-packages.yml
```diff
@@ -47,15 +47,31 @@ jobs:
...
-            fi
             echo "Checking package $p..."
             pushd "$p"
+            rustup component add clippy rustfmt
+            rustup target add wasm32-unknown-unknown
+
+            rustup component add rust-src --toolchain nightly-2022-10-30-x86_64-unknown-linux-gnu
+            cargo install cargo-dylint dylint-link
+            cargo install cargo-contract --vers ^2.0 --force --locked
+
```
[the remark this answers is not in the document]
deuszx: We've agreed to use the `docker-ink-dev` here, why are you not using it here?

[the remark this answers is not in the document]
fbielejec: migrating legacy workflows is not part of this ticket

contracts/button/lib.rs
```diff
@@ -56,6 +60,12 @@ pub mod button_game {
         when: BlockNumber,
     }
 
+    #[ink(event)]
+    pub struct Halted {}
```
deuszx, line 64: In genera, Rust allows for "unit structs":
```rust
struct Halted;
```
and when constructing it's just `Halted`. Can we do it?

contracts/button/lib.rs
```diff
@@ -112,16 +162,16 @@ pub mod button_game {
...
                     access_control,
                     ticket_token,
                     reward_token,
                     marketplace,
                     button_lifetime,
                     scoring,
                 ),
-                Err(why) => panic!("Could not initialize the contract {:?}", why),
+                false => panic!("Caller is not allowed to initialize this contract"),
             }
```
deuszx, line 175: nit: this can be replaced with more readable version IMO:
```rust
if !access_control.has_role(caller, required_role) {
  panic!("Caller is not allowed to initialize this contract")
}
Self::init(
                    access_control,
                    ticket_token,
                    reward_token,
                    marketplace,
                    button_lifetime,
                    scoring,
                )
```

and as a follow-up we might ticket adding a macro that checks the roles and panics if `false`.

contracts/shared_traits/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+mod haltable;
+
+pub use haltable::*;
```
deuszx, line 5: nit: This could be just
```rust
pub mod haltable;
```
and then on the use-side:
```
use shared_traits::haltable::{DefaultHaltable, Haltable}
```
instead of just
```rust
use shared_traits::Haltable
```
This looks as better package management that groups functionality together. If we add more things into `shared_traits` that are not haltable-related then it will become its own package.

contracts/button/lib.rs
```diff
@@ -56,6 +60,12 @@ pub mod button_game {
         when: BlockNumber,
     }
 
+    #[ink(event)]
+    pub struct Halted {}
+
+    #[ink(event)]
+    pub struct Resumed {}
```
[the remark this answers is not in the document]
fbielejec: IMO not yet, but hopefully coming: https://github.com/paritytech/ink/issues/809

contracts/shared_traits/src/lib.rs
fbielejec: I'm following https://www.notion.so/cardinalcryptography/Module-based-access-in-our-codebase-b08edefff9394f68938da09030c0bda6?pvs=4#4a539c6232e144b49e03ef94a6aa5e78

deuszx: Different thing - the notion ADR is about what and how to share. I'm talking about grouping the elements behind the logical modules. You could still achieve both here, for example by re-rexporting:
```rust
mod haltable_internal; // or whatever
pub mod haltable {
  use haltable_internal::{...}
}
```
which would still make it available through `use shared_traits::haltable::*`.

Not a blocker but I think we would eventually follow-up with something similar, like we did for all other repos eventually.

contracts/button/lib.rs
fbielejec: I'm a big fan of pattern matching - albeit which is limited in Rust when compared to [Clojure](https://github.com/clojure/core.match#example-usage), ymmv, I find it eons more readable than any branching logic.

contracts/button/lib.rs
fbielejec: needs literal syntax when instantiating. macros

fbielejec: because that's not ink! DSL

contracts/shared_traits/src/lib.rs
fbielejec: This is literally written down as a no-go in the ADR:
```
All modules are private (no pub mod foo), if we want something from a submodule to be accessible from outside of the module then we have to explicitly reexport it (e.g. pub use foo::Bar in mod.rs).
```
I don't care what the convention is as long as it's consistent. I don't see a point in discussing it here - the ADR can be re-visisted if it really itches someone.

contracts/button/lib.rs
```diff
@@ -374,12 +417,8 @@ pub mod button_game {
             Ok(())
         }
 
-        fn check_role(
-            access_control: &AccessControlRef,
-            account: AccountId,
-            role: Role,
-        ) -> ButtonResult<()> {
-            if access_control.has_role(account, role) {
+        fn check_role(&self, account: AccountId, role: Role) -> ButtonResult<()> {
```
mike1729, line 420: This method is used always to check if caller is an Admin. Could you specialize it to `check_caller_is_admin` and remove args `account` and `role`?

contracts/shared_traits/src/lib.rs
deuszx: You still don't understand - these are two different things:
1. Not making whole module `pub`
2. Re-exporting things from a module with a wrapper.
I'm not asking for the 1), but 2). `pub use foo::Bar` makes `Bar` being accessed as a top-level element:
```rust
mod foo {
    // I want this public.
    pub struct Foo;
    
    // And this private.
    pub struct FooPriv;
    
}

pub use foo::Foo;

mod bar {
    use crate::Foo;
}
```
vs
```rust
mod foo_internal {
    // I want this public.
    pub struct Foo;
    // And this private.    
    pub struct FooPriv;
}

pub mod foo_public {
    pub use crate::foo_internal::Foo;
}

use foo_public::Foo;
```

In the first version `Foo` is accessed directly from the top-level module while in the latter is behind logicaly-grouped `foo` package (in the case of this PR it would be `use shared_traits::haltable::Haltable` vs `use shared_traits::Haltable`).

It's completely independent from the ADR that you linked.

contracts/button/lib.rs
mike1729, line 1: Missing `is_halted` check for `press`.

contracts/simple_dex/lib.rs
mike1729, line 1: Checks if contract is halted are missing in methods.

mike1729 (changes requested):

contracts/shared_traits/src/haltable/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+        <Self as StorageAsRef>::data(self).halted
+    }
+
+    fn check_halted(&self) -> HaltableResult<()> {
+        if self.is_halted() {
+            return Err(HaltableError::InHaltedState);
+        }
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
fbielejec: maybe possible, hold on.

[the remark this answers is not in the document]
fbielejec: I added the blanket impl - but like in the DM, the caveat is all contracts are now subject to the unstable specialization feature...

contracts/button/lib.rs
fbielejec: I will tackle this next but in another PR (https://cardinal-cryptography.atlassian.net/browse/A0-2238) and use a very similar trait abstraction for these checks once we stabilize on what syntax we want

contracts/button/lib.rs
fbielejec: Good spot! Added

contracts/simple_dex/lib.rs
mike1729: How about  `deposit` and `withdraw`?

contracts/button/lib.rs
```diff
@@ -261,10 +299,7 @@ pub mod button_game {
             &mut self,
             new_button_lifetime: BlockNumber,
         ) -> ButtonResult<()> {
-            let caller = self.env().caller();
-            let this = self.env().account_id();
-            let required_role = Role::Admin(this);
-            ButtonGame::check_role(&self.access_control, caller, required_role)?;
+            self.check_role(self.env().caller(), Role::Admin(self.env().account_id()))?;
```
deuszx, line 302: We're not checking for halted state here b/c it can only be called by an admin?

contracts/simple_dex/lib.rs
fbielejec: I took a general approach that the tx accessible by an Admin are not subject to that halt (Admin sets them on halt anyway, and you might want access to them to fix things, salvage funds etc)

contracts/button/lib.rs
fbielejec: yes, see previous responses

contracts/simple_dex/lib.rs
mike1729: But `deposit` and `withdraw` may be called with lower access right than Admin, right?

fbielejec: it can be called with a different role (**LIQUIDITY_PROVIDER**) whether it has lower/higher access is semantics, but for practical purposes this still means an account that has elevated access rights. I reckon this call is definitely the one you want access too if things go south - to e.g. salvage funds if there is a vulnerability.

See how you like this solution - in normal operation it's the designated account that adds/removes liquidity, in a halted state it's an Admin that can access these calls.

contracts/marketplace/lib.rs
```diff
@@ -174,13 +223,25 @@ pub mod marketplace {
...
+        ///
+        /// Can only be called by an account with an Admin role
+        /// and only if the contract is currently halted
+        #[ink(message)]
+        pub fn set_auction_length(&mut self, new_auction_length: BlockNumber) -> Result<(), Error> {
+            // if DefaultHaltable::is_halted(self) {
+            self.check_role(self.env().caller(), self.admin())?;
+            self.auction_length = new_auction_length;
+            // return Ok(());
+            // }
```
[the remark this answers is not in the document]
fbielejec: @pmikolajczyk41 commented that we probably should allow this call only in halted state - this is updated to reflect this now

contracts/simple_dex/lib.rs
mike1729: That works :)

mike1729 (approved):

contracts/simple_dex/lib.rs
deuszx: In practice, won't `Admin` and `LiquidityProvider` be the same people? I.e. we will most probably not use this approach in a real DEX.

PR #1080, opened 2023-03-23
A0-2173: Circuit feature in liminal-ark-relations

pmikolajczyk41:
# Description

We introduce a new feature for `liminal-ark-relations` crate: `circuit` (together with its _std_ companion - `circuit-std`). This enables hiding all circuit-related code behind feature flag, which improves experience for code that depends only on relations objects (like smart contracts). This required a minor change to `liminal-ark-relations-proc-macro` lib as well.

Additionally, we expose inner modules of `liminal-ark-relations` instead of flattening it.

Note: Once this PR is approved, I will publish both crates manually and edit `Cargo.toml` files for `cliain` and `relations` to use newly pushed artifacts.

## Type of change

# Checklist:


maciejzelaszczyk (approved):
Very nice work indeed. I particularly like that you took the time to de-flatten module imports, which makes the structure of the crate more explicit.

relations/Cargo.toml
```diff
@@ -12,25 +12,28 @@ description = "An instantiation of the Poseidon SNARK-friendly hash function."
...
-blake2 = { version = "0.9.2", default-features = false }
-liminal-ark-poseidon = { version = "0.1.0", default-features = false, features = ["circuit"] }
-liminal-ark-relation-macro = { version = "0.1.1" }
+liminal-ark-poseidon = { version = "0.1.0", default-features = false }
+# todo: remove path after publishing
+liminal-ark-relation-macro = { version = "0.1.2", path = "src/proc_macro" }
 paste = { version = "1.0.11" }
 
+# Required for computation in a circuit.
+ark-crypto-primitives = { version = "^0.3.0", default-features = false, features = [ "r1cs" ], optional = true }
```
maciejzelaszczyk, line 25: Nit: perhaps in the future we could have a uniform convention by either using `features = ["dependency_feature"]` or `"dependency/feature"` in the `features` table? There might be some intricacies I am not seeing here, but it would be nice to have dependency features defined in one way.

maciejzelaszczyk: @pmikolajczyk41 Feel free to resolve this at any time.

mike1729 (approved):

relations/Cargo.toml
pmikolajczyk41: fair point, but I would set it up on some retro/general forum and then adjust all crates to the decision

PR #1095, opened 2023-03-28
A0-2256: Handle genesis header in sync

timorl:
# Description

The genesis header is a finalized header that doesn't have a justification. This PR adds a pseudojustification for the purposes of justification sync.

## Type of change

# Checklist:


finality-aleph/src/sync/substrate/mod.rs
```diff
@@ -68,11 +68,36 @@ impl<H: SubstrateHeader<Number = BlockNumber>> Header for H {
     }
 }
 
+/// Proper `AlephJustification` or a variant indicating virtual justification
+/// for the genesis block, which is the only block that can be the top finalized
+/// block with no proper justification.
+#[derive(Clone, Debug, Encode, Decode)]
+pub enum InnerJustification {
```
kostekIV, line 75: This is smth that is only for network usage, right? It will not go into the db?

finality-aleph/src/sync/substrate/verification/mod.rs
```diff
@@ -97,18 +95,26 @@ impl Display for VerificationError {
...
+    H: SubstrateHeader<Number = BlockNumber>,
 {
     type Error = VerificationError;
 
     fn verify(&mut self, justification: Justification<H>) -> Result<Justification<H>, Self::Error> {
         let header = &justification.header;
-        let verifier = self.get(*header.number())?;
-        verifier.verify_bytes(&justification.raw_justification, header.hash().encode())?;
-        Ok(justification)
+        match justification.clone().inner_justification {
```
deuszx, line 108: Do we need to clone here? `verify_bytes` takes `aleph_justification` via reference and we return `justification` - is compiler not smart enough to figure out we won't move it?

finality-aleph/src/sync/substrate/mod.rs
```diff
@@ -68,11 +68,36 @@ impl<H: SubstrateHeader<Number = BlockNumber>> Header for H {
...
+    AlephJustification(AlephJustification),
+    Genesis,
+}
+
 /// A justification, including the related header.
 #[derive(Clone, Debug, Encode, Decode)]
 pub struct Justification<H: SubstrateHeader<Number = BlockNumber>> {
```
deuszx, line 82: You don't necessarily need this trait bound here, correct? It can be just `H`.

deuszx (approved):
I don't think I know the overall syncing algorithm well enough to comment on whether this isn't breaking anything but it LGTM, no blockers.

A general comment on the design (it's probably too late to change it here but I wanted to share it nevertheless): In my previous project we built it in such a way that there was no "Genesis block" at all, it was rather a _genesis state_. Our thinking was that Genesis is something that a client (including a node) has to trust as its a bootstrapping state for the whole chain - if the genesis state was different, then the whole chain that follows would be a different fork as well. System was implemented such that the _genesis state_ had to be constructed locally from the chainspec. When syncing, joining node/client had to construct that state on its own and so the syncing never reached the Genesis block - the first block in the chain, right after the Genesis state, was already a block like any other. This has some security implications that relate to the _weak subjectivity_ - if client is offline for long enough it cannot tell, without external input, who it should trust. By providing this _genesis state_ it's required to (able to) decide for itself what's the "starting point" (in that state it also includes information who are the validators it should accept blocks from).

Additional advantage of this was that chain could be "restarted" at any point in time and use a snapshot of a chain (at some particular block) as the new _genesis state_ - for example during an "emergency restart" which is a rollback of a chain + some changes applied on top of it (like forceful validator ejection, state changes, etc).

It was not without flaws - when doing the "emergency restart" - that state had to be constructed and couldn't be sync'd.

finality-aleph/src/sync/substrate/mod.rs
timorl: Nope, the stuff going into the DB is what is in this context `AlephJustification` (plus I think a small constant stirng, because substrate?).

finality-aleph/src/sync/substrate/verification/cache.rs
```diff
@@ -236,12 +258,14 @@ mod tests {
...
         let finalization_info = MockFinalizationInfo { finalized_number };
         let authority_provider = MockAuthorityProvider::new(max_session_n);
+        let genesis_header = MockHeader::random_parentless(0);
 
         VerifierCache::new(
             SessionPeriod(SESSION_PERIOD),
             finalization_info,
             authority_provider,
             CACHE_SIZE,
+            genesis_header,
```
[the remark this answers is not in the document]
timorl: Nope, it's annoyingly hard to make one, because this problem only occurs in the substrate implementation of our traits (the internal code doesn't know one kind of justification is special), and to test the substrate implementation in this context we would have to produce a `Header` which is annoyingly hard to do for testing purposes. :/

finality-aleph/src/sync/substrate/mod.rs
deuszx: I thought it's considered one of the "Rust good practices" to not put type bounds on structs and it's more customary to put them on `impls` and/or functions.

finality-aleph/src/sync/substrate/verification/mod.rs
timorl: It required a tiny amount of help, but now it works and one clone eliminated. Good catch, thanks.

kostekIV (approved):

PR #1096, opened 2023-03-28
A0-2074: Enforce that all PRs start from Jira task number

Kosiare:
# Description

I have added an action that will block PR from merging is the PR title is not following this pattern: `^A0-[0-9]+: .*$`

## Type of change


mikogs (approved):
Can we rename `on-pull-request-check.yml` to `on-pull-request-change.yml`?  I was thinking of a word that doesn't indicate a commit but another action.

Marcin-Radecki (approved):

PR #1097, opened 2023-03-28
upgradeability

fbielejec:
# Description

Adds upgradeability to contracts

### DIFF
--- a/contracts/access_control/lib.rs
+++ b/contracts/access_control/lib.rs
@@ -14,7 +14,13 @@ type Hash = <DefaultEnvironment as Environment>::Hash;
 
 #[ink::contract]
 mod access_control {
-    use ink::{codegen::EmitEvent, reflect::ContractEventBase, storage::Mapping};
+    use ink::{
+        codegen::EmitEvent,
+        env::{set_code_hash, Error as InkEnvError},
+        prelude::{format, string::String},
+        reflect::ContractEventBase,
+        storage::Mapping,
+    };
     use scale::{Decode, Encode};
 
     use crate::roles::Role;
@@ -56,10 +62,17 @@ mod access_control {
     #[derive(Debug, PartialEq, Eq, Encode, Decode)]
     #[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
     pub enum AccessControlError {
+        InkEnvError(String),
         MissingRole(Role),
     }
 
-    /// Result type    
+    impl From<InkEnvError> for AccessControlError {
+        fn from(why: InkEnvError) -> Self {
+            Self::InkEnvError(format!("{:?}", why))
+        }
+    }
+
+    /// Result type
     pub type Result<T> = core::result::Result<T, AccessControlError>;
     /// Event type
     pub type Event = <AccessControl as ContractEventBase>::Type;
@@ -131,8 +144,7 @@ mod access_control {
         #[ink(message, selector = 4)]
         pub fn terminate(&mut self) -> Result<()> {
             let caller = self.env().caller();
-            let this = self.env().account_id();
-            self.check_role(caller, Role::Admin(this))?;
+            self.check_role(caller, Role::Admin(self.env().account_id()))?;
             self.env().terminate_contract(caller)
         }
 
@@ -144,6 +156,14 @@ mod access_control {
             if !self.has_role(account, role) {
                 return Err(AccessControlError::MissingRole(role));
             }
+            Ok(())
+        }
+
+        /// Upgrades contract code
+        #[ink(message, selector = 6)]
+        pub fn set_code(&mut self, code_hash: [u8; 32]) -> Result<()> {
+            self.check_role(self.env().caller(), Role::Admin(self.env().account_id()))?;
+            set_code_hash(&code_hash)?;
             Ok(())
         }
 
--- a/contracts/button/lib.rs
+++ b/contracts/button/lib.rs
@@ -10,7 +10,7 @@ pub mod button_game {
     use ink::storage::traits::StorageLayout;
     use ink::{
         codegen::{EmitEvent, Env},
-        env::{call::FromAccountId, CallFlags},
+        env::{call::FromAccountId, set_code_hash, CallFlags},
         prelude::vec,
         reflect::ContractEventBase,
         ToAccountId,
@@ -321,6 +321,14 @@ pub mod button_game {
             let caller = self.env().caller();
             self.check_role(caller, Role::Admin(self.env().account_id()))?;
             self.env().terminate_contract(caller)
+        }
+
+        /// Upgrades contract code
+        #[ink(message)]
+        pub fn set_code(&mut self, code_hash: [u8; 32]) -> ButtonResult<()> {
+            self.check_role(self.env().caller(), Role::Admin(self.env().account_id()))?;
+            set_code_hash(&code_hash)?;
+            Ok(())
         }
 
         //===================================================================================================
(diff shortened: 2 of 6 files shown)


deuszx:
I haven't read the PR but have you read the https://docs.openbrush.io/smart-contracts/upgradeable where they explain how to make storage layout upgradeable as well?

krzysztofziobro:
I think something like "versioning" storage with `upgradable_storage(VERSION_KEY)` macro might be a good idea, as (I think) it makes storage migration process more clear. But it shouldn't really matter that much, so it's up to you.

krzysztofziobro:
Another thing to consider is enabling upgrade + migration to be "atomic", by taking `Option<Selector>` in the upgrade message, so migration function's selector can be called right after the upgrade. It might help us avoid downtime, when all calls to contract produce errors, or might even potentially break something, depending on the nature of the migration.

fbielejec:
> This might be a trivial question, but why some of the added `set_code` messages get custom selectors and others don't?

cargo culted, some contracts custom, some resort to the defaults so I just followed whatever the convention was so far :shrug:

fbielejec:
> I think something like "versioning" storage with `upgradable_storage(VERSION_KEY)` macro might be a good idea, as (I think) it makes storage migration process more clear. But it shouldn't really matter that much, so it's up to you.

1) Added - not sure what I think of adding the reserved fields - intuitively it's OK I guess, I don't see us reaching for it too often. 
2) Token contracts - they all have just one proprietary field - access_control contract address, which has it's value "upgradeable" with a simple message, so I reckon it's not worth wrapping it in it' s own storage type... But YMMV and I can be easily persuaded otherwise.

contracts/access_control/lib.rs
```diff
@@ -25,10 +31,20 @@ mod access_control {
     pub const HAS_ROLE_SELECTOR: [u8; 4] = [0, 0, 0, 3];
     pub const CHECK_ROLE_SELECTOR: [u8; 4] = [0, 0, 0, 5];
 
-    #[ink(storage)]
-    pub struct AccessControl {
+    pub const STORAGE_KEY: u32 = openbrush::storage_unique_key!(Data);
```
krzysztofziobro, line 34: I'm personally not a huge fan of defining storage keys this way, as it's less transparent what's going on and how to migrate (you need to change `Data` struct name, I would just use constant keys). But it's kinda the default way, so I leave it to your judgement which way to choose.

krzysztofziobro:
> > I think something like "versioning" storage with `upgradable_storage(VERSION_KEY)` macro might be a good idea, as (I think) it makes storage migration process more clear. But it shouldn't really matter that much, so it's up to you.
> 
>     1. Added - not sure what I think of adding the reserved fields - intuitively it's OK I guess, I don't see us reaching for it too often.
> 
>     2. Token contracts - they all have just one proprietary field - access_control contract address, which has it's value "upgradeable" with a simple message, so I reckon it's not worth wrapping it in it' s own storage type... But YMMV and I can be easily persuaded otherwise.

I think that if it's likely the only field that will ever be there, then it's fine to leave it as is.

krzysztofziobro:
> Another thing to consider is enabling upgrade + migration to be "atomic", by taking `Option<Selector>` in the upgrade message, so migration function's selector can be called right after the upgrade. It might help us avoid downtime, when all calls to contract produce errors, or might even potentially break something, depending on the nature of the migration.
> 
> I don't follow - are you saying we should add an optional argument to `set_code`, that let's one invoke migration functionality in the same tx as setting contract pointer to a new hash?
> 
> Correct me if I'm wrong but `set_code` will always be invoked in the context of the "old" contract code, so it cannot self-migrate, since there are no stored instructions on how to?

When you are doing `set_code` then you have code hash available, so you can make a delegate call. 

If you anticipate that upgrades are only made after contract is stopped, then you can ignore this suggestion.

contracts/access_control/lib.rs
```diff
@@ -25,10 +31,20 @@ mod access_control {
...
+
+    #[derive(Debug)]
+    #[openbrush::upgradeable_storage(STORAGE_KEY)]
```
DamianStraszak, line 37: I'm trying to learn what this macro actually does. So what's the difference between having `#[openbrush::upgradeable_storage(STORAGE_KEY)]` here vs just `#[ink(storage)]` ?

krzysztofziobro: In the "original" contract it does not do that much - it places storage struct under specified storage key (it's clear where to migrate from). If it were in the upgraded version, it additionally allows reading/writing to the uninitialized fields (initializes with default on the first use), making migrations straightforward (at least when no Mappings are involved).

fbielejec: IMO that's the killer feature - it allows one to implicitly initialize the new upgraded contract

fbielejec:
> > Another thing to consider is enabling upgrade + migration to be "atomic", by taking `Option<Selector>` in the upgrade message, so migration function's selector can be called right after the upgrade. It might help us avoid downtime, when all calls to contract produce errors, or might even potentially break something, depending on the nature of the migration.
> > I don't follow - are you saying we should add an optional argument to `set_code`, that let's one invoke migration functionality in the same tx as setting contract pointer to a new hash?
> > Correct me if I'm wrong but `set_code` will always be invoked in the context of the "old" contract code, so it cannot self-migrate, since there are no stored instructions on how to?
> 
> When you are doing `set_code` then you have code hash available, so you can make a delegate call.
> 
> If you anticipate that upgrades are only made after contract is stopped, then you can ignore this suggestion.

I have no idea what/if the upgrades would look like :) IMHO it's best to make it general and open-ended. Check how you like this solution - as per your suggestion added a delegate call to "new me". Since it's a no ABI delegate-call the callback function will have to adhere to some conventions (like take no arguments, be access controlled etc), which i left as a comment

contracts/access_control/lib.rs
DamianStraszak: Something is wrong here. I have made quite a bit of tests here and even with all these macros, this is still using the standard storage key of `00000000`. My suspicion is that we are using the macro `#[openbrush::upgradeable_storage(STORAGE_KEY)]` incorrectly. (the annoying thing though is that the docs page is down today https://727-ventures.github.io/openbrush-contracts/ so its not possible to consult it). 
Did we ever test what it does, vs what is it supposed to do?

DamianStraszak: The way to confirm this is to inspect the `layout` part of the json metadata file. You can clearly see that `_reserved` uses the key of `00000000` no matter what "hardcoded" value we put as an argument to the macro.

This is also consistent with what lands on chain.

contracts/button/lib.rs
```diff
@@ -169,13 +200,13 @@ pub mod button_game {
         /// Deadline is the block number at which the game will end if there are no more participants
         #[ink(message)]
         pub fn deadline(&self) -> BlockNumber {
-            self.last_press + self.button_lifetime
+            self.data.get_or_default().last_press + self.data.get_or_default().button_lifetime
```
deuszx, line 203: My preferred choice would be to panic here, rather than returning the default. It seems dangerous to return some arbitrary (default) data to the client who can't tell whether they're some defaults or actual ones. Falling back to the defaults would also mean the data was not properly initialized, which I would consider a setup error.

DamianStraszak: I actually agree with @deuszx here -- self.data.get().unwrap() would be better, even though it looks ugly. Making it default can cover up some issues that we might otherwise be able to find during testing.

DamianStraszak (approved):

DamianStraszak:
The overall structure of upgradeability looks good to me now.

contracts/button/lib.rs
fbielejec: ok - btw do we know if repeated calls to data.get() are cached?

PR #1100, opened 2023-03-29
A0-2198: Extract copying binary and runtime to separate action

Marcin-Radecki:
# Description

* extract copying a file to a separate action
* re-use in two workflows

## Type of change

# Checklist:

### DIFF
--- a/.github/workflows/_store-production-node-and-runtime.yml
+++ b/.github/workflows/_store-production-node-and-runtime.yml
@@ -51,49 +51,27 @@ jobs:
           aws-region: ${{ env.AWS_REGION }}
 
       - name: Copy production binary to S3 Mainnet AWS bucket
+        uses: ./.github/actions/copy-file-to-s3
         if: >
           env.SECRETS_AWS_MAINNET_ACCESS_KEY_ID != '' &&
           env.SECRETS_AWS_MAINNET_SECRET_ACCESS_KEY != ''
-        shell: bash
-        env:
-          BINARY_DIR: target/production
-          BINARY_FILE: aleph-node
+        with:
+          binary-dir: target/production
+          binary-file: aleph-node
           # yamllint disable-line rule:line-length
-          S3BUCKET_PATH: builds/aleph-node/commits/${{ steps.get-ref-properties.outputs.sha }}/aleph-node
-          S3BUCKET_FILE: aleph-node-${{ steps.get-ref-properties.outputs.sha }}.tar.gz
-          S3BUCKET_URL: s3://${{ secrets.CI_MAINNET_S3BUCKET_NAME }}
-        run: |
-          aws s3api head-object \
-            --bucket '${{ secrets.CI_MAINNET_S3BUCKET_NAME }}' \
-            --key '${{ env.S3BUCKET_PATH }}/${{ env.S3BUCKET_FILE }}' || not_exist=true
-          if [[ "${not_exist}" ]]; then
-            tar -cvzf ${{ env.S3BUCKET_FILE }} -C ${{ env.BINARY_DIR }} ${{ env.BINARY_FILE }}
-            aws s3 cp ${{ env.S3BUCKET_FILE }} \
-              ${{ env.S3BUCKET_URL }}/${{ env.S3BUCKET_PATH }}/${{ env.S3BUCKET_FILE }}
-          else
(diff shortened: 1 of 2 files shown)


.github/workflows/_build-production-node-and-runtime.yml
```diff
@@ -65,49 +65,27 @@ jobs:
...
-            --key '${{ env.S3BUCKET_PATH }}/${{ env.S3BUCKET_FILE }}' || not_exist=true
-          if [[ "${not_exist}" ]]; then
-            tar -cvzf ${{ env.S3BUCKET_FILE }} -C ${{ env.BINARY_DIR }} ${{ env.BINARY_FILE }}
-            aws s3 cp ${{ env.S3BUCKET_FILE }} \
-              ${{ env.S3BUCKET_URL }}/${{ env.S3BUCKET_PATH }}/${{ env.S3BUCKET_FILE }}
-          else
-            echo "File ${{ env.S3BUCKET_FILE }} already exists on CI S3 bucket"
-          fi
+          s3-bucket-path: builds/aleph-node/commits/${{ steps.get-ref-properties.outputs.sha }}/aleph-node
+          s3-bucket-file: aleph-node-${{ steps.get-ref-properties.outputs.sha }}.tar.gz
```
mikogs, line 77: There's one thing that bothers me.  When I see `.tar.gz` at the end of that input, then I assume it's going to be compressed.  However, if it's not there then I would assume that the file is just going to be copied over.
I think we should deal with it somehow.

For example, 
* if `s3-bucket-file` has a suffix of `.tar.gz` (or `.tgz`) and the `binary-file` hasn't, then the file is compressed
* if it's `.zip` and the `binary-file` does not have a `.zip` extension then it's zipped (or we throw an error that we support .tar.gz only)
* otherwise it is just copied over

There is another solution to that as well, we can have a `compress` input set to `'true'` by default, which will toggle the compression, and if the file is not meant to be compressed the it has to be explicitly set to `'false'` (`not 'true'` - because that is a string, not real boolean).
If a file extension is `.tar.gz` or `.tgz` and we do not compress, we should print out a warning that a file that is being copied over has invalid extension.

.github/workflows/_build-production-node-and-runtime.yml
```diff
@@ -65,49 +65,27 @@ jobs:
...
-            --bucket '${{ secrets.CI_MAINNET_S3BUCKET_NAME }}' \
-            --key '${{ env.S3BUCKET_PATH }}/${{ env.S3BUCKET_FILE }}' || not_exist=true
-          if [[ "${not_exist}" ]]; then
-            tar -cvzf ${{ env.S3BUCKET_FILE }} -C ${{ env.BINARY_DIR }} ${{ env.BINARY_FILE }}
-            aws s3 cp ${{ env.S3BUCKET_FILE }} \
-              ${{ env.S3BUCKET_URL }}/${{ env.S3BUCKET_PATH }}/${{ env.S3BUCKET_FILE }}
-          else
-            echo "File ${{ env.S3BUCKET_FILE }} already exists on CI S3 bucket"
-          fi
+          s3-bucket-path: builds/aleph-node/commits/${{ steps.get-ref-properties.outputs.sha }}/aleph-node
```
mikogs, line 76: Something that not necessarily have to be done in this PR...  I was thinking if instead of passing that path in the workflow we could just have a more general action to "store artifact X from repo Y for commit Z" where you would pass X, Y and Z instead of path.  In that scenario, the action could be used in other repositories without worrying that the path is always `builds/Y/commits/Z/X`, as it's going to be concatenated by the action.
But then, this action would have to be renamed to `store-commit-artifact-in-s3` - or it can be a separate action re-using `copy-to-s3`.

.github/workflows/_build-production-node-and-runtime.yml
mikogs: When I think about it again, having `compress` is a sensible thing to do.  However, I'm wondering if the `.tar.gz` shouldn't be applied automatically to target filename if it's about to be compressed and the extension is not there (unless another `no-compression-suffix` input is set to `'false'` - again, `'true'` by default).

.github/workflows/_build-production-node-and-runtime.yml
Marcin-Radecki: Agree it would be value added, yet we don't have any more usecases to actually be covered by that new funcionality.

mikogs (approved):

PR #1104, opened 2023-03-30
A0-2241: Validate token amounts

pmikolajczyk41:
# Description

1. We use `u128` for `FrontendTokenAmount` (to be compatible with AlephZero `Balance`)
2. We use `TokenAmountVar` wrapper over `FpVar` that additionally ensures that the underlying value is at most `1<<128 - 1`
3. A testcase showing previous exploit has been added - after commenting `value.enforce_cmp(...)?` in `TokenAmountVar`, the proof in this testcase will be valid.
4. `snark_relation` macro preserves attributes for relation object types

Note: After 2xApprove, I will publish manually both macro and relations crates, update dependencies and then merge.

## Type of change

- I have added tests


mike1729 (approved):
What a nice fix for such a nasty bug! Great work!

PR #1101, opened 2023-03-29
A0-2229: refactor finalizer, request blocks and verifier to use BlockIdentifier

maciejnems:
# Description
This is next step in unifying database access. Next two steps would be:
* completely unifying HashNum with BlockId (it is not done in this PR, as it is a lot of changes in code) https://github.com/Cardinal-Cryptography/aleph-node/pull/1106
* aleph_node_rpc does not need to be dependent on `B: Block` it can simplify a lot of code https://github.com/Cardinal-Cryptography/aleph-node/pull/1102

I have not changed `BlockchainBackend` in any way even though it could also use `BlockIdentifier`. It will be removed in future when new sync protocol is added.

Change list:
* `BlockIdentifier` is extracted to `finality-aleph/src/lib.rs` from sync module
* `HashNum` implements `BlockIdentifier`
* `Finalizer` uses `BlockIdentifier`
* `Verifier` uses `BlockIdentifier`
* `RequestBlocks` uses `BlockIdentifier`
* Added implementation of `From<AlephJustification> for Justification`

## Type of change

- Refactor

PR #1105, opened 2023-03-30
Fix justification handler not to wait 500 ms

maciejnems:
# Description

Based on top of https://github.com/Cardinal-Cryptography/aleph-node/pull/1101. 

Now justification handler does not wait 500ms on each start, which caused `catch_up` in party not to work correctly. Also might fix some other bugs when we waited for verifier unnecessarily long.

## Type of change

- Bug fix


timorl:
It seems this has been obsoleted by removing literally all the code that it changed, so closing.

PR #1110, opened 2023-03-31
A0-2258: Added wasm target to rust-toolchain.toml

DamianStraszak:
# Description

Adds `targets = [ "wasm32-unknown-unknown" ]` to `rust-toolchain.toml` 

The point of this, is that if you just 1) install cargo, 2) clone repo, and 3) run `cargo build --release` then
- Currently we get the error `Rust WASM toolchain not installed, please install it!`
- After this change, it works right away, because the target is added automatically.


Marcin-Radecki (approved):

Marcin-Radecki:
One more comment: this changes the way some pipelines work, e.g. currently `cliain` or `e2e-client` are built without target wasm, and after this is merged they will be. This is not an issue though.

maciejzelaszczyk (approved):

PR #1113, opened 2023-04-03
A0-2261: Disable slack notification on push to main for button e2e test failure

Marcin-Radecki:


DamianStraszak (approved):

obrok (changes requested):
I disagree with this - we should spam until effective. I'm back, can look into the flakiness. If we're angry that the tests are flaky it would be better to remove them than to hide everything under a jira ticket - a ticket like that is easy to ignore forever.

PR #1114, opened 2023-04-03
A0-2187: Improve button e2e tests

obrok:
# Description

This should fix the most common cause of flakiness for these tests, that is when the two presses we compare occur at roughly equal intervals, giving equal scores instead of one or the other being bigger. It also fixes two other issues:

1. For `the_pressiah_cometh` it's the order of the presses that matters, not the times.
2. When resetting the button, there might be extra tickets in the button from other tests, this is now taken into account.

## Type of change


Marcin-Radecki (approved):

h4nsu (approved):
LGTM

PR #1119, opened 2023-04-04
A0-1848: e2e split

kostekIV:
# Description

E2e tests for committee split logic. We check whether the block finalization stops after 'killing' nodes which are supposed to be in the committee. For these tests the number of nodes increased to 7. 7 nodes produce blocks and 5 of them are in the finalization committee which makes it so that killing 2 of them stops finalization.

# Checklist:

- I have added tests


.github/workflows/on-pull-request-commit.yml
```diff
@@ -438,6 +453,93 @@ jobs:
           follow-up-finalization-check: true
         timeout-minutes: 10
 
+  run-e2e-committee-split-reserved-01:
+    needs: [build-test-node-image-and-e2e-client-image]
+    name: Run committee split test with node-0 and node-1 dead
```
timorl, line 458: Are the roles of these nodes different? Te description doesn't really tell me what the test tests and why it needs three variants.

e2e-tests/src/test/committee_split.rs
```diff
@@ -0,0 +1,151 @@
...
+use log::info;
+
+use crate::{
+    accounts::{get_validator_seed, get_validators_keys, NodeKeys},
+    config::setup_test,
+    rewards::set_invalid_keys_for_validator,
+};
+
+/// time needed for 5 out of 7 block producers to do 3 sessions.
+const SLEEP_DURATION: Duration = Duration::from_secs(126);
```
timorl, line 19: This is suspiciously specific. We cannot really rely on sessions taking a constant time, so it either requires some criteria checking or a time with a margin of error (if applicable).

e2e-tests/src/test/committee_split.rs
```diff
@@ -0,0 +1,151 @@
...
+            Some(new_validators[3..].to_vec()),
+            Some(seats),
+            TxStatus::InBlock,
+        )
+        .await?;
+
+    Ok(())
+}
+
+fn validator_address(index: u32) -> String {
```
timorl, line 50: Huh, we don't have such a function already? If so, could you put it in some common tools, since it seems very generically useful?

e2e-tests/src/test/committee_split.rs
```diff
@@ -0,0 +1,151 @@
...
+#[tokio::test]
+/// Check if reserved node-1 and node-2 are in the finality committee
+async fn split_test_reserved_12() -> anyhow::Result<()> {
+    split_disable(&[0, 1]).await
+}
+
+#[tokio::test]
+/// Check if reserved node-0 and node-2 are in the finality committee
+async fn split_test_reserved_02() -> anyhow::Result<()> {
+    split_disable(&[0, 1]).await
```
timorl, line 127: I think you forgot to adjust the numbers after copying. :P But also, relating this to my other comment about the meaning of these tests – what exactly are you trying to check here?

timorl (approved):
Some comments, but in general looks good.

e2e-tests/src/test/committee_split.rs
kostekIV: facepalm :P Here I am trying to check whether all of [node-0, node-1, node-2] plus one non-reserved node (hard to pinpoint which exactly in the tests) are actually in the finality committee.

e2e-tests/src/test/committee_split.rs
timorl: I would avoid weird-looking numbers at least – make it 130 and add an "about" or "at least" to the comment to make things clearer please.

e2e-tests/src/test/committee_split.rs
timorl: Hm, but these are in separate tests, so you are not really checking them together anyway... I guess running these three tests in a loop in a single function is not an option?

It really feels weird when these are separate tests, since they seem to be checking very specific combinations in a very specific setting, while covering all cases but _only_ in this setting. This looks to me as if something is going wrong. I would respect having one test with 2/3 nodes picked randomly more, but adding randomness is always a bit of a problem... :/ If you don't have a better idea feel free to leave it as-is (after fixing the copy problem ofc ;p), but maybe you can figure something reasonable out.

lesniak43 (approved):

e2e-tests/src/test/committee_split.rs
kostekIV: done

e2e-tests/src/test/committee_split.rs
kostekIV: done

PR #1122, opened 2023-04-05
A0-2182: Fix fast price drop when not enough decimals

obrok:
# Description

This is a cleaner version of the marketplace price computation that works well both for cases where the price drops by more and less than 1 pico per block. The previous version assumed that the price should drop by at least 1 pico per block, so if the price was very low (say 100 pico), then it would only take 100 blocks to reach the minimum price, instead of the advertised `auction_length` blocks.

Note that we still need to choose a suitable initial price for devnet/testnet/mainnet. I'm leaving it for next week, because I'm told that we're going to be changing the rewards to be on the order of 1 game token (10^12 pico) instead of 1 pico, so any setting needs to take that into account as well.

## Type of change

# Checklist:

- [x] I have added tests


deuszx:
The title says "fix fast price drop when not enough decimals" but I can't make it out of the code where the actual fix is - i.e. which part of the old code was incorrect.

obrok:
> The title says "fix fast price drop when not enough decimals" but I can't make it out of the code where the actual fix is - i.e. which part of the old code was incorrect.

See updated description.

deuszx (approved):

PR #1128, opened 2023-04-12
A0-2321: Add more control over featurenets

mikogs:


mikogs:
create-featurenet!

mikogs:
delete-featurenet!

.github/workflows/updatenet-update.yml
```diff
@@ -0,0 +1,88 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      name:
+        description: Name of deployment (updatenet)
+        required: true
+        type: string
+      destination:
+        description: Node image tag, commit sha short or "testnet" or "mainnet"
```
Marcin-Radecki, line 12: I think we don't need neither mainnet nor testnet here since the image we update to would be really an arbitrary node image, so commit sha would just do

.github/workflows/updatenet-update.yml
```diff
@@ -0,0 +1,88 @@
...
+        uses: bobheadxi/deployments@v1.1.0
+        id: deployment
+        with:
+          step: start
+          token: ${{ secrets.CI_GH_TOKEN }}
+          env: updatenet-${{ inputs.name }}
+          override: true
+
+      - name: Create featurenet data, app and deploy it
+        uses: ./.github/actions/create-featurenet
```
Marcin-Radecki, line 59: Does it rewrite whole FR from scratch, including wiping DB? Then unfortunately we need to rework it, this workflow should only do following on the given update net:
1. Stop old binary
2. Start a new binary  
nothing more

.github/workflows/_create-featurenet.yml
```diff
@@ -1,64 +1,53 @@
...
+  workflow_call:
+    inputs:
+      hotfix:
+        description: Set to 'true' if it is a hotfix environment
+        required: false
+        type: boolean
+        default: false
 
 env:
   RPC_TESTNET_URL: https://rpc.test.azero.dev
```
Marcin-Radecki, line 14: Not related to this PR, but I cannot see neither of those 4 envs used anywhere

Marcin-Radecki (changes requested):
This is solid work, I am happy to approve everything except .github/workflows/updatenet-update.yml  file, so maybe we move that file to another PR and merge this PR?

.github/workflows/updatenet-update.yml
mikogs: Actually it was happening.  However, the condition wasn't quite clear.  It's fixed now and there's an explicit `update` input to `create-featurenet` action.

Marcin-Radecki (approved):

.github/workflows/updatenet-update.yml
```diff
@@ -0,0 +1,89 @@
...
+    inputs:
+      name:
+        description: Name of deployment (updatenet)
+        required: true
+        type: string
+      destination:
+        description: Node image tag
+        required: true
+        type: string
+        default: testnet
```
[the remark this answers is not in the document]
mikogs: It's removed now

PR #1130, opened 2023-04-13
A0-1818: Prepare `aleph-client` for publishing

pmikolajczyk41:
# Description

Versioning `aleph-client` is a pretty tricky issue. On the one hand, in many side tools (like `ink-wrapper` or `shielder-relayer`) we would strongly prefer relying on `aleph-client` from registry than from git. On the other hand, semantic meaning of the `aleph-client`'s version is very unclear - the newest version is very probable to be incompatible with both testnet and mainnet. It is guaranteed to be compatible only with the attached runtime.

The issue how to tackle compatibility with production nets is somehow isolated. We might go for some feature-gating (which might be very onerous and hard to maintain) or some tagging model. For now, we want just publish code to unblock side tooling.

The only problem with publishing it now are git- and local dependencies:
 - local:
   - `pallet-baby-liminal` - this crate is not published, but the whole dependency served only for one simple type alias; I've copied it and removed dependency
   - `primitives` - this crate is not published (although it probably makes sense to publish it); it served for a couple of simple type aliases (I've copied them) and one function - `era_payout` - which I had to copy as well; imho this is awful
 - git:
   - `frame_support` and `sp_runtime` - both from our Substrate fork; I was able to remove them, thanks to reexports from `subxt`
   - `pallet-contracts-primitives` - from our Substrate fork; it provides us with `ContractExecResult` which cannot be imported from anywhere else, nor copied without copying a lot of stuff. Therefore we have to keep it. However, we cannot keep only git pointer to our fork. Instead, we have to point to `version = 0.7.0` on crates.io, which internally uses just Substrate code. However, we should be pretty safe here, since our changes on the fork don't touch this part.

# Other solutions
1. We do not publish `aleph-client` at all. Make @obrok et.al. your haters.
2. We publish `primitives` (very soon) and thus we don't have to copy all the type aliases and `era_payout` from there. This will require maintenance.
3. We publish all our Substrate dependencies from fork. Not so much more maintenance, probably huge gains.

# Testing

Try running `cargo publish --dry-run` in `aleph-client/` directory.

## Type of change


obrok (approved):
This seems OK to me as a first step, I'd recommend creating tickets for publishing primitives and (potentially) substrate immediately on merging this PR, so that it doesn't get lost.

Note that `cargo publish --dry-run` doesn't guarantee that the publish will succeed - there are some validations, that `crates` does on the server side which are not reflected in `--dry-run`. However, you've already filled in the various metadata fields, so you should be good.

Re publishing substrate - we'd have to publish it under another name, wouldn't we? Like `aleph-substrate`? Why do we even need our own fork again?

obrok:
Oh, one more thing, I'd recommend starting a CHANGELOG, maybe based on https://keepachangelog.com/en/1.1.0/

lesniak43 (approved):

Marcin-Radecki (approved):
* Is there a way to sneak what crates.io page would look like for published `aleph-client` package before we publish?
* we need a ticket to justify if we use creates.io version in `cliain` and other dev tool crates

PR #1131, opened 2023-04-14
A0-1816: Tweak invalid filenames in one of the workflows

mikogs:


.github/workflows/on-release-branch-commit.yml
```diff
@@ -28,7 +28,9 @@ jobs:
         run: |
           set +e
           fail=0
-          for i in deploy-testnet.yml deploy-mainnet.yml; do
```
Marcin-Radecki, line 31: Let's diff all the workflows

Marcin-Radecki (changes requested):

.github/workflows/on-release-branch-commit.yml
mikogs: This has been changed and the workflow is going to check all action and workflow .yml files.

Marcin-Radecki (approved):

PR #1135, opened 2023-04-18
Randomize selection

kostekIV:
# Description

Main change is in pallet-elections. Now every era before setting the EraValidators the order of the validators is shuffled based on era number.

Also:
* Expose runtime-api which allows to query the committee for some session as long as this session is from the current era. This should be used later in UI to calculate committees for the current era. This can be done in session `N` if session `N + 1` starts era. <- this can be used by UI to calculate committee for all of sessions in the next era!
* Bunch of adjustments to the e2e tests that relied on the specific order and selection of the committee

## Type of change

# Checklist:


pallets/committee-management/src/impls.rs
```diff
@@ -423,6 +437,63 @@ impl<T: Config> Pallet<T> {
...
+            // Shouldn't happen
+            None => {
+                return Err(SessionValidatorError::Other(
+                    "No known starting session for current era".encode(),
+                ))
+            }
+        };
+
+        // Historical session
+        if session < active_starting_index - 1 {
```
DamianStraszak, line 467: This is most likely off-by-one and should be `session < active_starting_index`...

pallets/committee-management/src/impls.rs
```diff
@@ -423,6 +437,63 @@ impl<T: Config> Pallet<T> {
...
+
+    /// Calculates committee for the given session.
+    /// If the era `E` starts in the session `a`, and ends in session `b` then from session `a-1`
+    /// to session `b-1` this function can answer question who will be in the committee in the era `E`.
+    /// In the last session of the era `E` this can be used to determine all of the sessions in the
+    /// era `E+1`.
+    pub fn session_committee_for_session(
+        session: SessionIndex,
+    ) -> Result<SessionCommittee<T::AccountId>, SessionValidatorError> {
+        let ae = match T::EraInfoProvider::active_era() {
```
DamianStraszak, line 449: I think there are some issues in this function related to our lovely off-by-ones :fearful: In particular it is confusing to me why we need to look what's the `current_session` at all. (I might be wrong here, if so then I'm sorry :|  I know it's painful to think about these things all the time :P)
Anyway, my understanding is that `current_era_validators()` is synchronized with `current_era()` and not `active_era()`. This means that if we query `ce = T::EraInfoProvider::current_era()` then compute `ce_session_start` and `ce_session_end` then we can just check:
- `ce_session_start<= session <= ce_session_end`
and if the above holds, then `Self::rotate_committee_inner(&era_validators, committee_seats, session)` should output the correct result (with `era_validators` and `committee_seats` computed as below).

DamianStraszak (changes requested):
Looks great, although I have left some small comments just to make sure :)

pallets/committee-management/src/impls.rs
kostekIV: This is actually because at the point of the session `active_starting_index - 1` we already have everything we need to calculate the committee for the era that `active_starting_index` starts

pallets/committee-management/src/impls.rs
kostekIV: We need to look at what current session is because we can only answer query for lets say era N in sessions `SS -1` to `ES - 1` where `SS` starts era `N` and `ES` ends it.

kostekIV: But the solution with CurrentEra instead is cleaner. I will go with it

primitives/src/lib.rs
```diff
@@ -201,6 +227,13 @@ pub enum ApiError {
     DecodeKey,
 }
 
+#[derive(Encode, Decode, PartialEq, Eq, Debug)]
+pub enum SessionValidatorError {
+    SessionTooMuchIntoFuture { upper_limit: SessionIndex },
```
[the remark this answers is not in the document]
kostekIV: done

bin/runtime/src/lib.rs
```diff
@@ -1011,6 +1002,12 @@ impl_runtime_apis! {
         fn next_session_finality_version() -> FinalityVersion {
             Aleph::next_session_finality_version()
         }
+
+        fn session_committee(
+            at: SessionIndex,
```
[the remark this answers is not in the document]
kostekIV: sure, done

pallets/committee-management/src/impls.rs
kostekIV: Reworked :)

pallets/committee-management/src/impls.rs
```diff
@@ -307,33 +304,50 @@ impl<T: Config> Pallet<T> {
         CurrentAndNextSessionValidatorsStorage::<T>::put(session_validators);
     }
 
-    pub(crate) fn rotate_committee(
+    pub(crate) fn rotate_committee_inner(
```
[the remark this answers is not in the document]
kostekIV: Rotate mutates the `CurrentAndNextSessionValidatorsStorage`. I would leave the naming as is but I may be biased here :P

pallets/committee-management/src/impls.rs
DamianStraszak: Yeah, but in that case I think you confused `session` with `current_session` :P But this is gone now :)

pallets/committee-management/src/impls.rs
```diff
@@ -423,6 +436,56 @@ impl<T: Config> Pallet<T> {
...
+            None => {
+                return Err(SessionValidatorError::Other(
+                    "No known starting session for current era".encode(),
+                ))
+            }
+        };
+
+        // Historical session
+        if session < current_starting_index {
+            return Err(SessionValidatorError::OldEra);
```
DamianStraszak, line 465: Mayba it would be better for the user to combine this error with `SessionValidatorError::SessionTooFarIntoFuture` and output something like `SessionValidatorError::SessionNotWithinRange { lower_limit: current_starting_index, upper_limit: planned_era_end}`

Makes sense?

pallets/committee-management/src/impls.rs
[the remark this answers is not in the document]
DamianStraszak: I think you refer to `rotate_committee` which at the end calls `store_session_validators` and it's fine. This name I would keep. But what I mean are the functions `rotate_committee_inner` and `rotate`.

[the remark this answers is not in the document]
kostekIV: yeo, I got confused here for a moment :P

pallets/committee-management/src/impls.rs
kostekIV: :upside_down_face: right

pallets/committee-management/src/impls.rs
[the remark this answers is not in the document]
kostekIV: Done

pallets/committee-management/src/impls.rs
kostekIV: Sure, done

DamianStraszak (approved):

primitives/src/lib.rs
```diff
@@ -197,6 +223,16 @@ pub enum ApiError {
     DecodeKey,
 }
 
+#[derive(Encode, Decode, PartialEq, Eq, Debug)]
+pub enum SessionValidatorError {
+    SessionNotWithinRange {
+        lower_limit: SessionIndex,
+        upper_limit: SessionIndex,
+    },
+    OldEra,
```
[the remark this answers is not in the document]
kostekIV: again, you are right :) removed

bin/runtime/src/lib.rs
```diff
@@ -1011,6 +1001,12 @@ impl_runtime_apis! {
         fn next_session_finality_version() -> FinalityVersion {
             Aleph::next_session_finality_version()
         }
+
+        fn session_committee(
+            session: SessionIndex,
```
fixxxedpoint, line 1006: Is this argument necessary? Aren't we interested only in current's session committee or at some given block where we are manually computing it from some session number?

primitives/src/lib.rs
```diff
@@ -197,6 +223,15 @@ pub enum ApiError {
     DecodeKey,
 }
 
+#[derive(Encode, Decode, PartialEq, Eq, Debug)]
+pub enum SessionValidatorError {
+    SessionNotWithinRange {
+        lower_limit: SessionIndex,
+        upper_limit: SessionIndex,
+    },
+    Other(Vec<u8>),
```
fixxxedpoint, line 232: Is there any reason for not using `String` here?

fixxxedpoint (approved):
LGTM

bin/runtime/src/lib.rs
kostekIV: thats the whole point of this function to be able to compute committee for **any** session in the current era

e2e-tests/src/rewards.rs
```diff
@@ -38,16 +38,20 @@ type RewardPoint = u32;
...
+    }
+
     let mut rng = rand::thread_rng();
     for con in &controller_connections {
         let mut invalid_keys = [0u8; 64];
         rng.fill(&mut invalid_keys);
 
-        // wait until our node is forced to use new keys, i.e. current session + 2
-        con.set_keys(invalid_keys.to_vec().into(), TxStatus::InBlock)
+        con.set_keys(invalid_keys.to_vec().into(), TxStatus::Submitted)
```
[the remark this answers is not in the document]
kostekIV: :+1:,  this is leftover after tinkering with e2e tests not working :P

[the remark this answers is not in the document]
kostekIV: ok done

PR #1136, opened 2023-04-18
A0-2350: rewards in full units

fbielejec:
- Rewards in pico units are not sexy, we make sure to always return at least 1 full unit.
- gas optimization when calculating linear
- introduce extra event for indexer to listen to. It will include ThePressiah rewards in the scoreboard

### DIFF
--- a/contracts/button/lib.rs
+++ b/contracts/button/lib.rs
@@ -30,6 +30,9 @@ pub mod button_game {
 
     use crate::errors::GameError;
 
+    pub const ONE_TOKEN: Balance = 1_000_000_000_000;
+    pub const ONE_HUNDRED_TOKENS: Balance = 1_000_000_000_000;
+
     /// Result type
     type ButtonResult<T> = core::result::Result<T, GameError>;
 
@@ -548,17 +551,17 @@ pub mod button_game {
                     amount,
                     0,
                     data.button_lifetime as Balance,
-                    1_000_000_000_000,
-                    100_000_000_000_000,
+                    ONE_TOKEN,
+                    ONE_HUNDRED_TOKENS,
                 ),
 
-                Scoring::ThePressiahCometh => amount.saturating_mul(1_000_000_000_000),
+                Scoring::ThePressiahCometh => amount.saturating_mul(ONE_TOKEN),
             };
 
             PSP22MintableRef::mint(&data.reward_token, to, scaled_amount)?;
 
             Self::emit_event(
-                Self::env(),
+                self.env(),
                 Event::RewardMinted(RewardMinted {
                     when: self.env().block_number(),
                     reward_token: data.reward_token,
--- a/contracts/button/lib.rs
+++ b/contracts/button/lib.rs
@@ -58,6 +58,19 @@ pub mod button_game {
         score: Balance,
     }
 
+    /// Event emitted when a reward token is minted to a players account
+    ///
+    /// Could be a regular player or the Pressiah
+    #[ink(event)]
+    #[derive(Debug)]
+    pub struct RewardMinted {
+        when: BlockNumber,
+        #[ink(topic)]
+        reward_token: AccountId,
+        to: AccountId,
+        amount: Balance,
+    }
+
     /// Event emitted when the finished game is reset and pressiah is rewarded
     #[ink(event)]
     #[derive(Debug)]
@@ -81,8 +94,6 @@ pub mod button_game {
         BackToTheFuture,
         /// The reward increases linearly with the number of participants
         ThePressiahCometh,
-        /// Placeholder for the default implementation
-        Default,
     }
 
     #[derive(Debug)]
@@ -499,7 +510,6 @@ pub mod button_game {
                 Scoring::EarlyBirdSpecial => deadline.saturating_sub(now) as Balance,
                 Scoring::BackToTheFuture => now.saturating_sub(last_press) as Balance,
                 Scoring::ThePressiahCometh => (presses + 1) as Balance,
-                Scoring::Default => panic!("Should never get here"),
             }
         }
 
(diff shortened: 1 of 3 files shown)


pmikolajczyk41 (approved):
everything looks okay to me, thus approve, but I kinda miss the point of this scaling by 10^12 - afaik, these 12 decimal places are relevant only to the chain's _native_ token

is this because tooling was displaying these rewards in an 'unpleasant' way, like it would do for low amounts of AZERO?

also, could this change introduce some numerical problems to rewards marketplace?

contracts/button/lib.rs
```diff
@@ -527,7 +537,23 @@ pub mod button_game {
         }
 
         fn mint_reward(&self, to: AccountId, amount: Balance) -> ButtonResult<()> {
-            PSP22MintableRef::mint(&self.data.get().unwrap().reward_token, to, amount)?;
+            let data = self.data.get().unwrap();
+
+            // scale the amount to always pay out full token units
+            let scaled_amount = amount.saturating_mul(1000000000000);
```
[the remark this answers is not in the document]
deuszx: IIUC, this number is hard-coded here with the assumption that it's the `token_decimals()` of the reward token. The alternative, to hardcoding, would be to read it on initialization/computation.

contracts/marketplace/lib.rs
```diff
@@ -377,7 +381,7 @@ pub mod marketplace {
             let data = self.data.get().unwrap();
             linear_decrease(
                 data.current_start_block.into(),
-                self.average_price().saturating_mul(data.sale_multiplier),
+                Self::_average_price(&data).saturating_mul(data.sale_multiplier),
```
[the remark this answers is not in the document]
fbielejec: sure - you do not pay for lifting and decoding this whole struct from the storage

contracts/button/lib.rs
```diff
@@ -58,6 +58,19 @@ pub mod button_game {
         score: Balance,
     }
 
+    /// Event emitted when a reward token is minted to a players account
+    ///
+    /// Could be a regular player or the Pressiah
+    #[ink(event)]
+    #[derive(Debug)]
+    pub struct RewardMinted {
+        when: BlockNumber,
```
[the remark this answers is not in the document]
fbielejec: > everything looks okay to me, thus approve, but I kinda miss the point of this scaling by 10^12 - afaik, these 12 decimal places are relevant only to the chain's _native_ token
> 
> is this because tooling was displaying these rewards in an 'unpleasant' way, like it would do for low amounts of AZERO?
> 
> also, could this change introduce some numerical problems to rewards marketplace?

Because rewards are paid out proportionally to the time-to-deadline (or time elapsed) and games live for ~8600 blocks or less, they are always in pico/nano units. Wallets aren't best suited to displaying those and it might generally be poorly understood that you got e.g 1e-8 of a reward. 

@obrok recently made some fixes to the linear price drop calculations so this should not have an impact.

fbielejec:
> everything looks okay to me, thus approve, but I kinda miss the point of this scaling by 10^12 - afaik, these 12 decimal places are relevant only to the chain's _native_ token
> 
> is this because tooling was displaying these rewards in an 'unpleasant' way, like it would do for low amounts of AZERO?
> 
> also, could this change introduce some numerical problems to rewards marketplace?

One more solution comes to my mind: one could map the scores (which live in [0, button_lifetime] to some domain (say [1e12, 100e12]), thus we would always pay out rewards that are in the same range, regardless of the game length. Caveat would be one extra calculation and the associated gas costs, but curious to hear opinions?

contracts/button/lib.rs
```diff
@@ -527,7 +537,23 @@ pub mod button_game {
         }
 
         fn mint_reward(&self, to: AccountId, amount: Balance) -> ButtonResult<()> {
-            PSP22MintableRef::mint(&self.data.get().unwrap().reward_token, to, amount)?;
+            let data = self.data.get().unwrap();
+
+            // scale the amount to always pay out full token units
+            let scaled_amount = amount.saturating_mul(1_000_000_000_000);
```
h4nsu, line 543: This should be a `const` maybe?

contracts/button/lib.rs
```diff
@@ -527,7 +537,23 @@ pub mod button_game {
...
+
+            PSP22MintableRef::mint(&data.reward_token, to, scaled_amount)?;
+
+            Self::emit_event(
```
h4nsu, line 547: Nit: I think mixing `self.` and `Self::` within one method is kinda ugly (and probably confusing to people who try to learn by reading such code). I'd propose to use `Self::` only where you really have to (constructor)

h4nsu (approved):

deuszx:
> > everything looks okay to me, thus approve, but I kinda miss the point of this scaling by 10^12 - afaik, these 12 decimal places are relevant only to the chain's _native_ token
> > is this because tooling was displaying these rewards in an 'unpleasant' way, like it would do for low amounts of AZERO?
> > also, could this change introduce some numerical problems to rewards marketplace?
> 
> One more solution comes to my mind: one could map the scores (which live in [0, button_lifetime] to some domain (say [1e12, 100e12]), thus we would always pay out rewards that are in the same range, regardless of the game length. Caveat would be one extra calculation and the associated gas costs, but curious to hear opinions?

Calculation complexity aside, there's also a "business questions" whether games with different lengths are more difficult/easier to play and should be rewarded differently. One could say that longer games always becomes the short one but not the vice versa - would that mean that the shorter games are more difficult to play?

I was also thinking that maybe the reward token should have 0 decimals and `RewardTokenRef::mint(1)` would award 1 whole unit of the reward token.

fbielejec:
> 1_000_000_000_000

It's not possible as DEX calculations need fractional units. I think the last commit fixes it anyway

contracts/button/lib.rs
fbielejec: IMHO it's not possible to use self for `emit_event` either - it's a static method (i.e. first argument is not `&self` )

contracts/button/lib.rs
fbielejec: Done

contracts/button/lib.rs
```diff
@@ -30,6 +30,9 @@ pub mod button_game {
 
     use crate::errors::GameError;
 
+    pub const ONE_TOKEN: Balance = 1_000_000_000_000;
+    pub const ONE_HUNDRED_TOKENS: Balance = 1_000_000_000_000;
```
pmikolajczyk41, line 34: looks like a single token :P

PR #1138, opened 2023-04-19
A0-2303: aleph-client add fn to get finalizer

kostekIV:


aleph-client/src/pallets/aleph.rs
```diff
@@ -81,6 +83,14 @@ impl<C: ConnectionApi> AlephApi for C {
 
         self.rpc_call(method.to_string(), params).await.unwrap()
     }
+
+    async fn emergency_finalizer(&self, at: Option<BlockHash>) -> Option<AccountId> {
+        let addrs = api::storage().aleph().emergency_finalizer();
+
+        self.get_storage_entry_maybe(&addrs, at)
+            .await
+            .map(|public| public.0 .0.into())
```
DamianStraszak, line 92: Why do we map to AccountId? This is just a "raw" ed25519 pubkey, it is not associated with any particular account

DamianStraszak (changes requested):

e2e-tests/src/test/emergency_finalizer.rs
```diff
@@ -0,0 +1,26 @@
...
+
+#[tokio::test]
+async fn set_emergency_finalizer_test() -> anyhow::Result<()> {
+    let config = setup_test();
+    let (finalizer, _seed) = AlephKeyPair::generate();
+    let public: AccountId = finalizer.public().0.into();
+    let root = config.create_root_connection().await;
+    let current_finalizer = root.as_connection().emergency_finalizer(None).await;
+
+    assert!(current_finalizer.is_none());
```
fixxxedpoint, line 17: nit: perhaps it would be a good idea to check just whether the retrieved one differs from the one we set. This way this test should be "more compatible" with other tests/doesn't assume it was run on a fresh chain. btw it's little bit confusing that a finalizer can be `None` but we can't set it to `None` (not related with this PR).

fixxxedpoint (approved):
lgtm. Left a small nit.

aleph-client/src/pallets/aleph.rs
kostekIV: I see you use raw bytes in your script. Would it be ok here to return [u8; 32]?

DamianStraszak: Yes. I think it's fine, because you normally query it to check against some other key (in whatever format).

DamianStraszak (approved):

e2e-tests/src/test/emergency_finalizer.rs
kostekIV: ok done. In my free time I will add option to set finalizer to None

aleph-client/src/pallets/aleph.rs
kostekIV: done

PR #1143, opened 2023-04-20
A0-1771: Configurable lenient threshold

kostekIV:


pallets/committee-management/src/impls.rs
```diff
@@ -138,12 +138,13 @@ fn calculate_adjusted_session_points(
     blocks_to_produce_per_session: u32,
     blocks_created: u32,
     total_possible_reward: u32,
+    lenient_threshold: Perquintill,
 ) -> u32 {
     let performance =
         Perquintill::from_rational(blocks_created as u64, blocks_to_produce_per_session as u64);
 
-    // when produced more than 90% expected blocks, get 100% possible reward for session
-    if performance >= LENIENT_THRESHOLD {
```
Marcin-Radecki, line 146: Please raise a task to update wallet https://github.com/Cardinal-Cryptography/apps/blob/alephzero/packages/page-staking/src/Performance/CurrentList.tsx#L48 code to use `committee_management.get_lenient_threshold_percentage` instead of hardcoded value

pallets/committee-management/src/lib.rs
```diff
@@ -28,31 +28,57 @@ mod migration;
...
 pub type TotalReward = u32;
 #[derive(Decode, Encode, TypeInfo, PartialEq, Eq)]
 pub struct ValidatorTotalRewards<T>(pub BTreeMap<T, TotalReward>);
 
-#[derive(Decode, Encode, TypeInfo)]
-struct CurrentAndNextSessionValidators<T> {
-    pub next: SessionValidators<T>,
-    pub current: SessionValidators<T>,
+#[derive(Decode, Encode, TypeInfo, Default)]
+pub struct ChangingValue<T> {
```
DamianStraszak, line 46: I like a lot what you did here, adding the ChangingValue. This is generally a great addition.

However... xD

I'm not so sure for this particular case this is necessary. I mean why don't we hold the lenient_threshold as `Perquintill` and just update it normally? I can see no disadvantage:
- It is simpler to code, and simpler to understand the code
- When using it, we don't need to wait a full session for the new value to kick in. Which in this case might not be a big deal, but still...

DamianStraszak (changes requested):

pallets/committee-management/src/lib.rs
kostekIV: I thought it might be unfair to others who had different threshold in the era :P

aleph-client/src/pallets/committee_management.rs
```diff
@@ -145,6 +160,14 @@ impl<C: ConnectionApi + AsConnection> CommitteeManagementApi for C {
             .at(&addrs)
             .map_err(|e| e.into())
     }
+
+    async fn get_lenient_threshold_percentage(&self, at: Option<BlockHash>) -> Option<Perquintill> {
+        let addrs = api::storage().committee_management().lenient_threshold();
+
+        self.get_storage_entry_maybe(&addrs, at)
+            .await
+            .map(|lt| Perquintill::from_perthousand(lt.current.0))
```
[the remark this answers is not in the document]
kostekIV: If I recall correctly internall represantation (`.0`) is kept as perthousand :|

pallets/committee-management/src/lib.rs
DamianStraszak: Hah, yeah, that's a valid reason :P. I didn't even realize it's changing on era boundaries, not session boundaries. Still, I think it's better to give more control to the "admin" here -- they can choose to do that on era change if they wish :) But normally we will probably want the change with no delay.

aleph-client/src/pallets/committee_management.rs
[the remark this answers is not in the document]
DamianStraszak: Uhh, this looks like proper hacking, to the extent that I have trouble verifying if it's correct or not. At this point the "type safety" which we get from the static type generation is actually harming us :| 

Is it possible to instead specify manually the type it should decode the value to? So we would specify the true `Perquintill` and not the autogenerated one :P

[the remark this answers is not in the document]
kostekIV: ok, done

pallets/committee-management/src/lib.rs
kostekIV: ok, done

aleph-client/src/pallets/committee_management.rs
```diff
@@ -145,6 +161,14 @@ impl<C: ConnectionApi + AsConnection> CommitteeManagementApi for C {
             .at(&addrs)
             .map_err(|e| e.into())
     }
+
+    async fn get_lenient_threshold_percentage(&self, at: Option<BlockHash>) -> Option<Perquintill> {
+        let addrs = api::storage().committee_management().lenient_threshold();
+
+        self.get_storage_entry_maybe(&addrs, at)
+            .await
+            .map(|lt| Perquintill::decode(&mut &*lt.encode()).unwrap())
```
DamianStraszak, line 170: This should ideally use `decode_all` for a little bit of safety.

DamianStraszak (approved):

lesniak43 (approved):

aleph-client/src/pallets/committee_management.rs
deuszx: Is it OK to unwrap and panic? We could log an error and fallback to some default instead?

kostekIV: if we will really panic here I don't know what default to really use? Plus I don't see how we will ever panic here but I may be proven wrong in the future

deuszx: I'd default to something where users don't take the hit - i.e. full leniency :P

DamianStraszak: The panic would probably mean that someone is using aleph-client incompatible with the chain version. I'm treating this library as something we use internally for tests, so I don't mind panicking.

deuszx: Ah, I didn't notice it's in `aleph_client` 🙈 I thought it's in the node. Agree then.

kostekIV: How would that help? In such cases, we would lie to them that they are getting full rewards while they could be actually losing them. Also, there is an accepted plan to divide aleph-client into chain agnostic and rest. This fn would go to the other part which would be for 'internal' use only such as tests etc. We have already reverted changes that made it possible to push it into crates.io so it is already only used in tests

kostekIV: added decode_all :)

PR #1144, opened 2023-04-20
A0-2023: Add pallet assets

pmikolajczyk41:
# Description

Integrating `pallet_assets` with `aleph-node`. Also, for security, we add `permissioned_assets` feature to ensure that only `root` can maintain (create) assets.

## Type of change

# Checklist:


bin/runtime/src/lib.rs
```diff
@@ -754,6 +757,53 @@ impl pallet_identity::Config for Runtime {
     type WeightInfo = pallet_identity::weights::SubstrateWeight<Self>;
 }
 
+parameter_types! {
+    /// We compute the deposit required for an account based on the size of the `AssetAccount`
+    /// struct.
+    pub const AssetAccountDeposit: Balance =
```
DamianStraszak, line 763: This is probably quite expensive because of using `LEGACY_DEPOSIT_PER_BYTE`, perhaps using something like `CONTRACT_DEPOSIT_PER_BYTE` would be better, but it's actually tricky to get it right... Here is some methodology for computing these fees https://www.notion.so/cardinalcryptography/Contracts-storage-fee-30f8019d129144a99698136d65f390cc#6aa42d8761bb448da9b6dbfaef19bbcc

DamianStraszak (changes requested):
The PR is nice -- great effort in documenting all the parameters of the integration. However, given the points in https://www.notion.so/cardinalcryptography/Assets-77de8257f4464601b311975dd43383fd we should probably wait with merging it till we make a final decision to do so... 

I will keep my review as "Request changes" until we discuss this in a suitable group.

mike1729 (approved):

PR #1151, opened 2023-04-24
A0-2154: Block abstraction

lesniak43:
# Description

`Block` trait for the purpose of the block sync, also new types in `primitives`.

## Type of change

# Checklist:


finality-aleph/src/sync/mock/mod.rs
```diff
@@ -93,6 +93,38 @@ impl Header for MockHeader {
...
+}
+
+impl MockBlock {
+    #[allow(dead_code)]
+    pub fn for_header(header: MockHeader) -> Self {
+        Self { header }
+    }
+}
+
+impl Header for MockBlock {
```
kostekIV, line 108: MockBlock -> MockHeader?

lesniak43: nah, I want the `MockBlock` to have `id` and `parent_id`, similar to `impl Header for MockJustification { ... ` in line 143

finality-aleph/src/sync/mock/mod.rs
```diff
@@ -4,7 +4,7 @@ use codec::{Decode, Encode};
 use sp_core::H256;
 
 use crate::{
-    sync::{ChainStatusNotification, Header, Justification as JustificationT},
+    sync::{Block as BlockT, ChainStatusNotification, Header, Justification as JustificationT},
```
timorl, line 7: Aaaa, `Header` but `JustificationT`, there was already an inconsistency here and you are only adding to it, but please, make this consistent. >.<

I soft prefer the version without `T`, since it's not necessary, but any consistent approach is acceptable.

timorl (approved):
I'm still worried about how the tree structure of both blocks and justifications having headers and typing constraints stemming from that, but I guess we will see once we get there and fix it then if necessary.

kostekIV (approved):

finality-aleph/src/sync/mock/mod.rs
lesniak43: fixed

PR #1152, opened 2023-04-24
A0-2390: Interface for requesting blocks

lesniak43:
# Description

Interface for requesting blocks, currently has no real functionality and is not used.

## Type of change

# Checklist:


finality-aleph/src/sync/service.rs
```diff
@@ -40,6 +41,8 @@ pub struct Service<
     chain_events: CE,
     justifications_from_user: mpsc::UnboundedReceiver<J::Unverified>,
     additional_justifications_from_user: mpsc::UnboundedReceiver<J::Unverified>,
+    _block_requests_from_user:
+        mpsc::UnboundedReceiver<<<J as Justification>::Header as Header>::Identifier>,
```
kostekIV, line 45: type alias for this receiver? We should also maybe add such things to style guides :P

finality-aleph/src/sync/service.rs
```diff
@@ -69,12 +81,20 @@ impl<
         finalizer: F,
         period: SessionPeriod,
         additional_justifications_from_user: mpsc::UnboundedReceiver<J::Unverified>,
-    ) -> Result<(Self, impl JustificationSubmissions<J> + Clone), HandlerError<J, CS, V, F>> {
+    ) -> Result<
+        (
+            Self,
+            impl JustificationSubmissions<J> + Clone,
+            impl RequestBlocks<<<J as Justification>::Header as Header>::Identifier>,
```
kostekIV, line 88: Hmm, what is the benefit of returning impl here over specific type?

kostekIV (approved):
lgtm :)

finality-aleph/src/sync/service.rs
lesniak43: just duplicating what was already here, but I guess this communicates the intention much better than saying "it's a channel", and there's only, tiny, compilation time cost, right?

@timorl - would you explain more pls?

finality-aleph/src/sync/service.rs
timorl: Some alias would really be nice. <_<

finality-aleph/src/sync/service.rs
timorl: The compiler ensures that external code will not depend on any other properties of the returned type, which makes replacing it later with something else much simpler and generally ensures that we are doing what we expect. The smaller the exposed API, the better.

finality-aleph/src/sync/service.rs
```diff
@@ -40,6 +41,8 @@ pub struct Service<
     chain_events: CE,
     justifications_from_user: mpsc::UnboundedReceiver<J::Unverified>,
     additional_justifications_from_user: mpsc::UnboundedReceiver<J::Unverified>,
+    _block_requests_from_user:
```
timorl, line 44: Hm, could you perhaps read from this channel instead and ignore the results? That way if you end up implementing the submissions side first you won't get a memory leak at least.

This might require either keeping the other side of the channel alive or fusing this side (unless it's fused by default?).

timorl (approved):

finality-aleph/src/sync/service.rs
lesniak43: I'll just mark A0-2409 as blocked by A0-2408.

kostekIV (approved):

PR #1153, opened 2023-04-25
A0-2424: Restore dependency on primitives

pmikolajczyk41:
# Description

We restore dependency on `primitives` in `aleph-client`, rollbacking some of the changes from #1130.

I've added full export for `primitives` as well, so that all types and constants are visible further, but if someone is against star-export, I'm open to change it.

## Type of change

- Rollback

# Checklist:

### DIFF
--- a/aleph-client/Cargo.toml
+++ b/aleph-client/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aleph_client"
-version = "3.1.0"
+version = "3.1.1"
 edition = "2021"
 authors = ["Cardinal"]
 documentation = "https://docs.rs/aleph_client"
@@ -30,6 +30,8 @@ serde = { version = "1.0", features = ["derive"] }
 # that there are subtle differences between local building and crates.io-based building. Hence we stick to just
 # versioned dependency.
 pallet-contracts-primitives = { version = "7.0.0" }
+
+primitives = { path = "../primitives" }
 
 [dev-dependencies]
 tokio = "1.21"
--- a/aleph-client/Cargo.toml
+++ b/aleph-client/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aleph_client"
-version = "3.1.1"
+version = "3.2.0"
 edition = "2021"
 authors = ["Cardinal"]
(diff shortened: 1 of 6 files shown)


aleph-client/src/pallets/treasury.rs
```diff
@@ -140,18 +141,6 @@ impl<C: AsConnection + Sync> TreasureApiExt for C {
         let sessions_per_era = self.get_session_per_era().await?;
         let millisecs_per_era =
             MILLISECS_PER_BLOCK * session_period as u64 * sessions_per_era as u64;
-        Ok(era_payout(millisecs_per_era))
+        Ok(era_payout(millisecs_per_era).1)
```
pmikolajczyk41, line 144: https://github.com/Cardinal-Cryptography/aleph-node/commit/a79ab070aaadec8b4ebfbf99ecf5f2299d962808#diff-aaaa08990f63b79987b4aa9682a918cadabae528f7713a928b9656103e803490

aleph-client/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aleph_client"
-version = "3.1.0"
```
Marcin-Radecki, line 3: 3.2.0 please :pray:

Marcin-Radecki (approved):

aleph-client/Cargo.toml
pmikolajczyk41: bumped

aleph-client/src/lib.rs
```diff
@@ -43,25 +43,10 @@ pub mod utility;
 /// Waiting for some events API.
 pub mod waiting;
 
+pub use ::primitives::*;
```
timorl, line 46: I usually don't like star imports, but it might actually make sense here. What probbably doesn't though is the leading `::`, no?

timorl (approved):
Thanks!

aleph-client/src/lib.rs
pmikolajczyk41: Actually here rust requires leading ::, because it could be ambiguous...

PR #1157, opened 2023-04-27
A0-2397: Improve key rotation UX in cliain

timorl:
# Description

Using `cliain` to change the keys of a validator was annoying, because of the format of the result of a `rotate_keys` call. I made it simpler.

## Type of change

# Checklist:

### DIFF
--- a/bin/cliain/src/keys.rs
+++ b/bin/cliain/src/keys.rs
@@ -38,9 +38,14 @@ pub async fn set_keys(connection: SignedConnection, new_keys: String) {
 }
 
 pub async fn rotate_keys(connection: Connection) {
-    let new_keys = connection.author_rotate_keys().await;
-
-    info!("Rotated keys: {:?}", new_keys);
+    match connection.author_rotate_keys().await {
+        Ok(new_keys) => info!(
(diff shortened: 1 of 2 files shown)


bin/cliain/src/keys.rs
```diff
@@ -38,9 +38,10 @@ pub async fn set_keys(connection: SignedConnection, new_keys: String) {
 }
 
 pub async fn rotate_keys(connection: Connection) {
-    let new_keys = connection.author_rotate_keys().await;
-
-    info!("Rotated keys: {:?}", new_keys);
+    match connection.author_rotate_keys().await {
+        Ok(new_keys) => info!("Keys rotated, use the following in set_keys: {}{}", new_keys.aura.0.0.encode_hex::<String>(), new_keys.aleph.0.0.encode_hex::<String>()),
```
kostekIV, line 42: nitt: did you run cargo fmt?

timorl: agh, dammit, you are too fast, I thought I would be able to sneak the forcepush under you Dx

timorl: //it's correct now

Marcin-Radecki (approved):

PR #1158, opened 2023-04-27
A0-2303: e2e

kostekIV:
Tests usage of the emergency finalizer


e2e-tests/src/test/emergency_finalizer.rs
```diff
@@ -24,3 +34,117 @@ async fn set_emergency_finalizer_test() -> anyhow::Result<()> {
...
+async fn finalize_from_to<C: AsConnection + Sync>(
+    connection: &C,
+    from: BlockHash,
+    to: BlockHash,
+    finalizer: AlephKeyPair,
+) -> anyhow::Result<()> {
+    let from = connection.get_block_number(from).await?.unwrap();
+    let to = connection.get_block_number(to).await?.unwrap();
+
+    for i in (from + 1)..to {
```
deuszx, line 47: THis won't finalize `to` - is that intended?

e2e-tests/src/test/emergency_finalizer.rs
```diff
@@ -24,3 +34,117 @@ async fn set_emergency_finalizer_test() -> anyhow::Result<()> {
...
+
+    let last_finalized = connection.get_finalized_block_hash().await?;
+    let last_finalized_before = connection.get_block_number(last_finalized).await?.unwrap();
+    sleep(Duration::from_secs(40));
+    let mut last_finalized = connection.get_finalized_block_hash().await?;
+    let last_finalized_now = connection.get_block_number(last_finalized).await?.unwrap();
+
+    assert!(
+        last_finalized_now - last_finalized_before <= 20,
+        "at most 20 blocks can be created after finalization stops"
```
timorl, line 124: That's a bit of a _non sequitur_ here, since you check finalized block, which, uh, shouldn't be finalized at all after finalization stops, right? I'm not sure what the condition here should be, but at least the justification doesn't make sense, so is likely to be confusing in the future.

e2e-tests/src/validators.rs
```diff
@@ -182,7 +182,7 @@ pub async fn get_controller_connections_to_nodes(
 /// gets ws address to `n-th` node
 pub fn validator_address(index: u32) -> String {
     const BASE: &str = "ws://127.0.0.1";
-    const FIRST_PORT: u32 = 9944;
+    const FIRST_PORT: u32 = 9943;
```
timorl, line 185: Why change this?

timorl (approved):
Thanks for this.

e2e-tests/src/test/emergency_finalizer.rs
kostekIV: oh, I will change that last_finalized_now to us the latest best_block instead.

timorl: Huh, should that actually work? I thought disabling validators only happened after a session or two (since it's a key change), and here we immediately expect it to have worked? Or is there a hidden (or just missed by me :P) wait somewhere there?

e2e-tests/src/test/emergency_finalizer.rs
```diff
@@ -24,3 +34,117 @@ async fn set_emergency_finalizer_test() -> anyhow::Result<()> {
...
+            Some(seats),
+            TxStatus::InBlock,
+        )
+        .await?;
+    disable_validators(&[0, 1]).await?;
+
+    Ok(finalizer)
+}
+
+/// Tests emergency finalizer. Runs on 6 nodes 0-6.
```
lesniak43, line 101: 7?

lesniak43 (approved):

e2e-tests/src/test/emergency_finalizer.rs
lesniak43: oh, I get it - in that case, please write:
* Runs on 6 nodes 0-5
* setup 0-3
* setup 2-5
* disable 0-1 (this is ok)

e2e-tests/src/test/emergency_finalizer.rs
kostekIV: there is a 2 session wait in disable function

e2e-tests/src/test/emergency_finalizer.rs
kostekIV: good catch :P fixed

e2e-tests/src/test/emergency_finalizer.rs
kostekIV: fixed

PR #1161, opened 2023-04-28
A0-2386: Refactor metrics in aleph-node

ggawryal:
# Description

Pushed Option inside aleph-node metrics, so that `None` check is not needed when using `report_block`. 

## Type of change
- Refactor

# Checklist:
- I have added a test


finality-aleph/src/data_io/data_provider.rs
```diff
@@ -366,8 +364,13 @@ mod tests {
...
         };
 
-        let (chain_tracker, data_provider) =
-            ChainTracker::new(select_chain, client, session_boundaries, config, None);
+        let (chain_tracker, data_provider) = ChainTracker::new(
+            select_chain,
+            client,
+            session_boundaries,
+            config,
+            Metrics::register(None, |_| {}),
```
kostekIV, line 372: if you will follow my earlier suggestion then this would be simply Metrics::new(None).

finality-aleph/src/metrics.rs
```diff
@@ -71,11 +71,26 @@ pub(crate) enum Checkpoint {
...
 pub struct Metrics<H: Key> {
-    inner: Arc<Mutex<Inner<H>>>,
+    inner: Option<Arc<Mutex<Inner<H>>>>,
 }
 
 impl<H: Key> Metrics<H> {
-    pub fn register(registry: &Registry) -> Result<Self, PrometheusError> {
+    pub fn register<F: FnOnce(PrometheusError)>(
+        registry: Option<&Registry>,
+        on_registration_failed: F,
```
kostekIV, line 80: Rather that invoking closures on failure it is better to return Result

finality-aleph/src/metrics.rs
```diff
@@ -71,11 +71,26 @@ pub(crate) enum Checkpoint {
...
+    ) -> Metrics<H> {
```
kostekIV, line 81: Register fn with option looks strange to me. Just rename and change the signature of this fn to `new(registry: Option<_>) -> Result<_,_>` to get rid of the failure callback, its a more rust-like way of doing such thngs.

finality-aleph/src/metrics.rs
```diff
@@ -71,11 +71,26 @@ pub(crate) enum Checkpoint {
...
+        let inner = match registry {
+            Some(r) => Self::try_register(r).map_err(on_registration_failed).ok(),
+            None => {
+                on_registration_failed(PrometheusError::Msg("registry not available".to_string()));
```
kostekIV, line 85: here you could log that we run without metrics enabled (most likely warn level)

kostekIV (changes requested):
Great work :) Left some comments about failure closure and signature of register fn

finality-aleph/src/metrics.rs
ggawryal: Good idea with renaming to `new`. I agree that callbacks are ugly, but the thing is, ideally I'd like to return both `Error` and `Metrics` in case of a failure. I thought of replacing the signature to return `(Metrics<H>, Option<PrometheusError>)`, but finally I've decided to move error handling to that function instead.

finality-aleph/src/data_io/data_provider.rs
ggawryal: Yup, now it's a little less confusing.

finality-aleph/src/metrics.rs
ggawryal: Just to make sure - logging on unit tests does nothing serious, right? If so, I think we can move error handling entirely to that function.

finality-aleph/src/metrics.rs
ggawryal: Thanks, nice to know.

finality-aleph/src/metrics.rs
```diff
@@ -71,11 +71,25 @@ pub(crate) enum Checkpoint {
...
 
 impl<H: Key> Metrics<H> {
-    pub fn register(registry: &Registry) -> Result<Self, PrometheusError> {
+    pub fn new(registry: Option<&Registry>) -> Metrics<H> {
+        let inner = match registry {
+            Some(r) => Self::try_register(r)
+                .map_err(|err| warn!("Failed to register Prometheus metrics\n{:?}", err))
+                .ok(),
+            None => {
+                warn!("Failed to register Prometheus metrics: registry not available");
```
kostekIV, line 84: add target to those logs. `aleph-metrics` to be precise. While you are at that you could extract it to const `const LOG_TARGET: &str = "aleph-metrics";` and replace other usages of raw `target: "aleph-metrics"`in this file :D

finality-aleph/src/metrics.rs
kostekIV: well, unit test can show logs if invoked with flag `-- --no-capture` so imho its good to have them anyway

kostekIV (approved):

finality-aleph/src/metrics.rs
```diff
@@ -71,11 +71,25 @@ pub(crate) enum Checkpoint {
 
 #[derive(Clone)]
 pub struct Metrics<H: Key> {
-    inner: Arc<Mutex<Inner<H>>>,
+    inner: Option<Arc<Mutex<Inner<H>>>>,
 }
 
 impl<H: Key> Metrics<H> {
-    pub fn register(registry: &Registry) -> Result<Self, PrometheusError> {
+    pub fn new(registry: Option<&Registry>) -> Metrics<H> {
```
timorl, line 78: I don't think it's particularly great to have logging happening here. I saw the discussion about `Result`, but I still think returning a result here is better. To keep the log about running without metrics I would add a `is_noop` method to `Metrics` and invoke it in the one place where we care. At the same time the error could be logged and the result produced using the `noop` constructor I suggested above.

timorl (approved):
Cool, I left a couple of comments.

finality-aleph/src/data_io/data_provider.rs
```diff
@@ -374,8 +372,13 @@ mod tests {
...
         };
 
-        let (chain_tracker, data_provider) =
-            ChainTracker::new(select_chain, client, session_boundaries, config, None);
+        let (chain_tracker, data_provider) = ChainTracker::new(
+            select_chain,
+            client,
+            session_boundaries,
+            config,
+            Metrics::new(None),
```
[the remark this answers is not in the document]
ggawryal: Let's do it (but this is just a unit test :))

finality-aleph/src/metrics.rs
ggawryal: Done, but now we are using `warn!` in a `service.rs` file, so we either need to make `LOG_TARGET` separately defined in `service.rs`, or make it crate-public in `metrics.rs` file.

aggregator/src/aggregator.rs
```diff
@@ -32,12 +32,12 @@ pub struct BlockSignatureAggregator<H: Hash + Copy, PMS, M: Metrics<H>> {
     signatures: HashMap<H, PMS>,
     hash_queue: VecDeque<H>,
     started_hashes: HashSet<H>,
-    metrics: Option<M>,
+    metrics: M,
```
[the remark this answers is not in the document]
timorl: Versioning in `finality-aleph` isn't particularly relevant as we never depend on it in another way than by path, so I don't care too much. You can bump the patch version if you want, but I wouldn't bother.

bin/node/src/service.rs
```diff
@@ -142,19 +140,20 @@ pub fn new_partial(
...
+    let metrics = match config.prometheus_registry() {
+        Some(register) => match Metrics::new(register) {
+            Ok(metrics) => metrics,
+            Err(e) => {
+                warn!("Failed to register Prometheus metrics:{:?}", e);
+                Metrics::noop()
+            }
+        },
+        None => {
+            info!("Running without the metrics not available");
```
timorl, line 152: Nit: this sentence doesn't quite parse. :P

timorl (approved):

finality-aleph/src/metrics.rs
kostekIV: in service.rs it should be some other target so lets not worry about it here. Gj

bin/node/src/service.rs
ggawryal: Thanks, updated.

PR #1102, opened 2023-03-29
A0-2240: refactor aleph node rpc

maciejnems:
# Description

Aleph Node Rpc does not need to be dependent on `B: Block` and also some other stuff. It is based on https://github.com/Cardinal-Cryptography/aleph-node/pull/1101. Only commit that matters is `refactor aleph_node_rpc`.

## Type of change

- Refactor

PR #1162, opened 2023-04-28
A0-2240: Refactor aleph node RPC

timorl:
# Description

Aleph Node Rpc does not need to be dependent on `B: Block`. This replaces and is heavily based on #1102, which got sufficiently obsoleted by other changes that it was simpler to write it from scratch.

## Type of change

Refactor

### DIFF
--- a/bin/node/src/service.rs
+++ b/bin/node/src/service.rs
@@ -31,8 +31,10 @@ use sp_runtime::{
 };
 
 use crate::{
-    aleph_cli::AlephCli, chain_spec::DEFAULT_BACKUP_FOLDER, executor::AlephExecutor,
-    rpc::FullDeps as RpcFullDeps,
+    aleph_cli::AlephCli,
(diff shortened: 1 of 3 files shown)


bin/node/src/service.rs
```diff
@@ -264,15 +266,15 @@ fn setup(
         let client = client.clone();
         let pool = transaction_pool.clone();
         Box::new(move |deny_unsafe, _| {
-            let deps: RpcFullDeps<Block, _, _, _> = crate::rpc::FullDeps {
```
kostekIV, line 267: lul nice one

lesniak43:
Nice. I'd consider using `opaque::Header` instead of `<Block as BlockT>::Header>` (or, even better, import `Header` from `primitives`), but the decision is yours.

lesniak43 (approved):

Yithis (approved):

PR #1164, opened 2023-04-28
A0-0000: in_given_out

fbielejec:
# Description

Quick request for frontend purposes


contracts/simple_dex/lib.rs
```diff
@@ -477,10 +477,39 @@ mod simple_dex {
 
             Ok(())
         }
+        /// Returns the swap trade input given a desired amount and assuming a curve with equal token weights
+        ///
+        /// A_in = B_i * ((B_o / (B_o - A_o)) - 1)
```
obrok, line 482: Question: This one ignores the swap fee - is it by design?
Suggestion: Is there any guarantee that if `out_given_in(x) = y`, then `in_given_out(y) = x`? I suspect there is going to be some inconsistency there, due to rounding. If the mismatch between the two is expected to be small, you could add a property test basically saying that if `out_given_in(x) = y`, then `in_given_out(y) = x +/- 1` or something.

obrok (approved):

contracts/simple_dex/lib.rs
fbielejec: I think we will simply remove the fee altogether - since it will never be non-zero.

+1 on the test - I'm adding for a handful of PRs built on top of this branch to be merged and will most def add some more coverage

h4nsu (approved):

contracts/button/lib.rs
```diff
@@ -299,16 +327,33 @@ pub mod button_game {
...
+            Ok(())
+        }
+
+        /// Rewards the Pressiah
+        ///
+        /// Does not reset any other state beyond the last_presser record
+        /// Can only be called after button's deadline
+        /// Can be called by any account
+        #[ink(message)]
+        pub fn reward_pressiah(&mut self) -> ButtonResult<()> {
```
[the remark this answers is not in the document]
fbielejec: to pay out last reward but not reset the game any more

contracts/simple_dex/lib.rs
fbielejec: Done, please take another look

obrok (approved):

contracts/simple_dex/lib.rs
```diff
@@ -606,20 +582,37 @@ mod simple_dex {
...
+
+            // 939587570196u128;
+            let amount_out =
+                SimpleDex::_out_given_in(expected_amount_in, balance_in, balance_out).unwrap();
+
+            let amount_in = SimpleDex::_in_given_out(amount_out, balance_in, balance_out).unwrap();
+
+            assert_eq!(amount_in, expected_amount_in - dust);
+        }
+
```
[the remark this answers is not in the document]
fbielejec: Done, for in_given_out proptests I had to increase the diff though - to be within 10 pico-units.
Not sure if a cause for concern, probably not.

PR #1167, opened 2023-05-03
A0-2455: refactor events

fbielejec:
# Description

- Adding more events in preparation for a richer indexer API
- Cleaned up events that will not be indexed
- Removed fields that should be on the overarching event obj anyway

### DIFF
--- a/contracts/button/lib.rs
+++ b/contracts/button/lib.rs
@@ -52,7 +52,7 @@ pub mod button_game {
     #[ink(event)]
     #[derive(Debug)]
     pub struct RewardMinted {
-        round: u64,
+        round: Round,
         #[ink(topic)]
         reward_token: AccountId,
         #[ink(topic)]
@@ -117,7 +117,7 @@ pub mod button_game {
         /// scoring strategy
         pub scoring: Scoring,
         /// current round number
-        pub round: u64,
+        pub round: Round,
         /// minimal reward denominated in reward tokens that can be minted to a player for single press
         pub min_reward: Balance,
         /// maximal reward denominated in reward tokens that can be minted to a player for single press
--- a/contracts/button/lib.rs
+++ b/contracts/button/lib.rs
@@ -26,7 +26,7 @@ pub mod button_game {
         traits::Storage,
     };
     use scale::{Decode, Encode};
-    use shared_traits::{Haltable, HaltableData, HaltableError, Internal, Selector};
+    use shared_traits::{Haltable, HaltableData, HaltableError, Internal, Round, Selector};
 
     use crate::errors::GameError;
 
@@ -42,7 +42,7 @@ pub mod button_game {
     pub struct ButtonPressed {
         #[ink(topic)]
         by: AccountId,
-        round: u64,
+        round: Round,
         score: Balance,
     }
 
@@ -73,9 +73,7 @@ pub mod button_game {
     /// Event emitted when the finished game is reset and pressiah is rewarded
     #[ink(event)]
     #[derive(Debug)]
-    pub struct ButtonReset {
-        by: AccountId,
-    }
+    pub struct ButtonReset {}
 
     #[ink(event)]
     pub struct Halted;
@@ -311,7 +309,6 @@ pub mod button_game {
                 self.env(),
                 Event::ButtonPressed(ButtonPressed {
                     by: caller,
-                    // when: now,
                     score,
                     round: data.round,
                 }),
@@ -333,12 +330,7 @@ pub mod button_game {
             self.reset_marketplace()?;
             self.reset_state()?;
 
-            Self::emit_event(
-                self.env(),
-                Event::ButtonReset(ButtonReset {
-                    by: self.env().caller(),
-                }),
-            );
+            Self::emit_event(self.env(), Event::ButtonReset(ButtonReset {}));
 
             Ok(())
         }
@@ -499,11 +491,12 @@ pub mod button_game {
         fn reward_pressiah(&self) -> ButtonResult<()> {
             if let Some(pressiah) = self.data.get().unwrap().last_presser {
                 let reward = self.pressiah_reward();
+                self.mint_reward(pressiah, reward)?;
+
                 Self::emit_event(
                     self.env(),
                     Event::PressiahFound(PressiahFound { pressiah, reward }),
                 );
-                self.mint_reward(pressiah, reward)?;
             };
 
             Ok(())
--- a/contracts/button/lib.rs
+++ b/contracts/button/lib.rs
@@ -573,7 +573,7 @@ pub mod button_game {
         }
 
         fn pressiah_reward(&self) -> Balance {
-            (self.data.get().unwrap().total_rewards / 4) as Balance
+            (self.data.get().unwrap().total_rewards / 2) as Balance
         }
 
         fn transfer_ticket(
--- a/contracts/button/lib.rs
+++ b/contracts/button/lib.rs
@@ -30,9 +30,6 @@ pub mod button_game {
 
     use crate::errors::GameError;
 
-    pub const ONE_TOKEN: Balance = 1_000_000_000_000;
-    pub const ONE_HUNDRED_TOKENS: Balance = 100_000_000_000_000;
-
     /// Result type
     type ButtonResult<T> = core::result::Result<T, GameError>;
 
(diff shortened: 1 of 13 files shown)


contracts/button/lib.rs
```diff
@@ -36,25 +36,13 @@ pub mod button_game {
...
-    }
-
     /// Event emitted when TheButton is pressed
     #[ink(event)]
     #[derive(Debug)]
     pub struct ButtonPressed {
         #[ink(topic)]
         by: AccountId,
-        when: BlockNumber,
+        round: u64,
```
kostekIV, line 45: nit: add typealias for round

contracts/button/lib.rs
```diff
@@ -312,8 +311,9 @@ pub mod button_game {
                 self.env(),
                 Event::ButtonPressed(ButtonPressed {
                     by: caller,
-                    when: now,
+                    // when: now,
```
kostekIV, line 314: remove this

contracts/button/lib.rs
```diff
@@ -497,13 +493,16 @@ pub mod button_game {
...
             self.data.set(&data);
 
-            Self::emit_event(self.env(), Event::GameReset(GameReset { when: now }));
             Ok(())
         }
 
         fn reward_pressiah(&self) -> ButtonResult<()> {
             if let Some(pressiah) = self.data.get().unwrap().last_presser {
                 let reward = self.pressiah_reward();
+                Self::emit_event(
```
kostekIV, line 502: hmmm, I am not sure how it should be exactly but `mint_reward` which is called later can fail. In such case even though the pressiah was found the reward will not be minted, right? Maybe emit this event if `mint_reward` returns ok variant

contracts/simple_dex/lib.rs
```diff
@@ -102,23 +102,31 @@ mod simple_dex {
     }
 
     #[ink(event)]
-    pub struct SwapPairAdded {
+    pub struct Withdrawn {
+        caller: AccountId,
```
kostekIV, line 106: maybe caller also should be marked as topic?

contracts/button/lib.rs
```diff
@@ -64,18 +52,29 @@ pub mod button_game {
...
+        reward: Balance,
     }
 
     /// Event emitted when the finished game is reset and pressiah is rewarded
     #[ink(event)]
     #[derive(Debug)]
-    pub struct GameReset {
-        when: BlockNumber,
+    pub struct ButtonReset {
+        by: AccountId,
```
krzysztofziobro, line 77: From what I understand contract should behave in the same way no matter who resets the button. Is it worth it to keep the caller in this event, or maybe we should just have an empty struct here?

contracts/button/lib.rs
fbielejec: outdated comment? Anyway - done

contracts/button/lib.rs
fbielejec: txs even if chained are atomic - if anything fails, the whole state is reverted, so it will never happen

contracts/simple_dex/lib.rs
fbielejec: What I'm thinking rn is that this whole tx (deposit) makes 0 sense - anyone can circumvent it by just transferring PSP22 to the dex address, in fact this is how we'll be adding liquidity.

This is a leftover form when the DEX was more complex and included tracking LP, but it makes no sense in this simplified design. 

 I'm removing it, along with the event

contracts/button/lib.rs
fbielejec: agreed, makes no sense - removing

contracts/wrapped_azero/lib.rs
```diff
@@ -111,7 +111,7 @@ pub mod wrapped_azero {
 
     #[ink(event)]
     #[derive(Debug)]
-    pub struct UnWrapped {
+    pub struct Unwrapped {
         caller: AccountId,
```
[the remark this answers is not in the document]
fbielejec: done, added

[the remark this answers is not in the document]
krzysztofziobro: Hmm, I don't see the change, have you pushed it?

contracts/button/lib.rs
kostekIV: thx for explanantion

contracts/wrapped_azero/lib.rs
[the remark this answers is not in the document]
fbielejec: thought I did, apparently not

krzysztofziobro (approved):

contracts/button/lib.rs
```diff
@@ -64,20 +52,29 @@ pub mod button_game {
     #[ink(event)]
     #[derive(Debug)]
     pub struct RewardMinted {
-        when: BlockNumber,
+        round: u64,
```
pmikolajczyk41, line 55: `round: Round`

contracts/marketplace/lib.rs
```diff
@@ -268,20 +274,16 @@ pub mod marketplace {
 
         /// Update the length of the auction.
         ///
-        /// Can only be called by an account with an Admin role
-        /// and only if the contract is currently halted
+        /// Requires `Role::Admin`.
         #[ink(message)]
         pub fn set_auction_length(&mut self, new_auction_length: BlockNumber) -> Result<(), Error> {
-            if self.is_halted() {
```
pmikolajczyk41, line 275: why do we allow for changing auction length without halting?

contracts/button/lib.rs
```diff
@@ -329,9 +326,13 @@ pub mod button_game {
         pub fn reset(&mut self) -> ButtonResult<()> {
             self.ensure_dead()?;
             self.reward_pressiah()?;
-            self.reset_state()?;
             self.transfer_tickets_to_marketplace()?;
-            self.reset_marketplace()
+            self.reset_marketplace()?;
+            self.reset_state()?;
```
pmikolajczyk41, line 331: does the order matter?

contracts/button/lib.rs
```diff
@@ -36,25 +36,13 @@ pub mod button_game {
...
-    }
-
     /// Event emitted when TheButton is pressed
     #[ink(event)]
     #[derive(Debug)]
     pub struct ButtonPressed {
         #[ink(topic)]
         by: AccountId,
-        when: BlockNumber,
+        round: Round,
```
pmikolajczyk41, line 45: `Round` is less granular than `BlockNumber`, right? why is it better then?

contracts/button/lib.rs
pmikolajczyk41: the same in 120th line (`Data` struct)

contracts/simple_dex/lib.rs
```diff
@@ -268,46 +262,6 @@ mod simple_dex {
             Ok(())
         }
 
-        /// Liquidity deposit
-        ///
-        /// Can only be performed by an account with a LiquidityProvider role
-        /// Caller needs to give at least the passed amount of allowance to the contract to spend the deposited tokens on his behalf
-        /// prior to executing this tx
-        #[ink(message)]
-        pub fn deposit(&mut self, deposits: Vec<(AccountId, Balance)>) -> Result<(), DexError> {
```
pmikolajczyk41, line 277: why do we remove depositing?

pmikolajczyk41 (approved):
lgtm, I just left a couple of comments, because I'm a bit out of context

contracts/marketplace/lib.rs
fbielejec: both halting and the setter are guarded by the same Role:Admin, so I take it's not necessary, the same account will perform both actions if ever, and as it will likely be under an emergency scenario it's implicit the contracts will be halted

contracts/button/lib.rs
fbielejec: dependes, in this particular case not

contracts/simple_dex/lib.rs
fbielejec: this is an artifact from when we still floated the idea of having a full-featured DEX with liquidity provider, but now this will never be used - only A0 liquidity will ever be added and this can be done with a simple PSP22 transfer to the contracts account

contracts/button/lib.rs
fbielejec: Block number is already available to the indexer, it's attached to every event. With a round we can e.g. create an endpoint that returns players scores in a given round.

contracts/button/lib.rs
fbielejec: done

PR #1170, opened 2023-05-09
A0-2478: Replace bash commands in featurenet workflows with a docker run

mikogs:


.github/workflows/updatenet-create.yml
```diff
@@ -22,7 +22,7 @@ jobs:
   create-updatenet:
     needs: [check-vars-and-secrets]
     name: Create updatenet
-    runs-on: ubuntu-20.04
+    runs-on: self-hosted
```
Marcin-Radecki, line 25: Can we use some tiny machine instead of big fish we use for compilation?

Marcin-Radecki (approved):

PR #1181, opened 2023-05-15
A0 2505 note

mike1729:
# Description

## Type of change

# Checklist:

- I have added tests


relations/jf/src/lib.rs
```diff
@@ -26,18 +27,16 @@ pub fn generate_srs<R: CryptoRng + RngCore>(
...
+pub trait Relation {
+    /// Include this relation in the circuit and return all variables that are needed for outside circuit.
+    fn generate_subcircuit(
+        &self,
+        circuit: &mut PlonkCircuit<CircuitField>,
+    ) -> PlonkResult<Vec<Variable>>;
+}
 
+/// Common API
+pub trait ProofSystem: Relation + Default {
```
deuszx, line 39: This sets of my spidey-senses: it feels like there shouldn't exist a default proof system or exist _some_ _single_ `Relation` instance for this to type-check. 

Instead of these trait bounds we could pass the `Relation` as an argument to methods where it's required ( like `generate_keys`).

relations/jf/src/note.rs
```diff
@@ -0,0 +1,143 @@
...
+    Deposit,
+    Spend,
+}
+
+#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
+pub struct NoteRelation {
+    pub note: Note,
+    pub nullifier: Nullifier,
+    pub token_id_var: Variable,
+    pub token_amount: TokenAmount,
```
pmikolajczyk41, line 21: personally I would avoid mixing `Variable` with frontend types - I think it really doesn't hurt to double-create token id variable: especially that we'd gain consistency

relations/jf/src/deposit.rs
```diff
@@ -1,10 +1,9 @@
...
+use jf_relation::{Circuit, PlonkCircuit, Variable};
 
 use crate::{
+    note::{NoteRelation, NoteType},
     shielder_types::{convert_hash, Note, Nullifier, TokenAmount, TokenId, Trapdoor},
-    CircuitField, Marshall, PlonkResult, Relation,
+    CircuitField, PlonkResult, ProofSystem, PublicInput, Relation,
 };
 
 #[derive(Copy, Clone, Eq, PartialEq, Hash, Default, Debug)]
```
pmikolajczyk41, line 9: actually, we need just a single field: `NoteRelation<Deposit>`

relations/jf/src/lib.rs
```diff
@@ -26,18 +27,16 @@ pub fn generate_srs<R: CryptoRng + RngCore>(
...
-    fn new(public_input: Self::PublicInput, private_input: Self::PrivateInput) -> Self;
-
-    /// Include this relation in the circuit.
-    fn generate_subcircuit(&self, circuit: &mut PlonkCircuit<CircuitField>) -> PlonkResult<()>;
+pub trait Relation {
+    /// Include this relation in the circuit and return all variables that are needed for outside circuit.
+    fn generate_subcircuit(
+        &self,
+        circuit: &mut PlonkCircuit<CircuitField>,
+    ) -> PlonkResult<Vec<Variable>>;
```
pmikolajczyk41, line 35: I guess that this will be a pain eventually - it's unstructured, var-length sequence of plain `usize`, which might bring some stupid errors; also, it's hard to tell which variables are needed for outside circuit in general;

I think that the ideal thing would be to return a mirror type with all fields of `Variable` type

relations/jf/src/note.rs
pmikolajczyk41: an alternative would be to allow for both options, i.e. to register either a `Variable` or `TokenId`; in the old code we had a general, type-safe builder-pattern: https://github.com/Cardinal-Cryptography/aleph-node/blob/main/relations/src/shielder/note_var.rs, but this is an overkill I think...

relations/jf/src/lib.rs
mike1729: I split in two since the methods from the second trait are not needed for NoteRelation.

pmikolajczyk41: but they come with default implementation, so no effort needed

relations/jf/src/lib.rs
mike1729: This return type here is needed e.g. in withdraw. I don't see a clear API how to combine subcircuits that are not self contained in clear API.

relations/jf/src/lib.rs
mike1729: There is mental effort and poor rustc has to generate that :P I don't insist, which version you prefer @deuszx? Going back to one trait or two traits and the second in version you proposed.

relations/jf/src/note.rs
mike1729: changed

relations/jf/src/lib.rs
mike1729: outdated

relations/jf/src/note.rs
mike1729: outdated

relations/jf/src/deposit.rs
mike1729: changed

relations/jf/src/lib.rs
mike1729: outdated

relations/jf/src/lib.rs
```diff
@@ -26,18 +27,12 @@ pub fn generate_srs<R: CryptoRng + RngCore>(
...
-    fn new(public_input: Self::PublicInput, private_input: Self::PrivateInput) -> Self;
-
+pub trait Relation: Default  {
     /// Include this relation in the circuit.
-    fn generate_subcircuit(&self, circuit: &mut PlonkCircuit<CircuitField>) -> PlonkResult<()>;
-
+    fn generate_subcircuit(
+        &self,
+        circuit: &mut PlonkCircuit<CircuitField>,
+    ) -> PlonkResult<()>;
```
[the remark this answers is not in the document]
mike1729: definitely, forgot to run it.

relations/jf/src/deposit.rs
```diff
@@ -0,0 +1,139 @@
...
+                token_amount: public.token_amount,
+                trapdoor: private.trapdoor,
+                nullifier: private.nullifier,
+                note_type: NoteType::Deposit,
+            },
+        }
+    }
+}
+
+#[derive(Default)]
```
deuszx, line 34: I find it weird that there exists a `Default` instance of `DepositPublicInput`. My understanding of the `Default` concept is that it exists for things that it's correct to supply that default instance, that it makes sense in some contexts. I can't imagine a case where there's a default public input to a relation, even more if that input is token ID, amount, note etc. A default instance of `Note` is probably invalid in any context, other than for generating keys.

relations/jf/src/deposit.rs
```diff
@@ -1,85 +1,67 @@
...
-    type PrivateInput = DepositPrivateInput;
-
-    fn new(public_input: Self::PublicInput, private_input: Self::PrivateInput) -> Self {
-        Self {
-            public: public_input,
-            private: private_input,
-        }
-    }
-
     fn generate_subcircuit(&self, circuit: &mut PlonkCircuit<CircuitField>) -> PlonkResult<()> {
```
deuszx, line 54: Can we move the ownership of the `circuit` to the method, mutate it internally and return as a result? I think it's cleaner and less bug-prone.

relations/jf/src/note.rs
```diff
@@ -0,0 +1,167 @@
...
+    pub token_id_var: Variable,
+    pub token_amount_var: Variable,
+    pub trapdoor_var: Variable,
+    pub nullifier_var: Variable,
+}
+
+pub trait NoteGadget {
+    fn create_note_variable(&mut self, note: &SourcedNote) -> PlonkResult<SourcedNoteVar>;
+    fn enforce_note_preimage(&mut self, note_var: SourcedNoteVar) -> PlonkResult<Variable>;
+}
```
deuszx, line 35: These will eventually be public traits and structs so will need documentation.

relations/jf/src/note.rs
```diff
@@ -0,0 +1,167 @@
...
+    shielder_types::{convert_hash, Note, Nullifier, TokenAmount, TokenId, Trapdoor},
+    CircuitField, PlonkResult, PublicInput,
+};
+
+pub enum NoteType {
+    Deposit,
+    Spend,
+}
+
+pub struct SourcedNote {
```
deuszx, line 15: What does _sourced_ mean?

relations/jf/src/note.rs
```diff
@@ -0,0 +1,167 @@
...
+        let zero_var = self.create_constant_variable(CircuitField::zero())?;
+
+        // Check that the note is valid.
+        let inputs: [usize; 6] = [
+            token_id_var,
+            token_amount_var,
+            trapdoor_var,
+            nullifier_var,
+            zero_var,
+            zero_var,
```
deuszx, line 87: Why do we need to add these two `zero_var`s ?

pmikolajczyk41: jellyfish's rescue hashing function takes 3n arguments... it's messed up and we will provide some wrappers soon

pmikolajczyk41: b/c similar problems will appear with Merkle trees

relations/jf/src/note.rs
pmikolajczyk41: my naming, sorry - "note" is just a hash; "sourced note" is a hash with its source, i.e. preimage

relations/jf/src/note.rs
deuszx: ok, thanks.

relations/jf/src/note.rs
pmikolajczyk41: 100% agree, but I wouldn't spend too much time now for docs, while the API is changing so rapidly - for now I would leave things to be clear at least for people working with the code and reviewers

relations/jf/src/deposit.rs
pmikolajczyk41: `&mut` is more jellyfish-consistent; and also this will be a problem for generating gadgets that return some variable bindings - we would have to return additionally circuit there

deuszx (approved):
Can't contribute w.r.t. to checking the correctness of the circuit/relation logic and have commented on the code instead. Apart from the outstanding issues, which either won't be changed or will be changed later, I have nothing more to add.

relations/jf/src/note.rs
```diff
@@ -0,0 +1,167 @@
...
+
+        let computed_note_var = RescueNativeGadget::<CircuitField>::rescue_sponge_no_padding(
+            self,
+            inputs.as_slice(),
+            1,
+        )?[0];
+
+        self.enforce_equal(note_var, computed_note_var)?;
+
+        Ok(token_amount_var)
```
pmikolajczyk41, line 98: we don't have to return this, because the caller already has `note_var: SourcedNoteVar`

relations/jf/src/deposit.rs
```diff
@@ -1,85 +1,67 @@
...
-            zero_var,
-        ];
-        let computed_note_var = RescueNativeGadget::<CircuitField>::rescue_sponge_no_padding(
-            circuit,
-            inputs.as_slice(),
-            1,
-        )?[0];
-
-        circuit.enforce_equal(note_var, computed_note_var)?;
+        let note = SourcedNote {
```
pmikolajczyk41, line 55: not needed, right? we can use directly `self.deposit_note`

pmikolajczyk41 (approved):

relations/jf/src/note.rs
mike1729: right, changed.

relations/jf/src/deposit.rs
mike1729: oh yes, this is from previous version, forgot to change that.

relations/jf/src/note.rs
deuszx: They don't b/c it's passed into the method with ownership.

pmikolajczyk41: it's `Copy`

deuszx: It _can_ be? https://github.com/Cardinal-Cryptography/aleph-node/pull/1181/files#diff-843190817ebe94d180908160036f30f6275895c0ddd5dd3a315ef8cf090c7ed9R24

pmikolajczyk41: apologies; should be :) @mike1729

mike1729: Variable is copy, I'm just writing withdraw rel and it's fine :)

PR #1182, opened 2023-05-15
 A0-3461: error variant to handle liquidity issues

fbielejec:
# Description

Added optional block info to the events.

Fixed e2e test suite

`in_given_out` is often used by traders and our DEX frontend to query the field inputs, here we make it return a semantic error variant early if the trade will not be supported anyway.

## Type of change

# Checklist:

- I have added tests


jalooc:
Wouldn't it make sense to add it to `out_given_in` as well?

fbielejec:
> Wouldn't it make sense to add it to `out_given_in` as well?

it won't throw during calculations, formula has addition in the divisor.

jalooc (approved):

mike1729 (approved):
Just two nits.

PR #1191, opened 2023-05-17
A0-2431: Add rolling update partition input to updatenets

mikogs:


.github/actions/create-featurenet/action.yml
```diff
@@ -129,7 +141,10 @@ runs:
         pr_image="${{ inputs.ecr-public-registry }}feature-env-aleph-node:${pr_image_tag}"
         ./Ops.sh create-featurenet \
           "${{ env.APP_NAME }}" \
-          "${pr_image}"
+          "${pr_image}" \
+          "false" \
+          "false" \
+          "${{ inputs.update == 'true' && inputs.rolling-update-partition || '0' }}"
```
Marcin-Radecki, line 147: We have in total three added args, but Ops.sh will have in total added 2?

.github/actions/create-featurenet/action.yml
```diff
@@ -47,6 +47,10 @@ inputs:
     description: "Set to 'true' if it is just featurenet update"
     required: false
     default: 'false'
+  rolling-update-partition:
+    description: "Partition value when doing rolling update"
```
Marcin-Radecki, line 51: Can we use this sentence from https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#partitions
> all Pods with an ordinal that is greater than or equal to the partition will be updated when the StatefulSet's .spec.template is updated. All Pods with an ordinal that is less than the partition will not be updated, and, even if they are deleted, they will be recreated at the previous version

rephrased e.g. as 

> all `aleph-node-validator-N` with an ordinal `N` that is greater than or equal to the partition will be updated. Default value: 0.

Also, this information needs to be put in Notion.

.github/actions/create-featurenet/action.yml
```diff
@@ -82,6 +86,14 @@ runs:
           echo "!!! Invalid feature net node image tag"
           exit 1
         fi
+        if [[
+          "${{ inputs.update }}" == "true" && \
+          ! "${{ inputs.rolling-update-partition }}" =~ ^[0-9]{1,2}$
```
Marcin-Radecki, line 91: Technically we have only ten validators on Devnet, so range 0-9 is sufficient

.github/workflows/_create-featurenet.yml
```diff
@@ -37,7 +37,7 @@ jobs:
   create-featurenet:
     needs: [check-vars-and-secrets]
     name: Create featurenet based on the PR
-    runs-on: ubuntu-20.04
+    runs-on: self-hosted
```
Marcin-Radecki, line 40: Can we create a tiny machine that runs these jobs?

Marcin-Radecki (approved):

.github/actions/create-featurenet/action.yml
mikogs: That is not going to cause any problems.  There is an another PR in another repository to update that Ops.sh script where partition is actually last argument.  In that future version of Ops.sh the `create-featurenet` command goes as follows:

    create-featurenet <feature-net-name> <aleph-node-image|'testnet'|'mainnet'> [create-hook='false'] [generate-data-in-s3='false'] [rolling-update-partition='0']"

.github/actions/create-featurenet/action.yml
mikogs: This is now updated and the description is as you suggested

mikogs: I will ensure that Notion is updated once all this is merged

.github/actions/create-featurenet/action.yml
mikogs: It's now changed to ^[0-9]$

Marcin-Radecki (approved):
Rework looks good

PR #1194, opened 2023-05-17
A0 2499 jf-withdraw

mike1729:
Implements withdraw relation in jellyfish.


relations/jf/src/withdraw.rs
```diff
@@ -0,0 +1,288 @@
...
+    let index_var = circuit.create_variable(leaf_index.into())?;
+    let proof_var = MerkleTreeGadget::create_membership_proof_variable(circuit, merkle_proof)?;
+    let root_var = MerkleTreeGadget::create_root_variable(circuit, convert_array(merkle_root))?;
+    circuit.set_variable_public(root_var)?;
+
+    MerkleTreeGadget::enforce_membership_proof(circuit, index_var, proof_var, root_var)
+        .map_err(Into::into)
+}
+
+#[cfg(test)]
```
pmikolajczyk41, line 178: I would also like to have a test for recipient/fee inclusion

relations/jf/src/withdraw.rs
```diff
@@ -0,0 +1,288 @@
...
+// TODO refactor when implementing DepositAndMerge
+type MerkleTree = RescueSparseMerkleTree<BigUint, CircuitField>;
+type MerkleTreeGadget = dyn jf_primitives::circuit::merkle_tree::MerkleTreeGadget<
+    MerkleTree,
+    MembershipProofVar = Merkle3AryMembershipProofVar,
+    DigestGadget = RescueDigestGadget,
+>;
+type MerkleProof = <MerkleTree as MerkleTreeScheme>::MembershipProof;
+
+fn check_merkle_proof(
```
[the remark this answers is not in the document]
mike1729: Unfortunately, there already is a gadget named `MerkleTreeGadget`, so I cannot do that.

relations/jf/src/withdraw.rs
```diff
@@ -0,0 +1,288 @@
...
+    pub whole_token_amount: TokenAmount,
+    pub deposit_token_amount: TokenAmount,
+}
+
+impl Default for WithdrawPrivateInput {
+    fn default() -> Self {
+        let uid = BigUint::from(0u64);
+        let elem = CircuitField::from(0u64);
+        let mt = RescueSparseMerkleTree::from_kv_set(TREE_HEIGTH, &[(uid.clone(), elem)]).unwrap();
+        let (_, merkle_proof) = mt.lookup(&uid).expect_ok().unwrap();
```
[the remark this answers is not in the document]
mike1729: Yeah, I tried but I got an error: 
`error[E0275]: overflow evaluating the requirement Fp<MontBackend<FrConfig, 4>, 4>: Clone`

relations/jf/src/withdraw.rs
```diff
@@ -0,0 +1,288 @@
...
+use crate::{
+    note::{NoteGadget, NoteType, SourcedNote},
+    shielder_types::{
+        convert_account, convert_array, Account, LeafIndex, MerkleRoot, Note, Nullifier,
+        TokenAmount, TokenId, Trapdoor,
+    },
+    CircuitField, PlonkResult, PublicInput, Relation,
+};
+
+pub struct WithdrawRelation {
```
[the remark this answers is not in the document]
mike1729: why?

relations/jf/src/withdraw.rs
```diff
@@ -0,0 +1,288 @@
...
+        public_input
+    }
+}
+
+impl Relation for WithdrawRelation {
+    fn generate_subcircuit(&self, circuit: &mut PlonkCircuit<CircuitField>) -> PlonkResult<()> {
+        let _fee_var = circuit.create_public_variable(self.fee.into())?;
+        let _recipient_var = circuit.create_public_variable(convert_account(self.recipient))?;
+        let token_amount_out_var = circuit.create_public_variable(self.token_amount_out.into())?;
+        circuit.enforce_leq_constant(token_amount_out_var, CircuitField::from(u128::MAX))?;
```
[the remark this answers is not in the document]
mike1729: Added.

relations/jf/src/withdraw.rs
mike1729: Added

relations/jf/src/note.rs
```diff
@@ -102,16 +107,19 @@ impl NoteGadget for PlonkCircuit<CircuitField> {
...
                     self.token_id.into(),
                     self.token_amount.into(),
                 ]
             }
+            NoteType::Spend => {
+                vec![convert_array(self.nullifier)]
+            }
+            NoteType::Redeposit => {
+                vec![convert_array(self.note), self.token_id.into()]
+            }
```
deuszx, line 122: Alternative design would be to split `SourceNote` into 3 separate structs: `Deposit`, `Spend`, `Redeposit` with an associated type `PublicInput` and get the typesafety wrt to different sets of circuit field elements.

relations/jf/src/withdraw.rs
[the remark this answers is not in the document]
deuszx: https://www.notion.so/cardinalcryptography/Basic-traits-derivation-and-field-access-for-public-structs-0aa22c4887024547b613db5754e0664a ?

relations/jf/src/withdraw.rs
```diff
@@ -0,0 +1,326 @@
...
+}
+
+impl Default for WithdrawRelation {
+    fn default() -> Self {
+        Self::new(Default::default(), Default::default())
+    }
+}
+
+#[derive(Default)]
+pub struct WithdrawPublicInput {
```
pmikolajczyk41, line 37: up to you, but I'm not so sure that this is the easier way than creating directly `WithdrawRelation` (we already kinda give up this `Public/PrivateInput` in the previous PR

however, if you prefer to stay with this, we must add another constraint to the circuit, i.e. that the `token_id`s are equal in both notes - notice, that adversary doesn't have to use `WithdrawRelation` to create a malformed (valid!) relation witness

pmikolajczyk41 (changes requested):

relations/jf/src/withdraw.rs
mike1729: Let me leave that as it is, and I'll try to figure out some other API during/after implementing Merge relation. Added the check.

relations/jf/src/withdraw.rs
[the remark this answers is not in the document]
mike1729: Awful.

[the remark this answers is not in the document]
mike1729: Added :(

relations/jf/src/note.rs
deuszx: Here, the return type is opaque to the caller who has no idea what it us, why it is etc. If there was a single type representing the public inputs for that specific relation then there's no confusion - i.e. you can't supply `WithdrawRelation#PublicInput` where `DepositRelation#PublicInput` is expected while you can do that with `Vec<CircuitField>` 🤷

mike1729: In general this is obviously true but in the way we use public input this doesn't seem to give much, right?

deuszx: I just think we should write "better" code always, not only when it makes sense.

Doesn't have to be addressed in this PR obvs.

relations/jf/src/note.rs
```diff
@@ -102,16 +110,19 @@ impl NoteGadget for PlonkCircuit<CircuitField> {
 impl PublicInput for SourcedNote {
```
deuszx, line 110: I'm not sure why there even exists an `impl PublicInput for SourcedNote`: my understanding was that public/private/full inputs are for _relations_ and `SourcedNote` is a value in that relation; relation == function, note == input to the relation. I understand that it might be easier to write it that way and require at the type-level that `forall A exists A: PublicInput` but from a "domain" perspective (ie in SNARK theory?) there is no public input for a note, correct?

deuszx (approved):

mike1729:
Note is not an input to the outside relation it's subrelation or subcircuit in math meaning.

PR #1195, opened 2023-05-17
A0-2433: update substrate to 0.9.39

kostekIV:
update to substrate 0.9.39.

The liminal build in particular the baby-liminal-extension was broken. I rm it as a dependency from runtime, plus commented out ci step checking it. CC @pmikolajczyk41.


pmikolajczyk41:
I suggest rebasing onto the latest commit on main, hopefully you'll find it useful :pray:

kostekIV:
> I suggest rebasing onto the latest commit on main, hopefully you'll find it useful pray

well, its required so I have to :|

bin/runtime/Cargo.toml
```diff
@@ -30,7 +30,7 @@ pallet-identity = { workspace = true }
 pallet-multisig = { workspace = true }
 pallet-nomination-pools = { workspace = true }
 pallet-nomination-pools-runtime-api = { workspace = true }
-pallet-randomness-collective-flip = { workspace = true }
+pallet-insecure-randomness-collective-flip = { workspace = true }
```
pmikolajczyk41, line 33: here: https://github.com/paritytech/substrate/tree/master/frame/insecure-randomness-collective-flip they are screaming that: **DO NOT USE IN PRODUCTION** - should we be worried? or it has always been like this

pmikolajczyk41 (approved):
lgtm, gg

bin/runtime/Cargo.toml
kostekIV: it has always been like this. afaik it is only used in the pallet_contracts and source of randomness

mike1729: Is it still? I believe there was an effort to remove it after last audit.

kostekIV: yep, this is deprecated so we dont have to worry about it being insecure. I can remove the randomness pallet from the runtime but lets do it in separate pr
image

pmikolajczyk41: sounds good for me, resolving

finality-aleph/src/party/manager/mod.rs
```diff
@@ -333,6 +330,11 @@ where
         };
 
         let last_block_of_previous_session = session_boundaries.first_block().saturating_sub(1);
+        let last_block_of_previous_session_hash = self
+            .client
+            .block_hash(last_block_of_previous_session)
+            .unwrap()
+            .unwrap();
```
mike1729, line 337: I'm afraid that two unwraps don't work as two minuses :(

mike1729 (changes requested):
Unwrap two unwraps.

finality-aleph/src/party/manager/mod.rs
kostekIV: Changed to expect :)

mike1729 (approved):

PR #1199, opened 2023-05-19
A0-2289: Use workspace dependencies

pmikolajczyk41:
# Description

 - we define some common package attributes in the workspace file and inherit them in members
 - we move (almost) all dependencies to the workspace file and inherit them in members
 - dependencies with legacy/current versions (in `finality-aleph`) are left at the package level
 - we remove script for renaming substrate branch - now everything lives in the single file

Notes:
 - currently rust doesn't work with (/ I'm too dumb) renamed and inherited packages, so there is no more `codec` abbreviation, we use everywhere `parity-scale-codec`; with `aleph_primitives` I have renamed import in the lib crate roots 

## Type of change

maintenance


deuszx (approved):

jellyfier/Cargo.toml
```diff
@@ -1,34 +1,34 @@
...
 categories = ["cryptography"]
-repository = "https://github.com/Cardinal-Cryptography/aleph-node"
 description = "Jellyfish PLONK verifier"
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+repository.workspace = true
 
 [dependencies]
+# we must use newer version of arkworks for jellyfish
```
woocash2, line 14: Should we understand this comment as "todo: upgrade to newer arkworks version"? 
Currently someone could interpret it as: "0.4.0 below is the newer version, and we must use it".

woocash2: Nvm, I noticed 0.3.0 in workspace, so probably the second meaning was intended.

woocash2 (approved):

PR #1200, opened 2023-05-19
A0-2517: Other major sync solution

timorl:
# Description

Now reports more blocks if it missed them. Also slows down major sync on purpose to make this actually work reasonably.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/sync/substrate/status_notifier.rs
+++ b/finality-aleph/src/sync/substrate/status_notifier.rs
@@ -1,21 +1,31 @@
-use std::fmt::{Display, Error as FmtError, Formatter};
+use std::{fmt::{Display, Error as FmtError, Formatter}, time::{Duration, Instant}};
 
 use aleph_primitives::BlockNumber;
 use futures::StreamExt;
 use sc_client_api::client::{FinalityNotifications, ImportNotifications};
 use sp_runtime::traits::{Block as BlockT, Header as SubstrateHeader};
-use tokio::select;
+use tokio::{select, time::sleep};
 
-use crate::sync::{ChainStatusNotification, ChainStatusNotifier};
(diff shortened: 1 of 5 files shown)


Marcin-Radecki (approved):

finality-aleph/src/sync/substrate/status_notifier.rs
```diff
@@ -59,19 +104,46 @@ where
...
+                    if number > self.last_reported {
+                        self.last_reported = number;
+                    }
+                }
+                self.trying_since = Instant::now();
                 maybe_block
                 .map(|block| ChainStatusNotification::BlockImported(block.header))
                 .ok_or(Error::ImportStreamClosed)
+            },
+            _ = sleep(Duration::from_secs(2).saturating_sub(Instant::now() - self.trying_since)) => {
```
DamianStraszak, line 144: I think `Instant::now() - self.trying_since` can panic. Better to use something which has `try` or `saturating` in the name.

DamianStraszak (changes requested):

finality-aleph/src/sync/substrate/status_notifier.rs
DamianStraszak: I don't see any unsafe subtraction in the current version so I guess it cannot panic anymore :) 

You can resolve the comment after you read it.

DamianStraszak (approved):

finality-aleph/src/sync/substrate/status_notifier.rs
DamianStraszak: Oh wait, I'm not sure what happens if you sleep `0`. Does it panic?

timorl: Nope, [the documentation](https://docs.rs/tokio/latest/tokio/time/fn.sleep.html#panics) doesn't mention it as a possibility when panicking.

Marcin-Radecki (approved):

PR #1205, opened 2023-05-21
A0-0000: restore script

kostekIV:
This script is still needed in dirs like e2e-tests, aleph-cliet and similar places


pmikolajczyk41 (approved):
sry

woocash2 (approved):

PR #1206, opened 2023-05-22
A0-2507: Add digest log for authority set change (runtime change)

krzysztofziobro:
# Description

Adds a digest log containing new aleph authority set (in a block where the change happens).

Important runtime change: modifies contents of some headers.

## Type of change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6250,9 +6250,9 @@ dependencies = [
 
 [[package]]
 name = "parity-scale-codec"
-version = "3.6.0"
+version = "3.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "430d26d62e66cbff6ae144e8eebd43c0235922dec7e3aa0d2016c32d4575bf45"
+checksum = "2287753623c76f953acd29d15d8100bcab84d29db78fb6f352adb3c53e83b967"
 dependencies = [
  "arrayvec 0.7.4",
  "bitvec",
@@ -6265,9 +6265,9 @@ dependencies = [
 
 [[package]]
 name = "parity-scale-codec-derive"
-version = "3.6.0"
+version = "3.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a1620b1e3fc72ebaee01ff56fca838bab537c5d093a18b3549c3bbea374aa0b6"
+checksum = "2b6937b5e67bfba3351b87b040d48352a2fcb6ad72f81855412ce97b45c8f110"
 dependencies = [
  "proc-macro-crate",
  "proc-macro2",
@@ -10528,9 +10528,9 @@ dependencies = [
 
 [[package]]
 name = "tracing-attributes"
-version = "0.1.24"
+version = "0.1.25"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0f57e3ca2a01450b1a921183a9c9cbfda207fd822cef4ccb00a65402cbba7a74"
+checksum = "8803eee176538f94ae9a14b55b2804eb7e1441f8210b1c31290b3bccdccff73b"
 dependencies = [
  "proc-macro2",
  "quote",


pallets/aleph/src/lib.rs
```diff
@@ -123,6 +129,18 @@ pub mod pallet {
     pub(super) type FinalityScheduledVersionChange<T: Config> =
         StorageValue<_, VersionChange, OptionQuery>;
 
+    #[pallet::hooks]
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        fn on_finalize(block_number: T::BlockNumber) {
+            if block_number % DEFAULT_SESSION_PERIOD.into() == (DEFAULT_SESSION_PERIOD - 1).into() {
```
obrok, line 135: Question: are sessions guaranteed to last `DEFAULT_SESSION_PERIOD` or are there circumstances when they might last longer or shorter?

obrok (approved):

pallets/aleph/src/lib.rs
DamianStraszak: Even emergency finalizer can't change that. The session length is always constant, and this constant is `DEFAULT_SESSION_PERIOD` (the value is hardcoded into the runtime, the only way to change it is the `short_session` feature flag, or changing it explicitly in the code).

pallets/aleph/src/lib.rs
```diff
@@ -123,6 +129,18 @@ pub mod pallet {
...
+                <frame_system::Pallet<T>>::deposit_log(DigestItem::Consensus(
```
DamianStraszak, line 136: I'm wondering how does it actually work. I.e. is this log somehow verified by the remaining validators :P? I would assume it is, but I guess it would be good to make sure, because nobody here is familiar with how these logs work...

DamianStraszak (commented):
The PR looks good, I have only asked some clarifying questions. 

An important thing to do is to make sure this PR is mentioned in release notes in the next release. Not sure how to achieve that though -- but make sure to add a note to the PR description saying that its an important change to the runtime.

pallets/aleph/src/lib.rs
krzysztofziobro: Right, finalizer does not change that, what I wrote does not really make sense

mike1729: The check above is fine but a little nicer would be to use https://paritytech.github.io/substrate/master/pallet_session/pallet/trait.Config.html#associatedtype.NextSessionRotation which in our case is `PeriodicSession` instantiated here:     type ShouldEndSession = pallet_session::PeriodicSessions<SessionPeriod, Offset>;

pallets/aleph/src/lib.rs
mike1729: Actually, somebody has a raw idea how they work :P The check is done in some convoluted way, i.e. other validators don't trust logs (they are not part of prehash for example). They execute the block and after executing it they check if the logs they got from execution (particularily from `on_finalize`) match the ones that was send with the block. Checks:
https://github.com/paritytech/substrate/blob/master/frame/executive/src/lib.rs#L588
https://github.com/paritytech/substrate/blob/master/frame/executive/src/lib.rs#L594

mike1729 (approved):
Approve but it would be nice to use existing API for session boundary.

DamianStraszak (approved):

bin/runtime/src/lib.rs
```diff
@@ -310,6 +310,7 @@ impl pallet_aleph::Config for Runtime {
     type AuthorityId = AlephId;
     type RuntimeEvent = RuntimeEvent;
     type SessionInfoProvider = SessionInfoImpl;
+    type NextSessionRotation = pallet_session::PeriodicSessions<SessionPeriod, Offset>;
```
mike1729, line 313: I propose to extend trait `SessionInfoProvider` instead of new config item.

pallets/aleph/src/lib.rs
```diff
@@ -123,6 +134,23 @@ pub mod pallet {
     pub(super) type FinalityScheduledVersionChange<T: Config> =
         StorageValue<_, VersionChange, OptionQuery>;
 
+    #[pallet::hooks]
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        fn on_finalize(block_number: T::BlockNumber) {
+            if let Some(session_change_block) =
+                T::NextSessionRotation::estimate_next_session_rotation(block_number).0
+            {
+                if session_change_block == block_number + 1u32.into() {
```
kostekIV, line 143: @DamianStraszak wont there be a problem here because of +-1 between abft and block production committee session block start?

DamianStraszak: I think it's OK, as long as this thing `T::NextSessionRotation::estimate_next_session_rotation(block_number).0` is indeed always returning `session_period*k`.

The alephBFT change would then be included in block `900*k + 899` and the Aura change in `900*k+900`, which is fine. We cannot include the alephBFT change in block `900k+900` because that's actually one block to late, and the new committee would already sign it.

DamianStraszak (approved):

bin/runtime/src/lib.rs
```diff
@@ -64,6 +64,8 @@ use sp_std::prelude::*;
 use sp_version::NativeVersion;
 use sp_version::RuntimeVersion;
 
+use crate::sp_api_hidden_includes_construct_runtime::hidden_include::traits::EstimateNextSessionRotation;
```
mike1729, line 67: You almost never want to use traits from `sp_api_hidden_includes_construct_runtime` even though compiler encourages you to do so ;) There are corresponding traits defined in `frame_support`. In this case, this one: https://paritytech.github.io/substrate/master/frame_support/traits/trait.EstimateNextSessionRotation.html#

pallets/aleph/src/mock.rs
```diff
@@ -16,6 +16,7 @@ use sp_runtime::{
 
 use super::*;
 use crate as pallet_aleph;
+use crate::mock::sp_api_hidden_includes_construct_runtime::hidden_include::traits::EstimateNextSessionRotation;
```
mike1729, line 19: https://paritytech.github.io/substrate/master/frame_support/traits/trait.EstimateNextSessionRotation.html#

primitives/src/lib.rs
```diff
@@ -311,8 +320,9 @@ impl<T> Default for SessionValidators<T> {
 }
 
 /// Information provider from `pallet_session`. Loose pallet coupling via traits.
-pub trait SessionInfoProvider {
+pub trait SessionInfoProvider<T: AtLeast32BitUnsigned> {
```
mike1729, line 323: With the correct trait this bound is not needed.

mike1729 (changes requested):
Final nits :)

mike1729 (approved):

PR #1207, opened 2023-05-22
A0-2494: Jellypallet

pmikolajczyk41:
# Description

 - move from using Groth16 from `arkworks` to PLONK from `jellyfish`. 
 - remove `jellyfier` host function - verification can be done directly in the pallet
 - we abandon verification benchmarks, and until shielder relations are ready, we are charging 0 weight

## Type of change

- I have added tests


deuszx (approved):

pallets/baby-liminal/src/lib.rs
```diff
@@ -334,71 +342,60 @@ pub mod pallet {
...
-            Self::deposit_event(Event::VerificationSucceeded);
-            Ok(())
+            match PlonkKzgSnark::verify::<StandardTranscript>(
+                &verification_key,
+                &public_input,
+                &proof,
+                None,
+            ) {
+                Ok(_) => {
+                    Self::deposit_event(Event::VerificationSucceeded);
```
fbielejec, line 393: perhaps this event could carry more info: proof and input or v-key identifier at the very least?

pallets/baby-liminal/src/lib.rs
```diff
@@ -334,71 +342,60 @@ pub mod pallet {
...
-            Self::_bare_verify::<Groth16>(verification_key_identifier, proof, public_input)
-        }
-
-        fn _bare_verify<S: VerifyingSystem>(
-            verification_key_identifier: VerificationKeyIdentifier,
-            proof: Vec<u8>,
-            public_input: Vec<u8>,
         ) -> Result<(), (Error<T>, Option<Weight>)> {
             let data_length_limit = T::MaximumDataLength::get() as usize;
             let data_length_excess = proof.len().saturating_sub(data_length_limit)
```
fbielejec, line 347: I don't understand this computation: specifically why subtract with saturation and but add both values with a "regular" arithmetic op?

Maybe use checked operations anyway, even if you need to introduce extra error variant

pallets/baby-liminal/src/lib.rs
pmikolajczyk41: both proof and input might be pretty big (kilobytes), so I just included the v-key identifier

pmikolajczyk41:
not merging yet, because locally, the node cannot start:
```
Error: Service(Client(VersionInvalid("host doesn't provide any imports from non-env module: __wbindgen_placeholder__:__wbindgen_describe")))
```

PR #1211, opened 2023-05-23
A0-2155: Block import v2

lesniak43:
# Description

Abstraction and implementation of `BlockImport`.

I literally have no idea if the arguments of `IncomingBlock` make sense, this needs to be checked later when we start using this code.

## Type of change


finality-aleph/src/lib.rs
```diff
@@ -269,20 +270,16 @@ impl<H: Header<Number = BlockNumber>> BlockIdentifier for BlockId<H> {
...
-    B: Block,
-    B::Header: Header<Number = BlockNumber>,
-    H: ExHashT,
-{
-    pub network: Arc<NetworkService<B, H>>,
+pub struct AlephConfig<C, SC, CS> {
+    pub network: Arc<NetworkService<AlephBlock, AlephHash>>,
     pub client: Arc<C>,
     pub chain_status: CS,
+    pub import_queue_handle: Box<dyn ImportQueueService<AlephBlock>>,
```
kostekIV, line 277: why Box<dyn> over generic?

finality-aleph/src/sync/substrate/mod.rs
```diff
@@ -20,6 +22,25 @@ pub use status_notifier::SubstrateChainStatusNotifier;
 pub use translator::Error as TranslateError;
 pub use verification::{SessionVerifier, SubstrateFinalizationInfo, VerifierCache};
 
+impl BlockImport<Block> for Box<dyn ImportQueueService<Block>> {
+    fn import_block(&mut self, block: Block) {
+        let origin = BlockOrigin::NetworkBroadcast;
+        let incoming_block = IncomingBlock::<Block> {
```
kostekIV, line 28: if we have no idea if this is correct let's wait with this pr and create a task for discovery and research.

kostekIV (changes requested):

finality-aleph/src/lib.rs
lesniak43: this is what we get from substrate

finality-aleph/src/sync/substrate/mod.rs
kostekIV: Its not about aura here but a block import, right? Isn't it the case that it might work in e2e test but we wont know why? Also, it wont tell us about the subtle meaning of some of these args.

lesniak43: I think that the actual verification is done by aura? Nevertheless, it's infeasible to check this here, I'd still suggest waiting for A0-2406 (it's the same epic).
> Isn't it the case that it might work in e2e test but we wont know why?
yes, but I do hope that our tests cover everything that we need to check...

kostekIV: Still, imo we need to know what is going on here. Even if this blocks another task it just mean it has at least the same priority as the blocked one

finality-aleph/src/lib.rs
timorl: Ugh, it would be nice to wrap it (as I think we are trying to avoid traits here?) at some point, because the type is quite ugly. This is not a blocker for this PR imo, but if you have an idea how to do this then please do.

finality-aleph/src/nodes.rs
```diff
@@ -149,12 +147,18 @@ where
         genesis_header,
     );
     let finalizer = AlephFinalizer::new(client.clone(), metrics.clone());
-    let (sync_service, justifications_for_sync, _) = match SyncService::new(
+
+    let (sync_service, justifications_for_sync, _): (
+        SyncService<Block, _, _, _, _, _, _, Box<dyn ImportQueueService<Block>>>,
```
timorl, line 152: Why is the second template arg necessary?

If it's because the inner `dyn` depends on `Block` then this is another argument for a wrapper at least, as mentioned in the previous comment.

finality-aleph/src/sync/service.rs
```diff
@@ -61,23 +66,27 @@ impl<BI: BlockIdentifier> RequestBlocks<BI> for mpsc::UnboundedSender<BI> {
...
         V: Verifier<J>,
         F: Finalizer<J>,
-    > Service<J, N, CE, CS, V, F>
+        BI: BlockImport<B>,
+    > Service<B, J, N, CE, CS, V, F, BI>
 {
     /// Create a new service using the provided network for communication.
     /// Also returns an interface for submitting additional justifications,
     /// and an interface for requesting blocks.
+    #[allow(clippy::too_many_arguments)]
```
timorl, line 82: Nooo~! D: Figure out some wrapping or something?

timorl (approved):
Only style comments.

finality-aleph/src/lib.rs
lesniak43: `AlephConfig` needs to disappear...

IDK, maybe we want a type alias in `primitives`, similar to `Block`, `Header` etc.? How should we call it then, `BlockImport`?

finality-aleph/src/nodes.rs
lesniak43: turns out both were not necessary...
fixed

finality-aleph/src/sync/service.rs
lesniak43: do you want `SyncConfig`, similar to `AlephConfig`?...

finality-aleph/src/sync/substrate/mod.rs
lesniak43: ok, I'll read the substrate code today, and try to figure sth out

finality-aleph/src/lib.rs
lesniak43: no, w8, this makes little sense - `run_validator_node` should use trait bounds from submodules, as it's an interface to the whole crate, let's discuss this properly

finality-aleph/src/sync/service.rs
lesniak43: I'd say let's wait till we have all the necessary building blocks, and only then try to group them, possibly even after refactoring `run_validator_node`

finality-aleph/src/sync/substrate/mod.rs
lesniak43: OR we could simply merge these two tasks - what do you think? I'd feel more comfortable with that, since I'd be able to, at least partially, check if I understand substrate api correctly

kostekIV: Yep, that would be nice :)

finality-aleph/src/sync/service.rs
timorl: I at least wouldn't want this to get forgotten, so a task+TODO pointing to the task here? Although `SyncConfig` would also be good imo.

finality-aleph/src/lib.rs
lesniak43: wrapped

finality-aleph/src/sync/service.rs
lesniak43: packed

finality-aleph/src/sync/service.rs
```diff
@@ -112,7 +148,7 @@ where
                 justifications_from_user,
                 additional_justifications_from_user,
                 _block_requests_from_user,
-                _block_importer,
+                _block_importer: block_importer,
```
timorl, line 151: I suspect you won't want that here, but rather within `Handler` in the future, so you could just pass `database_io` still packed there. But this can also be fixed in following PRs (or maybe I'm wrong), so not a blocker.

timorl (approved):
Feel free to resolve the only comment below, as it implies, if you don't want to do it.

finality-aleph/src/sync/service.rs
lesniak43: whooops, fixed

kostekIV (approved):

finality-aleph/src/sync/substrate/chain_status.rs
```diff
@@ -49,6 +50,13 @@ impl Display for Error {
                     hash
                 )
             }
+            MissingIndexedBody(hash) => {
```
timorl, line 53: You never construct this, or do you?

It would only make sense if you expected the option to always be `Some` for blocks in the DB, but I'm not sure that's correct.

timorl (approved):
Just one thing that I thing clippy should've caught...? So maybe I just missed something.

finality-aleph/src/sync/substrate/chain_status.rs
lesniak43: oh come on, I was just fixing this :P

timorl: 遅い!

PR #1220, opened 2023-05-25
A0-0000: Fix rng in native

kostekIV:


pallets/elections/src/impls.rs
```diff
@@ -13,7 +14,7 @@ where
     T: Config,
 {
     fn populate_next_era_validators_on_next_era_start(era: EraIndex) {
-        let mut rng = SmallRng::seed_from_u64(era as u64);
+        let mut rng = Pcg64Mcg::seed_from_u64(era as u64);
```
pmikolajczyk41, line 17: `Pcg32` as in the companion PR?

DamianStraszak (approved):
Ok, only make sure to address Piotr's comments :)

pallets/elections/src/impls.rs
Marcin-Radecki: yes, done

pmikolajczyk41 (approved):

PR #1230, opened 2023-05-29
A0-2405: Use every_imported_block notification

lesniak43:
# Description

We use the extended notification that will inform us about every imported block. We also can request justifications, hence `Forest` refresh becomes unnecessary.

## Type of change

# Checklist:


timorl (commented):
The removal of the stall ticker might be premature – we should check how this interacts with a sync from genesis in this form, and possibly port the sleep change from `11.0`.

lesniak43:
connected to testnet, synced 0.5M blocks in 30 minutes, no issues

timorl (changes requested):
Just a sec, there are some more concerns after all...

timorl (approved):
Alright, now it's g2g.

timorl (changes requested):

finality-aleph/src/sync/forest/mod.rs
```diff
@@ -101,26 +119,65 @@ impl<I: PeerId, J: Justification> VertexWithChildren<I, J> {
...
+        };
+        let mut deque = VecDeque::from([top_finalized]);
+        while let Some(hash) = deque.pop_front() {
+            let children = chain_status.children(hash).map_err(Error::ChainStatus)?;
+            for header in children.iter() {
+                match forest.update_body(header) {
+                    Err(Error::TooNew) => {
+                        warn!(
+                                target: LOG_TARGET,
+                                "There are more imported non-finalized blocks that can fit into the Forest: {}.", Error::<B, J, CS>::TooNew
```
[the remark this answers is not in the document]
lesniak43: and then what?

finality-aleph/src/sync/forest/mod.rs
```diff
@@ -101,26 +119,65 @@ impl<I: PeerId, J: Justification> VertexWithChildren<I, J> {
...
-            highest_justified,
+            highest_justified: top_finalized.clone(),
             justified_blocks: HashMap::new(),
-            root_id,
+            root_id: top_finalized.clone(),
             root_children: HashSet::new(),
             compost_bin: HashSet::new(),
+            _phantom: PhantomData,
+        };
+        let mut deque = VecDeque::from([top_finalized]);
```
kostekIV, line 162: pls extract this logic to fn

lesniak43: why? this is basically the only interesting thing happening inside `new`

finality-aleph/src/sync/forest/mod.rs
```diff
@@ -101,26 +119,65 @@ impl<I: PeerId, J: Justification> VertexWithChildren<I, J> {
 // ever need worst case scenario.
 const MAX_DEPTH: u32 = 1800;
 
-pub struct Forest<I: PeerId, J: Justification> {
+pub struct Forest<B, I, J, CS>
```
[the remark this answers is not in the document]
lesniak43: done

finality-aleph/src/sync/forest/mod.rs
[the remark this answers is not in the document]
lesniak43: done

finality-aleph/src/sync/forest/mod.rs
lesniak43: still in `new`, but looks much better now

finality-aleph/src/sync/forest/mod.rs
```diff
@@ -79,6 +79,35 @@ impl Display for Error {
...
+    CS: ChainStatus<B, J>,
+{
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), FmtError> {
+        match self {
+            InitializationError::Error(e) => match e {
+                Error::TooNew => write!(f, "Error during initialization. There are more imported non-finalized blocks in the Substrate database that can fit into the Forest. Purge the block database and restart the node."),
+                e => write!(f, "Error during initialization: {}", e),
+            },
+            InitializationError::ChainStatus(e) => write!(f, "ChainStatus error during initialization: {}", e),
+        }
```
timorl, line 107: The "during initialization" things here are repeating themselves with the wrapper in handler – in general when writing an error message you want it to contain only internal information, not about the expected context.

On a similar note – error messages should not end with `.`, as this will get repeated in the log message. It's also better if they don't start with capital letters.

So imo better versions of this messages are, in order:
1. "there are more imported non-finalized blocks in the database that can fit into the forest – purge the block database and restart the node"
2. "{}" (yes, just the underlying error)
3. "chain status error: {}"

timorl (approved):

finality-aleph/src/sync/forest/mod.rs
lesniak43: fixed

PR #1231, opened 2023-05-29
A0-2571: Sanity check for minimum session duration.

DamianStraszak:
# Description

Note that the cherrypick for A0-2571 has already been done in https://github.com/Cardinal-Cryptography/aleph-node/pull/1221 
This is only :
- cleaning up dead code that was the main reason for this bug to appear
- adding a sanity check that will save us from a similar mistake in the future

## Type of change

# Checklist:
- I have added tests

### DIFF
--- a/finality-aleph/src/abft/legacy.rs
+++ b/finality-aleph/src/abft/legacy.rs
@@ -4,7 +4,7 @@ use network_clique::SpawnHandleT;
 use sp_blockchain::HeaderBackend;
 use sp_runtime::traits::{Block, Header};
 
-use super::common::{unit_creation_delay_fn, MAX_ROUNDS};
+use super::common::{sanity_check_round_delays, unit_creation_delay_fn, MAX_ROUNDS};
 pub use crate::aleph_primitives::{BlockNumber, LEGACY_FINALITY_VERSION as VERSION};
 use crate::{
     abft::NetworkWrapper,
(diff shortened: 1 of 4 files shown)


lesniak43 (approved):

finality-aleph/src/abft/legacy.rs
```diff
@@ -33,6 +33,12 @@ where
     C: HeaderBackend<B> + Send + 'static,
     ADN: Network<LegacyNetworkData<B>> + 'static,
 {
+    // Remove this check once we implement one on the AlephBFT side (A0-2583).
+    // Checks that the total time of a session is at least 7 days.
+    sanity_check_round_delays(
+        config.max_round,
```
woocash2, line 39: Does `config.max_round` denote the highest possible round, or the number of rounds? If the first one then given that rounds start from `0`, there's one extra round not counted in `sanity_check_round_delays_inner`.
(this probably doesn't affect this check's outcome for the test cases anyway)

woocash2 (approved):
lgtm, left one tiny comment

finality-aleph/src/abft/legacy.rs
DamianStraszak: Good point. So actually I have now checked in AlephBFT and it seems this is the highest possible round. However in practice, during the last crash they reached only round 4999, which probably means that something is crashing and the last unit is not created. It's not a big deal, because we should never allow this high round to be reached anyway. 

Anyway, I have changed the loop from `0..max_round` to `0..=max_round` in the sanity_check code. This way it is consistent with the intention of AlephBFT.

PR #1233, opened 2023-05-29
A0-2451: Document cancel safety properly

ggawryal:
Added a few missing comments regarding cancellation safety in `finality-aleph` crate.
Also I manually inspected by looking at the code, that assuming all select!s in `aleph-bft` and channels from `futures` are cancellation safe, then all select!s in `finality-aleph` shouldn't cause any problems with cancellation.

## Type of change

# Checklist:


fixxxedpoint (approved):
lgtm

PR #1234, opened 2023-05-29
A0-2389: Getting blocks from the substrate database

lesniak43:
# Description

Extending trait `ChainStatus` and its implementation to introduce exporting blocks.

## Type of change

# Checklist:


finality-aleph/src/sync/mod.rs
```diff
@@ -64,6 +64,8 @@ pub trait Justification: Clone + Send + Sync + Debug + 'static {
     fn into_unverified(self) -> Self::Unverified;
 }
 
+type BlockIdForBlock<B> = <<B as Block>::Header as Header>::Identifier;
```
timorl, line 67: Is this really necessary? I would expect all places that have `Block` to also have `Justification`, so you can just use the other one.

If it is, then at least restore some symmetry making the other one `BlockIdForJustification` or change the name somehow. Maybe this could be `IdFor`, since the block is the argument?

timorl (approved):
Makes sense, some small-ish comments.

finality-aleph/src/justification/requester.rs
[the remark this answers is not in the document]
lesniak43: removed

finality-aleph/src/sync/mod.rs
```diff
@@ -125,7 +123,7 @@ where
     type Error: Display;
 
     /// The status of the block.
-    fn status_of(&self, id: BlockIdForBlock<B>) -> Result<BlockStatus<J>, Self::Error>;
+    fn status_of(&self, id: BlockIdFor<J>) -> Result<BlockStatus<J>, Self::Error>;
 
     /// Export a copy of the block.
     fn block(&self, id: <B::Header as Header>::Identifier) -> Result<Option<B>, Self::Error>;
```
timorl, line 129: This should also use `BlockIdFor` probably, for consistency.

finality-aleph/src/sync/service.rs
```diff
@@ -43,6 +44,19 @@ pub struct Service<
     _block_requests_from_user: mpsc::UnboundedReceiver<BlockIdFor<J>>,
 }
 
+impl<B, J, N, CE, CS, V, F> HandlerTypes for Service<B, J, N, CE, CS, V, F>
```
timorl, line 47: Oh no...

timorl (changes requested):

finality-aleph/src/sync/service.rs
lesniak43: oh yes!

timorl (approved):
Disgusting, but as long as you don't accept that blocks have justifications this is probably the best we can do. ;<

PR #1238, opened 2023-05-30
A0-2499: Plonk merge relation

fbielejec:
# Description

PLONK-ed relations


relations/jf/src/merge.rs
```diff
@@ -0,0 +1,378 @@
...
+        MergeRelation::new(public, private)
+    }
+
+    fn merge_constraints_correctness(relation: MergeRelation) -> bool {
+        let circuit = MergeRelation::generate_circuit(&relation).unwrap();
+
+        match circuit.check_circuit_satisfiability(&relation.public_input()) {
+            Ok(_) => true,
+            Err(why) => {
+                println!("circuit not satisfied: {}", why);
```
deuszx, line 321: I'd return the reason here from the method : `anyhow::Result` and use the `why` as the _reason_. Then, on the call-site you can unwrap or `?` and see it in the logs by default, without having to rerun with `-- --nocapture`.

fbielejec: yet sometimes a failure is the desired outcome. in that case assertion panics and you get println! statements only when some tests failed

deuszx: You can still do that. If you expect `Ok` just `assert!(res.is_ok())` or `assert!(res.is_err())` otherwise. You'll get the error statement if the test fails as well b/c test runner prints the stack (reason).

fbielejec: then it's just a different way of writing the same behavior, hardly an improvement?

deuszx: I think my version is just cleaner - bubbles up the error to the caller instead of hiding it and relying on some assumptions of the test framework.

It's a nit nevertheless so you'll decide.

relations/ark/src/shielder/withdraw.rs
```diff
@@ -4,9 +4,9 @@ use liminal_ark_relation_macro::snark_relation;
 ///
 /// It expresses the facts that:
 ///  - `new_note` is a prefix of the result of hashing together `token_id`, `whole_token_amount`,
-///    `old_trapdoor` and `old_nullifier`,
-///  - `old_note` is a prefix of the result of hashing together `token_id`, `new_token_amount`,
 ///    `new_trapdoor` and `new_nullifier`,
+///  - `old_note` is a prefix of the result of hashing together `token_id`, `new_token_amount`,
```
pmikolajczyk41, line 8:
```suggestion
///  - `old_note` is the result of hashing together `token_id`, `new_token_amount`,
```

pmikolajczyk41 (approved):
lgtm, however, there is awful amount of boilerplate around - any ideas for reducing it?

fbielejec:
> lgtm, however, there is awful amount of boilerplate around - any ideas for reducing it?



> lgtm, however, there is awful amount of boilerplate around - any ideas for reducing it?

I guess we could factor out some common functionality and maybe finally reach for some macros.

But I have different concerns - is it's not very intuitive how to define public inputs - one has to backtrack the relations, with several gadgets that are now spread across, and that both define the relation as well as register some inputs as public. IMHO these two should be separated

mike1729 (approved):

PR #1244, opened 2023-06-05
A0-2472: bump substrate to 0.9.42

kostekIV:
# Description
* Keystore is no longer async, this made it so that a large part of our code is also not async :)
* ~Fix to fork-off will be after one cherry-pick to our substrate fork that exposes hidden struct that needs to be deserialised~
*  fee test show that 1400 transfer no longer fit into block: need to investigate - weight for transfer increased for about 55%


mike1729 (approved):

finality-aleph/src/network/session/service.rs
```diff
@@ -319,7 +319,7 @@ where
                     pen,
                 };
                 let (actions, data_from_network) =
-                    self.manager.update_validator_session(pre_session).await?;
+                    self.manager.update_validator_session(pre_session)?;
```
timorl, line 322: So this function doesn't have to be `async` either now, does it?

finality-aleph/src/network/tcp.rs
```diff
@@ -43,7 +43,7 @@ impl SecretKey for AuthorityPen {
     type PublicKey = AuthorityIdWrapper;
 
     async fn sign(&self, message: &[u8]) -> Self::Signature {
-        AuthorityPen::sign(self, message).await
+        AuthorityPen::sign(self, message)
```
timorl, line 46: So we can make `SecretKey` sync now, right? It would be nice to do that in this PR, but if it turns out to be too much work to propagate all the changes then at least create a task.

bin/node/src/service.rs
```diff
@@ -109,12 +109,7 @@ pub fn new_partial(
         })
         .transpose()?;
 
-    let executor = AlephExecutor::new(
-        config.wasm_method,
-        config.default_heap_pages,
-        config.max_runtime_instances,
-        config.runtime_cache_size,
-    );
+    let executor = sc_service::new_native_or_wasm_executor(config);
```
timorl, line 112: I'm a bit confused, how does this create `AlephExecutor`? Or am I misreading that below we run a function that accepts an `AlephExecutor` as an argument?

timorl (approved):
Looks good, minor comments.

bin/node/src/service.rs
kostekIV: AlephExecutor is just a type alias:
```
pub type AlephExecutor = NativeElseWasmExecutor<ExecutorDispatch>;
```

finality-aleph/src/abft/crypto.rs
```diff
@@ -69,7 +69,7 @@ impl current_aleph_bft::Keychain for Keychain {
     }
 
     async fn sign(&self, msg: &[u8]) -> Signature {
-        Keychain::sign(self, msg).await
+        Keychain::sign(self, msg)
```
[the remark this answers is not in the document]
kostekIV: https://cardinal-cryptography.atlassian.net/browse/A0-2697

finality-aleph/src/network/session/service.rs
kostekIV: good catch, done

finality-aleph/src/network/tcp.rs
kostekIV: yep, done

PR #1248, opened 2023-06-09
A0-2624: Lower log levels of verification failures

timorl:
# Description

We expect verification of justifications to sometimes fail, even if everything is going well. In practice this is most pronounced when synchronizing old blocks, but in general it shouldn't be an error. This PR treats verification errors separately for the purposes of logging.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/sync/service.rs
+++ b/finality-aleph/src/sync/service.rs
@@ -188,7 +188,7 @@ where
                 ),
                 e => warn!(
                     target: LOG_TARGET,
-                    "Error handling sync state from {:?}: {}.", peer, e
+                    "Failed to handle sync state from {:?}: {}.", peer, e
                 ),
             },
         }
@@ -223,7 +223,7 @@ where
                     e => {
                         warn!(
                             target: LOG_TARGET,
-                            "Error while handling justification from {:?}: {}.",
+                            "Failed to handle justification from {:?}: {}.",
                             peer.map_or("user".to_string(), |id| format!("{:?}", id)),
                             e
                         );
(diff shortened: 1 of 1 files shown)


lesniak43:
I believe we should avoid using the word "error" in logs with lvl lower than `error`, especially at the very beginning of the message.

finality-aleph/src/sync/service.rs
```diff
@@ -181,10 +181,16 @@ where
...
-                "Error handling sync state from {:?}: {}.", peer, e
-            ),
+            Err(e) => match e {
+                HandlerError::Verifier(e) => debug!(
+                    target: LOG_TARGET,
+                    "Could not verify justification in sync state from {:?}: {}.", peer, e
+                ),
+                e => warn!(
+                    target: LOG_TARGET,
+                    "Error handling sync state from {:?}: {}.", peer, e
```
woocash2, line 191: I agree with @lesniak43. Maybe change to "Failed handling sync state ..."?

finality-aleph/src/sync/service.rs
```diff
@@ -204,15 +210,26 @@ where
...
+                            "Could not verify justification from {:?}: {}.",
+                            peer.map_or("user".to_string(), |id| format!("{:?}", id)),
+                            e
+                        );
+                        return;
+                    }
+                    e => {
+                        warn!(
+                            target: LOG_TARGET,
+                            "Error while handling justification from {:?}: {}.",
```
woocash2, line 226: "Failed handling justification ..."

woocash2 (changes requested):

lesniak43 (approved):

finality-aleph/src/sync/service.rs
woocash2: True... Maybe I'll create a mini-task and fix it.

woocash2 (approved):

PR #1258, opened 2023-06-15
A0-2692: fix for nix

fixxxedpoint:
# Description

Changes in rust toolchain configuration (renamed rust-toolchain file) and weird behavior of `cargo` while building with `--locked` and `--offline` flags, i.e. writing new files inside of `CARGO_HOME` was causing failures of the nix based build procedure described in our `BUILD.md` document. This PR fixes these issues.

## Type of change

# IMPORTANT:
How to test it?
```
docker build -t aleph-build -f nix/Dockerfile.build . && docker run -ti --volume=$(pwd):/node/build aleph-build
```
After this, results (aleph-node and runtime binaries) should be stored in the `result` folder.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -12,7 +12,7 @@
 , runTests ? false
 # forces naersk (helper tool for building rust projects under nix) to build in a single derivation, instead default way that uses deps and project derivations
 # it is used for building aleph-runtime (we don't want its dependencies to be build separately for a non-WASM architecture)
-, singleStep ? false
+, singleStep ? true
 # passed to rustc by cargo - it allows us to set the list of supported cpu features
 # we can use for example `-C target-cpu=native` which should produce a binary that is significantly faster than the one produced using `generic`
 # `generic` is the default `target-cpu` provided by cargo
--- a/default.nix
+++ b/default.nix
@@ -140,7 +140,8 @@ with nixpkgs; naersk.buildPackage rec {
   postConfigure = ''
       ${nixpkgs.lib.optionalString providedCargoHome
          ''
-           export CARGO_HOME=${cargoHome}
+           cp -r ${cargoHome} .cargo_home
+           export CARGO_HOME=.cargo_home
          ''
        }
   '';
--- a/default.nix
+++ b/default.nix
@@ -12,6 +12,9 @@
 , runTests ? false
 # forces naersk (helper tool for building rust projects under nix) to build in a single derivation, instead default way that uses deps and project derivations
 # it is used for building aleph-runtime (we don't want its dependencies to be build separately for a non-WASM architecture)
+# FIXME two-step build fails. naersk attempts creating a type of a mock project that mainly consists of no-op main/lib.rs and a list cargo dependencies copied from
+# the processed project. Unfortunately, it somehow fails to process our workspace configuration and crashes while building some of our crates (even we don't use
+# them later in the main build procedure).
 , singleStep ? true
 # passed to rustc by cargo - it allows us to set the list of supported cpu features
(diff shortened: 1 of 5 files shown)


default.nix
```diff
@@ -12,7 +12,7 @@
 , runTests ? false
 # forces naersk (helper tool for building rust projects under nix) to build in a single derivation, instead default way that uses deps and project derivations
 # it is used for building aleph-runtime (we don't want its dependencies to be build separately for a non-WASM architecture)
-, singleStep ? false
+, singleStep ? true
```
fixxxedpoint, line 15: Two step build (only dependencies -> whole project with prebuilt deps) is currently broken due to some `cargo` resolution magic. This is a workaround.

default.nix
```diff
@@ -140,7 +140,8 @@ with nixpkgs; naersk.buildPackage rec {
   postConfigure = ''
       ${nixpkgs.lib.optionalString providedCargoHome
          ''
-           export CARGO_HOME=${cargoHome}
+           cp -r ${cargoHome} .cargo_home
```
fixxxedpoint, line 143: Somehow `cargo` attempts to write inside of the `CARGO_HOME` folder, which previously was stored inside of the `/nix` folder and so it was read-only. This copies it into the build folder.

nix/rocksdb.nix
```diff
@@ -3,7 +3,7 @@ let
   # all of these values can be modified using the override method of this derivation, e.g. `customRocksDb.override { useSnappy = true; }`
   defaultArgs = {
       # defines which version of rocksdb should be downloaded from github
-      version = "6.29.3";
+      version = "7.9.2";
```
fixxxedpoint, line 6: Version of `rocksdb` aligned with `substrate` version `polkadot-v0.9.42`.

default.nix
timorl: You probably should change the comment as well?

default.nix
timorl: Also add this information to the comment above.

nix/versions.nix
```diff
@@ -44,9 +44,9 @@ rec {
   naersk =
     let
       naerskSrc = builtins.fetchTarball {
-        # link: https://github.com/nix-community/naersk/tree/2fc8ce9d3c025d59fee349c1f80be9785049d653
-        url = "https://github.com/nix-community/naersk/archive/2fc8ce9d3c025d59fee349c1f80be9785049d653.tar.gz";
-        sha256 = "1jhagazh69w7jfbrchhdss54salxc66ap1a1yd7xasc92vr0qsx4";
+        # link: https://github.com/nix-community/naersk/tree/d998160d6a076cfe8f9741e56aeec7e267e3e114
```
timorl, line 47: Why is this here? Is this a comment?

timorl (approved):
Looks good, but I don't know much about nix. <_<"

default.nix
fixxxedpoint: Just added a `FIXME` tag for it.

nix/versions.nix
fixxxedpoint: That was a link to github allowing you to investigate its repo state for this particular version of the library. I deleted it.

default.nix
timorl: That doesn't seem like enough information to understand what this was about even in a month. <_< But since you are essentially the only person who knows what's going on here anyway I guess you can decide, so feel free to resolve if you feel this is sufficient.

lesniak43 (approved):

default.nix
fixxxedpoint: Added more details. But anyway, running it with this flag set to `true` gives enough info.

PR #1259, opened 2023-06-19
A0-2406: Handle incoming blocks

lesniak43:
# Description

Puts the incoming blocks into Substrate import queue.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/sync/handler.rs
+++ b/finality-aleph/src/sync/handler.rs
@@ -255,7 +255,9 @@ where
 
     /// Handle a single block.
     pub fn handle_block(&mut self, block: B) {
-        self.block_importer.import_block(block)
+        if self.forest.importable(&block.header().id()) {
+            self.block_importer.import_block(block);
+        }
     }
 
     /// Inform the handler that a block has been imported.
--- a/finality-aleph/src/sync/handler.rs
+++ b/finality-aleph/src/sync/handler.rs
@@ -13,7 +13,10 @@ use crate::{
 };
 
 /// How many justifications we will send at most in response to an explicit query.
-const MAX_JUSTIFICATION_BATCH: usize = 100;
+pub const MAX_JUSTIFICATION_BATCH: usize = 100;
+
+/// How many blocks we will send at most in response to an explicit query.
+pub const MAX_BLOCK_BATCH: usize = 25;
 
 /// Handles for interacting with the blockchain database.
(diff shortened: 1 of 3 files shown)


finality-aleph/src/sync/service.rs
```diff
@@ -282,10 +289,11 @@ where
...
                 let state = request.state().clone();
                 self.handle_request(request, peer.clone());
-                self.handle_state(state, peer)
+                self.handle_state(state, peer);
             }
-            RequestResponse(_, justifications) => {
-                self.handle_justifications(justifications, Some(peer))
+            RequestResponse(blocks, justifications) => {
+                self.handle_justifications(justifications, Some(peer));
+                self.handle_blocks(blocks);
```
kostekIV, line 296: hmmm, shouldnt we first import block and only then handle the justifications? Also its quite confusing to me why the response has both blocks and justifications in body. Shouldn't it be the separate response for justifications and blocks?

finality-aleph/src/sync/handler.rs
```diff
@@ -250,6 +250,11 @@ where
         Ok(maybe_id)
     }
 
+    /// Handle a single block.
+    pub fn handle_block(&mut self, block: B) {
+        self.block_importer.import_block(block)
```
timorl, line 255: Uhhh, you don't want to just import any blocks that someone throws at you, you gotta check whether they are required, no?

finality-aleph/src/sync/service.rs
timorl: The order doesn't matter that much, as handling justifications only puts them in the forest, they get actually justified when we get info that blocks got imported.

timorl (changes requested):
I think you missed a very important part of the logic here. <_<

finality-aleph/src/sync/handler.rs
lesniak43: immensely reasonable, fixed

finality-aleph/src/sync/service.rs
lesniak43: also, the main idea is that the response represents "all what can be useful to us according to the peer, based on info from our request", hence single message variant

kostekIV: Imho its better for code readability to make distinction here (in enum variant). Anyway resolving, looks good as it is :)

kostekIV (approved):

PR #1261, opened 2023-06-19
A0-000: remove pallet randomness

kostekIV:
# Description
Lets slim down runtime a little bit :) This will need to have migration included before the next release. The migration is quite simple with the newly implement RemovePallet utility from substrate.

## Type of change


bin/runtime/src/lib.rs
```diff
@@ -701,9 +699,17 @@ parameter_types! {
     pub Schedule: pallet_contracts::Schedule<Runtime> = Default::default();
 }
 
+pub struct ZeroRandomness;
+
+impl Randomness<Hash, AlephBlockNumber> for ZeroRandomness {
+    fn random(_: &[u8]) -> (Hash, AlephBlockNumber) {
+        Default::default()
```
pmikolajczyk41, line 706: just to find out myself in the context: the whole change is because pallet contracts dropped support for randomness as a stable feature for contracts?

bin/runtime/src/lib.rs
```diff
@@ -759,7 +765,6 @@ construct_runtime!(
         UncheckedExtrinsic = UncheckedExtrinsic
     {
         System: frame_system,
-        RandomnessCollectiveFlip: pallet_insecure_randomness_collective_flip,
```
pmikolajczyk41, line 762: this pallet shift will probably make some current tools fail (unless they use new codec approach for dynamic encoding like subxt does); I would consider keeping the original indexing - this can be done by explicitly defining pallet indices within this `construct_runtime` macro (maybe it's sufficient to mark only `Scheduler`, I'm not sure)

I will summon @DamianStraszak here for support or roast

DamianStraszak: I don't know how the explicit indexing works, and if one can have gaps (I would assume so), but since it's quite standard for other chains to use it, I think it would be good to also do it. It requires research ofc... The question is can we add indexing + remove the pallet right away in one update? I think it should be possible, but we should be consistent with the current indexing, which is probably some default one?

pmikolajczyk41: yes yes, we can do it right away, with gaps etc

DamianStraszak: Right, then it sounds like a good idea :)

bin/runtime/src/lib.rs
kostekIV: Not really, the change is because we do not this pallet and in the contracts the support for randomness can be replaced by a dummy implementation since its dropped

bin/runtime/src/lib.rs
kostekIV: Done :P

DamianStraszak (changes requested):
I'm OK merging this, but I see this as a high-risk change, since I'm not that familiar with runtime internals and I'm not sure how should one go about removing pallets. Also we will be adding a new pallet in the next release which looks kind of risky. 

My recommendation would be one of:
1. Merge, but prepare a plan for testing this in jira. So that we are sure that after running such an update no pallet breaks or so. 
2. Merge only indexing of the pallets, and schedule removing it for later. In a release when we are not adding any pallets.

pmikolajczyk41 (approved):

PR #1263, opened 2023-06-21
A0-2711: Ink!++

fbielejec:
# Description

update existing contracts to ink! v4.2.1

## Type of change

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,5 @@
 [toolchain]
 channel = "nightly-2022-10-30"
 targets = [ "wasm32-unknown-unknown" ]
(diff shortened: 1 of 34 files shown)


rust-toolchain.toml
```diff
@@ -1,3 +1,5 @@
 [toolchain]
 channel = "nightly-2022-10-30"
 targets = [ "wasm32-unknown-unknown" ]
+components = [ "rustfmt", "clippy", "rust-src" ]
```
kostekIV, line 4: can we specify different channels for build and different for lints&check :crossed_fingers:? If so we can use stable for builds :D

PR #1267, opened 2023-06-27
A0-2732: Fix handling of incoming responses

lesniak43:
# Description

Fixing the flawed handling of incoming responses in the sync.
Further separating `Service` and `Handler` logic.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/sync/handler.rs
+++ b/finality-aleph/src/sync/handler.rs
@@ -1,5 +1,6 @@
 use core::marker::PhantomData;
 use std::fmt::{Debug, Display, Error as FmtError, Formatter};
+use std::iter;
 
 use crate::{
     session::{SessionBoundaryInfo, SessionId},
@@ -14,6 +15,9 @@ use crate::{
 
 /// How many justifications we will send at most in response to an explicit query.
 pub const MAX_JUSTIFICATION_BATCH: usize = 100;
+
+/// How many headers we will send at most in response to an explicit query.
+pub const MAX_HEADER_BATCH: usize = 500;
 
 /// How many blocks we will send at most in response to an explicit query.
 pub const MAX_BLOCK_BATCH: usize = 25;
@@ -100,7 +104,7 @@ impl<B: Block, J: Justification> SyncAction<B, J> {
     }
 
     fn request_response(justifications: Vec<J::Unverified>) -> Self {
-        SyncAction::Response(NetworkData::RequestResponse(Vec::new(), justifications))
+        SyncAction::Response(NetworkData::RequestResponse(justifications, Vec::new(), Vec::new()))
     }
 }
 
@@ -120,6 +124,9 @@ where
     Forest(ForestError),
     ForestInitialization(ForestInitializationError<B, J, CS>),
     MissingJustification,
+    BadRequestResponse,
+    BlockNotImportable,
+    HeaderNotRequired,
 }
 
 impl<B, J, CS, V, F> Display for Error<B, J, CS, V, F>
@@ -142,6 +149,9 @@ where
                 f,
                 "justification for the last block of a past session missing"
             ),
+            BadRequestResponse => write!(f, "incorrect request response"),
+            BlockNotImportable => write!(f, "cannot import a block that we do not consider required"),
+            HeaderNotRequired => write!(f, "cannot import a header that we do not consider required"),
         }
     }
 }
@@ -253,13 +263,6 @@ where
         Ok(maybe_id)
     }
 
-    /// Handle a single block.
-    pub fn handle_block(&mut self, block: B) {
-        if self.forest.importable(&block.header().id()) {
-            self.block_importer.import_block(block);
-        }
-    }
-
     /// Inform the handler that a block has been imported.
     pub fn block_imported(
         &mut self,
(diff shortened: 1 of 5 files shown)


finality-aleph/src/sync/handler.rs
```diff
@@ -684,7 +785,7 @@ mod tests {
     fn handles_request() {
         let (mut handler, backend, _keep) = setup();
         let initial_state = handler.state().expect("state works");
-        let peer = rand::random();
+        let peer: MockPeerId = rand::random();
```
timorl, line 788: Huh, you now need this typing?

finality-aleph/src/sync/service.rs
```diff
@@ -374,10 +357,20 @@ where
 
     /// Stay synchronized.
     pub async fn run(mut self) {
+        use NetworkData::*;
         loop {
             tokio::select! {
                 maybe_data = self.network.next() => match maybe_data {
-                    Ok((data, peer)) => self.handle_network_data(data, peer),
+                    Ok((data, peer)) => match data {
+                        StateBroadcast(state) => self.handle_state(state, peer),
```
timorl, line 365: Imo it's better to have this match as a method rather than so much nesting.

timorl (changes requested):
Two small comments in addition to what we discussed elsewhere.

finality-aleph/src/sync/handler.rs
```diff
@@ -79,28 +79,54 @@ where
 
 /// What actions can the handler recommend as a reaction to some data.
 #[derive(Clone, Debug)]
-pub enum SyncAction<B: Block, J: Justification> {
+pub enum SyncAction<B, J, I>
```
timorl, line 82: This is disgustingly wrong. I will not elaborate.

finality-aleph/src/sync/handler.rs
lesniak43: it's a mystery

finality-aleph/src/sync/service.rs
lesniak43: fixed

finality-aleph/src/sync/handler.rs
```diff
@@ -307,21 +302,111 @@ where
...
+        };
+        self.try_finalize()?;
+        Ok(maybe_id)
+    }
+
+    fn handle_justifications(
+        &mut self,
+        justifications: Vec<J::Unverified>,
+        maybe_peer: Option<I>,
+    ) -> (Option<BlockIdFor<J>>, Option<<Self as HandlerTypes>::Error>) {
```
kostekIV, line 339: here also we could return boolean. 
Maybe_id could be an id of the top_justification.  In the loop in the ok variant we could simply check if the justification.id.number Is higher than new_top_justification. wdyt?

finality-aleph/src/sync/handler.rs
```diff
@@ -307,21 +302,111 @@ where
...
+                .update_required_header(&header, Some(peer.clone()))
+            {
+                return (sync_action, Some(Error::Forest(e)));
+            }
+        }
+
+        // handle blocks
+        for block in blocks {
+            match self.forest.importable(&block.header().id()) {
+                true => self.block_importer.import_block(block),
```
kostekIV, line 404: shouldnt we call `forest.update_body` also here?

finality-aleph/src/sync/handler.rs
timorl: That doesn't work in case we encounter an invalid justification halfway through – then we would either return `true` (making us request something that might not make sense) or `false` (making us not request out top justification).

finality-aleph/src/sync/handler.rs
timorl: Not really, we depend on the events to let us know the import really succeeded. In particular this line might fail for multiple reasons.

finality-aleph/src/sync/handler.rs
```diff
@@ -465,12 +548,12 @@ mod tests {
...
         let peer = rand::random();
-        assert!(
+        assert!(matches!(
             handler
                 .handle_justification(justification.clone().into_unverified(), Some(peer))
-                .expect("correct justification")
-                == Some(justification.id())
-        );
+                .expect("correct justification"),
+            Some(id) if id == justification.id()
```
timorl, line 555: You can now go back to the equality check, it looks less mysterious.

finality-aleph/src/sync/handler.rs
```diff
@@ -561,7 +644,7 @@ mod tests {
             .expect("correct justification")
         {
             Some(id) => assert_eq!(id, header.id()),
-            None => panic!("expected an id, got nothing"),
+            _ => panic!("expected an id, got nothing"),
```
timorl, line 647: Avoid `_` when replacing only a single thing. <_<

finality-aleph/src/sync/handler.rs
timorl: Meaning?

finality-aleph/src/sync/service.rs
```diff
@@ -177,24 +175,20 @@ where
...
+        use HandleStateAction::*;
         trace!(
             target: LOG_TARGET,
             "Handling state {:?} received from {:?}.",
             state,
             peer
         );
         match self.handler.handle_state(state, peer.clone()) {
-            Ok(action) => self.perform_sync_action(action, peer),
+            Ok(action) => match action {
```
timorl, line 187: Could we go back to having this as a separate method?

timorl (approved):
Only some nits left.

I'm not sure the more concrete return types from `Handler` made stuff more readable, now `Service` has to deal with all the return types differently in multiple places. But that might just be some weird aesthetic preference on my part, so whatever.

finality-aleph/src/sync/handler.rs
lesniak43: ask the compiler...

kostekIV (approved):

finality-aleph/src/sync/handler.rs
lesniak43: welp, turns out the compiler says it's ok, my bad, fixed

finality-aleph/src/sync/handler.rs
lesniak43: let's leave this for later, when we reexamine the task system (requesting highest justified blocks)

finality-aleph/src/sync/service.rs
lesniak43: (after a short discussion) let's leave it that way

finality-aleph/src/sync/handler.rs
lesniak43: fixed

PR #1268, opened 2023-06-29
A0-000: Rolling update test

h4nsu:
# Description

This PR adds an interactive local tool which tests the procedure for updating the chain (rolling update of nodes + runtime update).
Also contains a handful of small improvements of the chainrunner Python tool.

## Type of change


local-tests/rolling_update.ipynb
```diff
@@ -0,0 +1,363 @@
...
+    "                public_validator_addresses=val_addrs,\n",
+    "                unit_creation_delay=500,\n",
+    "                execution='Native',\n",
+    "                rpc_cors='all',\n",
+    "                rpc_methods='Unsafe',\n",
+    "                state_pruning='archive'\n",
+    "               )\n",
+    "\n",
+    "addresses = [n.address() for n in chain]\n",
+    "chain.set_flags(bootnodes=''.join(addresses), public_addr=addresses)"
```
timorl, line 102: Could you instead run an additional nonvalidator node and use that as the bootnode? This is even closer to our actual setup.

h4nsu: As the only bootnode, or just one of many?

timorl: I would go with only, as this is closer to what is going on irl. Unless you want to restart/update it as part of the test, then add a second one imo.

Marcin-Radecki (changes requested):
Hi, this is a great tool to have this all commands in one place. The choice of Jupyter is great since we can interact with individual test phases. What I'd add though to the test plan is
* separate bootnode, as pointed out by @timorl ,
* runtime should be upgraded somewhere in the middle, to achieve goal of having some nodes on old version, and some nodes on new version
* updating [readme](https://github.com/Cardinal-Cryptography/aleph-node/blob/main/local-tests/README.md) on how to run this file, ie it requires Jupyter, what paths and lines need to be replaced etc.

DamianStraszak (approved):

h4nsu:
@Marcin-Radecki 
I've made some updates according to your suggestions. I decided not to put the runtime update in between node updates for better readability. In the end this is not a run-everything-at-once script, but more like a set of buttons and it's up to the test operator to decide, in which order they should be pressed.

local-tests/rolling_update.ipynb
h4nsu: Done. Both ways are possible

Marcin-Radecki (approved):
Very nice

PR #1278, opened 2023-07-07
A0-2778: Bound notification block sync channel

ggawryal:
# Description

Replaced two channels in gossip network (`mpsc_notification_stream_authentication` and `mpsc_notification_stream_block_sync`) from tracing_unbounded to bounded channels, discarding new messages when buffer limit was hit and logging warning. For convenience, I replaced both these channels instead of just `mpsc_notification_stream_block_sync`.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/network/gossip/service.rs
+++ b/finality-aleph/src/network/gossip/service.rs
@@ -5,14 +5,16 @@ use std::{
     hash::Hash,
 };
 
-use futures::{channel::mpsc, StreamExt};
+use futures::{
+    channel::{mpsc, mpsc::TrySendError},
+    StreamExt,
+};
 use log::{debug, error, info, trace, warn};
 use network_clique::SpawnHandleT;
 use rand::{seq::IteratorRandom, thread_rng};
-use sc_utils::mpsc::{tracing_unbounded, TracingUnboundedReceiver, TracingUnboundedSender};
 use tokio::time;
 
-const QUEUE_SIZE_WARNING: usize = 1_000;
+const MAX_QUEUE_SIZE: usize = 1_000;
 
 use crate::{
     network::{
@@ -41,9 +43,9 @@ pub struct Service<N: RawNetwork, AD: Data, BSD: Data> {
     messages_for_authentication_user: mpsc::UnboundedSender<(AD, N::PeerId)>,
     messages_for_block_sync_user: mpsc::UnboundedSender<(BSD, N::PeerId)>,
     authentication_connected_peers: HashSet<N::PeerId>,
-    authentication_peer_senders: HashMap<N::PeerId, TracingUnboundedSender<AD>>,
+    authentication_peer_senders: HashMap<N::PeerId, NamedFuturesBoundedSender<AD>>,
     block_sync_connected_peers: HashSet<N::PeerId>,
-    block_sync_peer_senders: HashMap<N::PeerId, TracingUnboundedSender<BSD>>,
+    block_sync_peer_senders: HashMap<N::PeerId, NamedFuturesBoundedSender<BSD>>,
     spawn_handle: SpawnHandle,
 }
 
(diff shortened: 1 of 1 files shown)


finality-aleph/src/network/gossip/service.rs
```diff
@@ -5,14 +5,16 @@ use std::{
...
+    StreamExt,
+};
 use log::{debug, error, info, trace, warn};
 use network_clique::SpawnHandleT;
 use rand::{seq::IteratorRandom, thread_rng};
-use sc_utils::mpsc::{tracing_unbounded, TracingUnboundedReceiver, TracingUnboundedSender};
 use tokio::time;
 
-const QUEUE_SIZE_WARNING: usize = 1_000;
+const MAX_QUEUE_SIZE: usize = 1_000;
```
woocash2, line 17: We've recently seen the limit of 1000 being hit - just a question, if we're gonna drop the messages, do we want to increase this limit?

finality-aleph/src/network/gossip/service.rs
```diff
@@ -104,6 +106,34 @@ impl<D: Data, P: Clone + Debug + Eq + Hash + Send + 'static> Network<D> for Serv
...
+struct NamedFuturesBoundedSender<T> {
+    name: &'static str,
+    tx: mpsc::Sender<T>,
+}
+
+impl<T> NamedFuturesBoundedSender<T> {
+    pub fn send(&mut self, msg: T) -> Result<(), TrySendError<T>> {
+        self.tx.try_send(msg).map_err(|e| {
+            if e.is_full() {
+                warn!(
```
timorl, line 118: Logging should rather happen outside, next to the other `if` you changed. In particular then you cannot (but also don't have to) use the name – from context you will know which one it is, and even better, to whom the dropped message was addressed. Since you don't have to do that you can just use a plain `Sender`, so you don't even need this struct, just construct a `mpsc::channel` and you are done.

tl;dr You put in too much effort, you need more laziness in your life. ;p

timorl (changes requested):

finality-aleph/src/network/gossip/service.rs
timorl: If anything probably decrease – neither block sync nor authentications should be sending that many messages to a single node ever.

ggawryal: I'd keep it as it is and let us make relevant benchmarks to determine optimal value there later, in the task related to bounding memory usage in the whole aleph-node - probably this value should be editable in some config anyway.

finality-aleph/src/network/gossip/service.rs
ggawryal: Good point, tried to abstract the underlying channel, but I agree it is a bit unnecessary.

finality-aleph/src/network/gossip/service.rs
```diff
@@ -200,11 +193,18 @@ impl<N: RawNetwork, AD: Data, BSD: Data> Service<N, AD, BSD> {
     fn send_to_authentication_peer(&mut self, data: AD, peer: N::PeerId) -> Result<(), SendError> {
         match self.get_authentication_sender(&peer) {
             Some(sender) => {
-                match sender.unbounded_send(data) {
+                match sender.try_send(data) {
                     Err(e) => {
                         // Receiver can also be dropped when thread cannot send to peer. In case receiver is dropped this entry will be removed by Event::NotificationStreamClosed
```
woocash2, line 198: is this comment relevant to `e.is_full()`?

finality-aleph/src/network/gossip/service.rs
```diff
@@ -200,11 +193,18 @@ impl<N: RawNetwork, AD: Data, BSD: Data> Service<N, AD, BSD> {
...
                         // No need to remove the entry here
-                        if e.is_closed() {
+                        if e.is_full() {
+                            warn!(
+                                target: "aleph-network",
+                                "Failed sending data in mpsc_notification_stream_authentication to peer because peer_sender receiver is full: {:?}",
```
woocash2, line 203: Do we wanna use these names `mpsc_notification_stream_authentication` if actually nothing's named like this anymore? We could have: "Failed sending data through authentication notification channel ..." or sth like that.

woocash2 (approved):
Added some small comments, other than that looks great

finality-aleph/src/network/gossip/service.rs
timorl: Yeah, no point keeping the weird name, the alternative you suggest is better.

finality-aleph/src/network/gossip/service.rs
timorl: Not really, it should be only before `is_disconnected`.

timorl (approved):
Looks good, assuming @woocash2 's comments will get resolved.

finality-aleph/src/network/gossip/service.rs
ggawryal: Good catch

finality-aleph/src/network/gossip/service.rs
ggawryal: Agreed & renamed

PR #1286, opened 2023-07-12
A0-1898 add more timeouts in clique network

fixxxedpoint:
# Description

Since by default tokio's send operations for tcp connections do not timeout, it was possible to exploit a node and force it to create any number of incoming connections with a single node (Replace mechanic not working as expected) without possibility to drop previous ones - task was unable to receive a signal from a dropped user_data channel and finish.

## Type of change


clique/src/protocols/mod.rs
```diff
@@ -59,6 +61,7 @@ impl<PK: PublicKey> Display for ProtocolError<PK> {
             NoParentConnection => write!(f, "cannot send result to service"),
             NoUserConnection => write!(f, "cannot send data to user"),
             NotAuthorized => write!(f, "peer not authorized"),
+            Timeout => write!(f, "network operation timed out"),
```
timorl, line 64: Since this only happens during sending perhaps rename to `SendTimeout` and change the message appropriately? Alternatively (although possibly worse?) reuse `CardiacArrest` since the time you are waiting is very related to that anyway.

timorl (approved):

clique/src/protocols/mod.rs
fixxxedpoint: Changed to `SendTimeout`.

kostekIV (approved):

PR #1287, opened 2023-07-14
A0-2391: revamp

kostekIV:
Include blocks in the response

## Type of change

### DIFF
--- a/finality-aleph/src/sync/handler/request_handler.rs
+++ b/finality-aleph/src/sync/handler/request_handler.rs
@@ -1,0 +1,394 @@
+use core::{default::Default, marker::PhantomData};
+use std::{
+    collections::VecDeque,
+    fmt::{Display, Formatter},
+};
+
+use primitives::BlockNumber;
+
+use crate::{
+    session::{SessionBoundaryInfo, SessionId},
+    sync::{
+        data::BranchKnowledge, handler::Request, Block, BlockIdFor, BlockStatus, ChainStatus,
+        FinalizationStatus, Header, Justification,
+    },
+    BlockIdentifier,
+};
+
+#[derive(Debug, Clone)]
+pub enum RequestHandlerError<T: Display> {
+    RootMismatch,
+    MissingParent,
+    BadState,
+    BadRequest,
+    ChainStatusBadState,
+    ChainStatusError(T),
+}
+
+impl<T: Display> From<T> for RequestHandlerError<T> {
+    fn from(value: T) -> Self {
+        RequestHandlerError::ChainStatusError(value)
+    }
+}
+
+impl<T: Display> Display for RequestHandlerError<T> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            RequestHandlerError::RootMismatch => write!(f, "root mismatch"),
+            RequestHandlerError::MissingParent => write!(f, "missing parent"),
+            RequestHandlerError::ChainStatusError(e) => write!(f, "{}", e),
+            RequestHandlerError::BadState => write!(f, "bad state"),
+            RequestHandlerError::BadRequest => write!(f, "bad request"),
+            RequestHandlerError::ChainStatusBadState => write!(f, "bad state of chain_status"),
+        }
+    }
+}
+
+type HandlerResult<T, Error> = Result<T, RequestHandlerError<Error>>;
+
+pub enum Chunk<B, J>
+where
+    B: Block,
+    J: Justification<Header = B::Header>,
+{
+    Justification(J::Unverified),
+    Blocks(Vec<B>),
+    Headers(Vec<J::Header>),
+}
+
+pub fn into_vecs<B, J>(chunks: Vec<Chunk<B, J>>) -> (Vec<B>, Vec<J::Unverified>, Vec<J::Header>)
+where
+    J: Justification,
+    B: Block<Header = J::Header>,
+{
+    let mut blocks = vec![];
+    let mut headers = vec![];
+    let mut justifications = vec![];
+
+    for chunk in chunks {
+        match chunk {
+            Chunk::Blocks(mut bs) => {
+                bs.extend(blocks);
+                blocks = bs;
+            }
+            Chunk::Justification(j) => justifications.push(j),
+            Chunk::Headers(h) => headers.extend(h),
+        }
+    }
+
+    (blocks, justifications, headers)
+}
+
+#[derive(Debug)]
+enum HeadOfChunk<B, J>
+where
+    B: Block,
+    J: Justification<Header = B::Header>,
+{
+    Justification(J, Option<B>, Option<J::Header>),
+    Header(J::Header),
+}
+
+#[derive(Default)]
+struct PreChunk<B, J>
+where
+    B: Block,
+    J: Justification<Header = B::Header>,
+{
+    pub just: Option<J>,
+    pub blocks: VecDeque<B>,
+    pub headers: Vec<J::Header>,
+}
+
+impl<B, J> PreChunk<B, J>
+where
+    B: Block,
+    J: Justification<Header = B::Header>,
+{
+    fn new() -> Self {
+        Self {
+            just: None,
+            blocks: VecDeque::new(),
+            headers: vec![],
+        }
+    }
+
+    fn to_chunks(self) -> Vec<Chunk<B, J>> {
+        let mut chunks = vec![];
+
+        if let Some(j) = self.just {
+            chunks.push(Chunk::Justification(j.into_unverified()));
+        }
+
+        if !self.headers.is_empty() {
+            chunks.push(Chunk::Headers(self.headers));
+        }
+
+        if !self.blocks.is_empty() {
+            chunks.push(Chunk::Blocks(self.blocks.into()));
+        }
+
+        chunks
+    }
+
+    fn add_block(&mut self, b: Option<B>) {
+        if let Some(b) = b {
+            self.blocks.push_front(b);
+        }
+    }
+
+    fn add_header(&mut self, h: Option<J::Header>) {
+        if let Some(h) = h {
+            self.headers.push(h);
+        }
+    }
+
+    fn add_justification(&mut self, j: Option<J>) {
+        self.just = j;
+    }
+}
+
+#[derive(PartialEq, Eq, Debug)]
+enum State {
+    EverythingButHeader,
+    Everything,
+    OnlyJustification,
+    Finished,
+}
+
+pub struct RequestHandler<'a, B, J, CS>
+where
+    B: Block,
+    J: Justification<Header = B::Header>,
+    CS: ChainStatus<B, J>,
+{
+    chain_status: &'a CS,
+    session_info: &'a SessionBoundaryInfo,
+    _phantom: PhantomData<(B, J)>,
+}
+
+impl<'a, B, J, CS> RequestHandler<'a, B, J, CS>
+where
+    B: Block,
+    J: Justification<Header = B::Header>,
+    CS: ChainStatus<B, J>,
+{
+    pub fn new(chain_status: &'a CS, session_info: &'a SessionBoundaryInfo) -> Self {
+        Self {
+            chain_status,
+            session_info,
+            _phantom: PhantomData,
+        }
+    }
+
+    fn upper_limit(&self, id: BlockIdFor<J>) -> BlockNumber {
+        let session = self.session_info.session_id_from_block_num(id.number());
+        self.session_info
+            .last_block_of_session(SessionId(session.0 + 1))
+    }
+
+    fn get_block(&self, id: BlockIdFor<J>) -> HandlerResult<Option<B>, CS::Error> {
+        let b = self.chain_status.block(id)?;
+
+        Ok(b)
+    }
+
+    fn get_justification_block_and_header(
+        &self,
+        id: BlockIdFor<J>,
+    ) -> HandlerResult<(Option<J>, Option<B>, Option<J::Header>), CS::Error> {
+        let (justification, block) = self.get_justification_and_block(id.clone())?;
+        let header = block.as_ref().map(|b| b.header().clone());
+        Ok((justification, block, header))
+    }
+
+    fn get_justification_and_block(
+        &self,
+        id: BlockIdFor<J>,
+    ) -> HandlerResult<(Option<J>, Option<B>), CS::Error> {
+        let justification = self.get_justification(id.clone())?;
+        let block = self.chain_status.block(id.clone())?;
+        Ok((justification, block))
+    }
+
+    fn get_justification(&self, id: BlockIdFor<J>) -> HandlerResult<Option<J>, CS::Error> {
+        match self.chain_status.status_of(id.clone())? {
+            BlockStatus::Justified(justification) => Ok(Some(justification)),
+            BlockStatus::Present(_) => Ok(None),
+            BlockStatus::Unknown => Err(RequestHandlerError::RootMismatch),
+        }
+    }
+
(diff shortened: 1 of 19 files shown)


finality-aleph/src/sync/handler/request_handler.rs
```diff
@@ -0,0 +1,423 @@
...
+            RequestHandlerError::ChainStatusBadState => write!(f, "bad state of chain_status"),
+        }
+    }
+}
+
+type HandlerResult<T, Error> = Result<T, RequestHandlerError<Error>>;
+type JustificationBlockAndHeader<J, B> =
+    (Option<J>, Option<B>, Option<<J as Justification>::Header>);
+type JustificationAndBlock<J, B> = (Option<J>, Option<B>);
+type StepResult<B, J> = (State, HeadOfChunk<B, J>, Vec<Chunk<B, J>>);
```
timorl, line 51: Imo it would be clearer if this contained a `PreChunk` rather than directly a vector of chunks, and only the caller did the conversion.

(There is a comment somewhere about making this a separate struct as well, but that comment is separate from this one.)

finality-aleph/src/sync/handler/request_handler.rs
```diff
@@ -0,0 +1,423 @@
...
+
+        let our_top_justification = self.chain_status.top_finalized()?;
+        let top_justification = request.state().top_justification();
+        let target = request.target_id();
+
+        let upper_limit = self.upper_limit(top_justification.id());
+
+        // request too far into future
+        if target.number() > upper_limit {
+            return Ok(Response::from_chunks(vec![]));
```
timorl, line 381: Uh, we probably shouldn't be sending empty responses over the network for absolutely no reason, now should we?

And btw, this should probably be doing something else anyway – you constructed the upper limit from their top justification, so if the target is too far then they didn't really want to request it. Either return a `BadRequest` error or see whether you can respond with something useful by just trying to respond to `upper_limit`.

finality-aleph/src/sync/handler/request_handler.rs
```diff
@@ -0,0 +1,423 @@
...
+        };
+
+        let our_top_justification_number = our_top_justification.header().id().number();
+
+        // set the target to either target, our last justified block or finalized block
+        // at upper_limit height.
+        let head = if target.number() > our_top_justification_number {
+            head
+        } else if upper_limit > our_top_justification_number {
+            let b = self.get_block(our_top_justification.header().id())?;
```
timorl, line 400: This makes me slightly suspicious – we might not want to send this block in the end (if they are asking only for justifications), so the fact we are taking it out of the DB here suggests maybe something went wrong. But meh, it might just be sufficiently simpler to do this way, so this comment is just a mark so I can check after the other changes whether it now makes sense to me. Feel free to ignore.

finality-aleph/src/sync/handler/request_handler.rs
```diff
@@ -0,0 +1,423 @@
...
+            HeadOfChunk::Justification(our_top_justification, b, None)
+        } else {
+            match self.chain_status.finalized_at(upper_limit)? {
+                FinalizationStatus::FinalizedWithJustification(j) => {
+                    let b = self.get_block(j.header().id())?;
+                    HeadOfChunk::Justification(j, b, None)
+                }
+                FinalizationStatus::FinalizedByDescendant(header) => HeadOfChunk::Header(header),
```
timorl, line 408: This doesn't make sense, we cannot send such a chunk. On the other hand, this should never happen, as `upper_limit` is the end of a session, right? Maybe just return an error although, ugh, it will never really get returned...

finality-aleph/src/sync/handler/request_handler.rs
```diff
@@ -0,0 +1,423 @@
...
+
+impl<T: Display> Display for RequestHandlerError<T> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            RequestHandlerError::RootMismatch => write!(f, "root mismatch"),
+            RequestHandlerError::MissingParent => write!(f, "missing parent"),
+            RequestHandlerError::ChainStatusError(e) => write!(f, "{}", e),
+            RequestHandlerError::BadState => write!(f, "bad state"),
+            RequestHandlerError::BadRequest => write!(f, "bad request"),
+            RequestHandlerError::ChainStatusBadState => write!(f, "bad state of chain_status"),
```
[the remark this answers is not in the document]
timorl: In particular the last one should even be called something else, as the only case in which it gets thrown is very concrete (nonfinalized end of session block below top finalized?).

finality-aleph/src/sync/handler/request_handler.rs
```diff
@@ -0,0 +1,423 @@
...
+            HeadOfChunk::Justification(our_top_justification, b, None)
+        } else {
+            match self.chain_status.finalized_at(upper_limit)? {
+                FinalizationStatus::FinalizedWithJustification(j) => {
+                    let b = self.get_block(j.header().id())?;
+                    HeadOfChunk::Justification(j, b, None)
+                }
+                FinalizationStatus::FinalizedByDescendant(header) => HeadOfChunk::Header(header),
+                FinalizationStatus::NotFinalized => {
+                    return Err(RequestHandlerError::ChainStatusBadState)
```
timorl, line 410: Oh, actually you can rename this error and solve the above by returning it there as well: `UnjusitfiedAtEndOfSession` or something.

finality-aleph/src/sync/handler/request_handler.rs
```diff
@@ -0,0 +1,423 @@
...
+    Headers(Vec<J::Header>),
+}
+
+pub struct Response<B, J>
+where
+    B: Block,
+    J: Justification<Header = B::Header>,
+{
+    pub chunks: Vec<Chunk<B, J>>,
+    pub block_to_request: Option<BlockIdFor<J>>,
```
timorl, line 69: Eh, these seem to be exclusive, so this should be an `enum`, no?

timorl (changes requested):

finality-aleph/src/sync/handler/request_handler.rs
```diff
@@ -0,0 +1,423 @@
...
+        let mut pre_chunk = PreChunk::new();
+        let mut head_chunk = from;
+        let mut state = state;
+
+        loop {
+            let header = match &head_chunk {
+                HeadOfChunk::Justification(j, b, header) => {
+                    pre_chunk.add_block(b.clone());
+                    pre_chunk.add_header(header.clone());
+                    pre_chunk.add_justification(Some(j.clone()));
```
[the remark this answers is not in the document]
kostekIV: here header can be None so we can always use `add_header(j.header())`

lesniak43 (approved):

finality-aleph/src/sync/handler/request_handler.rs
```diff
@@ -0,0 +1,423 @@
...
+            state = new_state;
+            head = new_head;
+        }
+
+        Ok(chunks)
+    }
+
+    pub fn response(self, request: Request<J>) -> HandlerResult<Response<B, J>, CS::Error> {
+        if !request.is_valid() {
+            return Err(RequestHandlerError::BadRequest);
```
lesniak43, line 370: we don't need this check, right? maybe remove it completely?

lesniak43 (approved):

finality-aleph/src/sync/handler/request_handler.rs
```diff
@@ -0,0 +1,423 @@
...
+
+            chunks.extend(chunks_from_step);
+            state = new_state;
+            head = new_head;
+        }
+
+        Ok(chunks)
+    }
+
+    pub fn response(self, request: Request<J>) -> HandlerResult<Response<B, J>, CS::Error> {
```
lesniak43, line 368: I wouldn't call it `response`, as it also contains info about requesting a block. I'd also rename `Response` to sth. like `RequestHandleResult`

finality-aleph/src/sync/handler/request_handler.rs
lesniak43: I love that it works, I hate the reasons...

lesniak43 (approved):

finality-aleph/src/sync/handler/request_handler.rs
```diff
@@ -0,0 +1,423 @@
...
+
+    fn chunks(
+        self,
+        head: HeadOfChunk<B, J>,
+        branch_knowledge: BranchKnowledge<J>,
+        to: BlockIdFor<J>,
+    ) -> HandlerResult<Vec<Chunk<B, J>>, CS::Error> {
+        let mut chunks = vec![];
+
+        let mut head = head;
```
lesniak43, line 353: add `mut` to the signature?

finality-aleph/src/sync/handler/request_handler.rs
```diff
@@ -0,0 +1,423 @@
...
+        state: State,
+        from: HeadOfChunk<B, J>,
+        to: &BlockIdFor<J>,
+        branch_knowledge: &BranchKnowledge<J>,
+    ) -> HandlerResult<StepResult<B, J>, CS::Error> {
+        let mut pre_chunk = PreChunk::new();
+        let mut head_chunk = from;
+        let mut state = state;
+
+        loop {
```
lesniak43, line 275:
```
    fn step(
        &self,
        mut state: State,
        mut head_chunk: HeadOfChunk<B, J>,
        to: &BlockIdFor<J>,
        branch_knowledge: &BranchKnowledge<J>,
    ) -> HandlerResult<StepResult<B, J>, CS::Error> {
        let mut pre_chunk = PreChunk::new();

        loop {

```

finality-aleph/src/sync/handler/request_handler.rs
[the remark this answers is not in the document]
timorl: The point is you don't want to add it at all – you don't want to send it, as it's already contained within the justification.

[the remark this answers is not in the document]
kostekIV: oh you are right, I somehow missed that :P

finality-aleph/src/sync/handler/request_handler.rs
```diff
@@ -0,0 +1,423 @@
...
+{
+    let mut blocks = vec![];
+    let mut headers = vec![];
+    let mut justifications = vec![];
+
+    for chunk in chunks {
+        match chunk {
+            Chunk::Blocks(mut bs) => {
+                bs.extend(blocks);
+                blocks = bs;
```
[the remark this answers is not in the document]
kostekIV: fixed

finality-aleph/src/sync/handler/request_handler.rs
```diff
@@ -0,0 +1,423 @@
...
+            RequestHandlerError::BadState => write!(f, "bad state"),
+            RequestHandlerError::BadRequest => write!(f, "bad request"),
+            RequestHandlerError::ChainStatusBadState => write!(f, "bad state of chain_status"),
+        }
+    }
+}
+
+type HandlerResult<T, Error> = Result<T, RequestHandlerError<Error>>;
+type JustificationBlockAndHeader<J, B> =
+    (Option<J>, Option<B>, Option<<J as Justification>::Header>);
```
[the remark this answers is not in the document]
kostekIV: these types  are no longer here

finality-aleph/src/sync/mod.rs
```diff
@@ -158,6 +158,9 @@ where
     /// Export a copy of the block.
     fn block(&self, id: BlockIdFor<J>) -> Result<Option<B>, Self::Error>;
 
+    /// Export a copy of the header.
+    fn header(&self, id: BlockIdFor<J>) -> Result<Option<J::Header>, Self::Error>;
```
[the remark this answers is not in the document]
kostekIV: header removed

finality-aleph/src/sync/handler/request_handler.rs
```diff
@@ -0,0 +1,423 @@
...
+                    pre_chunk.add_block(b.clone());
+                    pre_chunk.add_header(header.clone());
+                    pre_chunk.add_justification(Some(j.clone()));
+                    self.parent_header(j.header())?
+                }
+                HeadOfChunk::Header(header) => header.clone(),
+            };
+
+            state = match branch_knowledge {
+                BranchKnowledge::TopImported(id) if *id == header.id() => State::OnlyJustification,
```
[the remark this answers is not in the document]
kostekIV: removed

finality-aleph/src/sync/handler/request_handler.rs
```diff
@@ -0,0 +1,423 @@
...
+
+        loop {
+            let header = match &head_chunk {
+                HeadOfChunk::Justification(j, b, header) => {
+                    pre_chunk.add_block(b.clone());
+                    pre_chunk.add_header(header.clone());
+                    pre_chunk.add_justification(Some(j.clone()));
+                    self.parent_header(j.header())?
+                }
+                HeadOfChunk::Header(header) => header.clone(),
```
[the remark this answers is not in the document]
kostekIV: no longer here

finality-aleph/src/sync/handler/request_handler.rs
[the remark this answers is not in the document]
kostekIV: ok, now htere is no interior of the loop, everything is in the condition :D

finality-aleph/src/sync/handler/request_handler.rs
```diff
@@ -0,0 +1,423 @@
...
+    fn step(
+        &self,
+        state: State,
+        from: HeadOfChunk<B, J>,
+        to: &BlockIdFor<J>,
+        branch_knowledge: &BranchKnowledge<J>,
+    ) -> HandlerResult<StepResult<B, J>, CS::Error> {
+        let mut pre_chunk = PreChunk::new();
+        let mut head_chunk = from;
+        let mut state = state;
```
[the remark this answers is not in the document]
kostekIV: done

finality-aleph/src/sync/handler/request_handler.rs
kostekIV: done

finality-aleph/src/sync/handler/request_handler.rs
kostekIV: now we return Action for 3 possibilities.

finality-aleph/src/sync/handler/request_handler.rs
kostekIV: done, by StepResult logic

finality-aleph/src/sync/handler/request_handler.rs
kostekIV: now this variant throws error which should not ever be returned

finality-aleph/src/sync/handler/request_handler.rs
kostekIV: done

finality-aleph/src/sync/handler/request_handler.rs
kostekIV: done

finality-aleph/src/sync/handler/request_handler.rs
kostekIV: done

finality-aleph/src/sync/handler/request_handler.rs
kostekIV: now its call action!

finality-aleph/src/sync/handler/request_handler.rs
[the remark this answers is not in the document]
kostekIV: done

ggawryal (approved):
Gonna trust you that timorl's requests were addressed :p

PR #1295, opened 2023-07-18
A0-2770: Handler unit tests, part one

lesniak43:
# Description

Preparing code for an upcoming series of unit tests:
* check session number in mock `Verifier` - merge with mock `Backend`, as it knows the current session
* allow `MockBlock` to be "incorrect"
* `Forest` should be deep enough
* don't expose `Forest` in `Handler` API
* update `Handler` docs

## Type of change

# Checklist:


finality-aleph/src/sync/forest/mod.rs
```diff
@@ -133,6 +134,11 @@ impl<I: PeerId, J: Justification> VertexWithChildren<I, J> {
 // How deep can the forest be, vaguely based on two sessions ahead, which is the most we expect to
 // ever need worst case scenario.
 const MAX_DEPTH: u32 = 1800;
+#[allow(clippy::assertions_on_constants)]
```
kostekIV, line 137: you can use https://docs.rs/static_assertions/latest/static_assertions/macro.const_assert.html here

finality-aleph/src/sync/mock/backend.rs
```diff
@@ -352,3 +361,42 @@ impl ChainStatus<MockBlock, MockJustification> for Backend {
...
+        let current_session = storage
+            .session_boundary_info
+            .session_id_from_block_num(top_number);
+        let justification_session = storage
+            .session_boundary_info
+            .session_id_from_block_num(justification.id().number);
+        if !(justification_session == current_session
+            || justification_session == current_session.next())
+        {
+            return Err(Self::Error::IncorrectSession);
```
kostekIV, line 395: Why cant you just verify if the justification is correct for older sessions?

finality-aleph/src/sync/forest/mod.rs
kostekIV: I meant it for the line below, not sure if this will help with the clippy :P

finality-aleph/src/sync/mock/backend.rs
kostekIV: I think it might be too restrictive in test scenarios. But some wiggle room with cache size would be better here :+1:

lesniak43: but test scenarios should reflect the real case...

kostekIV: Lets tie this to the `VERIFIER_CACHE_SIZE ` :)

finality-aleph/src/sync/mock/backend.rs
```diff
@@ -254,7 +233,37 @@ impl Finalizer<MockJustification> for Backend {
...
+
+        if storage.blockchain.contains_key(&block.id())
+            || !storage.blockchain.contains_key(&parent_id)
+            || block.id().number() != parent_id.number() + 1
+        {
+            return;
+        }
+
+        if block.id().number() > storage.best_block.number()
+            && is_predecessor(
```
woocash2, line 255: Does it mean that the sequence `best_block(t)` where `t` is time forms a chain? It's not true in the general case right? It's possible that the next best block will appear on a different branch. Is it just in these tests that  we assume such behavior?

finality-aleph/src/sync/mock/backend.rs
```diff
@@ -352,3 +361,42 @@ impl ChainStatus<MockBlock, MockJustification> for Backend {
...
+    ) -> Result<MockJustification, Self::Error> {
+        let top_number = self.top_finalized().unwrap().header.id.number;
+        let storage = self.inner.lock();
+        let current_session = storage
+            .session_boundary_info
+            .session_id_from_block_num(top_number);
+        let justification_session = storage
+            .session_boundary_info
+            .session_id_from_block_num(justification.id().number);
+        if !(justification_session == current_session
```
woocash2, line 392: Personal preference: can we apply De Morgan's here :stuck_out_tongue:?  Would be more immediate to understand this condition imho.

finality-aleph/src/sync/mock/backend.rs
kostekIV: good catch, I guess before we really define what the best_block means to us this does not matter :P

finality-aleph/src/sync/mock/backend.rs
```diff
@@ -254,7 +233,37 @@ impl Finalizer<MockJustification> for Backend {
...
+        }
+
+        let mut storage = self.inner.lock();
+
+        let parent_id = match block.parent_id() {
+            Some(id) => id,
+            None => return,
+        };
+
+        if storage.blockchain.contains_key(&block.id())
```
kostekIV, line 247: we could add a check that the block being imported is not below the last finalized

finality-aleph/src/sync/forest/mod.rs
lesniak43: we don't use this crate in `finality-aleph`, do we want to add it just for this?

kostekIV: up to you, imho I think its nice addition and we could start using it in similar cases to this one. At the same time your approach is also fine

lesniak43: ok, then let's wait for the second case :P

finality-aleph/src/sync/mock/backend.rs
lesniak43: done

finality-aleph/src/sync/mock/backend.rs
lesniak43: I've removed the whole best block logic, now the trait method just returns an error

we will reintroduce proper implementation when we know what exactly we want to mock...

finality-aleph/src/sync/mock/backend.rs
lesniak43: done

finality-aleph/src/sync/mock/backend.rs
```diff
@@ -95,41 +81,40 @@ impl Backend {
...
+                }
+                chunk.insert(id.clone());
+                id = storage
+                    .blockchain
+                    .get(&id)
+                    .expect("should be there")
+                    .header()
+                    .parent_id()
+                    .expect("genesis should be marked");
+            }
```
kostekIV, line 113: I think this is overcomplicated. The easy way to do this is:
```rust
        let last_finalized = self.top_finalized()?;
        let mut storage = self.inner.lock();

        let finalized = &storage.finalized;
        let mut to_prune = vec![];
        for id in storage.blockchain.keys() {
            if id.number <= last_finalized.header.id.number && !finalized.contains(id) {
                to_prune.push(id)
            }
        }
```

lesniak43: this can leave some partially pruned branches hanging in the air (literally...)

kostekIV: The hanging branches will be cleared later

lesniak43: Ok, the question is how does the Substrate db behave in such situation. I was under the impression that the forks should become inaccessible immediately.

kostekIV: hmm then we can do smth like this:
```rust
        let last_finalized = self.top_finalized()?;
        let mut storage = self.inner.lock();

        let finalized = &storage.finalized;
        let mut to_prune = vec![];
        for id in storage.blockchain.keys() {
            if id.number <= last_finalized.header.id.number && !finalized.contains(id) {
                to_prune.push(id);
            } else if  id.number > last_finalized.header.id.number && !is_predecessor(id, last_finalized) {
               to_prune.push(id)
           }
        }
```
its not the most efficient way to do it but its more readable ;)

finality-aleph/src/sync/mock/backend.rs
kostekIV: check lgtm

finality-aleph/src/sync/mock/backend.rs
lesniak43: fixed, but you'll pay the electricity bill...

kostekIV (approved):

finality-aleph/src/sync/mock/backend.rs
woocash2: I think even this can be simplified to:
```rust
if !finalized.contains(id) && !is_predecessor(id, last_finalized) {
    to_prune.push(id)
}
```
right?

lesniak43: yup, fixed
I've even used `filter`!

woocash2 (approved):

PR #1296, opened 2023-07-19
A0-2849: Moved aleph production artifacts to other bucket

Marcin-Radecki:
# Description

Those S3 artifacts are used by CI only and they can be kept in one bucket.

## Type of change


Marcin-Radecki:
This is no longer required

PR #1300, opened 2023-07-24
A0-2835: Slice large msgs into smaller ones 

kostekIV:
# Description

## Type of change

# Checklist:

- I have added tests


finality-aleph/src/sync/service.rs
```diff
@@ -293,7 +294,9 @@ where
         );
         match self.handler.handle_request(request) {
             Ok(Action::Response(response_items)) => {
-                self.send_to(NetworkData::RequestResponse(response_items), peer);
+                for chunk in MsgLimiter::new(&response_items) {
+                    self.send_to(NetworkData::RequestResponse(chunk.to_vec()), peer.clone());
```
deuszx, line 298: Does it correctly handle lost parts and requests them when needed? How does sub-part verification look like? i.e. can a malicious node spam the receiver by sending messages claiming to be a subpart of a bigger message?

kostekIV: not really applicable here. We just chunk large msg into smaller ones where the prefix is understandable by the receiver. The spam here would behave exactly the same as without it. We also do not need the sub-part verification

lesniak43: the idea is that we hope the chunks will be delivered in order - if not, the receiver will process as many as possible, and then send another request

lesniak43: the message we're talking about is a vector of "items", and we verify each item separately

e.g. if there were 5 chunks 100 items each, and the third chunk gets lost, we then request the missing 300 items, because we were able to import 2x100 items

we do not check if a message was requested, we simply process it

we also don't have any spam protection right now

kostekIV:
I know. I answered your question to this specific situation in which we do
not care about all those things you listed above

pon., 24 lip 2023, 14:39 użytkownik deuszx ***@***.***>
napisał:

finality-aleph/src/sync/message_limiter.rs
```diff
@@ -0,0 +1,181 @@
+use parity_scale_codec::Encode;
+
+use crate::sync::data::MAX_SYNC_MESSAGE_SIZE;
+
+pub const MSG_BYTES_LIMIT: usize = MAX_SYNC_MESSAGE_SIZE as usize;
```
lesniak43, line 5: do we need two separate consts?

finality-aleph/src/sync/message_limiter.rs
```diff
@@ -0,0 +1,181 @@
...
+
+        let idx = indexes.partition_point(|&idx| {
+            let encoded_size = self.msg[self.start_index..idx].encoded_size();
+            encoded_size <= LIMIT
+        }) - 1; // minus 1 since this is the first index where encoded size is larger than LIMIT
+
+        let idx = idx + self.start_index;
+
+        if idx == self.start_index {
+            None
```
lesniak43, line 47: this means that a single item is larger than the message size limit, right? if so, this should definitely be a critical error

lesniak43 (approved):

finality-aleph/src/sync/message_limiter.rs
lesniak43: I think we need to set the limits properly this time... Please check if e.g. this is accurate: https://substrate.stackexchange.com/questions/6999/what-is-the-maximum-size-of-a-block

If possible, let's import relevant consts from Substrate, and use them only to introduce some static assertions (we define the limits, and the compiler tells us if they decide to change sth - this is imho the best approach).

kostekIV: We could use only one constant :)

finality-aleph/src/sync/message_limiter.rs
kostekIV: imho we could even just panic here as it would be really bad if one item is larger than the message size limit as you have pointed it out

lesniak43: I'd say we return an error, catch and display it somewhere higher as "critical", and hope that this particular block won't get finalized and we'll prune it soon...

finality-aleph/src/sync/message_limiter.rs
kostekIV: I left this here because it serve as convenient place for cast to usize

finality-aleph/src/sync/message_limiter.rs
kostekIV: ok, done

finality-aleph/src/sync/message_limiter.rs
lesniak43: terrible :D

finality-aleph/src/sync/message_limiter.rs
```diff
@@ -0,0 +1,188 @@
...
+        self.start_index = end_idx;
+
+        Ok(Some(&self.msg[start_index..end_idx]))
+    }
+
+    fn bs_by_encode(&self) -> Result<usize, Error> {
+        let indexes = &self.indexes[self.start_index..];
+
+        let idx = indexes.partition_point(|&idx| {
+            let encoded_size = self.msg[self.start_index..idx].encoded_size();
```
fixxxedpoint, line 55: Does it `panic` when called one more time after returning last element? Maybe we can make it keep returning `None` or move back to previous `Iterator` implementation.

finality-aleph/src/sync/service.rs
```diff
@@ -293,7 +294,22 @@ where
         );
         match self.handler.handle_request(request) {
             Ok(Action::Response(response_items)) => {
-                self.send_to(NetworkData::RequestResponse(response_items), peer);
+                let mut limiter = MsgLimiter::new(&response_items);
```
fixxxedpoint, line 297: What do you think about following the pattern of `VersionedWrapper` and making it implementation detail of the `Network` implementation?

fixxxedpoint (changes requested):

finality-aleph/src/sync/message_limiter.rs
kostekIV: it keeps returning None or Err(e)

finality-aleph/src/sync/service.rs
kostekIV: Im not sure what you exactly mean here

finality-aleph/src/sync/message_limiter.rs
```diff
@@ -0,0 +1,188 @@
...
+
+impl Display for Error {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Error::ItemTooBig => write!(f, "Single item takes more than the limit"),
+        }
+    }
+}
+
+impl<'a, D: Encode, const LIMIT: usize> Limiter<'a, D, LIMIT> {
```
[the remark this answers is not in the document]
kostekIV: for me this is fine, we would need to remember aboutadditional byte (or 2?) needed for encoding the vec size and I probably wouldnt mutate the msg but overall I like this approach much more. @lesniak43 wdyt, bs was because we wanted to clever and have no assumption about encoding but it looks like this is much better option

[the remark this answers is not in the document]
kostekIV: done

finality-aleph/src/sync/service.rs
kostekIV: not done, as discussed offline :P

finality-aleph/src/sync/message_limiter.rs
```diff
@@ -32,31 +31,32 @@ impl<'a, D: Encode, const LIMIT: usize> Limiter<'a, D, LIMIT> {
...
+    fn find_idx_of_largest_prefix(&self) -> Result<usize, Error> {
+        let mut idx = self.start_index;
+        let mut encoded_sum = 0;
+        while idx < self.msg.len() && encoded_sum <= LIMIT {
+            encoded_sum += self.msg[idx].encoded_size();
+            idx += 1;
+        }
 
-        let idx = idx + self.start_index;
+        while idx > self.start_index && self.msg[self.start_index..idx].encoded_size() > LIMIT {
```
fixxxedpoint, line 57: Add a short comment describing this estimation.

fixxxedpoint (changes requested):

finality-aleph/src/sync/message_limiter.rs
kostekIV: done

fixxxedpoint (approved):

PR #1301, opened 2023-07-24
Local test for block sync in presence of reorgs

woocash2:
# Description
A test which ensures that block sync works in presence of forks.
The test implements the following scenario:
1. split validators into two halves: `X, Y` and let them produce blocks for a while
2. kill `X`, let `Y` produce blocks for a while
3. kill `Y`, recover `X` and let it produce blocks for a while
4. recover `Y`, both `X` and `Y` should discover a fork and one of them should perform a reorg

## Type of change

local test

### DIFF
--- a/local-tests/chainrunner/chain.py
+++ b/local-tests/chainrunner/chain.py
@@ -169,13 +169,25 @@ class Chain:
         chainspec = join(self.path, 'chainspec.json')
         snapshot = snapshot_file if snapshot_file else join(self.path, 'snapshot.json')
         cmd = [check_file(forkoff_path), '--ws-rpc-endpoint', ws_endpoint,
-                '--initial-spec-path', chainspec,
-                '--snapshot-path', snapshot,
-                '--combined-spec-path', forked]
+               '--initial-spec-path', chainspec,
+               '--snapshot-path', snapshot,
+               '--combined-spec-path', forked]
         if snapshot_file:
             cmd.append('--use-snapshot-file')
         subprocess.run(cmd, check=True)
         self.set_chainspec(forked)
+
+    def get_highest_imported(self, nodes=None):
+        """Return the highest block id imported by all the selected nodes"""
+        nodes = nodes or range(len(self.nodes))
+        nodes = [self.nodes[i] for i in nodes]
+        return min([n.highest_block()[0] for n in nodes], default=-1)
+
+    def get_highest_finalized(self, nodes=None):
+        """Return the highest block id finalized by all the selected nodes"""
+        nodes = nodes or range(len(self.nodes))
+        nodes = [self.nodes[i] for i in nodes]
+        return min([n.highest_block()[1] for n in nodes], default=-1)
 
     def wait_for_finalization(self, old_finalized, nodes=None, timeout=600, finalized_delta=3, catchup=True, catchup_delta=10):
         """Wait for finalization to catch up with the newest blocks. Requires providing the number
@@ -185,12 +197,13 @@ class Chain:
         If `catchup` is True, wait until finalization catches up with the newly produced blocks
         (within `catchup_delta` blocks). 'timeout' (in seconds) is a global timeout for the whole method
         to execute. Raise TimeoutError if finalization fails to recover within the given timeout."""
-        nodes = [self.nodes[i] for i in nodes] if nodes else self.nodes
+        nodes = nodes or range(len(self.nodes))
         deadline = time.time() + timeout
-        while any((n.highest_block()[1] <= old_finalized + finalized_delta) for n in nodes):
+        while self.get_highest_finalized(nodes) <= old_finalized + finalized_delta:
             time.sleep(5)
             if time.time() > deadline:
                 raise TimeoutError(f'Block finalization stalled after {timeout} seconds')
+        nodes = [self.nodes[i] for i in nodes]
         if catchup:
             def lags(node):
                 r, f = node.highest_block()
@@ -200,6 +213,15 @@ class Chain:
                 if time.time() > deadline:
                     print(f'Finalization restored, but failed to catch up with recent blocks within {timeout} seconds')
                     break
+
+    def wait_for_block_id_imported(self, block_id, nodes=None, timeout=600):
+        """Wait until all the selected nodes imported a block of id `block_id`"""
+        nodes = nodes or range(len(self.nodes))
+        deadline = time.time() + timeout
+        while self.get_highest_imported(nodes) < block_id:
+            time.sleep(1)
+            if time.time() > deadline:
+                raise TimeoutError(f'Block production stalled after {timeout} seconds')
 
     def wait_for_authorities(self, nodes=None, timeout=600):
         """Wait for the selected `nodes` (all validator nodes if None) to connect to all known authorities.
--- a/local-tests/test_force_reorg.py
+++ b/local-tests/test_force_reorg.py
@@ -1,0 +1,68 @@
+#!/bin/env python
+import os
+from os.path import abspath, join
+from time import ctime
+from chainrunner import Chain, Seq, generate_keys
+
+
+def printt(s): print(ctime() + ' | ' + s)
+
+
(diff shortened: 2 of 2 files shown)


local-tests/chainrunner/chain.py
```diff
@@ -169,14 +169,26 @@ def fork(self, forkoff_path, ws_endpoint, snapshot_file=None):
...
+               '--initial-spec-path', chainspec,
+               '--snapshot-path', snapshot,
+               '--combined-spec-path', forked]
         if snapshot_file:
             cmd.append('--use-snapshot-file')
         subprocess.run(cmd, check=True)
         self.set_chainspec(forked)
 
+    def get_highest_imported(self, nodes=None):
+        """Return the highest block id imported by all the selected nodes"""
```
lesniak43, line 181: let's not say `id` if it's only height - two nodes may import different blocks of the same height, and this function will still return something

local-tests/chainrunner/chain.py
```diff
@@ -169,14 +169,26 @@ def fork(self, forkoff_path, ws_endpoint, snapshot_file=None):
...
+        nodes = nodes or range(len(self.nodes))
```
lesniak43, line 182: `nodes = range(len(self.nodes)) if nodes is None else nodes`

check what happens if you pass `nodes=[]`

local-tests/chainrunner/chain.py
```diff
@@ -169,14 +169,26 @@ def fork(self, forkoff_path, ws_endpoint, snapshot_file=None):
...
+        nodes = [self.nodes[i] for i in nodes]
+        return min([n.highest_block()[0] for n in nodes], default=-1)
+
+    def get_highest_finalized(self, nodes=None):
+        """Return the highest block id finalized by all the selected nodes"""
+        nodes = nodes or range(len(self.nodes))
```
lesniak43, line 188: same as above

local-tests/chainrunner/chain.py
```diff
@@ -185,12 +197,13 @@ def wait_for_finalization(self, old_finalized, nodes=None, timeout=600, finalize
         If `catchup` is True, wait until finalization catches up with the newly produced blocks
         (within `catchup_delta` blocks). 'timeout' (in seconds) is a global timeout for the whole method
         to execute. Raise TimeoutError if finalization fails to recover within the given timeout."""
-        nodes = [self.nodes[i] for i in nodes] if nodes else self.nodes
+        nodes = nodes or range(len(self.nodes))
```
lesniak43, line 200: same as above

local-tests/chainrunner/chain.py
```diff
@@ -201,6 +214,15 @@ def lags(node):
                     print(f'Finalization restored, but failed to catch up with recent blocks within {timeout} seconds')
                     break
 
+    def wait_for_block_id_imported(self, block_id, nodes=None, timeout=600):
+        """Wait until all the selected nodes imported a block of id `block_id`"""
+        nodes = nodes or range(len(self.nodes))
```
lesniak43, line 219: same as above

local-tests/chainrunner/chain.py
```diff
@@ -201,6 +214,15 @@ def lags(node):
                     print(f'Finalization restored, but failed to catch up with recent blocks within {timeout} seconds')
                     break
 
+    def wait_for_block_id_imported(self, block_id, nodes=None, timeout=600):
+        """Wait until all the selected nodes imported a block of id `block_id`"""
```
lesniak43, line 218: same as above

local-tests/chainrunner/chain.py
```diff
@@ -201,6 +214,15 @@ def lags(node):
                     print(f'Finalization restored, but failed to catch up with recent blocks within {timeout} seconds')
                     break
 
+    def wait_for_block_id_imported(self, block_id, nodes=None, timeout=600):
```
lesniak43, line 217: `s/id/height`

local-tests/chainrunner/chain.py
woocash2: We return `-1`. If we strictly interpret the name of the function, it should return "the maximum `k` such that for each `x` in `nodes` `x` imported `k`" and I agree that it's a bit misleading as such `k` doesn't exist. But I find returning `-1` as an acceptable default behavior.

local-tests/chainrunner/chain.py
```diff
@@ -169,14 +169,26 @@ def fork(self, forkoff_path, ws_endpoint, snapshot_file=None):
...
         self.set_chainspec(forked)
 
+    def get_highest_imported(self, nodes=None):
+        """Return the highest block id imported by all the selected nodes"""
+        nodes = nodes or range(len(self.nodes))
+        nodes = [self.nodes[i] for i in nodes]
+        return min([n.highest_block()[0] for n in nodes], default=-1)
+
+    def get_highest_finalized(self, nodes=None):
+        """Return the highest block id finalized by all the selected nodes"""
```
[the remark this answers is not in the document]
woocash2: For finalization I think it's ok as we can't have two finalized blocks at the same height

local-tests/chainrunner/chain.py
lesniak43: check again :P

lesniak43: (And I like the idea of returning `-1`, cause it's somewhat consistent with the convention that max of an empty set is minus infinity - here `-1` will be smaller than any possible block height, so you conveniently keep this property of `-inf`. Returning `None` would also be fine.)

woocash2: a yeah, the full range, tricky

woocash2: Actually after some thought I'd rather convert both `None` and the empty list to the full range and describe it in the docstring. I would really want `None` to get converted, and treating `[]` differently here seems unnecessary to me here.

local-tests/chainrunner/chain.py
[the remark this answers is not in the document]
woocash2: ok, let me change it to height

local-tests/chainrunner/chain.py
lesniak43: `None` gets converted because it's a python convention of providing default value if you can't do this in the method signature.

I don't think converting `[]` to full range is a good idea, as I don't see the use case other than "by accident". If you dislike the idea of returning `-1`, then I'd suggest either returning `None`, or raising an exception.

woocash2: ok, let it be -1

h4nsu (approved):

local-tests/test_force_reorg.py
```diff
@@ -0,0 +1,68 @@
...
+                rpc_port=Seq(9933),
+                unit_creation_delay=200,
+                execution='Native')
+
+addresses = [n.address() for n in chain]
+validator_addresses = [n.validator_address() for n in chain]
+chain.set_flags(bootnodes=addresses[1])
+chain.set_flags_validator(public_addr=addresses, public_validator_addresses=validator_addresses)
+
+BLOCKS_PER_STAGE = 50
```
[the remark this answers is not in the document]
woocash2: I'll do that manually and leave 50 here as it is

local-tests/test_force_reorg.py
```diff
@@ -0,0 +1,68 @@
...
+
+printt('Starting nodes: ' + ' '.join([str(n) for n in part2]))
+chain.start('aleph', nodes=part2)
+chain.wait_for_finalization(0, nodes=part2)
+f2 = chain.get_highest_finalized(nodes=part2)
+chain.wait_for_block_id_imported(f2 + BLOCKS_PER_STAGE, nodes=part2)
+
+printt('Starting nodes: ' + ' '.join([str(n) for n in part1]))
+chain.start('aleph', nodes=part1)
+f = chain.get_highest_finalized()
```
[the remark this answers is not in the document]
woocash2: removed

h4nsu (approved):
G2G

PR #1302, opened 2023-07-24
A0-2864: Obrok/protobridge

fbielejec:
# Description

This PR add the light client pallets

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -412,6 +412,9 @@ name = "aleph-runtime"
 version = "0.9.1"
 dependencies = [
  "baby-liminal-extension",
+ "bp-parachains",
+ "bp-runtime",
+ "bp-westend",
  "frame-benchmarking",
  "frame-executive",
  "frame-support 4.0.0-dev (git+https://github.com/Cardinal-Cryptography/substrate.git?branch=aleph-v0.9.41)",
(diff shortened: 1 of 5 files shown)


Cargo.lock
```diff
@@ -423,6 +426,8 @@ dependencies = [
  "pallet-authorship",
  "pallet-baby-liminal",
  "pallet-balances",
+ "pallet-bridge-grandpa",
+ "pallet-bridge-parachains",
```
Marcin-Radecki, line 430: How it's possible `Cargo.lock` does not contain amy changes to `wasmtime` 8.X package? As this https://github.com/Cardinal-Cryptography/parity-bridges-common/commit/125c099f7a1e915b126bc24778240446f65d571e bumps uses different branch substrate fork branch. In general `cargo` should resolve `wasmtime` correctly to 8.X in the end (e.g. in C++ world this was undefined behaviour, ie if lib A uses dep D in different version than lib B, then if you compile binary with lib A and B can lead to nasty bugs). I;m thinking in general whether this is good approach to mix different in `aleph-node` substrate forks on different branches. Can we use `obrok/aleph-v0.9.41-wasmtime8` explicetely in ``aleph-node` to not hide this dependency?

PR #1304, opened 2023-07-25
A0-2753: Test add pr ci

woocash2:
# Description

## Type of change

# Checklist:

- I have added tests


.github/workflows/on-pull-request-commit.yml
```diff
@@ -32,10 +32,44 @@ jobs:
...
+    uses: ./.github/workflows/_check-vars-and-secrets.yml
+    secrets: inherit
+
   build-production-node-and-runtime:
-    name: Build production node and runtime
+    needs: [check-vars-and-secrets]
+    name: Build production node and runtime artifacts (PR version)
     uses: ./.github/workflows/_build-production-node-and-runtime.yml
 
+  test-force-reorg:
```
Marcin-Radecki, line 45: If it's the same test (ie same logic and input params, then it can be either in nightly or in PR pipeline, not both, to not waste computational power. I'd favor nightly as the timeout is 10 minutes. We should keep PR pipeline each e2e test less than 10 minutes for now, and I think we should go even more like 5 minute limit.

PR #1323, opened 2023-08-02
A0-2942: Remove dead statuses from README

pmikolajczyk41:
# Description

Since we are no longer running unit/e2e tests on merge to the `main` branch (apart from button-related, but this is temporary afaik), we should get rid of invalid links from `README.md` (and thus repo's main page).

## Type of change


Marcin-Radecki (approved):
In future, we should bring those statuses back but for **nightly** build.

PR #1324, opened 2023-08-02
A0-1846: Removed migrations from release-11 branch

Marcin-Radecki:
# Description

Remove migrations that were introduced in `r-11.3` release.

### DIFF
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -34,7 +34,7 @@ use frame_system::{EnsureRoot, EnsureSignedBy};
 #[cfg(feature = "try-runtime")]
 use frame_try_runtime::UpgradeCheckSelect;
 pub use pallet_balances::Call as BalancesCall;
-use pallet_committee_management::{PrefixMigration, SessionAndEraManager};
+use pallet_committee_management::SessionAndEraManager;
 pub use pallet_timestamp::Call as TimestampCall;
 use pallet_transaction_payment::{CurrencyAdapter, Multiplier, TargetedFeeAdjustment};
 pub use primitives::Balance;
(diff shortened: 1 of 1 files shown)


kostekIV (commented):
LGTM, there are changes in the scripts/synthetic-network that are not related to this.

pallets/elections/src/migration.rs
```diff
@@ -1,162 +0,0 @@
...
-    log,
-    pallet_prelude::{StorageVersion, TypeInfo},
-    traits::OnRuntimeUpgrade,
-};
-use primitives::CommitteeSeats;
-use sp_core::Get;
-#[cfg(feature = "try-runtime")]
-use {frame_support::ensure, pallets_support::ensure_storage_version, sp_std::vec::Vec};
-
-use crate::{CommitteeSize, Config, NextEraCommitteeSize, Pallet, LOG_TARGET};
```
DamianStraszak, line 13: I don't think we should remove migrations from our code. We should just detach them from the runtime...

DamianStraszak (changes requested):

pallets/elections/src/migration.rs
Marcin-Radecki: That also I thought, yet I don't see now in `aleph-node` repo any other migrations, hance I thought to follow that rule and remove them. Let me check what substrate does with that

Marcin-Radecki:
> LGTM, there are changes in the scripts/synthetic-network that are not related to the migrations tho.

True, will revert

pallets/elections/src/migration.rs
DamianStraszak: Yes, the reason is we had this pallet hook based migrations before, which were removed to clean things up. But it seems the convention is to keep migrations, and just attach them when needed.

kostekIV: I have removed them quite some time ago. We have them in the git history if they will be ever needed (?). I have removed them because maintaining the old migrations is kinda meaningless and makes it harder to develop pallets.

Marcin-Radecki: Actually, in Substrate they indeed keep migrations. I'll will bring them back then

DamianStraszak (approved):

kostekIV (approved):

PR #1333, opened 2023-08-07
A0-2770: Handler unit tests part 4

lesniak43:
# Description

Testing finalizing with justification gaps.

## Type of change

# Checklist:

- I have added tests


finality-aleph/src/sync/mod.rs
```diff
@@ -78,7 +78,7 @@ pub trait Justification: Clone + Send + Sync + Debug + 'static {
 
 /// A verifier of justifications.
 pub trait Verifier<J: Justification> {
-    type Error: Display;
+    type Error: Debug + Display;
```
kostekIV, line 81: hmmmmmm, I think we want this to be only display to disallow using expects and unwraps in production code :P

finality-aleph/src/sync/handler/mod.rs
```diff
@@ -708,6 +718,137 @@ mod tests {
...
+                .collect(),
+            true,
+            true,
+            false,
+        );
+        let (maybe_id, maybe_error) = handler.handle_request_response(response_items, peer_id);
+        assert!(maybe_id.is_none(), "should not import justification");
+        assert!(maybe_error.is_none(), "should work");
+        let mut counter = 0;
+        while let Ok(BlockImported(header)) = notifier.next().await {
```
timorl, line 797: These `while`s could probably be abstracted into a function like `ensure_branch_imported` (called twice here ofc).

finality-aleph/src/sync/handler/mod.rs
```diff
@@ -708,6 +718,137 @@ mod tests {
...
+    async fn finalizes_with_justification_gaps() {
+        let (mut handler, _backend, mut notifier, genesis) = setup();
+        let mut bottom = genesis;
+        let peer_id = 0;
+        for session in 0.. {
+            let top = SESSION_BOUNDARY_INFO.last_block_of_session(SessionId(session));
+            let branch = grow_light_branch_till(&mut handler, &bottom, &top, peer_id);
+            bottom = branch.last().expect("should not be empty").id();
+            // import blocks
+            let response_items = branch_response(branch.clone(), true, true, false);
```
timorl, line 731: Ugh, three bools as arguments... could you make this into a single `enum` instead? Or, if you really need all possible combinations, a small struct which would make this more readable.

finality-aleph/src/sync/mod.rs
timorl: Eh, I don't feel this is a particularly strong argument. I would also prefer this to not be here, but if the lack makes testing much harder then imo it can stay.

timorl (approved):
Only some readability.

finality-aleph/src/sync/mod.rs
lesniak43: we're our worst enemies...
fixed

kostekIV (approved):

finality-aleph/src/sync/handler/mod.rs
lesniak43: done

finality-aleph/src/sync/handler/mod.rs
lesniak43: struct it is

PR #1334, opened 2023-08-08
A0-2978: Accept more responses in sync Handler

lesniak43:
# Description

`sync::Handler` should not drop request responses containing already imported blocks. This PR changes the import strategy to simply skip old/imported blocks.

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/finality-aleph/src/sync/handler/mod.rs
+++ b/finality-aleph/src/sync/handler/mod.rs
@@ -161,6 +161,7 @@ where
     RequestHandlerError(RequestHandlerError<J, CS::Error>),
     MissingJustification,
     BlockNotImportable,
+    HeaderNotRequired,
 }
 
 impl<B, J, CS, V, F> Display for Error<B, J, CS, V, F>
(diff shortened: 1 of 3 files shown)


kostekIV (approved):

finality-aleph/src/sync/handler/mod.rs
```diff
@@ -741,6 +749,108 @@ mod tests {
...
+        assert!(maybe_id.is_some());
+        assert!(maybe_error.is_none());
+        mark_branch_imported(&mut handler, &mut notifier, &branch).await;
+        let (maybe_id, maybe_error) = handler.handle_request_response(response, 8);
+        assert!(maybe_id.is_none());
+        assert!(maybe_error.is_none());
+    }
+
+    #[tokio::test]
+    async fn accepts_long_response_after_handling_short_one() {
```
timorl, line 791: This name is kinda misleading, you do much more than such a basic test. Perhaps split this test into two (or rather add one that actually does what this name implies and rename this one).

timorl (approved):

finality-aleph/src/sync/handler/mod.rs
lesniak43: true, fixed

PR #1335, opened 2023-08-08
A0-2730: bump subxt to 0.28

ggawryal:
# Description

Bump subxt from 0.25 to 0.28.0 [(Changelog)](https://github.com/paritytech/subxt/blob/master/CHANGELOG.md). 
Subxt is slowly removing the dependency on substrate (making it optional) in order to reduce wasm build size, so there are some breaking changes. Most importantly, a `suxbt::utils::AccountId32` type is introduced, which is basically a stripped `sp_runtime::AccountId32`.

[EDIT 1] We've decided to continue using substrate's AccountId, so in subxt codegen, we had to substitute subxt::utils::AccountId into that type. Unfortunately, this means that we have to wrap it into `subxt::utils::Static<>`. Higher level code, outside `aleph_zero.rs` can, and is returning plain unwrapped AccountId. Subxt also introduced `suxbt::utils::MultiAddress` in place of `sp_runtime::MultiAddress`, but currently switching to the former for us is not causing any compilation errors, so we should be safe to use it. 

Version 0.28 is the highest compatible with our substrate's version (next one uses newer `wasmtime`).

I've run only unit and e2e tests in aleph-node - please let me know if I need to test something else (ui, manual etc).

## Type of change

# Checklist:


aleph-client/src/lib.rs
```diff
@@ -150,3 +157,9 @@ where
 {
     AccountId::from(keypair.public())
 }
+
+/// Converts an SS58Check address string to `AccountId`.
+/// * `ss58check` - &str with SS58Check address
+pub fn account_from_ss58check(ss58check: &str) -> Result<AccountId, PublicError> {
+    sp_runtime::AccountId32::from_ss58check(ss58check).map(|address| address.into())
+}
```
ggawryal, line 165: Helper function, because `subxt::utils::AccountId32` doesn't implement `from_ss58check`

aleph-client/src/pallets/multisig.rs
```diff
@@ -243,6 +247,38 @@ impl ContextState for Ongoing {}
 pub enum Closed {}
 impl ContextState for Closed {}
 
+// Wrapper for AccountId that implements Hash.
```
ggawryal, line 250: Previously we had `pub fn approvers(&self) -> &HashSet<AccountId>`, but now AccountId is not `Hash`, so I've chosen to modify the code this way.

e2e-tests/src/ban.rs
```diff
@@ -1,4 +1,4 @@
-use std::collections::HashSet;
+use std::collections::BTreeSet;
```
ggawryal, line 1: Same problem with `Hash`, but for tests imo we can apply that ugly quick fix - let me know what do you think

aleph-client/src/connections.rs
```diff
@@ -249,16 +252,18 @@ impl<C: AsConnection + Sync> ConnectionApi for C {
...
+    ) -> Option<T> {
         info!(target: "aleph-client", "accessing storage at {}::{} at block {:?}", addrs.pallet_name(), addrs.entry_name(), at);
-        self.as_connection()
-            .as_client()
-            .storage()
-            .fetch(addrs, at)
-            .await
-            .expect("Should access storage")
+
+        let storage = self.as_connection().as_client().storage();
```
ggawryal, line 260: Api change in subxt in 0.28 - storage.at(Option<block_hash>) has changed to `.at(block_hash)` and `.at_latest()`. Should we also change our api.

aleph-client/src/pallets/staking.rs
```diff
@@ -424,13 +419,14 @@ impl<C: AsConnection + Sync> StakingRawApi for C {
...
         let key_addrs = api::storage().staking().eras_stakers_root();
         let mut key = key_addrs.to_root_bytes();
-        StorageMapKey::new(era, StorageHasher::Twox64Concat).to_bytes(&mut key);
-        self.as_connection()
-            .as_client()
-            .storage()
-            .fetch_keys(&key, 10, None, at)
-            .await
-            .map_err(|e| e.into())
+        key.extend(subxt::ext::sp_core::twox_64(&era.encode()));
```
ggawryal, line 422: Perhaps ugly, but this is the way:  https://github.com/paritytech/subxt/blob/43809ee5928a77eed5438f7d61f24fe0183f970a/testing/integration-tests/src/full_client/storage/mod.rs#L78.

kostekIV (changes requested):
There are some problems related to the change of AccountId from substrate one. Pls do:
1. substitute type for substrate one (you may need to use [Static](https://docs.rs/subxt/0.28.0/subxt/utils/struct.Static.html)
2. update readme for codegen cmd to include this
3. In public API return AccountId not Static<AccountId>

aleph-client/Cargo.toml
```diff
@@ -20,7 +20,7 @@ thiserror = "1.0"
 serde_json = { version = "1.0.94" }
 contract-transcode = "2.1.0"
 ink_metadata = { version = "=4.0.1" }
-subxt = "0.25.0"
+subxt = { version = "0.28.0", features = ["substrate-compat"] }
```
pmikolajczyk41, line 23: do we have to be `substrate-compat`? what are we using from this feature?

aleph-client/src/pallets/staking.rs
```diff
@@ -424,13 +419,14 @@ impl<C: AsConnection + Sync> StakingRawApi for C {
...
+        key.extend(&era.encode());
```
pmikolajczyk41, line 423: why do we extend the key with era twice?

aleph-client/src/pallets/staking.rs
```diff
@@ -441,12 +437,14 @@ impl<C: AsConnection + Sync> StakingRawApi for C {
...
-        StorageMapKey::new(era, StorageHasher::Twox64Concat).to_bytes(&mut key);
+        key.extend(subxt::ext::sp_core::twox_64(&era.encode()));
+        key.extend(&era.encode());
         accounts
             .iter()
             .map(|account| {
                 let mut key = key.clone();
-                StorageMapKey::new(account, StorageHasher::Twox64Concat).to_bytes(&mut key);
+                key.extend(subxt::ext::sp_core::twox_64(&account.encode()));
+                key.extend(&account.encode());
```
pmikolajczyk41, line 447: maybe it'd be convenient to create some utility function that takes `key`, `subkey` and performs:
```rust
key.extend(subxt::ext::sp_core::twox_64(&subkey.encode()));
key.extend(&subkey.encode()));
```

e2e-tests/src/ban.rs
pmikolajczyk41: I guess we are safe, modulo problems with repeated validators, which actually should never occur...

aleph-client/src/lib.rs
ggawryal: Removed, no longer needed

aleph-client/src/pallets/multisig.rs
ggawryal: Rolled back as we're using substrate's AccountId instead

e2e-tests/src/ban.rs
ggawryal: Rolled back

aleph-client/src/pallets/staking.rs
ggawryal: If I'm not mistaken, `EraIndex` and `AccountId` should be encoded using `twox64_concat`

ggawryal:
Switched to substrate's `AccountId` as suggested. There is `--subsititute-type` flag in codegen which should be used for that purpose, but unfortunately it seems that it cannot override default substitutions (particularly sp_core::crypto::AccountId -> subxt::utils::AccountId32), so I had to use `sed`. The problem is resolved in [subxt 0.29](https://github.com/paritytech/subxt/pull/936), but to use it we would have to bump substrate by one version.

aleph-client/docker/subxt-integration-entrypoint.sh
```diff
@@ -1,13 +1,14 @@
 #!/usr/bin/env bash
 
-subxt codegen --derive Clone --derive Debug --derive Eq --derive PartialEq | rustfmt --edition=2021 --config-path aleph-node/rustfmt.toml > aleph_zero.rs
+subxt codegen --derive Clone --derive Debug --derive PartialEq --derive Eq \
+  | sed 's/::[ ]*subxt[ ]*::[ ]*utils[ ]*::[ ]*AccountId32/::subxt::utils::Static<::subxt::ext::sp_core::crypto::AccountId32>/g' \
```
pmikolajczyk41, line 4: cli should have `substitute-type` option for that

aleph-client/src/lib.rs
```diff
@@ -68,8 +78,20 @@ pub use connections::{
 };
 
 /// An alias for a configuration of live chain, e.g. block index type, hash type.
-type AlephConfig = PolkadotConfig;
-type ParamsBuilder = subxt::tx::PolkadotExtrinsicParamsBuilder<SubstrateConfig>;
+pub enum AlephConfig {}
+
+impl Config for AlephConfig {
```
pmikolajczyk41, line 83: what is the difference now between `AlephConfig` and `PolkadotConfig`? in `AccountId`?

pmikolajczyk41 (approved):

aleph-client/docker/subxt-integration-entrypoint.sh
ggawryal: See [comment above](https://github.com/Cardinal-Cryptography/aleph-node/pull/1335#issuecomment-1681913775) - I could be wrong, but tried every possible path to `AccountId` I was able to come up with and had no luck

aleph-client/src/lib.rs
ggawryal: Good point, it's AccountId, Address and Signature - refactored to emphasise that

aleph-client/docker/subxt-integration-entrypoint.sh
kostekIV: add a jira task to fix it when we will have subxt > 0.29

kostekIV (approved):

aleph-client/docker/subxt-integration-entrypoint.sh
ggawryal: https://cardinal-cryptography.atlassian.net/browse/A0-3050

ggawryal: Nevermind, I figured out how to substitute, removing that task

PR #1336, opened 2023-08-08
A0-2133: Fix nix

h4nsu:


nix/Dockerfile.build
```diff
@@ -5,7 +5,7 @@ RUN nix-env -i patchelf && \
 
 COPY nix/ /node/nix-files/nix/
 RUN chmod +x /node/nix-files/nix/nix-build.sh
-COPY default.nix shell.nix rust-toolchain /node/nix-files/
+COPY default.nix shell.nix /node/nix-files/
```
deuszx, line 8: Copying rust-toolchain is required for build reproducibility, isn't it?

h4nsu: @fixxxedpoint please help ;P

fixxxedpoint: Version of rust is fixed in `nix/versions.nix`. It was made this way after some nightly versions of rust were unable to install correctly. Perhaps current stable version of rust that we are using is ok.

deuszx: So prior to these changes, builds in nix would use the overridden Rust from `rust-toolchain` instead of one from nix package?

fixxxedpoint (approved):
LGTM. Could you also add that missing command to `BUILD.md` in section `Build with nix` (sudo docker build -t aleph-build -f nix/Dockerfile.build .)

nix/Dockerfile.build
fixxxedpoint: Fixed Rust version was introduced in commit `887b51f56c9deac70c10425f0735c1795408a0dc`.

fixxxedpoint (approved):

Marcin-Radecki (approved):

PR #1341, opened 2023-08-09
A0-2882: Export clique network metrics to prometheus

ggawryal:
# Description

## Type of change

# Checklist:


ggawryal:
Tested using `./run_nodes.sh` (we can set up grafana locally - https://docs.substrate.io/tutorials/build-a-blockchain/monitor-node-metrics/ - I recall somebody asked that question recently) and metrics appear, though it doesn't help much as they are all expected to be zero.

clique/src/lib.rs
```diff
@@ -119,6 +119,13 @@ pub trait Dialer<A: Data>: Clone + Send + 'static {
     async fn connect(&mut self, address: A) -> Result<Self::Connection, Self::Error>;
 }
 
+pub trait NetworkCliqueMetrics {
+    fn set_present_incoming_connections(&self, present: u64);
+    fn set_missing_incoming_connections(&self, missing: u64);
+    fn set_present_outgoing_connections(&self, present: u64);
+    fn set_missing_outgoing_connections(&self, missing: u64);
```
kostekIV, line 126: consistent with status and metrics in finality-aleph
```suggestion
    fn set_incoming_connections(&self, present: u64);
    fn set_missing_incoming_connections(&self, missing: u64);
    fn set_outgoing_connections(&self, present: u64);
    fn set_missing_outgoing_connections(&self, missing: u64);
```

clique/src/mock.rs
```diff
@@ -667,3 +667,12 @@ pub struct MockPrelims<D> {
     pub result_from_outgoing: UnboundedReceiver<ResultForService<MockPublicKey, D>>,
     pub authorization_requests: mpsc::UnboundedReceiver<(MockPublicKey, oneshot::Sender<bool>)>,
 }
+
+pub struct MockMetrics {}
```
kostekIV, line 671: terrible nit to use less characters :p
```suggestion
pub struct MockMetrics;
```

clique/src/service.rs
```diff
@@ -108,6 +124,7 @@ where
         listener: NL,
         secret_key: SK,
         spawn_handle: SH,
+        metrics: Option<M>,
```
kostekIV, line 127: add impl Metrics for Option<M: Metrics>

clique/src/service.rs
```diff
@@ -246,6 +264,9 @@ where
                 },
                 // periodically reporting what we are trying to do
                 _ = status_ticker.tick() => {
+                    if let Some(ref metrics) = &self.metrics {
```
kostekIV, line 267: then here you can just use self.manager.update_metrics(self.metrics)

kostekIV (approved):
looks good,

clique/src/lib.rs
ggawryal: Done, but I'm already missing that equal alignment

clique/src/service.rs
ggawryal: Good point

clique/src/mock.rs
ggawryal: Renamed to `NoopMetrics` and moved to `metrics.rs` file so that it can be used as a default implementation also for non-testing code.

finality-aleph/src/metrics.rs
```diff
@@ -233,11 +292,18 @@ impl<H: Key> Metrics<H> {
                 registry,
             )?,
             network_send_times,
+            validator_network_metrics,
         })));
 
         Ok(Metrics { inner })
     }
 
+    pub fn validator_network_metrics_cloned(&self) -> Option<impl NetworkCliqueMetrics> {
```
fixxxedpoint, line 301: Do you need that `Option` since you have `impl` for `Option<M>` anyway?

clique/src/manager/mod.rs
```diff
@@ -225,6 +227,14 @@ impl<PK: PublicKey + PeerId, A: Data, D: Data> Manager<PK, A, D> {
             .map_err(|_| SendError::ConnectionClosed)
     }
 
+    pub fn update_metrics<M: NetworkCliqueMetrics>(&self, metrics: &M) {
```
fixxxedpoint, line 230: Make this a method of `ManagerStatus` instead and call it here. Or perhaps that `ManagerStatus` should be rather returned by that `status_report` method and it should be handled in `service` instead.

finality-aleph/src/metrics.rs
fixxxedpoint: And remove that `cloned` from its name. Anyone using this api don't care it it's cloned.

fixxxedpoint (changes requested):

finality-aleph/src/metrics.rs
ggawryal: Yeah, we don't need Option there

clique/src/manager/mod.rs
ggawryal: Done

fixxxedpoint (approved):

PR #1342, opened 2023-08-09
A0-2846: removed purge-alephbft-backup from the purge-backup command

fixxxedpoint:
# Description

Removed a sub-command of the purge-chain command that was deleting files of the alephbft backup.

## Type of change

After this change alephbft's backup can be removed only by manually deleting its root folder inside `--base-path`.


kostekIV (approved):
:+1:

timorl (approved):

PR #1345, opened 2023-08-10
A0-2771: late sync 

kostekIV:
# Description

Add test of semi major sync to the nightly pipeline. Its run for around 2h :D


local-tests/test_major_sync.py
```diff
@@ -0,0 +1,66 @@
...
+chain.set_flags(bootnodes=addresses[0])
+chain.set_flags_validator(public_addr=addresses,
+                          public_validator_addresses=validator_addresses)
+
+chain.set_flags_validator('validator')
+
+printt('Starting the chain')
+chain.start('aleph', nodes=[0, 1, 2, 3])
+
+# sleep 2h to produce around 5/6 * 7200 blocks > 3 * 1800 (max forest depth).
```
timorl, line 45: Hm, it would likely be better to have a DB snapshot with such a state already set up, and here at most check whether it works at all (produce a small amount of blocks and check finalization). Not a blocker for this PR, but something we should consider talking with infra about. Could you make a task for that? Unless you think you can organize it quickly ofc.

timorl (approved):
One non-blocking comment.

local-tests/test_major_sync.py
kostekIV: it would be better, w8 for @Marcin-Radecki input before creating a ticket. there is also something like export/import blocks commands in cli that could be helpful here

Marcin-Radecki: Unfortunately, this is a blocker to this PR as the amount of GH-free minutes would be spent after 17 days (2000 minutes / 120). 
@MikolajGasior  can we spin a tiny self-hosted machine for this test? What would be its cost (PM us please)

Marcin-Radecki (changes requested):

local-tests/test_major_sync.py
timorl: If this is enough of a problem then I would definitely suggest solving it properly, i.e. generating a snapshot (once/rarely?) and only downloading it in the test. If the snapshot is from ~2h then this would download almost immediately, thus likely removing any problem with GitHub minutes.

kostekIV: Ok done, I have zipped workdir with 6k > blocks in db plus this one 'late' node starts from zero

Marcin-Radecki (approved):

PR #1346, opened 2023-08-11
A0-2998: Put 'true' in double quote in one the conditions to check if that helps

MikolajGasior:
# Description
Even though input as a boolean, its value in condition is cast to string.

### DIFF
--- a/.github/workflows/_check-vars-and-secrets.yml
+++ b/.github/workflows/_check-vars-and-secrets.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Check vars
-        if: inputs.check-vars == true
+        if: ${{ inputs.check-vars == "true" }}
         run: |
           if [[ \
             -z "${{ vars.ECR_PUBLIC_HOST }}" || \
@@ -37,7 +37,7 @@ jobs:
           fi
 
       - name: Check secrets
-        if: inputs.check-secrets == true
+        if: ${{ inputs.check-secrets == "true" }}
         run: |
           if [[ \
             -z "${{ secrets.ARGOCD_DEVNET_HOST }}" || \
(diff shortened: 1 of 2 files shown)


CezaryKierzyk (approved):

.github/workflows/_check-vars-and-secrets.yml
```diff
@@ -21,9 +26,32 @@ jobs:
     name: Check available vars and secrets
     runs-on: ubuntu-20.04
     steps:
+      - name: Debug input values
```
timorl, line 29: This shouldn't stay here, right?

timorl (approved):

.github/workflows/_check-vars-and-secrets.yml
```diff
@@ -21,9 +26,32 @@ jobs:
...
-        if: inputs.check-vars == true
         run: |
+          if [[ "${{ inputs.check-vars }}" != "true" ]]; then
+            echo "Skipping vars check"
+            exit 0
+          fi
+
           if [[ \
             -z "${{ vars.ECR_PUBLIC_HOST }}" || \
             -z "${{ vars.ECR_PUBLIC_REGISTRY }}" || \
```
ggawryal, line 57: Do we want to check `"${{ vars.RUST_TOOLCHAIN_NIGHTLY }}"` in both `dev-vars` and `vars`?

ggawryal (approved):

.github/workflows/_check-vars-and-secrets.yml
MikolajGasior: Removed

.github/workflows/_check-vars-and-secrets.yml
MikolajGasior: Oh boy, yes, of course!  It's removed now.

PR #1355, opened 2023-08-16
Disable button integration tests

obrok:
# Description

Now that the button is deployed, these are just causing noise when they flake out. In the future, we probably should archive all of the button code into a separate repo.

## Type of change


deuszx:
Does this remove TheButton e2e tests entirely from the CI? i.e. they will not run on "merge to main" nor when triggered with a label?

PR #1360, opened 2023-08-21
A0-3056: add reorg test to nightly

kostekIV:
# Description

## Type of change

# Checklist:

- I have added tests


local-tests/chainrunner/node.py
```diff
@@ -76,14 +80,42 @@ def greplog(self, regexp):
             log = f.read()
         return re.findall(regexp, log)
 
+    def highest_hash(self):
+        highest_block = self.rpc('chain_getBlockHash', None)
+        if isinstance(highest_block, Ok):
+            return highest_block.result
+        else:
+            None
```
woocash2, line 88: nit: not sure if `None` without `return` works in python. `None` will still be returned from this fn so just a nit

local-tests/chainrunner/node.py
```diff
@@ -76,14 +80,42 @@ def greplog(self, regexp):
...
+        """Find the height of the most recent block.
         Return two ints: highest block and highest finalized block."""
-        results = self.greplog(r'best: #(\d+) .+ finalized #(\d+)')
-        if results:
-            a, b = results[-1]
-            return int(a), int(b)
-        return -1, -1
+        try:
+            highest_block_hash = self.highest_hash()
+            highest_finalized_hash = self.highest_finalized_hash()
```
woocash2, line 109: we're getting the highest block first and then the highest finalized. When doing things in this order it's possible that highest finalized > highest known block :stuck_out_tongue:. I don't think we use it anywhere so feel free to resolve if it doesn't bother you.

woocash2 (approved):
Making rpc calls instead scrapping logs seems cooler :+1:

.github/workflows/nightly-normal-session-e2e-tests.yml
```diff
@@ -210,6 +210,26 @@ jobs:
           ALEPH_NODE_BINARY: ../target/release/aleph-node
         run: ./.github/scripts/test_major_sync.sh
 
+  run-force-reorg-test:
+    needs: [build-production-node-and-runtime]
+    name: Run force reorgs test
+    runs-on: self-hosted
```
Marcin-Radecki, line 216: Can you please use this syntax `[self-hosted, Linux, X64, small] to make sure small runer is used

If small is not enough then please let me know what specs this test needs in terms of CPU / RAM, we'll provide appropriate machine (but this will need some time to set it up)

e.g. as in https://github.com/Cardinal-Cryptography/aleph-node/blob/054225cbab6c06c1af436739fe550dad97a13d35/.github/workflows/_check-code-formatting.yml#L10

local-tests/chainrunner/node.py
kostekIV: yep, fixed

.github/workflows/nightly-normal-session-e2e-tests.yml
kostekIV: Ok, it passed on medium instance. Thx

.github/scripts/test_force_reorg.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
fixxxedpoint, line 1: use `#!/usr/bin/env bash`

.github/scripts/test_force_reorg.sh
```diff
@@ -0,0 +1,19 @@
...
+
+set -eu
+
+pushd local-tests/
+
+if [ ! -f "$ALEPH_NODE_BINARY" ]; then
```
fixxxedpoint, line 7: use some default value for `ALEPH_NODE_BINARY`. Echo the name of the missing var below?

.github/workflows/nightly-normal-session-e2e-tests.yml
```diff
@@ -210,6 +210,26 @@ jobs:
           ALEPH_NODE_BINARY: ../target/release/aleph-node
         run: ./.github/scripts/test_major_sync.sh
 
+  run-force-reorg-test:
+    needs: [build-production-node-and-runtime]
+    name: Run force reorgs test
+    runs-on: [self-hosted, Linux, X64, medium]
```
fixxxedpoint, line 216: It means it can be only run on a  `self-hosted` runner?

fixxxedpoint (approved):
lgtm

.github/workflows/nightly-normal-session-e2e-tests.yml
kostekIV: default github runner and self-hosted small instance were not powerful enough for this test (8 nodes running at the same time)

.github/scripts/test_force_reorg.sh
kostekIV: done

fixxxedpoint (approved):

PR #1362, opened 2023-08-21
A0-3058: Factor out clique metrics

timorl:
# Description

First part of the metrics cleanup, for now in the clique network.

## Type of change

# Checklist:

### DIFF
--- a/clique/src/metrics.rs
+++ b/clique/src/metrics.rs
@@ -2,19 +2,19 @@ use substrate_prometheus_endpoint::{register, Gauge, PrometheusError, Registry, 
 
 #[derive(Clone)]
 pub enum Metrics {
-    Noop,
-    Metrics {
+    Prometheus {
         incoming_connections: Gauge<U64>,
         missing_incoming_connections: Gauge<U64>,
         outgoing_connections: Gauge<U64>,
         missing_outgoing_connections: Gauge<U64>,
     },
+    Noop,
 }
 
 impl Metrics {
     pub fn new(registry: Option<Registry>) -> Result<Self, PrometheusError> {
         match registry {
-            Some(registry) => Ok(Metrics::Metrics {
+            Some(registry) => Ok(Metrics::Prometheus {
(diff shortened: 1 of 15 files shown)


clique/src/metrics.rs
```diff
@@ -1,41 +1,94 @@
...
-            m.set_missing_incoming_connections(missing)
-        }
+    pub fn noop() -> Self {
+        Metrics::Noop
     }
 
-    fn set_outgoing_connections(&self, present: u64) {
-        if let Some(m) = self {
-            m.set_outgoing_connections(present)
+    pub fn set_incoming_connections(&self, present: u64) {
```
kostekIV, line 55: We could also use an enum and have one fn set_connections_change:
```rust
Enum XXX {
 IncomingConnections(u64),
 ...,
}
```
names are just for this example pls pick better ones  than mine :P

clique/src/metrics.rs
```diff
@@ -1,41 +1,94 @@
...
+
+#[derive(Clone)]
+pub enum Metrics {
+    Noop,
+    Metrics {
+        incoming_connections: Gauge<U64>,
+        missing_incoming_connections: Gauge<U64>,
+        outgoing_connections: Gauge<U64>,
+        missing_outgoing_connections: Gauge<U64>,
+    },
```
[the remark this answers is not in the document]
timorl: Yeah, makes sense, done. Although I don't quite get why `Noop` being below is better, but whatever.

clique/src/metrics.rs
timorl: Ugh, I think I'll just do this properly – I left it like that because it's all being called in one place anyway, which is very weird from a metrics perspective. I'll rewrite it so that it actually tracks the metrics live and then the enum makes somewhat more sense.

kostekIV: Nice

woocash2 (approved):
lgtm

PR #1364, opened 2023-08-22
A0-3059: Factor out sync metrics

timorl:
# Description

Factor out the sync metrics from the centralized metrics, plus some changes to how they are organized. Since the metric names were not uniform anyway I changed them – this means there will need to be a companion PR on the Grafana side.

## Type of change

# Checklist:

- I haven't made neccessary updates to the Infrastructure, but I will

### DIFF
--- a/finality-aleph/src/sync/metrics.rs
+++ b/finality-aleph/src/sync/metrics.rs
@@ -76,38 +76,37 @@ pub enum Metrics {
 
 impl Metrics {
     pub fn new(registry: Option<Registry>) -> Result<Self, PrometheusError> {
-        match registry {
-            Some(registry) => {
-                let mut calls = HashMap::new();
-                let mut errors = HashMap::new();
-                for event in ALL_EVENTS {
-                    calls.insert(
-                        event,
-                        register(
-                            Counter::new(
-                                format!("aleph_sync_{}", event.name()),
-                                format!("number of times {} has been called", event.name()),
-                            )?,
-                            &registry,
-                        )?,
-                    );
-                }
-                for event in ERRORING_EVENTS {
-                    errors.insert(
-                        event,
-                        register(
-                            Counter::new(
-                                format!("aleph_sync_{}_error", event.name()),
-                                format!("number of times {} has returned an error", event.name()),
-                            )?,
-                            &registry,
-                        )?,
-                    );
-                }
-                Ok(Metrics::Prometheus { calls, errors })
-            }
-            None => Ok(Metrics::Noop),
+        let registry = match registry {
+            Some(registry) => registry,
+            None => return Ok(Metrics::Noop),
+        };
+        let mut calls = HashMap::new();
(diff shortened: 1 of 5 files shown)


finality-aleph/src/sync/metrics.rs
```diff
@@ -0,0 +1,132 @@
...
+    Prometheus {
+        calls: HashMap<Event, Counter<U64>>,
+        errors: HashMap<Event, Counter<U64>>,
+    },
+    Noop,
+}
+
+impl Metrics {
+    pub fn new(registry: Option<Registry>) -> Result<Self, PrometheusError> {
+        match registry {
```
kostekIV, line 79: nit to reduce nesting:
``` 
let registry = match registry {
    Some(registry) => registry,
    None => return Ok(Noop),
};
```

kostekIV (approved):
Nice :)

finality-aleph/src/sync/metrics.rs
timorl: Sure, done.

finality-aleph/src/sync/service.rs
```diff
@@ -104,6 +102,13 @@ where
         let broadcast_ticker = Ticker::new(BROADCAST_PERIOD, BROADCAST_COOLDOWN);
         let (justifications_for_sync, justifications_from_user) = mpsc::unbounded();
         let (block_requests_for_sync, block_requests_from_user) = mpsc::unbounded();
+        let metrics = match Metrics::new(metrics_registry) {
+            Ok(metrics) => metrics,
+            Err(e) => {
+                warn!(target: LOG_TARGET, "Failed to create metrics: {}.", e);
+                Metrics::noop()
```
fixxxedpoint, line 109: Maybe just fail? That creates a possibility of running a node without metrics for some time without noticing it.

fixxxedpoint (commented):
LGTM

finality-aleph/src/sync/service.rs
timorl: Hah, running without metrics should be the _easiest_ thing to notice though, especially if you have alerting for them. :D

I feel like returning an error here requires writing a lot of boilerplate, while the actual noop fallback is quite acceptable for metrics.

fixxxedpoint: I don't understand. This constructor already returns some error and when it occurs we just immediately panic. You can extend it or switch with `anyhow::Result`? Feel free to resolve this comment.

fixxxedpoint (approved):

finality-aleph/src/sync/service.rs
timorl: The error it returns is from `Handler`, which cannot fail on metrics, so I would have to create a separate error just for this. We don't currently use `anyhow`. Also, this would be inconsistent with the changes in `clique`, and there we don't return any error, so probably another one would have to be added there.

PR #1366, opened 2023-08-23
A0-3053: Add logging to e2e button tests

obrok:
# Description

I made the tests run on a single node (I think some of the issues, notably the timeouts were related to this) and added some logging around the spots I think can still be flaky to get more info later. This might not solve everything, but I think it's an improvement.

## Type of change


deuszx (approved):

fbielejec (approved):
we gotta start thinking about refactoring it all out to it's own repository. Code is easy to move, but the workflows not.

PR #1367, opened 2023-08-25
A0-3083: Refactor ticker for chain extensions

timorl:
# Description

These are changes that are needed to use ticker with the periodic requests for chain extensions which we will be implementing soon. In particular this _should not_ change any current behaviour of the node, thus it's just a refactor.

Comments on whether this interface is clear or not are very welcome.

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/finality-aleph/src/sync/ticker.rs
+++ b/finality-aleph/src/sync/ticker.rs
@@ -1,39 +1,37 @@
 use tokio::time::{sleep, Duration, Instant};
 
 /// This struct is used for rate limiting as an on-demand ticker. It can be used for ticking
-/// at least once `max_timeout` but not more than once every `min_timeout`.
-/// Example usage would be to use `wait` method in main select loop and `try_tick` whenever
-/// you would like to tick sooner in another branch of select.
+/// at most after `max_timeout` but no sooner than after `min_timeout`.
+/// It will always tick after `max_timeout` and never after that unless reset earlier.
+/// Example usage would be to use the `wait` method in main select loop and
+/// `try_tick` whenever you would like to tick sooner in another branch of select,
+/// resetting whenever the rate limited action actually occurs.
 pub struct Ticker {
-    last_tick: Instant,
+    last_reset: Instant,
     current_timeout: Duration,
     max_timeout: Duration,
     min_timeout: Duration,
 }
 
 impl Ticker {
-    /// Returns new Ticker struct. Behaves as if last tick happened during creation of Ticker.
-    /// Requires `max_timeout` >= `min_timeout`.
+    /// Returns new Ticker struct. Enforces `max_timeout` >= `min_timeout`.
     pub fn new(mut max_timeout: Duration, min_timeout: Duration) -> Self {
         if max_timeout < min_timeout {
             max_timeout = min_timeout;
         };
         Self {
-            last_tick: Instant::now(),
+            last_reset: Instant::now(),
             current_timeout: max_timeout,
             max_timeout,
             min_timeout,
         }
     }
 
-    /// Returns whether at least `min_timeout` time elapsed since the last tick.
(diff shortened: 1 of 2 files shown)


finality-aleph/src/sync/ticker.rs
```diff
@@ -1,39 +1,37 @@
 use tokio::time::{sleep, Duration, Instant};
 
 /// This struct is used for rate limiting as an on-demand ticker. It can be used for ticking
-/// at least once `max_timeout` but not more than once every `min_timeout`.
-/// Example usage would be to use `wait` method in main select loop and `try_tick` whenever
-/// you would like to tick sooner in another branch of select.
+/// at most after `max_timeout` but no sooner than after `min_timeout`.
+/// It will always tick after `max_timeout` and never after that unless reset earlier.
```
kostekIV, line 5:
```suggestion
/// It will always tick before `max_timeout` and never after that unless reset earlier.
```
?

timorl: Ugh, that's also not what I meant, but trying to write what I meant turned out to bee too hard, so I just deleted this line.

The documentation next to the methods should be good enough.

kostekIV (commented):
Could you restore `wait_and_reset` fn for usecases where the user doesn't want to remember about resetting the timeout? This is a much less error prone approach. Why is the extraction of the reset is needed?

finality-aleph/src/sync/ticker.rs
kostekIV: Yep, it was confusing as hell :P

finality-aleph/src/sync/ticker.rs
```diff
@@ -1,39 +1,36 @@
...
-/// at least once `max_timeout` but not more than once every `min_timeout`.
-/// Example usage would be to use `wait` method in main select loop and `try_tick` whenever
-/// you would like to tick sooner in another branch of select.
+/// at most after `max_timeout` but no sooner than after `min_timeout`.
+/// Example usage would be to use the `wait` method in main select loop and
+/// `try_tick` whenever you would like to tick sooner in another branch of select,
+/// resetting whenever the rate limited action actually occurs.
 pub struct Ticker {
-    last_tick: Instant,
+    last_reset: Instant,
```
kostekIV, line 9: while we are changing this could we also rename current_timeout to some better name :<

timorl:
The extraction of reset is needed because we will now have "failed ticks", i.e. the ticker ticks, but it turns out we don't actually want to send any message. We don't want to reset the ticker in this situation, since if anything suggests that we now want to send a message, we want to tick again immediately.

Making `wait` into `wait_and_reset` (or just keeping the name `wait_and_tick`) was something I even considered, the only issue I have with it now is that we either add weird boilerplate or always reset twice when we reach `max_timeout`. If you think the latter is acceptable then I can definitely change it, it only seemed somewhat annoying to me. To be clear, then we wouldn't even need a separate `wait` method – all the uses we have for this have a "forced" action when `wait` passes, so we would reset always anyway.

timorl:
Hm, changed significantly, especially since while addressing comments I noticed this wouldn't've been enough for the application it is supposed for. <_<"

Should be clearer what is happening now, although we have a double reset if `max_timeout` actually occurs.

finality-aleph/src/sync/ticker.rs
```diff
@@ -1,13 +1,18 @@
 use tokio::time::{sleep, Duration, Instant};
 
+enum State {
```
kostekIV, line 3: `Mode` seems better here

kostekIV (approved):

finality-aleph/src/sync/ticker.rs
timorl: Done.

woocash2 (approved):
lgtm

PR #1373, opened 2023-08-29
A0-3097: Degenerify block hashes

timorl:
# Description

We were theoretically parametrized by the block hash type, but in practice some parts of `finality-aleph` already fixed it based on `primitives`. This PR succumbs to that reality, thus simplifying quite a couple parametrizations here and there. The next step might be to use `BlockId` even more directly deeper (e.g. in `sync`), thus eliminating one more trait and lowering probably unnecessary parametrizations.

Also rename Metrics to BlockMetrics, since that was tiny in addition.

## Type of change

Refactor.

# Checklist:


finality-aleph/src/data_io/data_store.rs
```diff
@@ -33,82 +33,28 @@ use crate::{
...
 type MessageId = u64;
 
-#[derive(Clone, Debug)]
-pub enum ChainEvent<B>
-where
-    B: BlockT,
-    B::Header: HeaderT<Number = BlockNumber>,
-{
-    Imported(IdentifierFor<B>),
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
```
kostekIV, line 41: <3

kostekIV (approved):
gorgeous

finality-aleph/src/data_io/proposal.rs
```diff
@@ -338,17 +287,13 @@ mod tests {
             SessionBoundaryInfo::new(SessionPeriod(20)).boundaries_for_session(SessionId(0));
 
         let branch = vec![H256::default(); MAX_DATA_BRANCH_LEN];
-        let proposal = UnvalidatedAlephProposal::<TBlock>::new(
-            branch,
-            (MAX_DATA_BRANCH_LEN + 1) as BlockNumber,
-        );
+        let proposal =
+            UnvalidatedAlephProposal::new(branch, (MAX_DATA_BRANCH_LEN + 1) as BlockNumber);
```
woocash2, line 291: not for this pr probly but it oughta be `Invalidated` I guess

woocash2 (approved):

PR #1377, opened 2023-09-04
A0-2935: Investigate flaky e2e tests

ggawryal:
# Description

Changes in e2e-tests and aleph-client only.
* Improved finality version change test [fixes A0-2972] - now it schedules change in (current_session+1) for (current_session+3) instead of in the 1st session for 4th session - it requires one session less and should pass even when ran after 2nd session.
* More checks and better error for era payout test [partially fixes A0-2973] - previously a problem could occur, where one validator was not producing blocks, and because of that session durations were longer than expected. Added block timestamp check to detect such a situation, and report better error message int that case.
* Add logging for finalization test [to make further investigation on A0-2974 easier]

The actual problem in failing e2e-tests is that the chain might sometimes launch incorrectly, with some validators not being able to connect to other peers, causing a finalization stall. One possible solution to this I could imagine is to run the finalization test just after launching the chain and respawn the chain if the test timed out after 60s. Let me know what do you think about it.

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/aleph-client/Cargo.toml
+++ b/aleph-client/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aleph_client"
-version = "3.5.1"
+version = "3.6.0"
 edition = "2021"
 authors = ["Cardinal"]
 documentation = "https://docs.rs/aleph_client"
--- a/aleph-client/Cargo.toml
+++ b/aleph-client/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aleph_client"
-version = "3.5.0"
+version = "3.5.1"
 edition = "2021"
 authors = ["Cardinal"]
 documentation = "https://docs.rs/aleph_client"
--- a/e2e-tests/src/test/era_payout.rs
+++ b/e2e-tests/src/test/era_payout.rs
@@ -1,6 +1,8 @@
 use aleph_client::{
-    pallets::staking::{StakingApi, StakingSudoApi},
-    pallets::timestamp::TimestampApi,
+    pallets::{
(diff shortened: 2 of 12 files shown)


e2e-tests/src/test/era_payout.rs
```diff
@@ -19,32 +23,45 @@ pub async fn era_payouts_calculated_correctly() -> anyhow::Result<()> {
...
-    let end = expected_payout + one_block;
-    let within_delta = start <= payout && payout <= end;
-    assert!(
-        within_delta,
-        "payout should fall within range: [{start}, {end}] but was {payout}"
-    );
-}
-
-async fn wait_to_second_era<C: StakingApi + WaitingExt>(connection: &C) -> EraIndex {
+async fn wait_to_kth_era<C: StakingApi + WaitingExt>(
```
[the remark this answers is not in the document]
ggawryal: Indeed, replaced

aleph-client/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aleph_client"
-version = "3.5.0"
+version = "3.5.1"
```
timorl, line 3: You are adding at least one new functionality (timestamps), so bump the minor version.

timorl (approved):
Cool, nice to have less flaky tests even if not yet ideal.

maciejzelaszczyk (approved):
Nice effort!

aleph-client/Cargo.toml
ggawryal: Done

PR #1382, opened 2023-09-07
A0-3191: update aleph-client

fbielejec:
# Description

- Exposing function needed in bridge guardian
- don't flood default logs


krzysztofziobro (approved):

obrok (approved):
Or you could `ink-wrapper` instead - it has type-safe event translation.

fbielejec:
> Or you could `ink-wrapper` instead - it has type-safe event translation.

past the POC we most probably will

PR #1385, opened 2023-09-08
A0-3191: make some fields public

fbielejec:
# Description

Making some fields in the client public, needed by the guardian

### DIFF
--- a/aleph-client/src/connections.rs
+++ b/aleph-client/src/connections.rs
@@ -24,13 +24,16 @@ use crate::{
 /// Capable of communicating with a live Aleph chain.
 #[derive(Clone)]
 pub struct Connection {
+    /// inner subxt type
     pub client: SubxtClient,
 }
 
 /// Any connection that is signed by some key.
 #[derive(Clone)]
 pub struct SignedConnection {
+    /// vanilla connection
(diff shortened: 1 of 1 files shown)


DamianStraszak (approved):

aleph-client/src/connections.rs
```diff
@@ -24,14 +24,17 @@ use crate::{
...
 pub struct Connection {
-    client: SubxtClient,
+    /// inner subxt type
+    pub client: SubxtClient,
 }
 
 /// Any connection that is signed by some key.
 #[derive(Clone)]
 pub struct SignedConnection {
-    connection: Connection,
```
krzysztofziobro, line 33: Why can't you use `signed_connection.as_connection()`, `signed_connection.signer()` and `connection.as_client()` ?
https://github.com/Cardinal-Cryptography/aleph-node/blob/main/aleph-client/src/connections.rs#L220
https://github.com/Cardinal-Cryptography/aleph-node/blob/main/aleph-client/src/connections.rs#L175
https://github.com/Cardinal-Cryptography/aleph-node/blob/main/aleph-client/src/connections.rs#L387

krzysztofziobro (approved):

PR #1389, opened 2023-09-11
A0-3008: basic block metrics

ggawryal:
# Description

Adding two new metrics: `aleph_top_finalized_block` and `aleph_best_block`, alternatives to `substrate_block_height{status="best/finalized"}`. Main difference is that new metrics are updated in sync in `handle_chain_event`, whereas substrate ones were pulled from client every 5s.

Also refactored `TimedBlockMetrics` a little bit.

After merging, we can create a new prometheus job that would scrape just these 2 metrics in 1s intervals, guarantying that in healthy situations |metric_value_visible_in_grafana - actual_metric_value| <= 1 (see attached screenshots from local chain and local grafana instances)

## Type of change

# Checklist:
None

# Attachments:
Screenshot from 2023-09-12 17-15-36


finality-aleph/src/import.rs
```diff
@@ -54,15 +54,19 @@ where
...
         self.metrics
             .report_block(post_hash, Instant::now(), Checkpoint::Importing);
 
         let result = self.inner.import_block(block).await;
 
-        if let Ok(ImportResult::Imported(_)) = &result {
+        if let Ok(ImportResult::Imported(ref aux)) = &result {
             self.metrics
                 .report_block(post_hash, Instant::now(), Checkpoint::Imported);
+            if aux.is_new_best {
```
ggawryal, line 67: At first glance, it could be the case that we are importing blockA which becomes the best, then we import blockB which also becomes the best, but because of the concurrency they execute in different order in line 67, resulting in blockA being reported as best. If I'm not mistaken, this shouldn't be a problem for us, because we are never importing two blocks in one branch at the same time, and when importing two blocks from different branches we should quickly reject one branch.

finality-aleph/src/metrics.rs
```diff
@@ -127,31 +128,85 @@ pub enum Checkpoint {
...
+                highest_finalized.inc_by(delta);
+            }
+        }
     }
+}
 
+#[derive(Clone)]
+pub struct BlockMetrics {
+    pub top_block: TopBlockMetrics,
+    timed: Option<Arc<Mutex<TimedBlockMetrics>>>,
```
kostekIV, line 183: You don't need an option here. add Noop variant to TimedBlockMetrics. You also don't need Arc<Mutex<..>> at all, counters and gauges have some atomics in them. Those changes will help with concurrency issues.

ggawryal: I wanted to leave this refactoring to the timed metrics slo task, but I can do this right now. Tho, I have to keep Arc<Mutex<..>> for lru cache in metrics.

finality-aleph/src/import.rs
ggawryal: Changed to use best block from client, which is uglier, but should be safe.

finality-aleph/src/metrics.rs
ggawryal: Also, removing mutex should help with performance when importing many blocks, but if I'm not mistaken the concurrency issue I mentioned could have theoretically still occurred, so I fixed it in other way.

timorl:
Hm, the description does not match the contents of the PR.

Anyway, a bigger problem is that we are going to change the meaning of best block soon-ish (A0-3037) – the source-of-truth for that notion will be in sync rather than in the DB. Thus it would probably be better if this code already actually ran in sync rather than the other places – although figuring out whether there is a new "best block" might be much harder for now. (I think you can leave a placeholder with some silly logic like "higher than previous best", since it will be replaced before the next release anyway.) I'm less sure about finalization also fitting there, but if it did we could have these metrics restricted to `sync` rather than created in `bin/service`.

ggawryal:
> Hm, the description does not match the contents of the PR.
> 
> Anyway, a bigger problem is that we are going to change the meaning of best block soon-ish (A0-3037) – the source-of-truth for that notion will be in sync rather than in the DB. Thus it would probably be better if this code already actually ran in sync rather than the other places – although figuring out whether there is a new "best block" might be much harder for now. (I think you can leave a placeholder with some silly logic like "higher than previous best", since it will be replaced before the next release anyway.) I'm less sure about finalization also fitting there, but if it did we could have these metrics restricted to `sync` rather than created in `bin/service`.

@timorl Where exactly in `sync` should we update these metrics? How about trying updating them in `Forest::update_body` and `Forest::try_update_highest_justified` (and possibly renaming metrics to `forest_(best_block|highest_finalized)`)?

timorl:
I wouldn't put them in there, but rather in `Handler`. In particular, `update_body` will soon be returning whether something became the new best block, so that should make it simpler. Anyway, both of these kinda fit in `handle_chain_event`.

I am, however, open to other suggestions, especially since this is relatively "late" as far as the node is concerned. To be clear, I don't think this is a problem, but I might be lacking some perspective.

finality-aleph/src/sync/service.rs
```diff
@@ -119,7 +130,7 @@ where
                 justifications_from_user,
                 additional_justifications_from_user,
                 block_requests_from_user,
-                metrics,
+                metrics: (metrics, top_block_metrics),
```
timorl, line 133: Just make them one  struct, i.e. add the top block metrics to the previous struct. If you want the logic to be more separated then make it have two substructs within the `Prometheus` variant.

Also, since I'm already commenting despite not being a reviewer, maybe consider keeping the `noop` constructor? It's somewhat cleaner to not depend on the (imo internal implementation) detail that noop is implemented as an enum variant.

ggawryal: Sure, combined them. The substruct solution would require having methods in outer struct, or else pattern matching, so chosen simplest option. Bonus: I've just learned that in substrate, sometimes the following pattern is used:
```
pub fn report(&self, do_this: impl FnOnce(&Metrics)) {
    if let Some(metrics) = self.0.as_ref() {
	do_this(metrics);
    }
}
```
But calling this later `self.metrics.report(|metrics| metrics.some_fn())` looks perhaps more complicated than it should.

finality-aleph/src/metrics.rs
```diff
@@ -81,37 +97,43 @@ impl Inner {
...
-                    }
-                };
-                self.gauges
-                    .get(&checkpoint_type)
+        if let BlockMetrics::Prometheus {
+            prev,
+            gauges,
+            starts,
+        } = self
+        {
```
kostekIV, line 105: nit:
```suggestion
let (prev, gauges, start) = match self {
        BlockMetrics::Prometheus {
            prev,
            gauges,
            starts,
        } => (...),
        _ => return,
};
```

finality-aleph/src/sync/metrics.rs
```diff
@@ -106,26 +112,69 @@ impl Metrics {
...
-            if let Some(counter) = errors.get(&event) {
+        if let Metrics::Prometheus { event_errors, .. } = self {
+            if let Some(counter) = event_errors.get(&event) {
                 counter.inc();
             }
         }
     }
+
+    pub fn update_best_block(&self, number: BlockNumber) {
+        match self {
```
kostekIV, line 147: nit: 
use if let here

finality-aleph/src/sync/metrics.rs
```diff
@@ -106,26 +112,69 @@ impl Metrics {
...
+
+    pub fn get_best_block(&self) -> BlockNumber {
+        match self {
+            Metrics::Noop => 0,
+            Metrics::Prometheus { best_block, .. } => best_block.get().try_into().unwrap(),
+        }
+    }
+
+    pub fn update_top_finalized_block(&self, number: BlockNumber) {
+        match self {
```
kostekIV, line 161: here also, or return early in the Noop variant to reduce nesting

finality-aleph/src/sync/service.rs
```diff
@@ -395,14 +397,21 @@ where
                         target: LOG_TARGET,
                         "Error marking block as imported: {}.", e
                     )
+                } else {
+                    // TODO: use is_best_block info from Forest
+                    if number > self.metrics.get_best_block() {
+                        self.metrics.update_best_block(number);
```
kostekIV, line 403: push this logic into metrics code

finality-aleph/src/sync/metrics.rs
```diff
@@ -106,26 +112,69 @@ impl Metrics {
...
     }
+
+    pub fn update_best_block(&self, number: BlockNumber) {
+        match self {
+            Metrics::Noop => {}
+            Metrics::Prometheus { best_block, .. } => best_block.set(number as u64),
+        }
+    }
+
+    pub fn get_best_block(&self) -> BlockNumber {
```
kostekIV, line 153: this looks strange. This should be removed from the metrics code

kostekIV (changes requested):

finality-aleph/src/metrics.rs
ggawryal: Here we replace `if let` with `match`, and later `match` with `if let` :smiley:? Nevertheless, I think struct deconstruction with match + early return when Noop is both prettier and has smaller nesting

finality-aleph/src/sync/metrics.rs
ggawryal: done

finality-aleph/src/sync/metrics.rs
ggawryal: Added early return; cargo fmt has some bad taste and  formats `if let Struct { quite_a_long_name, ...}` into 4 lines, last of which is `} => {`.

finality-aleph/src/sync/service.rs
ggawryal: agreed & done, hope the todo will be resolved quite soon anyway

finality-aleph/src/sync/metrics.rs
ggawryal: removed

finality-aleph/src/sync/metrics.rs
```diff
@@ -68,8 +72,10 @@ const ERRORING_EVENTS: [Event; 9] = [
 
 pub enum Metrics {
     Prometheus {
-        calls: HashMap<Event, Counter<U64>>,
-        errors: HashMap<Event, Counter<U64>>,
+        event_calls: HashMap<Event, Counter<U64>>,
+        event_errors: HashMap<Event, Counter<U64>>,
+        top_finalized_block: Counter<U64>,
```
kostekIV, line 77: if you would change it to gauge it would eliminate the need to do some math for the inc_by. Anyway you are checking in the update fn if the number is increasing so counter IMHO is not needed

kostekIV (approved):

finality-aleph/src/sync/metrics.rs
kostekIV: its not a bad taste, rather the limit of the line length

finality-aleph/src/sync/metrics.rs
ggawryal: Hmm, using Gauge is easier, but the value of best finalized block doesn't decrease under normal conditions (I think even when sth went terribly wrong substrate cannot revert finalized blocks), so semantically it should be a counter.

kostekIV: yep, exactly. It will never decrease (and if it will well we will have other  problems) so we can use gauge and not worry about this property

fixxxedpoint (approved):
lgtm

finality-aleph/src/sync/metrics.rs
ggawryal: ok changed

PR #1407, opened 2023-09-19
A0-3112: Update python framework with RPC port changes

Marcin-Radecki:
# Description

We forgot to update python framework with recent changes that removed WS_PORT. Hence, the long session nightly test failed.

## Type of change

### DIFF
--- a/local-tests/chainrunner/node.py
+++ b/local-tests/chainrunner/node.py
@@ -68,14 +68,6 @@ class Node:
             raise KeyError("RPC port unknown, please set rpc_port flag")
         return port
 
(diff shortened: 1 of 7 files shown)


obrok (approved):
We seem to have a couple mechanisms like this for starting a network - is this the preferred one?

Marcin-Radecki:
> We seem to have a couple mechanisms like this for starting a network - is this the preferred one?

True, at some point we should invest in only one. Can't say whether it is the preferred one, for tinkering I use bash and this framework only for CI tests.

PR #1414, opened 2023-09-25
A0-3085: Add periodic extension request

timorleph:
# Description

Add periodic chain extension request to sync.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/sync/handler/mod.rs
+++ b/finality-aleph/src/sync/handler/mod.rs
@@ -9,20 +9,20 @@ use std::{
 use crate::{
     session::{SessionBoundaryInfo, SessionId},
     sync::{
-        data::{NetworkData, Request, State},
+        data::{BranchKnowledge, NetworkData, Request, State},
         forest::{
-            Error as ForestError, Forest, InitializationError as ForestInitializationError,
-            Interest,
+            Error as ForestError, ExtensionRequest, Forest,
+            InitializationError as ForestInitializationError, Interest,
         },
-        handler::request_handler::{RequestHandler, RequestHandlerError},
-        Block, BlockIdFor, BlockImport, ChainStatus, Finalizer, Header, Justification, PeerId,
-        Verifier,
+        handler::request_handler::RequestHandler,
+        Block, BlockIdFor, BlockImport, BlockStatus, ChainStatus, Finalizer, Header, Justification,
+        PeerId, UnverifiedJustification, Verifier,
     },
     BlockIdentifier, BlockNumber,
 };
 
 mod request_handler;
-pub use request_handler::Action;
+pub use request_handler::{Action, RequestHandlerError};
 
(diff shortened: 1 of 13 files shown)


finality-aleph/src/sync/handler/mod.rs
```diff
@@ -468,40 +466,66 @@ where
         })
     }
 
+    /// Handle a chain extension request.
+    ///
+    /// First treats it as a request for our favourite block with their favourite block
+    /// as the top imported.
+    /// If that fails due to their favourite block not being a descendant of our,
```
lesniak43, line 473: -> our of their?

finality-aleph/src/sync/handler/mod.rs
```diff
@@ -511,21 +535,21 @@ where
...
 
         for justification in iter::once(justification).chain(maybe_justification) {
-            maybe_id = match self.handle_justification(justification, Some(peer.clone())) {
-                Ok(id) => id,
-                Err(e) => return (maybe_id, Some(e)),
-            };
+            new_highest = match self.handle_justification(justification, Some(peer.clone())) {
+                Ok(new_highest) => new_highest,
+                Err(e) => return (new_highest, Some(e)),
+            } || new_highest;
```
lesniak43, line 545: Why is it necessary now? Was there a bug?

finality-aleph/src/sync/handler/mod.rs
```diff
@@ -538,41 +562,49 @@ where
...
+        // Lets us import descendands of importable blocks, useful for favourite blocks.
+        let mut last_imported_block: Option<BlockIdFor<J>> = None;
         for item in response_items {
             match item {
                 ResponseItem::Justification(j) => {
                     match self.handle_justification(j, Some(peer.clone())) {
-                        Ok(Some(id)) => highest_justified = Some(id),
-                        Err(e) => return (highest_justified, Some(e)),
-                        _ => {}
+                        Ok(highest) => new_highest = new_highest || highest,
```
lesniak43, line 573: same question here

lesniak43 (approved):
lgtm, but please double-check everything, as this is quite a large PR

finality-aleph/src/sync/handler/mod.rs
timorleph: Yuup, thanks, fixed.

finality-aleph/src/sync/handler/mod.rs
timorleph: `handle_justification` no longer returns an ID, so it couldn't've stayed as it was. Although I might be misunderstanding your question.

Iirc the underlying modification was made because we are not using the ID anymore, because we don't create a specific task for it, so the ticker only attempts firing, and later gathers the information for the request anyway.

fixxxedpoint (approved):
LGTM, but it was a lot to read.

PR #1416, opened 2023-09-26
A0-3037: Sync information for Aura

timorleph:
# Description

Make Aura use the information from our sync to decide on block production (whether we are in "major sync").

Some of the choices here might be controversial, so I would welcome opinions about them. <_<

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/sync_oracle.rs
+++ b/finality-aleph/src/sync_oracle.rs
@@ -32,6 +32,7 @@ impl SyncOracle {
 
     pub fn update_behind(&self, behind: u32) {
         self.behind.store(behind, Ordering::Relaxed);
+        *self.last_update.lock() = Instant::now();
     }
 }
 
--- a/finality-aleph/src/sync_oracle.rs
+++ b/finality-aleph/src/sync_oracle.rs
@@ -1,0 +1,50 @@
+use std::{
+    sync::{
+        atomic::{AtomicU32, Ordering},
+        Arc,
+    },
+    time::{Duration, Instant},
+};
+
+use parking_lot::Mutex;
+use sp_consensus::SyncOracle as SyncOracleT;
+
+/// A sync oracle implementation tracking how far behind the highest known justification the node is.
+/// It defines being in major sync as knowing of any justification of an unknown block.
+/// It defines being offline as not getting any update for at least 6 seconds (or never at all).
+#[derive(Clone)]
+pub struct SyncOracle {
+    behind: Arc<AtomicU32>,
+    last_update: Arc<Mutex<Instant>>,
+}
+
+impl SyncOracle {
+    pub fn new() -> Self {
+        SyncOracle {
+            // This should never exceed 1800 due to the structure of the forest, thus 9000 is a decent marker of being uninitialized.
+            behind: Arc::new(AtomicU32::new(9001)),
+            last_update: Arc::new(Mutex::new(Instant::now())),
+        }
+    }
+
+    pub fn update_behind(&self, behind: u32) {
+        self.behind.store(behind, Ordering::Relaxed);
+    }
+}
+
+impl Default for SyncOracle {
+    fn default() -> Self {
+        SyncOracle::new()
+    }
+}
+
+impl SyncOracleT for SyncOracle {
+    fn is_major_syncing(&self) -> bool {
+        self.behind.load(Ordering::Relaxed) > 0
+    }
+
+    fn is_offline(&self) -> bool {
+        self.last_update.lock().elapsed() > Duration::from_secs(6)
+            || self.behind.load(Ordering::Relaxed) > 9000
+    }
+}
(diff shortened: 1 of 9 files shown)


finality-aleph/src/sync_oracle.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+impl SyncOracleT for SyncOracle {
+    fn is_major_syncing(&self) -> bool {
+        self.behind.load(Ordering::Relaxed) > 0
+    }
+
+    fn is_offline(&self) -> bool {
+        self.last_update.lock().elapsed() > Duration::from_secs(6)
+            || self.behind.load(Ordering::Relaxed) > 9000
```
kostekIV, line 48: nit: name both constatns used in this file 9000 and 6s

finality-aleph/src/sync_oracle.rs
```diff
@@ -0,0 +1,50 @@
...
+        Arc,
+    },
+    time::{Duration, Instant},
+};
+
+use parking_lot::Mutex;
+use sp_consensus::SyncOracle as SyncOracleT;
+
+/// A sync oracle implementation tracking how far behind the highest known justification the node is.
+/// It defines being in major sync as knowing of any justification of an unknown block.
```
kostekIV, line 13: I think this is 
1. too strong: we will be too often in the  major sync because of one unknown justification from the next block ( I envision this leading to a lot of switches between major and normal mode)
2. too weak? what about the situation when we have all of the justifications but are behind in not yet finalized branch of blocks?

woocash2: 1. I'm not aware of other details but considering just block production it seems good to me that we stop producing blocks (by entering major sync) whenever there's a justified block above our top finalized, bc they'd be rejected anyway.
2. If I understand correctly this should trigger major sync, right?

timorleph: In this case the major sync state only impacts block production. Thus, my reasoning was that:
1. If we build a block on top of a block that we know is already not the best finalized block we are surely wasting work.
2. If we don't know which of the blocks will get finalized, then building on anything has a chance of doing something productive.
I think the only issue might happen during an actual catchup sync if for whatever reason we don't currently know of any future justifications. In this situation with the current definition we might end up creating some ancient useless block. I don't think this is a major problem (I might be missing something), but it would definitely require a lot of work to improve.

finality-aleph/src/sync_oracle.rs
timorleph: Good point, improved.

finality-aleph/src/sync_oracle.rs
kostekIV: 1. But we dont know yet if it is not the best finalized block, we don't compare hashes and paths here, right? We can have the best block that is not finalized yet and is built on another not finalized block etc.

kostekIV: And I feel like this being too strong can delay or even stop us from producing blocks.

timorleph: You mean a situation where we have the blocks, but are missing a justification in the path to finalize them? That should be really rare and fix itself very quickly. Although, if you have a decent idea that would fix that problem that would be nice.

Oh, to be clear, this is the only situation in which this can happen, since otherwise we finalize as soon as we receive the justification. There is no point at which we have both the justification and the block but we don't finalize it (other than finalization holes, as explained above).

How can this stop us from producing blocks if we are not lagging behind the finalized block (in which case our production would be wasted anyway)?

timorleph: Hm, welp, this actually doesn't work on a fresh chain, annoying... I'll try to figure out why.

kostekIV: @timorleph The issue I have here is that we changed quite a bit a definition of the major sync. Before it meant that we are in the process of downloading at least `MAJOR_SYNC_BLOCKS`  (whatever that const is). Now it means we have **one** justification that we know of that is missing from us.

timorleph: Major sync state is only examined when deciding whether to produce blocks, so changing the definition only impacts that. Thus, changing the definition is definitely intended, since the new one works much better for our usecase.

kostekIV: Its also used in ready endpoint (not yet used) and one place in finality-aleph (catch up fn). Im not sure if not producing a block is always better. But feel free to resolve

kostekIV (approved):

finality-aleph/src/sync_oracle.rs
timorleph: Hmmmm, I'm not sure about the endpoint, but the catch-up function in finality uses the old notion of major syncing. This is not great, so I created a separate task for fixing that.

woocash2 (approved):

PR #1418, opened 2023-09-28
A0-3249: Prepare release-12 RC commit

Marcin-Radecki:
# Description

This change
* bumps `spec_version` and `tx_version` - latter since quite a lot pallet API changed 
* bumps node and runtime version to `0.12.0`
* enable contract migrations

## Type of change

# Checklist:

### DIFF
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -29,6 +29,7 @@ use frame_support::{
     weights::constants::WEIGHT_REF_TIME_PER_MILLIS,
     PalletId,
 };
+use frame_support::pallet_prelude::Get;
 use frame_system::{EnsureRoot, EnsureSignedBy};
 #[cfg(feature = "try-runtime")]
 use frame_try_runtime::UpgradeCheckSelect;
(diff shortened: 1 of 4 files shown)


h4nsu (approved):
This made me wondering - why is `main` on 9.0/56? Shouldn't we be merging these version bumping commits back to `main`?

Marcin-Radecki:
> This made me wondering - why is `main` on 9.0/56? Shouldn't we be merging these version bumping commits back to `main`?

Good point, ideally we should have another commit on `main` just after `release-X` is cut which bumps node to sth like `0.13.0-develop`, and our wallet shows it as `0.13.0-develop-<sha>`. I can prepare something like this (but without migrations enabled ofc)

Marcin-Radecki:
`try-runtime` was able to test NominationPools migration, but no contracts, as tehre is no support yet in try-runtime for multi-block migration.
[try-runtime.log](https://github.com/Cardinal-Cryptography/aleph-node/files/12748114/try-runtime.log)

bin/runtime/src/lib.rs
```diff
@@ -832,6 +833,24 @@ pub type SignedExtra = (
     frame_system::CheckWeight<Runtime>,
     pallet_transaction_payment::ChargeTransactionPayment<Runtime>,
 );
+
+pub struct NominationPoolsMigrationV5OldPallet;
```
kostekIV, line 837: nit: rename to smth like `MaxGlobalCommision`

kostekIV (approved):

kostekIV (approved):

bin/runtime/src/lib.rs
Marcin-Radecki: done

PR #1422, opened 2023-10-02
A0-3254: Uniform major sync

timorleph:
# Description

This makes our code use the same definition of "major sync" (provided by our block sync) everywhere, rather than just for Aura. It also changes the definition to refer to being nontrivially behind in justifications within a period of time. This change might result in some blocks being produced futilely (in contrast to our previous impl from last week, but not necessarily to the substrate one), but keeping it simple is worth the (probably in practice tiny) inefficiency.

## Type of change

Uhhh... no idea.

# Checklist:


lesniak43 (approved):

finality-aleph/src/sync_oracle.rs
```diff
@@ -1,38 +1,42 @@
...
-// This should never exceed 1800 due to the structure of the forest, thus 9000 is a decent marker of being uninitialized.
-const UNINITIALIZED_THRESHOLD: u32 = 9000;
 const OFFLINE_THRESHOLD: Duration = Duration::from_secs(6);
+const FAR_BEHIND_THRESHOLD: u32 = 15;
+const MAJOR_SYNC_THRESHOLD: Duration = Duration::from_secs(10);
 
-/// A sync oracle implementation tracking how far behind the highest known justification the node is.
-/// It defines being in major sync as knowing of any justification of an unknown block.
+/// A sync oracle implementation tracking how recently the node was far behind the highest known justification.
+/// It defines being in major sync as being more than 15 blocks behind the highest known justification less than 10 seconds ago.
```
woocash2, line 14: I don't understand the idea behind this logical condition. If you could explain I'd be grateful.
Accepting not to block tho.

woocash2 (approved):

finality-aleph/src/sync_oracle.rs
timorleph: While we are actually downloading the chain (so "major sync" in the intuitive sense) we keep getting justifications pretty far in the future (more than 15 blocks definitely). We do get these before we get the related blocks, as a signal that we should request these blocks. We get them often enough that 10s is more than enough time to ensure that we are no longer in this state. Thus this will reliably be true for the initial downloading of the chain. At the same time, when we are working normally we should only get justifications at most one above our top finalized block, thus never triggering this condition. The 15 is to give a generous margin of error (same as the 10s) to this whole reasoning. Edge cases, where the node keeps falling behind about 15 justifications, are broken anyway, so we don't particularly care which way they fall, as they will not work properly anyway.

Let me know whether/which parts of this explanation would be useful as part of the comment here, so I can edit it in. Any comments on improving the presentation are also appreciated. ^^

woocash2: Doesn't this condition imply that sometimes we will unnecessarily be in the state of major sync?
For example we fall behind 15 blocks bc of some network lag, we enter major sync, but we manage to immediately catch up. Now we are up to date with the blocks & justifications but we have to wait these 10 seconds to start producing again. Isn't it wasteful?

timorleph: This should be very rare, and if it happens then the main practical difference is that we don't produce blocks for 10 more seconds – this is still at most one missed block (and ~2x the probability of actually hitting it), so if it was really a fluke then it shouldn't hurt anyone, and if it happens more often, then this is already unacceptable instability, so we shouldn't care. The other consequences of spending more time in major sync are even less consequential.

woocash2: aight, understood

PR #1423, opened 2023-10-02
A0-2831: Sync rate limiting

ggawryal:
# Description

* Adjusted peer sender queue size constant in gossip network from 1000 to 16 per peer.
* Rolled back metrics for total number of sends and receives in those queues, after removal in #1278 (previously exposed via `substrate_unbounded_channel_len{entity=}`

Picked constant is a ballpark based on the following checks:
* As per grafana, the queue size metrics is almost always 0, there are some punctual spikes with average queue sizes over all peer senders equal to 2-3.
* On a local chain with 4 nodes, the queue size grows above 7 once in a few minutes, but then all messages are being popped from queue almost instantly, so that metrics value is being reported as zero.
* By running local tests, both finalization and block production goes smoothly even after lowering max queue size to 1, dropping messages on receiver's side with some probability, and adding an artificial delay.

## Type of change

# Checklist:


finality-aleph/src/network/gossip/service.rs
```diff
@@ -250,21 +268,35 @@ impl<N: RawNetwork, AD: Data, BSD: Data> Service<N, AD, BSD> {
...
                     Err(e) => {
-                        // Receiver can also be dropped when thread cannot send to peer. In case receiver is dropped this entry will be removed by Event::NotificationStreamClosed
-                        // No need to remove the entry here
                         if e.is_full() {
-                            warn!(
-                                target: LOG_TARGET,
-                                "Failed sending data in block sync notification channel to peer because peer_sender receiver is full: {:?}",
-                                peer
-                            );
+                            let peer_and_protocol = (peer.clone(), Protocol::BlockSync);
```
lesniak43, line 272: maybe extract the following block to a separate method, as it's very similar to the one 40 lines above?

lesniak43 (approved):
Nice job! Please make sure that we don't forget to include the metrics in Grafana.

fixxxedpoint (approved):
lgtm

PR #1426, opened 2023-10-03
A0-3032: Handle own blocks in sync

timorleph:
# Description

Blocks created by us should go through sync and get broadcast as soon as we create them.

## Type of change

# Checklist:

(I might need to add tests, but I wanted to have this mostly done before my days off.)


finality-aleph/src/import.rs
```diff
@@ -217,3 +221,85 @@ where
...
+
+#[async_trait::async_trait]
+impl<I> BlockImport<Block> for RedirectingBlockImport<I>
+where
+    I: BlockImport<Block> + Clone + Send,
+{
+    type Error = RedirectingImportError<I::Error>;
+    type Transaction = I::Transaction;
+
+    async fn check_block(
```
woocash2, line 278: Why is it not used anywhere (why)?

finality-aleph/src/import.rs
```diff
@@ -217,3 +221,85 @@ where
...
+        let BlockImportParams { body, .. } = block;
+
+        let extrinsics = body.ok_or(RedirectingImportError::MissingBody)?;
+
+        self.blocks_tx
+            .unbounded_send(Block { header, extrinsics })
+            .map_err(|_| RedirectingImportError::ChannelClosed)?;
+
+        // We claim it was successfully imported and no further action is necessary.
+        // This is likely inaccurate, but again, should be enough for Aura.
```
woocash2, line 302: Why is it enough for Aura? (I sense that this may be related to my question above)

finality-aleph/src/import.rs
timorleph: I don't think we (or rather Aura) use it, but just in case it does this implementation should actually work, since it just forwards to the `Client`.

finality-aleph/src/import.rs
timorleph: Aura likely expects all blocks it created to work anyway, and definitely doesn't do anything smart when this fails. It also always has to check the `Client` for the parent block anyway, so it cannot depend on this working when creating.

lesniak43: this looks dangerous, so pls pls make sure that Aura does not misbehave if we lie here

lesniak43 (approved):

finality-aleph/src/import.rs
timorleph: I checked, the only things that happen on errors is logging+telemetry, and the result optionally handles justification requests through legacy sync – the value we return does not trigger that, and even if it did we don't care, as we handle our requests ourselves.

finality-aleph/src/sync/service.rs
```diff
@@ -632,6 +645,13 @@ where
                     },
                     None => warn!(target: LOG_TARGET, "Channel with internal block request from user closed."),
                 },
+                maybe_own_block = self.blocks_from_creator.next() => match maybe_own_block {
+                    Some(block) => {
+                        debug!(target: LOG_TARGET, "Received new onw block: {:?}.", block.header().id());
```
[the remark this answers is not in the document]
timorleph: Thanks, fixed.

woocash2 (approved):
I approve not to block, but I would need more context to understand better how Aura interacts with this code. I'd be grateful for a short explanation or some reference.

I understand that we have a wrapper for `block_import` which is adjusted for Aura. I get that `check_block` is aligned with `inner.check_block`, but `import_block` doesn't `check_block` which I perceive as some simplification which Aura won't break.

Tho I don't get why it doesn't break with Aura, and why we need this wrapper in the first place. Was there an issue that Aura sometimes didn't work with the original `block_import`?

timorleph:
The issue this is dealing with is passing the blocks Aura produces through sync, so that they can be broadcast. The wrapper intercepts created blocks and first passes them through our sync, which only afterwards forwards them to the proper import queue.

Aura only makes very limited use of the results returned by the import queue we give it (as I outlined in one of the responses above), so intercepting the newly created blocks this way cannot impact its logic. I haven't checked whether `check_block` is used anywhere, but it does not modify the wrapped object and the current implementation returns the exact same thing as it did previously, so it should still work. The implementations of `import_block` do not usually call `check_block` in the first place, so that shouldn't be too weird – anyway, we let Aura perform all the necessary checks later, when it passes through the other import queue after going through our sync.

PR #1428, opened 2023-10-06
A0-3036: Verify headers

lesniak43:
# Description

## Type of change

# Checklist:

- I have added tests


bin/node/src/service.rs
```diff
@@ -158,6 +158,9 @@ pub fn new_partial(
 
     let slot_duration = sc_consensus_aura::slot_duration(&*client)?;
 
+    // DO NOT change Aura parameters without updating the finality-aleph sync accordingly,
+    // in particular the code responsible for verifying incoming Headers, as it is supposed
+    // to duplicate parts of Aura internal logic
     let import_queue = sc_consensus_aura::import_queue::<AuraPair, _, _, _, _, _>(
```
kostekIV, line 164: does all paprameteres contributes to verifier logic?

finality-aleph/src/session_map.rs
```diff
@@ -74,19 +79,32 @@ where
...
+            self.block_hash(block_number)?,
+        )
+        .ok()
+    }
+
     fn authority_data(&self, block_number: BlockNumber) -> Option<SessionAuthorityData> {
         let block_hash = self.block_hash(block_number)?;
         match self.client.runtime_api().authority_data(block_hash) {
             Ok(data) => Some(data),
-            Err(_) => self
```
kostekIV, line 86: hmm, why is this better?

finality-aleph/src/sync/handler/mod.rs
```diff
@@ -580,6 +588,10 @@ where
                     }
                 }
                 ResponseItem::Header(h) => {
+                    let h = match self.verify_header(h) {
+                        Ok(h) => h,
+                        Err(e) => return (new_highest, Some(e)),
+                    };
                     if self.forest.skippable(&h.id()) {
                         continue;
                     }
```
kostekIV, line 597: is this order right? Just want to make sure it is.

finality-aleph/src/sync/mock/backend.rs
```diff
@@ -100,9 +101,14 @@ impl Backend {
         Self {
             inner: storage,
             notification_sender,
+            headers_are_correct: Arc::new(Mutex::new(true)),
         }
     }
 
+    pub fn start_discarding_headers(&mut self) {
+        *self.headers_are_correct.lock() = false;
```
kostekIV, line 109: any way to switch it back into accepting headers?

finality-aleph/src/sync/mock/backend.rs
```diff
@@ -30,6 +30,7 @@ struct BackendStorage {
 pub struct Backend {
     inner: Arc<Mutex<BackendStorage>>,
     notification_sender: UnboundedSender<MockNotification>,
+    headers_are_correct: Arc<Mutex<bool>>,
```
kostekIV, line 33: nit: rename to `accepting_headers` or smth like this

finality-aleph/src/sync/substrate/verification/cache.rs
```diff
@@ -128,9 +156,68 @@ where
     /// Prune all sessions with a number smaller than `session_id`
     fn prune(&mut self, session_id: SessionId) {
         self.sessions.retain(|&id, _| id >= session_id);
+        // off by one!
+        // let's retain one more session, just to be safe
```
kostekIV, line 160: hmm I dont like this :P could you do the same with self.sessions and update function comment

finality-aleph/src/sync/substrate/verification/cache.rs
```diff
@@ -128,9 +156,68 @@ where
...
+
+    /// Returns the list of Aura authorities for a given block number. Updates cache if necessary.
+    /// This method assumes that the queued Aura authorities will indeed become Aura authorities
+    /// in the next session.
+    pub fn get_aura_authorities(
+        &mut self,
+        number: BlockNumber,
+    ) -> Result<&Vec<AuraId>, CacheError> {
+        if number == 0 {
+            return Err(CacheError::NoAuraAutoritiesForGenesis);
```
kostekIV, line 190: Shouldnt we check at least that header at zero match genesis header?

finality-aleph/src/sync/substrate/verification/cache.rs
```diff
@@ -128,9 +156,68 @@ where
...
+        let session_id = self.session_info.session_id_from_block_num(number);
+        if session_id < self.lower_bound {
+            return Err(CacheError::SessionTooOld(session_id, self.lower_bound));
+        }
+        let upper_bound = SessionId(
+            self.session_info
+                .session_id_from_block_num(self.finalization_info.finalized_number())
+                .0,
+        );
+        if session_id > upper_bound {
```
kostekIV, line 201: cant we check also the next session? We should have confirmed authorities for the `n + 1` session if the first block of session `n` is finalized.

finality-aleph/src/sync/substrate/verification/mod.rs
```diff
@@ -94,4 +129,52 @@ where
...
+        let parent_number = header.number() - 1;
+        let slot = find_pre_digest::<Block, AuthoritySignature>(&header)
+            .map_err(|e| Self::Error::HeaderVerification(PreDigestLookupError(e)))?;
+        // duplicate code, watch out!
+        let slot_now = Slot::from_timestamp(
+            sp_timestamp::Timestamp::current(),
+            sp_consensus_slots::SlotDuration::from_millis(MILLISECS_PER_BLOCK),
+        );
+        // timorl's offset
+        if slot > slot_now + 10 {
```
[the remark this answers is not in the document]
kostekIV: and explain why 10 :P

kostekIV (commented):
Good code :) I have left some comments. Mostly nits and request for explanantions

finality-aleph/src/sync/handler/mod.rs
timorleph: The other way has to be correct and it's a minor optimization, so it would be nice to switch around, unless `UnverifiedHeader` doesn't have `id()`, in which case adding it is probably not worth it.

finality-aleph/src/sync/mock/backend.rs
timorleph: This is such a weird way of doing this, cannot you add something to headers instead? Even an `is_correct: bool` in `MockHeader` would work, wouldn't it?

finality-aleph/src/sync/substrate/verification/cache.rs
timorleph: Also, we pass `parent_number` to this function, which means we will actually run this code for block number one, which should have authorities – is this a bug or am I missing something?

timorleph: Tbh I would leave the call as is and remove this error, while getting the session of the next block below – otherwise this will be an error that never gets returned as we never call this method for 0, right?

finality-aleph/src/sync/substrate/verification/mod.rs
```diff
@@ -94,4 +129,52 @@ where
...
+            .digest_mut()
+            .pop()
+            .ok_or(Self::Error::HeaderVerification(MissingSeal))?;
+        let sig = seal
+            .as_aura_seal()
+            .ok_or(Self::Error::HeaderVerification(IncorrectSeal))?;
+        let authorities = self
+            .get_aura_authorities(parent_number)
+            .map_err(|_| Self::Error::HeaderVerification(MissingAuthorityData))?;
+        // duplicate code, watch out!
```
timorleph, line 166: This is a bit cryptic, duplicate with what exactly?

timorleph (changes requested):
Nice, although I think there is a bug?

bin/node/src/service.rs
lesniak43: no, but I don't want anyone to touch them unless they know how Aura works and how our sync duplicates parts of its logic

finality-aleph/src/session_map.rs
lesniak43: it's necessary cause both `AlephSessionApi` and `AuraApi` have a method called `authorities`, right?

finality-aleph/src/sync/mock/backend.rs
lesniak43: currently this is not required by any test

finality-aleph/src/sync/substrate/verification/cache.rs
lesniak43: ok, the comment is "slightly" misleading :) I actually want to have one more session in `aura_authorities`, and the reason is sth like that: we set `CACHE_SIZE = 2`, so I'd expect that we can verify every header/justification from the current or previous session; since header verification requires access to its parent, we need to keep the additional past session to be able to verify the first header of the previous session - you still don't like it? if so, explain! :D if you do, I'll update the comment accordingly

finality-aleph/src/sync/substrate/verification/cache.rs
lesniak43: it's checked in `VerifierCache::verify_header`, here we only have access to the id
currently this error will never be returned

finality-aleph/src/sync/substrate/verification/mod.rs
timorleph: Do you mean "assuming round robin"? That's hardly an explanation. <_<

finality-aleph/src/sync/substrate/verification/cache.rs
lesniak43: ooookay, this is buggy - let's do it your way

finality-aleph/src/session_map.rs
kostekIV: you re right

finality-aleph/src/sync/mock/backend.rs
kostekIV: +1 to timrl[eph] suggestion

finality-aleph/src/sync/substrate/verification/mod.rs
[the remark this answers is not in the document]
lesniak43: done

finality-aleph/src/sync/handler/mod.rs
lesniak43: fixed

finality-aleph/src/sync/substrate/verification/cache.rs
lesniak43: so I've changed it back to the more reasonable behaviour (`CACHE_SIZE` defines cache size...), and changed the const from 2 to 3 everywhere
also, I've added a comment that this is necessary due to Aura, and slightly modified one test that assumed `CACHE_SIZE == 2`
please take a look at the commit to make sure that it looks fine now

finality-aleph/src/sync/substrate/verification/cache.rs
lesniak43: I think I've fixed it, but please take a look at the full commit just to be sure

finality-aleph/src/sync/substrate/verification/cache.rs
lesniak43: fixed

finality-aleph/src/sync/substrate/verification/mod.rs
lesniak43: could you suggest something more descriptive then?

finality-aleph/src/sync/mock/backend.rs
lesniak43: rewritten

kostekIV: Cool, @timorleph leaving this conversation up to you to resolve

kostekIV (approved):

finality-aleph/src/sync/substrate/verification/mod.rs
timorleph: Unfortunately I have no idea what this (or actually most of the comments about duplication in this function) means – that's why I'm asking this question. xd

If it's only about duplication with Aura, then perhaps just having a comment above the function about this that details either the differences or the duplication (whichever is shorter) would be clearer?

Speaking of, "verify the signature" a couple lines below is completely redundant, the code is pretty clear about that, please just remove that one.

lesniak43: fixed

finality-aleph/src/sync/substrate/verification/cache.rs
```diff
@@ -163,7 +188,36 @@ where
...
+                    &self.session_info,
+                )
+                .ok_or(CacheError::UnknownAuraAuthorities(session_id))?;
+                vacant.insert(authorities)
+            }
+        })
+    }
+
+    /// Returns session verifier for block number if available. Updates cache if necessary.
+    pub fn get(&mut self, number: BlockNumber) -> Result<&SessionVerifier, CacheError> {
```
timorleph, line 218: There is some annoying code duplication here, although not trivial to remove – did you consider abstracting some stuff into a sub-struct (wrapping a single `HashMap`? Or downloading both at once and having only one map?

finality-aleph/src/sync/substrate/verification/cache.rs
```diff
@@ -163,7 +188,36 @@ where
...
+                    session_id,
+                    &self.session_info,
+                )
+                .ok_or(CacheError::UnknownAuraAuthorities(session_id))?;
+                vacant.insert(authorities)
+            }
+        })
+    }
+
+    /// Returns session verifier for block number if available. Updates cache if necessary.
```
timorleph, line 217: Please add something to this comment that it should be called on the block number of the _parent_ of the block being verified – you have this commented below, but this should also be documented here, especially since it's different than above.

timorleph (approved):
Nice.

finality-aleph/src/sync/substrate/verification/cache.rs
lesniak43: fixed the comments (not sure if you really meant this one though)

finality-aleph/src/sync/substrate/verification/cache.rs
lesniak43: I've refactored this, now should be fine - please check the commit

finality-aleph/src/sync/substrate/verification/cache.rs
```diff
@@ -150,41 +169,58 @@ where
             return Err(CacheError::SessionInFuture(session_id, upper_bound));
         }
 
+        // Prune old session data
```
timorleph, line 172: Move into separate function as it used to be, then you don't need this comment.

timorleph (approved):
Lovely, one code quality regression tho. ^^"

finality-aleph/src/sync/substrate/verification/cache.rs
lesniak43: so be it

PR #1439, opened 2023-10-09
A0-3281: Limit vertex size

timorleph:
# Description

We should have limited vertex size, to not run out of memory if new people keep sending us sync messages.

## Type of change

# Checklist:


ggawryal (approved):

lesniak43 (approved):
exquisite

PR #1441, opened 2023-10-11
A0-3298: Move clique network tests to nightly

Marcin-Radecki:
# Description

* Moved check runtime determinism to nightly. Motivation is that we don't need to check it that often, and it burns computation power. It is enough to check it every 24 hours in nightly.
* In PR pipeline, just check the production binary instead of compiling it. This is because we do not use production binary in this pipeline, just we want to know whether it compiles. Especially it's hard for linking time as it's 70% of compilation time (due to LTO). Hence I moved actual build to nightly as well.
* Move clique network tests to nightly. This is because they are long-running integration tests 
* Now the outlier is check excluded packages, but we'll deal with it in the next PRs.
* Split `clippy` and `cargo test` into two paraller jobs
* There's one required check pending in this PR but this is because of naming in repo settings. I'll deal with it just before this PR is merged.

New workflow runs: 
https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/6481519250
https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/6481519240

## Type of change

# Checklist:

### DIFF
--- a/.github/workflows/_check-production-node-and-runtime.yml
+++ b/.github/workflows/_check-production-node-and-runtime.yml
@@ -1,0 +1,30 @@
+---
+# This workflow checks production version of aleph-node and aleph-runtime, ie
+# only compilation and not linking, which takes most of the time
+name: Check production node and runtime
+on:
+  workflow_call:
+
+jobs:
+  main:
+    name: Build production node and runtime
+    runs-on: [self-hosted, Linux, X64, large]
+    env:
+      RUST_BACKTRACE: full
+      RUSTC_WRAPPER: sccache
+    steps:
+      - name: Checkout aleph-node source code
+        uses: actions/checkout@v4
+
+      - name: Call action get-ref-properties
+        id: get-ref-properties
+        # yamllint disable-line rule:line-length
+        uses: Cardinal-Cryptography/github-actions/get-ref-properties@v2
+
+      - name: Install Rust toolchain
+        uses: Cardinal-Cryptography/github-actions/install-rust-toolchain@v2
+        with:
+          targets: wasm32-unknown-unknown
+
+      - name: Build production binary and runtime
+        run: cargo check --profile production -p aleph-node
--- a/.github/workflows/_check-production-node-and-runtime.yml
+++ b/.github/workflows/_check-production-node-and-runtime.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   main:
-    name: Build production node and runtime
+    name: Check production node and runtime
     runs-on: [self-hosted, Linux, X64, large]
     env:
       RUST_BACKTRACE: full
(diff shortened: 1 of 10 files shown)


.github/workflows/_check-production-node-and-runtime.yml
```diff
@@ -0,0 +1,30 @@
...
+        id: get-ref-properties
+        # yamllint disable-line rule:line-length
+        uses: Cardinal-Cryptography/github-actions/get-ref-properties@v2
+
+      - name: Install Rust toolchain
+        uses: Cardinal-Cryptography/github-actions/install-rust-toolchain@v2
+        with:
+          targets: wasm32-unknown-unknown
+
+      - name: Build production binary and runtime
```
timorleph, line 29: This should probably have a better name, since it doesn't build. Same for the comment at the beginning of this file, as it doesn't even compile. :D

.github/workflows/_check-production-node-and-runtime.yml
```diff
@@ -0,0 +1,30 @@
...
+        run: cargo check --profile production -p aleph-node
```
timorleph, line 30: Hm, all the other files seem to be using `actions-rs/cargo`, while here the command is explicit, is this on purpose?

timorleph (approved):
Faster pipelines, yay! Some nits.

.github/workflows/_check-production-node-and-runtime.yml
Marcin-Radecki: I think it's the other way round, most of examples use bare commands as we use our action to install rust toolchain. We use actions-rs/cargo when invoking `cargo test` ie when `wasm32-unknown-unknown` is not needed. Good point though, we will unify those rust build actions in further PR.

.github/workflows/_unit-tests-and-static-checks.yml
```diff
@@ -28,8 +28,22 @@ jobs:
...
+
+      - name: Install Rust Toolchain
+        uses: Cardinal-Cryptography/github-actions/install-rust-toolchain@v2
+
       - name: Run Unit Test Suite
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --features "liminal-try-runtime liminal-runtime-benchmarks"
+          # yamllint disable-line rule:line-length
```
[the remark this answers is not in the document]
Marcin-Radecki: Removed

.github/workflows/_check-production-node-and-runtime.yml
Marcin-Radecki: Good point, fixed

timorleph (approved):

MikolajGasior (approved):

PR #1446, opened 2023-10-17
A0-3310: Replace double quotes with single quotes

MikolajGasior:
# Description

This PR changes double quotes to single quotes in sub-workflows.  When double quote is not necessary then it should not be used.

## Type of change

Changes the workflow so it needs to be properly reviews as it is not entirely covered by tests.  However, if any errors then we'll spot it straight away.

### DIFF
--- a/.github/workflows/_check-excluded-packages.yml
+++ b/.github/workflows/_check-excluded-packages.yml
@@ -37,6 +37,7 @@ jobs:
         env:
           RUST_TOOLCHAIN: ${{ steps.aleph-client-toolchain-channel.outputs.value }}
           RUST_CONTRACTS_TOOLCHAIN: ${{ steps.contracts-toolchain-channel.outputs.value }}
+        # yamllint disable rule:line-length
         run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/_check-excluded-packages.yml
```diff
@@ -72,13 +73,13 @@ jobs:
           # a version bump
           #
           # cd baby-liminal-extension/
-          # ${CHECK}
```
Marcin-Radecki, line 75: I'm curious, how did it work before

Marcin-Radecki (approved):

.github/workflows/_check-excluded-packages.yml
MikolajGasior: Well, it's an exported variable so it worked just fine.  In this change it actually doesn't make too much difference but let's just make this `${{ env.VARNAME }}` syntax default everywhere.  Not in this line, but it's safer that way and there is no need to expand it with double quote.

Kosiare (approved):

CezaryKierzyk (approved):

PR #1447, opened 2023-10-18
A0-3094: Validators network details rpc call

ggawryal:
# Description

Adding an RPC call to get network details of validators. This only pulls and caches network data that was previously observed by a node, without modifying or adding any new information flow between nodes in the network.

Related PRs: #1448 #1449

<details>
  <summary> RPC query output on local network </summary>
  
  ```curl -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "alephNode_unstable_validatorNetworkInfo", "params": []}' http://localhost:9948/```
  ```
 {
    "jsonrpc": "2.0",
    "result": {
        "5F4H97f7nQovyrbiq4ZetaaviNwThSVcFobcA5aGab6167dK": {
            "session": 1,
            "network_level_address": "127.0.0.1:30346",
            "validator_network_peer_id": "5CwWuuXMCvcSC7hqeMdFSfNFzVGXCAtdSGAMjwV1nLwaqtqE"
        },
        "5GBNeWRhZc2jXu7D55rBimKYDk8PGk8itRYFTPfC8RJLKG5o": {
            "session": 1,
            "network_level_address": "127.0.0.1:30344",
            "validator_network_peer_id": "5H7yZo8EzYkSsZn7MgC4f9vaVYQLgNaS4kEZ29Q7WtiYMVF8"
        },
        "5Dfis6XL8J2P6JHUnUtArnFWndn62SydeP8ee8sG2ky9nfm9": {
            "session": 1,
            "network_level_address": "127.0.0.1:30345",
            "validator_network_peer_id": "5HZFGbT6tsY2iV2YT927wnN4q2TxmPT67SUJE99TqGQsjpGV"
        },
        "5D34dL5prEUaGNQtPPZ3yN5Y6BnkfXunKXXz6fo7ZJbLwRRH": {
            "session": 1,
            "network_level_address": "127.0.0.1:30343",
            "validator_network_peer_id": "5E1YcJyx4u3kQ4UiRPgPXPSvb2vbcCR2DsxS927GM6x8wz85"
        }
    },
    "id": 1
}                        
  ```

</details>

Has been run also on testnet, slowly collects network details for authorities in each session (without `AccountId` because of the old runtime version).

# Type of change

# Checklist:


finality-aleph/src/network/address_cache.rs
```diff
@@ -0,0 +1,180 @@
...
+                NonZeroUsize::try_from(VALIDATOR_ADDRESS_CACHE_SIZE).unwrap(),
+            ))),
+        }
+    }
+
+    pub fn insert(&self, validator_stash: AccountId, info: ValidatorAddressingInfo) {
+        self.data.lock().put(validator_stash, info);
+    }
+
+    pub fn read(&self) -> HashMap<AccountId, ValidatorAddressingInfo> {
```
DamianStraszak, line 67: I think it's better to call it `snapshot`, `get_clone` or some such. Otherwise it might be confused with the `read` method from `RwLock` https://amanieu.github.io/parking_lot/parking_lot/struct.RwLock.html#method.read

bin/node/src/aleph_node_rpc.rs
```diff
@@ -248,6 +272,69 @@ where
...
+    async fn validator_network_info(
+        &self,
+    ) -> RpcResult<HashMap<AccountId, ValidatorAddressingInfo>> {
+        let mut info = self.validator_address_cache.read();
+
+        // This uses unstable method from substrate's API, but there's probably no other easy way
+        // of doing this. On the other hand, the p2p peer_id is only for debuging purposes,
+        // so in case of future substrate API change this if statement can be temporarily safely removed.
+        if let Ok(network_state) = self.network.network_state().await {
+            attach_p2p_network_info_to_validator_addressing_info(&mut info, network_state);
```
DamianStraszak, line 285: I think it's OK to drop this, unless you think it gives good results...

Is it the case that using a combination of the new RPC call (without `potential_p2p_network_additional_info`) and the old `system_unstable_networkState` one could recover roughly the same info?

bin/node/src/aleph_node_rpc.rs
```diff
@@ -156,6 +168,11 @@ pub trait AlephNodeApi<BE> {
     ///
     #[method(name = "ready")]
     fn ready(&self) -> RpcResult<bool>;
+
+    #[method(name = "validatorNetworkInfo")]
+    async fn validator_network_info(
```
DamianStraszak, line 173: There is a notion of `Safe` and `Unsafe` RPC calls in substrate, and how to tell which is which was always a big mystery to me. Anyway, the `system_unstable_networkState` is unsafe, probably because it might output lots of data. I'm wondering whether we could/should make this unsafe too. Being unsafe has this disadvantage that our RPC nodes have only safe methods exposed, so things might be annoying :|

DamianStraszak (approved):
Very nice!

bin/node/src/aleph_node_rpc.rs
timorleph: I think marking it "Unsafe" is a good idea, we haven't analyzed the data returned for vulnerabilities (it's mostly public, but maybe something can be inferred from which exact data is cached?) and no one who cannot run their own node would even be interested in the output anyway. This will make it slightly less convenient to use, but it shouldn't be that much of a problem?

finality-aleph/src/network/address_cache.rs
```diff
@@ -0,0 +1,180 @@
...
+        info: ValidatorAddressingInfo,
+    ) {
+        let block = self
+            .session_boundary_info
+            .boundaries_for_session(session)
+            .first_block();
+
+        if let Some(authority_data) = self.authority_provider.authority_data(block) {
+            let aleph_key = authority_data.authorities()[validator_index.0].clone();
+            if let Some(validator_stash) = self
```
kostekIV, line 128: nit: I would not add _stash here. We do not use stash notion outside of pallet staking related logic

kostekIV, line 128: make fn instead that return an owner for a given block and index.
`fn owner_at_block(block, index) -> Option<Owner>`

finality-aleph/src/network/address_cache.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+/// Network details for a given validator. This data is purely informational
+/// and can change over time, even within a single session.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ValidatorAddressingInfo {
+    /// Network level address of the validator, i.e. IP address
+    pub network_level_address: String,
+    /// PeerId of the validator used in validator (clique) network
+    pub validator_network_peer_id: String,
```
kostekIV, line 28: nit: change String to PeerId

finality-aleph/src/network/address_cache.rs
kostekIV: or instead of block use session also

finality-aleph/src/network/tcp.rs
```diff
@@ -106,6 +110,14 @@ impl AddressingInformation for SignedTcpAddressingInformation {
         self.peer_id()
             .verify(&self.addressing_information.encode(), &self.signature)
     }
+
+    fn lower_level_address(&self) -> String {
+        iter::once(self.addressing_information.primary_address.clone())
+            .chain(self.addressing_information.other_addresses.clone())
+            .filter_map(|address| Some(address.parse::<SocketAddr>().ok()?.ip().to_string()))
+            .next()
+            .unwrap_or("unknown".to_string())
```
kostekIV, line 119: return Option<String> instead of "unknown".

bin/node/src/aleph_node_rpc.rs
ggawryal: Combining these two rpc calls gives exactly the same info. I think I can remove p2p peerid, move the joining part to some python script and put it somewhere in our internal documentation.

DamianStraszak: Yes, so maybe you can indeed drop p2p peerid. 

There is also this issue that sometimes nodes use domain names and not ips, which complicates matters :)

finality-aleph/src/network/address_cache.rs
```diff
@@ -0,0 +1,180 @@
...
+    pub version_string: Option<String>,
+}
+
+/// Stores most recent information about validator addresses.
+#[derive(Debug, Clone)]
+pub struct ValidatorAddressCache {
+    data: Arc<Mutex<LruCache<AccountId, ValidatorAddressingInfo>>>,
+}
+
+const VALIDATOR_ADDRESS_CACHE_SIZE: usize = 300;
```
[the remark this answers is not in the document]
ggawryal: Got it, addressed in #1449

finality-aleph/src/network/address_cache.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+pub trait ValidatorAddressCacheUpdater {
+    /// In session `SessionIndex`, validator `NodeIndex` was using addresses specified in `most_recent_info`.
+    /// A session and validator_index identify the validator uniquely.
+    fn update(
+        &self,
+        session: SessionId,
+        validator_index: NodeIndex,
+        most_recent_info: ValidatorAddressingInfo,
```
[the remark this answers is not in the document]
ggawryal: done in #1449

finality-aleph/src/network/address_cache.rs
```diff
@@ -0,0 +1,180 @@
...
+        Self::new()
+    }
+}
+
+pub trait ValidatorAddressCacheUpdater {
+    /// In session `SessionIndex`, validator `NodeIndex` was using addresses specified in `most_recent_info`.
+    /// A session and validator_index identify the validator uniquely.
+    fn update(
+        &self,
+        session: SessionId,
```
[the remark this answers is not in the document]
ggawryal: good catch, removed in #1449

finality-aleph/src/network/address_cache.rs
ggawryal: renamed in #1449

ggawryal: Refactored as you suggested in #1449, however, after making updating logic optional and introducing an enum this has both up and downsides :/

finality-aleph/src/network/address_cache.rs
ggawryal: I'd rather not do it, moving discussion to #1449

finality-aleph/src/network/tcp.rs
ggawryal: Done in #1449

bin/node/src/aleph_node_rpc.rs
ggawryal: Dropped

bin/node/src/aleph_node_rpc.rs
```diff
@@ -175,16 +194,21 @@ where
         justification_translator: JustificationTranslator,
         client: Arc<Client>,
         sync_oracle: SO,
+        validator_address_cache: ValidatorAddressCache,
+        network: Arc<NetworkService<Block, Hash>>,
```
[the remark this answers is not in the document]
ggawryal: After removing substrate peerid network is not needed at all

bin/node/src/aleph_node_rpc.rs
ggawryal: Now marked as unsafe. Substrate says that `system_unstable_networkState` is unsafe mainly because its output api is not final yet, but they don't work on stabilizing it actively for quite some time.
I tried not to add new flags to cli, so right now that logic is run only for non-validators with rpc enabled.

bin/node/src/aleph_node_rpc.rs
```diff
@@ -156,6 +168,11 @@ pub trait AlephNodeApi<BE> {
     ///
     #[method(name = "ready")]
     fn ready(&self) -> RpcResult<bool>;
+
+    #[method(name = "validatorNetworkInfo")]
```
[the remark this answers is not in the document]
ggawryal: Added

finality-aleph/src/network/address_cache.rs
ggawryal: Renamed to `snapshot` and cloning.
> returning readonly reference

Not sure if I understand this, but I'm not aware of any way of converting `&LruCache<>` to `&Hashmap<>`

bin/node/src/aleph_node_rpc.rs
DamianStraszak: In our infra:
- RPC nodes (archivists) have only Safe RPC methods enabled,
- validators have also Unsafe methods enabled (this is safe, because they are not reachable from outside anyway)

I think ideally we would like this to be possible to call on at least some of nodes in our infra :P And from what you are saying 
> right now that logic is run only for non-validators with rpc enabled.

it would not?

ggawryal: You're right. Changed to be run only on nodes with `rpc` and `telemetry` enabled, and while the latter is not strongly connected to this, it is at least partially imho and can be switched off from `cli`. 
Ideally, I think we should have a separate config file for all such values and consts that have too specific role to have a separate `cli` flag, but can be sometimes useful to be modified. I'll consult what the rest of the team thinks, but even if we agree to that solution, it would be a separate task.

DamianStraszak: I think `telemetry` is quite unrelated, let's not mix it in, because that adds a strange constraint. 

The other thing I'm not quite sure is:

```
   let validator_address_cache = match (&config.role, config.telemetry_endpoints.is_some()) {
        (sc_network::config::Role::Full, true) => Some(ValidatorAddressCache::new()),
```
above what does it really mean `sc_network::config::Role::Full`, i.e., for instance in our infra:
- are our archivists full?
- are our validators full?

ggawryal: So the only option I see is that I'm adding a cli flag. 
Also removed criterion based on `Role`, but fyi there are two roles:
* Role::Authority for nodes that can participate in consensus (our validators)
* Role::Full for nodes that don't (our archivists)

It's not that easy however, because there is also another struct named `ObservedRole`, which refers to other nodes from our point of view, and it also includes light role (like light clients). To my understanding, we don't care about us being light clients there because `aleph-node` codebase is not supposed to support being light clients (we'd need a separate repo for that).

DamianStraszak: I think the current version is good, but there is one downside of having an additional flag -- a change must be introduced to our infra as well, which imho makes it troublesome. So there are two options:
1. Get rid of the flag, and enable this for all nodes. 
2. Keep the current version, but then need to coordinate with infra, on what changes must be made along the release of this. 

If I had to decide, I would just go with 1. because it's simplest, and the overhead is negligible anyway -- @timorleph @kostekIV ?

kostekIV: Option 1 sounds good enough for me although imho 2 would be ideal

ggawryal: Hmm, how about option 1.5: leave the flag, but enable by default?

DamianStraszak: Yes, I like it :)

DamianStraszak: Fee free to resolve the comment yourself @ggawryal 

I'm ok with merging

PR #1448, opened 2023-10-19
A0-3363: Runtime api for aleph key owner

ggawryal:
# Description

Add runtime api for getting owner of aleph key.

## Type of change

# Checklist:

### DIFF
--- a/primitives/src/lib.rs
+++ b/primitives/src/lib.rs
@@ -283,6 +283,8 @@ sp_api::decl_runtime_apis! {
             session: SessionIndex
         ) -> Result<SessionCommittee<AccountId>, SessionValidatorError>;
         fn next_session_aura_authorities() -> Vec<AuraAuthorityId>;
+        /// Allows to get owner's AccountId for an Aleph key used in the current session.
+        fn aleph_key_owner(key: AuthorityId) -> Option<AccountId>;
     }
 }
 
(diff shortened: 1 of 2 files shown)


primitives/src/lib.rs
```diff
@@ -283,6 +283,8 @@ sp_api::decl_runtime_apis! {
             session: SessionIndex
         ) -> Result<SessionCommittee<AccountId>, SessionValidatorError>;
         fn next_session_aura_authorities() -> Vec<AuraAuthorityId>;
+        /// Allows to get owner's AccountId for an Aleph key used in the current session.
+        fn aleph_key_owner(key: AuthorityId) -> Option<AccountId>;
```
deuszx, line 287: Is the `aleph_` prefix necessary? I understand that it's b/c this is from the "aleph sesion" but I haven't heard such concept before (maybe my lack of knowledge) and it seems that no other method in this API (`AlephSessionAPI`) is prefixed with it either.

deuszx (approved):
Feel free to ignore if I'm wrong.

kostekIV (approved):

primitives/src/lib.rs
```diff
@@ -283,6 +283,8 @@ sp_api::decl_runtime_apis! {
             session: SessionIndex
         ) -> Result<SessionCommittee<AccountId>, SessionValidatorError>;
         fn next_session_aura_authorities() -> Vec<AuraAuthorityId>;
+        /// Allows to get owner's AccountId for an Aleph key used in the current session.
```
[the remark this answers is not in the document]
ggawryal: You're right, I was thinking from the `pallet_session`'s pov, in which this is called `aleph_key`, as opposed to `aura_key`. I agree that your suggestion is more clear generally.

primitives/src/lib.rs
ggawryal: Yup, I wanted to distinguish `aleph_key` from `aura_key`, but this is rather unnecessary in `aleph` api.

PR #1449, opened 2023-10-20
A0-3364: cache network details of validators

ggawryal:
# Description

Adding structures that will be later helpful to construct RPC call to get network details of validators. This is a part 2/3 of the whole task, see #1447). This only pulls and caches network data that was previously observed by a node, without modifying or adding any new information flow between nodes in the network.

Follow up to #1448 

## Type of change

# Checklist:

- I have added tests


finality-aleph/src/network/address_cache.rs
```diff
@@ -0,0 +1,184 @@
...
+
+/// Network details for a given validator in a given session.
+#[derive(Debug, Clone)]
+pub struct ValidatorAddressingInfo {
+    /// Session to which given information applies.
+    pub session: SessionId,
+    /// Network level address of the validator, i.e. IP address (for validator network)
+    pub network_level_address: Option<String>,
+    /// PeerId of the validator used in validator (clique) network
+    pub validator_network_peer_id: String,
```
ggawryal, line 28: > nit: change String to PeerId

Validator network `PeerId` is a trait, implemented for tcp network addressing info, so we'd have to add a parameter to that struct (and others in that file), and even worse, since this struct will be passed to rpc, we'd have to specialize network type in runtime instead of `finality-aleph`.

kostekIV: ok, fair

finality-aleph/src/network/session/manager.rs
```diff
@@ -308,20 +324,29 @@ impl<NI: NetworkIdentity, D: Data> Manager<NI, D> {
...
         match self.sessions.get_mut(&session_id) {
             Some(Session {
                 handler, discovery, ..
             }) => {
                 let (maybe_address, maybe_message) =
                     discovery.handle_authentication(message, handler);
-                let maybe_command = match (maybe_address, handler.is_validator()) {
-                    (Some(address), true) => {
+                let mut maybe_command = None;
+                if let Some(address) = maybe_address {
```
kostekIV, line 335: is this correct? Wont it catch some nodes that do not have a validator role?

kostekIV (commented):
Looks great, one last comment

finality-aleph/src/network/session/manager.rs
ggawryal: In my understanding `handler.is_validator()` is with respect to self, so we should ignore it. And `discovery.handle_authentication` returns `Some(address)` iff an authentication has been validated successfully, and its author is a validator (and we are not that validator) in a given session, so I think this is correct.

kostekIV (approved):

clique/src/lib.rs
```diff
@@ -58,6 +58,9 @@ pub trait AddressingInformation: Debug + Hash + Codec + Clone + Eq + Send + Sync
 
     /// Verify the information.
     fn verify(&self) -> bool;
+
+    // Address used by the underlying protocol, for debugging purposes.
+    fn lower_level_address(&self) -> Option<String>;
```
timorleph, line 63: Either just call it `address` and have the comment say that it's _only_ used for debugging/logging (slightly more forcefully than now), or call it something that corresponds to the function better (I'm not sure what it would be, I only have extremely generic ideas like `additional_data`, which make it very clear that this does not impact behaviour, but are otherwise atrocious; just pointing out this option if you have some good idea).

finality-aleph/src/network/tcp.rs
```diff
@@ -106,6 +110,15 @@ impl AddressingInformation for SignedTcpAddressingInformation {
         self.peer_id()
             .verify(&self.addressing_information.encode(), &self.signature)
     }
+
+    // Returns first address (without port) from addressing information that parses to IP address,
+    // without performing dns resolution.
+    fn lower_level_address(&self) -> Option<String> {
+        iter::once(self.addressing_information.primary_address.clone())
+            .chain(self.addressing_information.other_addresses.clone())
+            .filter_map(|address| Some(address.parse::<SocketAddr>().ok()?.ip().to_string()))
```
timorleph, line 119: Uhhhh, I thought this was wrong since we support DNS addresses, not just plain IP, but then I checked and we don't. o__o''' Well, there should probably be a task for improving that?

finality-aleph/src/network/address_cache.rs
```diff
@@ -0,0 +1,184 @@
...
+    data: Arc<Mutex<LruCache<AccountId, ValidatorAddressingInfo>>>,
+}
+
+const VALIDATOR_ADDRESS_CACHE_SIZE: usize = 1000;
+
+impl ValidatorAddressCache {
+    pub fn new() -> Self {
+        Self {
+            data: Arc::new(Mutex::new(LruCache::new(
+                NonZeroUsize::try_from(VALIDATOR_ADDRESS_CACHE_SIZE).unwrap(),
```
timorleph, line 43: Nit: `expect("the cache size is a non-zero constant")`.

finality-aleph/src/network/address_cache.rs
```diff
@@ -0,0 +1,184 @@
...
+                ..
+            },
+        ) = (self.owner_at_session(info.session, validator_index), self)
+        {
+            validator_address_cache.insert(validator, info)
+        }
+    }
+}
+
+pub struct KeyOwnerInfoProviderImpl<C, B, BE>
```
timorleph, line 137: Since you are producing a struct here anyway, perhaps include session boundaries and the authority provider here as well (at least the latter is already implemented for this kind of client, innit?).

Alternatively, if you really want to keep these as separate pieces in the implementation above then maybe you can remove this struct and do the `impl` for just `Arc<C>`? Or is it not possible with the additional constraints?

finality-aleph/src/network/address_cache.rs
```diff
@@ -0,0 +1,184 @@
...
+impl<C, B, BE> KeyOwnerInfoProvider for KeyOwnerInfoProviderImpl<C, B, BE>
+where
+    C: ClientForAleph<B, BE> + Send + Sync + 'static,
+    C::Api: crate::aleph_primitives::AlephSessionApi<B>,
+    B: Block<Hash = BlockHash>,
+    B::Header: Header<Number = BlockNumber>,
+    BE: Backend<B> + 'static,
+{
+    fn aleph_key_owner(&self, block_number: BlockNumber, key: AuthorityId) -> Option<AccountId> {
+        let block_hash = self.client.block_hash(block_number).ok()??;
```
timorleph, line 173: This is technically risky, as getting a block hash for an arbitrary number might give you one that is not finalized yet, and thus the data is unreliable. I'm pretty sure this is not a problem due to how this is being called, but this depends on the code in `owner_at_session`. If you add more things to this struct as I suggested in the previous comment I think this will be fine (since it is correct based on the local code), but if it remains a separate struct  please add a comment here about only running this on finalized blocks.

finality-aleph/src/network/address_cache.rs
```diff
@@ -0,0 +1,184 @@
...
+        self.client.runtime_api().key_owner(block_hash, key).ok()?
+    }
+}
+
+#[cfg(test)]
+pub struct MockValidatorCacheUpdater;
```
timorleph, line 179: Maybe instead expose the `Noop` variant through a `noop()` constructor (if you want it can also be marked `#[cfg(test)]`)? If we already have a noop we don't need a separate noop mock.

finality-aleph/src/network/address_cache.rs
timorleph: Also minor point – we try to keep the substrate-specific impls of these kinds of traits in separate files. I'm not sure if this is the best idea here (might make stuff more confusing for no good reason), but if you see an obvious way to make it happen that would be nice.

bin/node/src/service.rs
```diff
@@ -399,6 +399,8 @@ pub fn new_authority(
             .unwrap_or(usize::MAX),
     };
 
+    let validator_address_cache = None;
```
timorleph, line 402: Hmm, why is this `Option`al?

timorleph (changes requested):
Cool stuff, some comments, although many of them ended up being almost one comment written in multiple places. :P

finality-aleph/src/network/tcp.rs
ggawryal: Here I wanted to use the former, since as a caller of `lower_level_address` I'd expect its implementation to not use relatively expensive calls.

timorleph: Ah, you are right, it does that pretty deep in the code... Hm, could you then not filter for IP here? why are you doing that in the first place?

bin/node/src/service.rs
ggawryal: Most users won't be interested in getting these network details, and we will need it only for rpc nodes. The decision if we create a `Noop` updater or the real one must be done in `bin/`, but the updater should be created in `finality-aleph`, so I've chosen to make `validator_address_cache` (which also must be created in `bin/`, as it will be used in rpc) optional and select the appropriate variant of updater based on that.

finality-aleph/src/network/tcp.rs
timorleph: Ah, I only saw the first response initially. Well, you could also assume the addresses are already correct and just return the primary one, no? It's a `String` already.

ggawryal: Sure, I will do that. Previously I needed some canonical form because I was joining by IP address to get substrate's network details, but we've decided that we don't need that, so I can simplify also that.

finality-aleph/src/network/address_cache.rs
ggawryal: Updated

finality-aleph/src/network/address_cache.rs
ggawryal: Done, although I had to add dummy impl for `ValidatorIndexToAccountIdConverter`, because that noop variant is parametrized with that trait.

finality-aleph/src/network/address_cache.rs
ggawryal: After including additional structs I think we are safe as you said, because we are querying session keys from the first block of a given session, which must have been set at least in previous session.

finality-aleph/src/network/address_cache.rs
```diff
@@ -0,0 +1,107 @@
...
+
+#[cfg(test)]
+pub mod test {
+    use crate::{
+        idx_to_account::MockConverter,
+        network::address_cache::{validator_address_cache_updater, ValidatorAddressCacheUpdater},
+    };
+
+    pub fn noop_updater() -> impl ValidatorAddressCacheUpdater {
+        validator_address_cache_updater(None, MockConverter)
```
timorleph, line 105: This is a somewhat convoluted way of doing this, why not just have it return `ValidatorAddressCacheUpdaterImpl::Noop` explicitly?

timorleph (approved):
Minor comments left.

finality-aleph/src/network/address_cache.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+impl Default for ValidatorAddressCache {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+pub trait ValidatorAddressCacheUpdater {
```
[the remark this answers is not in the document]
ggawryal: Impl is parametrized with `ValidatorIndexToAccountIdConverter` and trait is not, so imho it's better to have higer level classes in `network/` parametrized by `Updater` rater than `Converter`

finality-aleph/src/network/address_cache.rs
ggawryal: updated

finality-aleph/src/network/address_cache.rs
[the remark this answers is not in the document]
timorleph: Weird preference considering how much boilerplate it generates, but sure.

finality-aleph/src/idx_to_account.rs
```diff
@@ -0,0 +1,83 @@
...
+where
+    C: ClientForAleph<B, BE> + Send + Sync + 'static,
+    C::Api: crate::aleph_primitives::AlephSessionApi<B> + AuraApi<B, AuraId>,
+    B: Block<Hash = BlockHash>,
+    BE: Backend<B> + 'static,
+{
+    client: Arc<C>,
+    session_boundary_info: SessionBoundaryInfo,
+    authority_provider: AuthorityProviderImpl<C, B, BE>,
+    _phantom: PhantomData<(B, BE)>,
```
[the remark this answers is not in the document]
ggawryal: Good catch

PR #1457, opened 2023-10-24
A0-3256: Check for equivocations

lesniak43:
# Description

Detecting Aura equivocations, `panic!` when self-equivocating

## Type of change

# Checklist:

- I have added tests


timorleph (approved):
Looks good, one question – is the `None` variant for `AccountId` only there because of runtime APIs not supporting this in the past? Can this be fixed with our new planned approach to this API, and if yes is there a task for that?

finality-aleph/src/sync/service.rs
```diff
@@ -313,6 +313,15 @@ where
         }
     }
 
+    fn process_equivocation_proofs(&self, proofs: Vec<V::EquivocationProof>) {
+        for proof in proofs {
+            warn!(target: LOG_TARGET, "Equivocation detected: {proof}");
+            if proof.are_we_equivocating() {
+                panic!("We are equivocating, which is ILLEGAL - shutting down the node. This is probably caused by running two instances of the node with the same set of credentials. Make sure that you are running ONLY ONE instance of the node. If the problem persists, contact the Aleph Zero developers on Discord.");
```
kostekIV, line 320: we really should have just a page with a contact info and just past a link here -.-

finality-aleph/src/sync/substrate/verification/cache.rs
```diff
@@ -76,6 +76,7 @@ where
     H: Header,
 {
     cached_data: HashMap<SessionId, CachedData>,
+    pub equivocation_cache: HashMap<u64, (H, bool)>,
```
kostekIV, line 79: does it need to be a pub?

finality-aleph/src/sync/service.rs
lesniak43: I guess that we might create a task for this? Although I've talked with @DamianStraszak and he says this should be fine.

DamianStraszak: I guess it's fine, it's informative and practical. I don't know if we have any page with such contact information.

kostekIV: A task would be also about creating such a page. For this pr here its fine for now

lesniak43:
@timorleph yes, and no - I'll remember to create the task as soon as we have the related epic (so that we don't have to prioritize the task right now)

finality-aleph/src/sync/substrate/verification/cache.rs
lesniak43: ofc not, refactored the code slightly

kostekIV (approved):

PR #1460, opened 2023-10-27
A0-3390: Prepare for proposal compatibility

timorleph:
# Description

Split `data_io` into two parts in a way which will make changing the type of proposals much simpler. This is mostly moving/copying files (`legacy` is essentially a copy of `data_io`), but some small additional refactors also happened to make this simpler and the change less invasive.

## Type of change

# Checklist:


lesniak43 (approved):
nothing changed, good job!

finality-aleph/src/party/manager/task.rs
```diff
@@ -47,3 +48,34 @@ impl Task {
...
+    pub session_id: u32,
+}
+
+#[async_trait::async_trait]
+pub trait Runnable: Send + 'static {
+    async fn run(self, exit: oneshot::Receiver<()>);
+}
+
+/// Runs the given task within a single session.
+pub fn task<R: Runnable>(subtask_common: TaskCommon, runnable: R, name: &'static str) -> Task {
```
kostekIV, line 65: nit: I would put the name parameter inside the `Runnable` trait but it is purely cosmetic, feel free to ifnore

kostekIV (approved):
looks good to me

finality-aleph/src/party/manager/task.rs
timorleph: Hmm, I thought about this a bit, but I think it's better to leave the naming decision to the invoking party. It did make me notice one silly thing in nearby code though. <_<"

finality-aleph/src/data_io/mod.rs
```diff
@@ -9,6 +9,9 @@ mod data_store;
 mod proposal;
 mod status_provider;
 
+/// This is only temporary so we can change the proposal type once. Should be removed after that is done, and only the current version should be used.
```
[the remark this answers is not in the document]
timorleph: Now we do, and I added a reference to it here, thanks for pointing that out.

PR #1463, opened 2023-10-30
A0-3430: Update ink deps to latest ink4-compatible releases

deuszx:
# Description

Updates ink-releated dependencies in `aleph-client`. There were no changes in the API or in the code, purely dependency bump. This update is necessary to align `aleph-client` with `ink-wrapper` which works with `ink` and `aleph-client` as a backend.

The versions of dependencies used are going to be the last compatible with ink4

## Type of change

- Dependency versions update.

# Checklist:

- I have bumped aleph-client version
- I have bumped cliain verison

### DIFF
--- a/aleph-client/Cargo.lock
+++ b/aleph-client/Cargo.lock
@@ -105,7 +105,7 @@ dependencies = [
 
 [[package]]
 name = "aleph_client"
-version = "3.7.1"
+version = "3.7.2"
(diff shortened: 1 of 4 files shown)


kostekIV (approved):
If the ci will pass looks good to me

woocash2 (approved):

PR #1465, opened 2023-10-30
A0-3450: New runtime api

kostekIV:


lesniak43 (approved):

bin/runtime/src/lib.rs
```diff
@@ -1017,7 +1017,7 @@ impl_runtime_apis! {
             CommitteeManagement::predict_session_committee_for_session(session)
         }
 
-        fn next_session_aura_authorities() -> Vec<AuraId> {
+         fn next_session_aura_authorities() -> Vec<AuraId> {
```
timorleph, line 1020: Huh, this seems wrong...?

finality-aleph/src/runtime_api.rs
```diff
@@ -0,0 +1,110 @@
...
+
+        let encoded = match self
+            .client
+            .storage(at_block, &sc_client_api::StorageKey(storage_key))
+        {
+            Ok(Some(e)) => e,
+            _ => return Err(ApiError::NoStorage(pallet.to_string(), item.to_string())),
+        };
+
+        D::decode(&mut encoded.0.as_ref()).map_err(|_| ApiError::DecodeError)
```
timorleph, line 70: Shouldn't this be `decode_all`?

finality-aleph/src/runtime_api.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+}
+
+#[derive(Clone, Debug)]
+pub enum ApiError {
+    NoStorage(String, String),
+    DecodeError,
```
timorleph, line 77: This should probably wrap around the decode error rather than discard it.

timorleph (approved):
Even if this doesn't end up completely conforming to the ADR it's a good starting PR anyway.

bin/runtime/src/lib.rs
kostekIV: good catch, fixed :P

finality-aleph/src/runtime_api.rs
kostekIV: once agin great catch :) fixed

PR #1468, opened 2023-10-31
Fix for too many nonfinalized blocks at the start of the node

kostekIV:


timorleph (approved):
Lovely. Maybe add a comment about this likely not being needed after A0-2984?

kostekIV:
> Lovely. Maybe add a comment about this likely not being needed after A0-2984?

added

bin/node/src/service.rs
```diff
@@ -64,6 +64,21 @@ impl<N: BaseArithmetic> BackoffAuthoringBlocksStrategy<N> for LimitNonfinalized
...
+fn revert_unfinalized_blocks(client: &FullClient) {
+    let last_finalized = client.info().finalized_number;
+    let best_block = client.info().best_number;
+    // 20 so there are no problems with finality session
+    const REVERT_THRESHOLD: u32 = 20;
+
+    let blocks_count_to_revert = best_block.saturating_sub(last_finalized + REVERT_THRESHOLD);
+
+    debug!("Reverting {} not finalized blocks", blocks_count_to_revert);
+    let new_best = client
```
lesniak43, line 77: I think it returns the number of blocks that were reverted, not the new best block number.

lesniak43 (approved):

bin/node/src/service.rs
```diff
@@ -64,6 +64,22 @@ impl<N: BaseArithmetic> BackoffAuthoringBlocksStrategy<N> for LimitNonfinalized
...
+/// This will not be needed after A0-2984 is done
+fn revert_unfinalized_blocks(client: &FullClient) {
+    let last_finalized = client.info().finalized_number;
+    let best_block = client.info().best_number;
+    // 20 so there are no problems with finality session
+    const REVERT_THRESHOLD: u32 = 20;
+
+    let blocks_count_to_revert = best_block.saturating_sub(last_finalized + REVERT_THRESHOLD);
+
+    debug!("Reverting {} not finalized blocks", blocks_count_to_revert);
```
DamianStraszak, line 76: Could you make this `info`? And also add `last_finalized`, `best_block` to the log message?

bin/node/src/service.rs
```diff
@@ -64,6 +64,22 @@ impl<N: BaseArithmetic> BackoffAuthoringBlocksStrategy<N> for LimitNonfinalized
...
+    let new_best = client
+        .revert(blocks_count_to_revert)
+        .expect("Error while reverting blocks: ");
+    debug!("Reverted chain to block #{}", new_best);
```
DamianStraszak, line 80: Same here, please make this `info`

bin/node/src/service.rs
DamianStraszak: Also, let's wrap this into `if blocks_count_to_revert > 0`

bin/node/src/service.rs
kostekIV: good catch, fixed

bin/node/src/service.rs
kostekIV: done

bin/node/src/service.rs
kostekIV: done

PR #1469, opened 2023-10-31
A0-3422: Fix major sync finalization problems

timorleph:
# Description

There was a race making finalization sometimes stop if it fell too far behind the head of the imported chain, this works around that problem. It also slows down some requests, to lower the amount of log spamming during major sync (still spammy tho).

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/sync/handler/mod.rs
+++ b/finality-aleph/src/sync/handler/mod.rs
@@ -435,7 +435,7 @@ where
         header: J::Header,
     ) -> Result<(), <Self as HandlerTypes>::Error> {
         if let Err(e) = self.forest.update_body(&header) {
-            if matches!(e, ForestError::TooNew) {
+            if matches!(e, ForestError::TooNew | ForestError::ParentNotImported) {
                 self.missed_import_data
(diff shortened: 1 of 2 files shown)


finality-aleph/src/sync/handler/mod.rs
```diff
@@ -435,7 +435,7 @@ where
         header: J::Header,
     ) -> Result<(), <Self as HandlerTypes>::Error> {
         if let Err(e) = self.forest.update_body(&header) {
-            if matches!(e, ForestError::TooNew) {
```
kostekIV, line 438: question: in case of ParentNotImported dont we want to do `header.id().number()`? I dont remember how exactly missed_data works

timorleph: The abstraction is that it remembers the highest block it failed to import and refetches at most up to that. This was supposed to work with just stuff missed because of being too new, but this only adds the missing parent problem, treating it exactly the same – it also saves the number.

kostekIV (approved):

PR #1470, opened 2023-10-31
A0-3422: Deal with too many nonfinalized blocks on startup

timorleph:
# Description

Make sync handler understand having too many nonfinalized blocks on startup instead of panicking.

## Type of change

# Checklist:

### DIFF
--- a/bin/runtime/Cargo.toml
+++ b/bin/runtime/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aleph-runtime"
-version = "0.12.0"
+version = "0.12.1"
 license = "GPL-3.0-or-later"
 authors.workspace = true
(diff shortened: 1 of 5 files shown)


bin/runtime/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aleph-runtime"
-version = "0.12.0"
```
Marcin-Radecki, line 3: My bad, I forgot this hotfix we do not touch runtime, so ideally we remove this before merge. If this might delay things though, it's fine to leave as it is.

Marcin-Radecki (approved):
Approving on the behalf it was tested.

kostekIV (approved):

PR #1471, opened 2023-11-02
A0-3310: Tidy up double quotes

MikolajGasior:
# Description

Removes unnecessary double quotes in the workflow files.  Also, tidies up some variable names and make use of `${{ env.VAR }}` everywhere, for the consistency and security.

## Type of change

Please review carefully.  We cannot break deployment workflows.

# Checklist:

### DIFF
--- a/.github/actions/get-docker-image-names/action.yml
+++ b/.github/actions/get-docker-image-names/action.yml
@@ -61,6 +61,7 @@ runs:
         COMMIT_TAG: ${{ steps.get-ref-properties.outputs.tag }}
         ECR: ${{ inputs.ecr-repo }}
         DOCKERHUB: ${{ inputs.dockerhub-repo }}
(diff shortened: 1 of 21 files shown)


Kosiare (approved):

Marcin-Radecki (approved):
While I agree with enclosing all github expressions in single quotes, I think it'll be better if we stay with double quotes in `bash` snippets. This is because, in general in bash it is better to use double quotes, from our bash style guide:

> This prevents reinterpretation of all special characters within the quoted string — except $, ` (backquote), and \ (escape).
Keeping $ as a special character within double quotes permits referencing a quoted variable  ("$variable"), that is, replacing the variable with its value.

so now if we stay in single quotes in bash snippets, and someone will add bash some variable `${var}`, it will be directly interpreted as `${var}` string

PR #1472, opened 2023-11-02
A0-3432: run_nodes.sh revamp, in particular run bootnode as RPC node

Marcin-Radecki:
# Description

This PR revamps `scripts/run_nodes.sh` so that
* there are only two categories of nodes: RPC node and validator node
* boot node is always the first RPC node
* number of either node type is configurable, but sum must not exceed 10,
* script cmd line args was reworked which is a breaking change
* by default, script ideas all the following: builds `aleph-node`, bootstraps chain, purges db, removes AlephBFT backups. User can opt-out from either of these, but obviously some of them must be present if others are etc, so the script checks all that
* script now logs with timestamps and colors, which is justifies as the script is not intended to run in CI
* as the script is a front card of our repo and most likely the first point of contact with the new user, it contains now a compact intro
to chain bootstrap in general
* `jq` is now needed, though is not hard blocker for me to revert to previous behaviour, now it is just cleaner in bash to parse output of `aleph-node key inspect`
* base path default was changed from `/tmp/` to local `./run-nodes-sh`, this is because the user must have write permissions in the repo he or she just cloned, and it's better to have a base path that is not shared with any other files/dirs.
* arguments were synced with `docker/docker_entrypoint.sh`; they are indeed duplicated now but it does not make sense to extract them to common .sh as at some point `run_nodes.sh` might become obsolete due to dockerization

## Type of change


scripts/run_nodes.sh
```diff
@@ -1,123 +1,289 @@
-#!/bin/bash
+#!/usr/bin/env bash
+
+# This script runs locally aleph-node consensus, ie set of aleph-node processes that are either
+# validators (create blocks) or RPC nodes (read only nodes). By default, such consensus consist of
+# one RPC node that is a bootnode, and 6 validator nodes.
```
ggawryal, line 5: Why do we increase validators to 6? I think in most scenarios, 4 is sufficient, no?

scripts/run_nodes.sh
```diff
@@ -1,123 +1,289 @@
...
+    "${ALEPH_NODE}" bootstrap-node --base-path "${BASE_PATH}/${rpc_node_account_id}" --account-id "${rpc_node_account_id}" --chain-type local > /dev/null
+  done
+
+  if [[ "${DONT_REMOVE_ABFT_BACKUPS}" == "true" ]]; then
+    non_empty_backups=$(find "${BASE_PATH}" -type d -name "*backup-stash*" -not -empty)
+    if [[ -n "${non_empty_backups}" ]]; then
+      warning "Found following non-empty ABFT backups in base path:"
+      warning "${non_empty_backups}"
+      warning "In 99% you want them to be removed when bootstraping chain"
+      warning "Re-run this script with without flag --dont-remove-abtf-backups if you want them to be removed."
```
fixxxedpoint, line 252: `with without`

scripts/run_nodes.sh
```diff
@@ -1,123 +1,289 @@
...
+        if specified, use given base path (keystore, db, AlephBFT backups)
+        if not specified, base path is ./run-nodes-local
+    [--dont-bootstrap]
+      set if you don't want to bootstrap chain, ie generate keystore and chainspec
+    [--dont-build]
+       set if you do not want to build testing aleph-node binary
+    [--dont-delete-db]
+      set to not delete database
+    [--dont-remove-abtf-backups]
+      set to not delete AlephBFT backups; by default there are removed since
```
ggawryal, line 69:
```suggestion
      set to not delete AlephBFT backups; by default they are removed since
```

scripts/run_nodes.sh
ggawryal, line 252:
```suggestion
      warning "Re-run this script without flag --dont-remove-abtf-backups if you want them to be removed."
```

scripts/run_nodes.sh
```diff
@@ -1,123 +1,289 @@
...
-    v) N_VALIDATORS=${OPTARG};;
-    n) N_NON_VALIDATORS=${OPTARG};;
-    b) BUILD_ALEPH_NODE=${OPTARG};;
-    p) BASE_PATH=${OPTARG};;
-    l) N_LISTENERES=${OPTARG};;
-    r) BOOTSTRAP=${OPTARG};;
-    *)
+VALIDATORS=${VALIDATORS:-6}
+RPC_NODES=${RPC_NODES:-1}
+BASE_PATH=${BASE_PATH:-"./run-nodes-local"}
```
ggawryal, line 78: Please use relative address with respect to the script location instead of `./`, or assert before creating any backup etc., that the user is in main `aleph-node` directory

scripts/run_nodes.sh
```diff
@@ -1,123 +1,289 @@
...
+    validator_account_id="${validator_account_ids[$i]}"
+    "${ALEPH_NODE}" purge-chain --base-path "${BASE_PATH}/${validator_account_id}" --chain "${BASE_PATH}/chainspec.json" -y > /dev/null 2>&1
+  done
+fi
 
-for i in $(seq 0 "$(( N_VALIDATORS + N_NON_VALIDATORS - 1 ))"); do
-  run_node "$i" true
+if [[ -z "${DONT_REMOVE_ABFT_BACKUPS}" ]]; then
+  info "Removing AlephBFT backups."
+  find "${BASE_PATH}" -type d -name "backup-stash" | xargs rm -rf
```
fixxxedpoint, line 276: Can we use a loop here instead (also for sake of consistency)? It has potential for removing too many (e.g. someones backup of a backup) or too few folders.

fixxxedpoint (changes requested):

scripts/run_nodes.sh
ggawryal, line 276: Since we use `rm -rf` with `find` in some user-specified dir, could we at least add `-mindepth 2 -maxdepth 2`? Also, a note that may be worth writing is that this removes all backups, not only for `RPC_NODES + VALIDATORS` from the current run.

ggawryal (commented):
Great job, I like the idea of keeping that script well-documented and user-friendly as a first contact point with the repo. Left minor comments. 
Also, maybe @fixxxedpoint will know more context, there was some issue why we disabled purging backups by default (A0-2846). Nevertheless, I think we should be good unless user does something very strange.

scripts/run_nodes.sh
ggawryal: Or use loop as suggested by **fixxxedpoint**

scripts/run_nodes.sh
Marcin-Radecki: To make committee rotation feature actually applied, as default `CommitteeSize` is 4 reserved and 0 non-reserved

scripts/run_nodes.sh
```diff
@@ -1,123 +1,289 @@
...
+    -laleph-network=debug
+    -lnetwork-clique=debug
+    -laleph-finality=debug
+    -laleph-justification=debug
+    -laleph-data-store=debug
+    -laleph-updater=debug
+    -laleph-metrics=debug
+  )
 
+   ./target/release/aleph-node "${node_args[@]}"  2> "${node_name}.log" > /dev/null &
```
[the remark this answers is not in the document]
Marcin-Radecki: Good point, done

scripts/run_nodes.sh
```diff
@@ -1,123 +1,289 @@
...
 
-if $BUILD_ALEPH_NODE ; then
-  cargo build --release -p aleph-node --features "short_session enable_treasury_proposals"
+# ------------------------- input checks ----------------------------------
+
+if [[ "${VALIDATORS}" -lt 4 ]]; then
+  error "Number of validators should be at least 4!"
+fi
+if [[ "${RPC_NODES}" -lt 1 ]]; then
+  error "Number of validators should be at least 1!"
```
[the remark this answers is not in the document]
Marcin-Radecki: Done

scripts/run_nodes.sh
Marcin-Radecki: Fixed

scripts/run_nodes.sh
Marcin-Radecki: Fixed

scripts/run_nodes.sh
```diff
@@ -1,123 +1,289 @@
...
+#
+# They are 3 set of ports you need to have opened and free on your machine in order to run consensus:
+#  * RPC port - range [9944; 9954) - used for JSON RPC protocol
+#  * P2p port - range [30333; 30343) - used for P2P peer network
+#  * Validator port - range [30343; 30353) - used for consensus mechanism (AlephBFT)
+#
+# You need to have installed following prerequisites in order to use that script:
+#   * jq
+#
+# This script also accepts env variables instead arguments, see --help for details. All arguments
```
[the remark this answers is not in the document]
Marcin-Radecki: Fixed

scripts/run_nodes.sh
Marcin-Radecki: Fixed

scripts/run_nodes.sh
Marcin-Radecki: Good point, fixed

scripts/run_nodes.sh
Marcin-Radecki: Done

fixxxedpoint: Do we still need to use that `find` here then? If it is somehow unnecessary difficult without it then leave it. Since you have account-ids you could build those paths by yourself? IMHO it is nicer (and sometimes safer) when a clean-up script "knows" explicitly what needs to be removed, instead of searching/guessing.

Marcin-Radecki: Good point, fixed

ggawryal (approved):

fixxxedpoint (approved):
Well done!

PR #1473, opened 2023-11-02
A0-2533: Integrate abft

woocash2:
# Description

Use the newest `AlephBFT` version in `aleph-node`. The newest `AlephBFT` provides a changed `rmc` and `alerter`, but most importantly, the **Full unit backup** feature.
* Swap versions of `aleph-bft` packages in .toml files
* Bump `LEGACY_FINALITY_VERSION` in `primitives`
* Adjust `finality-aleph` with respect to the different abft versions
* Adjust `aggregator` with respect to newer `rmc`
* Add local test for the "all die, not all recover" scenario

## Type of change

# Checklist:

- I have added tests
- Performed Nightly workflows: [e2e-tests](https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/6800048531), [integration-tests](https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/6800050394), [logic-e2e](https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/6800051897), [normal-session-e2e](https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/6800054306)
- Ran a featurenet on 10 nodes for 3h and observed proper finalization
- Ran `test_recover_after_abft_update` local test
- Locally tested the full scenario of updating aleph-node and on-chain finality version


aggregator/src/aggregator.rs
```diff
@@ -160,48 +149,44 @@ impl<
...
         }
-        self.multicast
-            .start_multicast(SignableHash::new(hash))
-            .await;
+        if let Some(multisigned) = self.rmc_service.start_rmc(hash) {
+            self.multisigned_events.push_back(multisigned);
+        }
     }
 
     async fn wait_for_next_signature(&mut self) -> IOResult {
```
woocash2, line 157: Explanation:
The interface of `rmc` has changed. 
Previous interface which treated obtaining multisignatures from rmc as an `async` process:
* send network messages (signed or multisigned hashes) to / from `rmc` by channels
* start the rmc protocol by calling a `void` function
* obtain `multisigned` hashes by polling on `next_multisigned_hash()`
Current interface, based on the idea that multisignatures can be created only after appending a single signature (obtaining multisigs is not `async` anymore):
* pass network messages to `rmc` by calling `process_message` which may return a `multisigned` hash
* receive messages from rmc for network by polling on `next_message`
* start rmc protocol by calling a function which may return `multisigned`
The new `wait_for_next_signature` was adjusted to match the new interface of `rmc` in a way that the interface of aggregator was preserved.

aggregator/src/multicast.rs
```diff
@@ -1,64 +0,0 @@
-//! A set of abstractions for dealing with `ReliableMulticast` in a more testable
```
woocash2, line 1: The multicast trait was an unnecessary wrapper on `rmc` with minimal own logic.
After the `rmc` interface update it is no longer compatible for wrapping `rmc`. Decided to get rid of it.

finality-aleph/Cargo.toml
```diff
@@ -11,14 +11,14 @@ repository.workspace = true
 # fixed version to 'freeze' some types used in abft, mainly `SignatureSet` used in justification and signature aggregation
 aleph-bft-crypto = { workspace = true }
```
woocash2, line 13: `current` becomes `legacy` and the newest available become `current`

finality-aleph/Cargo.toml
```diff
@@ -11,14 +11,14 @@ repository.workspace = true
...
-legacy-aleph-bft-rmc = { package = "aleph-bft-rmc", version = "0.5" }
+current-aleph-bft = { package = "aleph-bft", version = "0.33" }
+current-aleph-bft-rmc = { package = "aleph-bft-rmc", version = "0.11" }
+legacy-aleph-bft = { package = "aleph-bft", version = "0.20" }
+legacy-aleph-bft-rmc = { package = "aleph-bft-rmc", version = "0.6" }
 
 network-clique = { workspace = true }
 primitives = { workspace = true }
-legacy-aleph-aggregator = { package = "aggregator", git = "https://github.com/Cardinal-Cryptography/aleph-node.git", tag = "aggregator-v0.2.1" }
+legacy-aleph-aggregator = { package = "aggregator", git = "https://github.com/Cardinal-Cryptography/aleph-node.git", tag = "r-12.0" }
```
woocash2, line 21: `current` aggregator has to become legacy. Using tag from release 12 here for the new legacy aggregator - probably better to create a new tag on the current `main` branch called `aggregator-something`

finality-aleph/src/abft/crypto.rs
```diff
@@ -68,7 +68,7 @@ impl current_aleph_bft::Keychain for Keychain {
         Keychain::node_count(self).into()
     }
 
-    async fn sign(&self, msg: &[u8]) -> Signature {
+    fn sign(&self, msg: &[u8]) -> Signature {
```
woocash2, line 71: In one of recent ABFT prs, the `MultiKeychain` became fully sync.

finality-aleph/src/abft/crypto.rs
```diff
@@ -105,8 +105,8 @@ impl current_aleph_bft::MultiKeychain for Keychain {
         index: current_aleph_bft::NodeIndex,
     ) -> Self::PartialMultisignature {
         current_aleph_bft::PartialMultisignature::add_signature(
-            SignatureSet(legacy_aleph_bft::SignatureSet::with_size(
-                legacy_aleph_bft::Keychain::node_count(self),
+            SignatureSet(aleph_bft_crypto::SignatureSet::with_size(
```
woocash2, line 108: I noticed that in both `current` and `legacy` implementations of `MultiKeychain`, legacy signature sets were used (which simply used older `aleph-bft-crypto`). I understood that we want the same signing interface regardless of legacy / current. As there were changes in `aleph-bft-crypto`, I decided to use the newest `aleph-bft-crypto` for that (which currently matches `current_aleph_bft::SignatureSet`).

finality-aleph/src/abft/mod.rs
```diff
@@ -84,7 +84,7 @@ impl<S: legacy_aleph_bft::Signature> legacy_aleph_bft::PartialMultisignature for
     }
 }
 
-impl<S: legacy_aleph_bft::Signature> current_aleph_bft::PartialMultisignature for SignatureSet<S> {
+impl<S: current_aleph_bft::Signature> current_aleph_bft::PartialMultisignature for SignatureSet<S> {
```
woocash2, line 87: Not sure what should be here. It also compiles with `legacy_aleph_bft::Signature` here, and in line 75 we can as well put `current_aleph_bft::Signature`. I'm guessing that they're just the same.

local-tests/test_recover.py
```diff
@@ -0,0 +1,64 @@
+#!/bin/env python
```
woocash2, line 1: No idea how to set finality version to `CURRENT` in the test, so left a comment below that this test should be ran only after switching the finality version. Tho I ran it manually setting the finality version.
* fails with `LEGACY` (as expected)
* works with `CURRENT` (as expected)

finality-aleph/Cargo.toml
kostekIV: yep, its not the greates dev-experience but pls prepare pr which bumps the aggregator version and then create a tag for that

finality-aleph/src/abft/crypto.rs
kostekIV: This depends, what kind of changes where there in aleph-bft-crypto? ANy changes to the signatureset encoding/decoding?

local-tests/test_recover.py
kostekIV: Is this test related to changing abft versins? If not pls add it in another pr

kostekIV: sidenote: use scheduleFinalityVersion extrinsic from pallet Aleph to change the finality version.

finality-aleph/src/aggregation/mod.rs
```diff
@@ -41,23 +41,16 @@ pub type LegacyAggregator<'a, N> = legacy_aleph_aggregator::IO<
...
-    BlockHash,
-    CurrentRmcNetworkData,
-    NetworkWrapper<CurrentRmcNetworkData, N>,
-    SignatureSet<Signature>,
-    CurrentRmc<'a>,
->;
+pub type CurrentAggregator<N> =
+    current_aleph_aggregator::IO<BlockHash, NetworkWrapper<CurrentRmcNetworkData, N>, Keychain>;
 
+#[allow(clippy::large_enum_variant)]
```
kostekIV, line 47: Is Box usage here cumbersome?

kostekIV: I guess this is also something we can ignore as this enum is not used that mich

finality-aleph/src/abft/current.rs
```diff
@@ -81,9 +76,10 @@ pub fn create_aleph_config(
...
-    let mut config = default_config(n_members.into(), node_id.into(), session_id.0 as u64);
-    config.delay_config.unit_creation_delay = unit_creation_delay_fn(unit_creation_delay);
-    config.max_round = MAX_ROUNDS;
-
-    config
+    let mut delay_config = default_delay_config();
+    delay_config.unit_creation_delay = unit_creation_delay_fn(unit_creation_delay);
+    match create_config(n_members.into(), node_id.into(), session_id.0 as u64, MAX_ROUNDS, delay_config, Duration::from_millis(SESSION_LEN_LOWER_BOUND_MS as u64)) {
+        Ok(config) => config,
+        Err(_) => panic!("Incorrect setting of delays. Make sure the total AlephBFT session time is at least {} ms.", SESSION_LEN_LOWER_BOUND_MS),
```
kostekIV, line 83: Shouldnt the error have this msg?

finality-aleph/Cargo.toml
woocash2: Hm, can't I just create a tag on the current `main`?

woocash2: Actually, I think it's best to create a tag for aggregator at the last commit which altered it. This would be A0-3009: (&A0-3020) Timed block metrics (#1411)

Marcin-Radecki: Sorry, are we talking about creating a tag in `aleph-node`? We should be doing that for release-thing and only on the release branches

Marcin-Radecki: If we're tatlking about `aleph-bft` repo, maybe it's the time to tackle release cycle process in that repo

primitives/src/lib.rs
```diff
@@ -101,7 +101,7 @@ pub const DEFAULT_FINALITY_VERSION: Version = 0;
 /// Current version of abft.
 pub const CURRENT_FINALITY_VERSION: u16 = LEGACY_FINALITY_VERSION + 1;
 /// Legacy version of abft.
-pub const LEGACY_FINALITY_VERSION: u16 = 1;
```
Marcin-Radecki, line 104: Does it affect the release procedure anyhow? Ilke we need to issue manualy some ext?

finality-aleph/Cargo.toml
woocash2: Oh, I've already created it [here](https://github.com/Cardinal-Cryptography/aleph-node/releases/tag/aggregator-v0.5.0). I saw there are some tags in the past dedicated to aggregator, so I did it in analogy to them.

After some thought, I believe we could use the `r12` tag for aggregator, or create `aggregator-v0.4.0` tag equivalent to `r12`.

local-tests/test_recover.py
woocash2: The test verifies that the node with new abft is resilient in case of a particular scenario.
How can we use `scheduleFinalityVersion` in a python local test?

finality-aleph/src/abft/current.rs
woocash2: Unfortunately error doesn't contain any message. Tho, `AlephBFT` will generate a separate `error!` log with some relevant message. I'd leave it as it is, as this panic contains a message more relevant to `aleph-node` user - a suggestion about how exactly to fix their config.

primitives/src/lib.rs
woocash2: I'd say that running update-nets from release branches is one thing to do, and I believe that it would be meaningful to run a kind of update-net in which we start from `main` and only a **part** of nodes (say half) is updated.
Maybe @kostekIV or @timorleph would have some other suggestions.

Explanation: version to use is stored on the chain and currently it is `2` which is translated to using `CURRENT` version. I altered the codebase so that the `LEGACY` version corresponds now to what was the `CURRENT` version. So changing `LEGACY` from `1` to `2` **shouldn't cause any difference in node's behavior**.

finality-aleph/src/abft/crypto.rs
woocash2: I see no such changes

finality-aleph/src/abft/mod.rs
woocash2: They kinda are the same. I guess the compiler unwinds the below definition which is the same in legacy and current and doesn't distinguish between them.
```
pub trait Signature: Debug + Clone + Codec + Send + Sync + Eq + 'static {}
```

finality-aleph/src/aggregation/mod.rs
timorleph: It's rather surprizing to me that this showed up now – is it because something now uses a cloned keychain rather than a reference or something like that?

primitives/src/lib.rs
timorleph: To answer the question directly – yes, we will need to set the finality version on-chain as probably the final part of the update, after we make sure ~all external nodes are updated as well.

For testing, I thought there were already some tests checking this type of update, but I might be wrong; if there are none, there should be a task for creating them. For manual testing we will have to run this somehow as well.

finality-aleph/Cargo.toml
woocash2: I removed `aggregator-v0.5.0` and introduced `aggregator-v0.4.0` which is the same as `r-12.0`

aggregator/src/multicast.rs
woocash2: I don't know `Multicast`'s original purpose. `rmc` changed so much that I'm not sure whether adjusting `Multicast` to `rmc` so that the interface of `Multicast` is preserved is feasible. It was easier for me to handle new `rmc` in `aggregator` directly. Overall this code has gotten much simpler so idk. Feel like adjusting `Multicast` for the new `rmc` would be lying to ourserves and not using the simplicity of the new `rmc`. Even `aggregator` could be simplified but I didn't want to change too much.

finality-aleph/src/abft/crypto.rs
woocash2: It doesn't compile with anything being `legacy`. This is due to line 39 in `abft/mod` in which we declare `SignatureSet` parametrized by `aleph_bft_crypto::SignatureSet`. I mentioned that previously, `legacy_aleph_bft` prefix was used in both `Current` and `Legacy` multikeychains, which was equivalent to `aleph_bft_crypto` prefix, as legacy abft used `aleph-bft-crypto` from workspace. Now, in this pr we update `aleph-bft-crypto` to the newest version, which is equivalent to version used by `current_aleph_bft`. So we need to use it, or change the `SignatureSet` but it doesn't seem good.

finality-aleph/src/aggregation/mod.rs
woocash2: Current aggregator directly uses the rmc service instead a `Multicast` trait. Maybe this is the reason. I can wrap around box

finality-aleph/Cargo.toml
woocash2: At the end, removed custom aggregator tag and used r-12.1 tag.

primitives/src/lib.rs
Marcin-Radecki: Ok, noted. Thanks.

Marcin-Radecki (approved):
Let's make local-tests/test_recover_after_abft_update.py in CI if possible please :pray: 

Approve from CI/release perspective

finality-aleph/Cargo.toml
Marcin-Radecki: cool thanks

local-tests/test_recover.py
woocash2: I changed the test so now it:
* Bumps the abft version using extrinsic at some session
* runs the nodes for twice as many sessions
* performs the recover scenario afterwards (kills all nodes, recovers all but one, and expects that the living nodes finalize)

I noticed [here](https://github.com/Cardinal-Cryptography/aleph-node/blob/main/pallets/aleph/src/lib.rs#L310) that we always start the chain with legacy abft version, so updating abft using extrinsic will always be valid (we'll always go from legacy to current). This way this test will work now and even if we update abft further in the future. For this version of the code, calling this extrinsic is required to pass the test.

finality-aleph/src/abft/current.rs
kostekIV: imho its responsibility that belongs to abft

finality-aleph/src/abft/crypto.rs
woocash2: all good here

local-tests/test_recover_after_abft_update.py
```diff
@@ -0,0 +1,68 @@
...
+from os.path import abspath, join
+from time import ctime, sleep
+from chainrunner import Chain, Seq, generate_keys
+
+
+def printt(s): print(ctime() + ' | ' + s)
+
+
+'''
+Make sure to complile the binary with --features short_session
```
kostekIV, line 12:
```suggestion
Make sure to compile the binary with --features short_session
```

local-tests/test_recover_after_abft_update.py
```diff
@@ -0,0 +1,68 @@
...
+'''
+
+# Path to working directory, where chainspec, logs and nodes' dbs are written:
+workdir = abspath(os.getenv('WORKDIR', '/tmp/workdir'))
+# Path to the aleph-node binary (important DON'T use short-session feature):
+binary = abspath(os.getenv('ALEPH_NODE_BINARY', join(workdir, 'aleph-node')))
```
kostekIV, line 18: contradicting comment to the one at line 12

kostekIV (approved):
finals nits

PR #1481, opened 2023-11-08
A0-3143: Major sync test does not use pre-baked workdir.

Marcin-Radecki:
# Description

This PR removes pre-baked `workdir` file used in major sync test. Motivations are:
* that `workdir` was generated on 11.X version. It has happened on the `main` branch in which there was a change that changed database schema in a way that finalization does not work. This test is supposed to validate whether node is able to sync in major sync mode, ie when gap is greater then 2 sessions, and not major update scenario which it does now
* `workdir` contributes to 20% of whole `aleph-node` size and it bloats git history
* as it is now, we need to regenerate `workdir` every time such breaking change occurrs, after this PR is merge we won't have to do it

Also, this PR removes some tech debt which is:
* removed not used code for python tests, which functionality is covererd by major sync and force reorgs tests
* added common bash script for running python tests

## Type of change

## Testing

- normal session nightly pipeline pass https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/6798163085/job/18482394829


kostekIV (approved):
<3

local-tests/chainrunner/chain.py
```diff
@@ -212,6 +212,7 @@ def wait_for_finalization(self, old_finalized, nodes=None, timeout=600, finalize
         deadline = time.time() + timeout
         while self.get_highest_finalized(nodes) <= old_finalized + finalized_delta:
             time.sleep(5)
+            print('wait_for_finalization: the highest finalized is ', self.get_highest_finalized(nodes))
```
woocash2, line 215: use logging?

woocash2 (approved):
Not having the greatest feeling about `test_catch_up` and `test_multiple_restarts`, although `test_major_sync` is more massive and expects the same result (some node catching up to other nodes) as the removed tests, the removed tests checked slightly different scenarios (for example in `test_multiple_restarts` the node is restarted many times and expected to catch up each time). While agreeing the value of these tests is low, I think it is non-zero :smiley:. Anyways, remove them if you're convinced that it's fine.

local-tests/chainrunner/chain.py
Marcin-Radecki: Yes but in other PRs please - this python test framework is a decent tool but it needs a slightly revamp, so we can tackle that problem there.

PR #1482, opened 2023-11-08
A0-3284: update to 1.1.0

kostekIV:
# Description
Substrate bump


Cargo.toml
```diff
@@ -90,91 +90,91 @@ tiny-bip39 = { version = "1.0" }
 tokio = { version = "1.32" }
 rand_pcg = { version = "0.3.1", default-features = false }
 
-frame-benchmarking = { git = "https://github.com/Cardinal-Cryptography/substrate.git", branch = "aleph-v1.0.0", default-features = false }
```
Marcin-Radecki, line 93: New repo <3 

Can you please create a ticket to make `https://github.com/Cardinal-Cryptography/substrate.git` read only. when 1.1.0 is released to production? Till then we need both repos open in case of backporting a hotfix to 1.0.0

kostekIV: sure, here you go https://cardinal-cryptography.atlassian.net/browse/A0-3468

Marcin-Radecki: Thanks

woocash2 (approved):

bin/runtime/src/lib.rs
```diff
@@ -699,6 +701,7 @@ parameter_types! {
     // Maximum size of the lazy deletion queue of terminated contracts.
     pub const DeletionQueueDepth: u32 = 128;
     pub Schedule: pallet_contracts::Schedule<Runtime> = Default::default();
+    pub CodeHashLockupDepositPercent: Perbill = Perbill::from_percent(30);
```
h4nsu, line 704: Where is the value 30% coming from?

h4nsu (approved):

bin/runtime/src/lib.rs
kostekIV: the docs for `CodeHashLockupDepositPercent`
```
/// The percentage of the storage deposit that should be held for using a code hash.
/// Instantiating a contract, or calling `add_delegate_dependency` (for contract delegation) protects
/// the code from being removed. In order to prevent abuse these actions are protected with
/// a percentage of the code deposit.
```
The 30% is the one used in substrate node runtime. I think we want to think about what value we would like here.

pmikolajczyk41 (approved):

bin/runtime/src/lib.rs
kostekIV: Resolved off github

PR #1483, opened 2023-11-09
A0-3010: Metrics for reorgs

ggawryal:
# Description

Adding metrics for (arbitrary) reorgs, as per initial slo doc. As for the slo we care about the external user experience and what's visible to they, I've decided to implement it in a separate thread, that uses notification stream from the backend, the same as RPC does. I've also moved best and top_finalized block to that task for the same reason.
After releasing we will have to update grafana to show `aleph_reorgs` histogram somehow.

## Type of change

# Checklist:

- I have added tests


finality-aleph/src/metrics/chain_status.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    fn report_reorg(&self, length: BlockNumber) {
+        if let ChainStatusMetrics::Prometheus { reorgs, .. } = self {
+            reorgs.observe(length as f64);
+        }
+    }
+}
+
+pub async fn start_chain_state_metrics_job_in_current_thread<
```
timorleph, line 76: This name is _ever so slightly_ overspecified. I would go with just `run`, but `run_chain_state_metrics` is also alright.

finality-aleph/src/metrics/mod.rs
```diff
@@ -0,0 +1,6 @@
+mod chain_status;
+mod timing;
+
+pub use chain_status::start_chain_state_metrics_job_in_current_thread;
+pub use timing::{Checkpoint, TimingBlockMetrics};
+pub(crate) const LOG_TARGET: &str = "aleph-metrics";
```
timorleph, line 6: No `pub(crate)` plx, I think this can just be without any `pub`, since it's only used inside of this module, no?

finality-aleph/src/sync/mod.rs
```diff
@@ -28,7 +28,7 @@ pub use substrate::{
     SubstrateChainStatus, SubstrateChainStatusNotifier, SubstrateFinalizationInfo, VerifierCache,
 };
 
-const LOG_TARGET: &str = "aleph-block-sync";
+use crate::metrics::LOG_TARGET;
```
timorleph, line 31: Uhh, no, this shouldn't be using the metrics log target.

timorleph (changes requested):
Nice! The log target thing seems very wrong though, what happened there?

finality-aleph/src/metrics/chain_status.rs
```diff
@@ -0,0 +1,239 @@
...
+    backend: &BE,
+    prev_best: Option<HE>,
+    best: HE,
+) -> Option<HE::Number> {
+    let prev_best = prev_best?;
+    if best.hash() == prev_best.hash() || *best.parent_hash() == prev_best.hash() {
+        // Quit early when no change or the best is a child of the previous best.
+        return None;
+    }
+    let lca = lowest_common_ancestor(backend, best.hash(), prev_best.hash()).ok()?;
```
lesniak43, line 145: highest?

lesniak43: oh no, it's a substrate method... nvm then

lesniak43 (approved):

timorleph (approved):

ggawryal:
I was thinking of adding also logging for reorgs with their exact length, but this would duplicate reorg log from substrate - https://github.com/Cardinal-Cryptography/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/client/informant/src/lib.rs#L119. It doesn't seem to be possible to be switched off without modifying substrate code. Anyway, I think it would be good to have these two close to each other. What do you think?

finality-aleph/src/metrics/chain_status.rs
ggawryal: True, changed

timorleph:
If you want to add such log it shouldn't be above `debug` level, precisely because of the duplication with the substrate log. The additional length info is not that valuable.

finality-aleph/src/sync/mod.rs
ggawryal: :facepalm: must've removed that accidentally and then auto import chosen the only possible and incorrect version

PR #1484, opened 2023-11-09
A0-3393: add proxy to runtime

kostekIV:
# Description

Add pallet proxy to runtime with 3 proxy type available:
* Any - proxy can make any call
* Staking - proxy can make calls related to staking (bond, nominate etc)
* Nontrasfer - calls that do not make any transfer of funds

## Type of change

# Checklist:

- I have added tests


aleph-client/src/pallets/proxy.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use crate::{
+    aleph_runtime::{ProxyType, RuntimeCall},
+    api, AccountId, SignedConnectionApi, TxInfo, TxStatus,
+};
+
+/// any object that implements pallet proxy api
+#[async_trait::async_trait]
+pub trait ProxyUserApi {
+    /// API for [`proxy`](https://paritytech.github.io/polkadot-sdk/master/pallet_proxy/pallet/struct.Pallet.html#method.proxy) call.
```
deuszx, line 11: This comment is just a link to the original one. In the browser, like azero.dev, when you choose the pallet/endpoint it will show the comment. How does it work here? Which comment will appear there?
Screenshot 2023-11-15 at 13 19 59

ggawryal (commented):
Good stuff

e2e-tests/src/test/proxy.rs
```diff
@@ -0,0 +1,262 @@
...
+        };
+        match status {
+            Success => {
+                if res.result.is_err() {
+                    panic!("Result of the call should be ok, {:?} {:?}", res, call);
+                }
+
+                let bi = match events.find_first::<BatchInterrupted>() {
+                    Ok(Some(bi)) => bi,
+                    _ => return,
```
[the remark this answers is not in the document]
kostekIV: good catch, fixed

e2e-tests/src/test/proxy.rs
```diff
@@ -0,0 +1,262 @@
...
+                };
+
+                panic!("No batch interrupted event should be present {:?}", bi);
```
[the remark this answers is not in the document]
kostekIV: assert under the hood panics. I thought there is not really that much to assert here, rather we found ourself in the state that indicates the test have failed

bin/runtime/src/lib.rs
```diff
@@ -750,6 +751,84 @@ impl pallet_identity::Config for Runtime {
...
+}
+impl InstanceFilter<RuntimeCall> for ProxyType {
+    fn filter(&self, c: &RuntimeCall) -> bool {
+        match self {
+            ProxyType::Any => true,
+            ProxyType::NonTransfer => !matches!(
+                c,
+                RuntimeCall::Balances(..)
+                    | RuntimeCall::Vesting(pallet_vesting::Call::vested_transfer { .. })
+            ),
```
DamianStraszak, line 795: This is quite dangerous. If we ever add a pallet that can do any type of transfers we must remember to update this... Even now, I think this is incorrect. Because you can make transfers using `Contract` calls. 

We need a plan on how to make this safe.

DamianStraszak: In general any proxy that works by blacklisting, instead of whitelisting calls is dangerous.

bin/runtime/src/lib.rs
```diff
@@ -750,6 +751,84 @@ impl pallet_identity::Config for Runtime {
...
+                    c,
+                    RuntimeCall::Staking(..)
+                        | RuntimeCall::Session(..)
+                        | RuntimeCall::Utility(..)
+                        | RuntimeCall::NominationPools(..)
+                )
+            }
+        }
+    }
+    fn is_superset(&self, o: &Self) -> bool {
```
DamianStraszak, line 807: What is this responsible for?

DamianStraszak (changes requested):

bin/runtime/src/lib.rs
kostekIV: There is also an option to not add the non-trasnfer proxy

DamianStraszak: I think it's best to define it via whitelisting...

deuszx: And no "capture-all" statements - if we add a new variant we will get a compilation error, rather than a runtime surprise.

Marcin-Radecki: What about we do same as per 
https://github.com/paritytech/polkadot/blob/master/runtime/polkadot/src/lib.rs#L960-L998 ?

Marcin-Radecki: Sorry, this is link to the archived repo, but we can check how polkadot-sdk repo does it

kostekIV: Ok, non-transfer changed to 'whitelisted' variant

ggawryal (approved):

bin/runtime/src/lib.rs
```diff
@@ -758,6 +759,101 @@ impl pallet_identity::Config for Runtime {
...
+                    | RuntimeCall::Aleph(..)
+                    | RuntimeCall::Elections(..)
+                    | RuntimeCall::Treasury(..)
+                    | RuntimeCall::Vesting(pallet_vesting::Call::vest { .. })
+                    | RuntimeCall::Vesting(pallet_vesting::Call::vest_other { .. })
+                    | RuntimeCall::Vesting(pallet_vesting::Call::merge_schedules { .. })
+                    | RuntimeCall::Utility(..)
+                    | RuntimeCall::Multisig(..)
+                    | RuntimeCall::Sudo(..)
+                    | RuntimeCall::Contracts(..)
```
DamianStraszak, line 815: This should not be allowed, because you can transfer money to a contract.
Also, I would simply get rid of all the pallets that you probably shouldn't call as a user:
- Scheduler,
- Timestamp,
- Aleph,
- Elections,
- Sudo,
- CommitteeManagement

DamianStraszak (changes requested):

bin/runtime/src/lib.rs

bin/runtime/src/lib.rs
kostekIV: In pallet proxy its used in one place.
```
// Proxy call cannot add or remove a proxy with more permissions than it already has.
```
So proxy calls that add or remove proxy (A) of some type have to be called with a proxy (B) such that `B.is_superset(A) == true`

DamianStraszak: I propose to write this like:

```
    fn is_superset(&self, o: &Self) -> bool {
        match (self, o) {
            (x, y) if x == y => true,
            (ProxyType::Any, ProxyType::NonTransfer) => true,
            (ProxyType::Any, ProxyType::Staking) => true,
            (ProxyType::NonTransfer, ProxyType::Staking) => true,
            _ => false,
        }
    }
```

This way it's not possible to make any mistake when adding a new proxy type in the future.

bin/runtime/src/lib.rs
```diff
@@ -758,6 +759,93 @@ impl pallet_identity::Config for Runtime {
...
+                    | RuntimeCall::Session(..)
+                    | RuntimeCall::Treasury(..)
+                    | RuntimeCall::Vesting(pallet_vesting::Call::vest { .. })
+                    | RuntimeCall::Vesting(pallet_vesting::Call::vest_other { .. })
+                    | RuntimeCall::Vesting(pallet_vesting::Call::merge_schedules { .. })
+                    | RuntimeCall::Utility(..)
+                    | RuntimeCall::Multisig(..)
+                    | RuntimeCall::NominationPools(..)
+                    | RuntimeCall::Identity(..)
+                    | RuntimeCall::Proxy(..)
```
DamianStraszak, line 811: I read the code of the pallet a bit and I would propose removing `Proxy` from the list. Weird shit can happen if you allow proxies here. And we don't want to get bitten in the ass for no benefit.

kostekIV: ok, removed

bin/runtime/src/lib.rs
kostekIV: Ok, I have rewritten it but in a little bit different way. Removed capture all arm and `(x, y) if x == y => true`.
First 2 arms to comfirm to `Any` specification
```
            (ProxyType::Any, _) => true,
            (_, ProxyType::Any) => false,
```
And the rest is explicitly written.

DamianStraszak (approved):

PR #1489, opened 2023-11-14
A0-3225: Automatic update net tests from Testnet to main and from Mainnet to main

Marcin-Radecki:
# Description

This PR adds two automatic update net tests which baseline is really simple
1. Create internal net from scratch based on production chain image (Mainnet or Testnet), 7 nodes
2. Check for finalization
3. Update binary to current aleph-node `main`
4. Check for finalization

and runs two tests (mainnet, Testnet) in paraller. Each they around 15 minutes.

Those flows can be extended. For example, we might want to update only half of the consensus first, check for finalization, then update other half of consensus etc. 

Those tests were added to nightly and they report to slack only on failure.

## Type of change

## Testing

* https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/6860640237

## Related PRs

* https://github.com/Cardinal-Cryptography/github-actions/pull/30


woocash2 (approved):
Good initiative :+1:

.github/workflows/nightly-update-net-tests.yml
```diff
@@ -0,0 +1,150 @@
...
+      - name: Checkout aleph-node repository
+        uses: actions/checkout@v4
+        with:
+          path: aleph-node
+          ref: main
+          repository: Cardinal-Cryptography/aleph-node
+          token: ${{ secrets.CI_GH_TOKEN }}
+          fetch-depth: 0
+
+      - name: Get last two commit hashes
```
bitsnops, line 44: We're only taking one commit hash here, fix the step title

bitsnops (approved):

PR #1490, opened 2023-11-14
A0-2244: Remove old junk backups

woocash2:
### Description
Changed the backup removal code to remove everything from the backup stash folder except the subfolder corresponding to the current session.

### Testing
Tested manually - ran for many short sessions, verified finalization, verified backup folder contents

### DIFF
--- a/finality-aleph/Cargo.toml
+++ b/finality-aleph/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "finality-aleph"
-version = "0.10.0"
+version = "0.10.1"
 license = "Apache 2.0"
 authors.workspace = true
 edition.workspace = true
--- a/finality-aleph/Cargo.toml
+++ b/finality-aleph/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "finality-aleph"
-version = "0.10.1"
+version = "0.10.0"
 license = "Apache 2.0"
 authors.workspace = true
 edition.workspace = true
--- a/finality-aleph/src/party/backup.rs
+++ b/finality-aleph/src/party/backup.rs
@@ -118,14 +118,14 @@ pub fn rotate(
     Ok((backup_saver, backup_loader))
 }
 
-/// Removes the backup directory for a session.
+/// Removes the backup directory for all sessions except the current session.
 ///
 /// `backup_path` is the path to the backup directory (i.e. the argument to `--backup-saving-path`).
 /// If it is `None`, nothing is done.
 ///
-/// Any filesystem errors are logged and dropped.
+/// Any filesystem errors are returned.
 ///
-/// This should be done after the end of the session.
+/// This should be done at the beginning of the new session.
 pub fn remove_old_backups(path: Option<PathBuf>, current_session: u32) -> io::Result<()> {
     if let Some(path) = path {
         for read_dir in fs::read_dir(path.clone())? {
--- a/finality-aleph/src/party/backup.rs
+++ b/finality-aleph/src/party/backup.rs
@@ -9,6 +9,7 @@ use std::{
 
 use log::debug;
 
+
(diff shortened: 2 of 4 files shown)


finality-aleph/src/party/backup.rs
```diff
@@ -118,27 +118,22 @@ pub fn rotate(
...
-            debug!(target: "aleph-party", "Removed backup for session {}", session_id);
-        }
-        Err(err) => {
-            if err.kind() != io::ErrorKind::NotFound {
-                warn!(target: "aleph-party", "Error cleaning up backup for session {}: {}", session_id, err);
+/// This should be done at the beginning of the new session.
+pub fn remove_old_backups(path: Option<PathBuf>, current_session: u32) -> io::Result<()> {
+    if let Some(path) = path {
+        for read_dir in fs::read_dir(path.clone())? {
+            let item = read_dir?;
```
ggawryal, line 132: We shouldn't get errors there in practice, but theoretically `if let Ok(item) = read_dir) {}` is more correct there.

finality-aleph/src/party/backup.rs
```diff
@@ -118,27 +118,22 @@ pub fn rotate(
...
+            if item.path() != path.join(current_session.to_string()) {
+                fs::remove_dir_all(item.path())?;
```
ggawryal, line 134: Same there - perhaps clearing as much as possible and returning an arbitrary error that occurred is better than early quit.

finality-aleph/src/party/backup.rs
```diff
@@ -118,27 +118,22 @@ pub fn rotate(
...
-    };
-    match fs::remove_dir_all(path) {
-        Ok(()) => {
-            debug!(target: "aleph-party", "Removed backup for session {}", session_id);
-        }
-        Err(err) => {
-            if err.kind() != io::ErrorKind::NotFound {
-                warn!(target: "aleph-party", "Error cleaning up backup for session {}: {}", session_id, err);
+/// This should be done at the beginning of the new session.
+pub fn remove_old_backups(path: Option<PathBuf>, current_session: u32) -> io::Result<()> {
```
ggawryal, line 129: Nit: How about `remove_other_backups` or such, as we don't explicitly assume there won't be any backups of newer sessions?

ggawryal (approved):
Looks good, left minor comments

finality-aleph/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "finality-aleph"
-version = "0.10.0"
+version = "0.10.1"
```
timorleph, line 3: I think we avoid bumping this version in most PRs, so please do so here as well, for consistency.

finality-aleph/src/party/backup.rs
timorleph: It would actually be better if this only removed old sessions – we might not want to accidentally delete a backup if we had to e.g. start from a snapshot because of DB corruption or something.

finality-aleph/src/party/backup.rs
timorleph: Hm, I dunno, this logging an error on failure and not doing anything else seems good to me actually. Trying to power through with data deletion tends to be unnecessarily risky.

timorleph (approved):
Looks good.

finality-aleph/src/party/backup.rs
ggawryal: I can agree, we can leave it as it is and not overcomplicate error handling when it doesn't give clear advantages

woocash2: ok, leaving as it is

finality-aleph/src/party/backup.rs
woocash2: done

ggawryal (approved):

PR #1492, opened 2023-11-16
A0-3499: Archive logs in case of e2e failure

Marcin-Radecki:
# Description

* Enable RUST_LOG=debug in e2e.
* dump all logs from docker-compose and archive them in a tarball, but only for failed e2e
* Upload this tarball to GH artifacts

Considerations:

Log growth is around 200 MB per node per 1 hour so given that some nightly tests run 1 hour and there are max 10 nodes, we have 2GB per test, which is affordable as GH runners have a quota of 14 GB, and our self-hosted runners have at least that much. Also, compressed logs should be around 100 MBs max, so for the case of GH artifacts quota, we should be good. For public repos, we don’t pay anything for this quota anyway.

There's some trickery when calling e2e client docker, as there's no way to enforce timeout on step level in  GH composite actions, so I needed to do a workaround to apply custom timeout mechanism when calling e2e client in docker. 

## Type of change

## Testing

* success case - no logs uploaded https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/6889159800/job/18740144103
* a genuine timeouts in one e2e tests https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/6904198011 .See that at the bottom there's tgz attached from failed e2e case

### DIFF
--- a/.github/actions/run-e2e-test/action.yml
+++ b/.github/actions/run-e2e-test/action.yml
@@ -124,21 +124,31 @@ runs:
           contracts/scripts/clean.sh
         fi
 
+    - name: Get log tarball file name
+      # if: ${{ failure() }}
+      if: ${{ always() }}
+      id: get-log-tarball-file-name
+      shell: bash
+      run: |
+        test_case_escaped=$(echo ${{ inputs.test-case }} | sed 's/::/-/g')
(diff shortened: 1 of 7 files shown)


obrok (approved):
:+1:

PR #1499, opened 2023-11-21
A0-3514: Don't log old backups not existing

lesniak43:
# Description

Do not try to remove backups if they're not there.

## Type of change


timorleph (approved):

finality-aleph/src/party/backup.rs
```diff
@@ -128,6 +128,9 @@ pub fn rotate(
 /// This should be done at the beginning of the new session.
 pub fn remove_old_backups(path: Option<PathBuf>, current_session: u32) -> io::Result<()> {
     if let Some(path) = path {
+        if !path.exists() {
+            return Ok(());
+        }
         for read_dir in fs::read_dir(path)? {
```
woocash2, line 134: I see a problem here - if the folder gets deleted between lines 133 and 134 we still throw an error :P

lesniak43: will it get deleted?...

woocash2: Not in reality, but in theory someone could hit `rm -r backup_stash` and cpu would have to expropriate aleph-node just after line 133 to make it happen. :stuck_out_tongue: 

I would consider 2 opitons:
- merge this pr, and treat the above scenario as unreal & if it happens it is very weird so throwing error isn't a bad thing to do,
- use `if let Ok(contents) = read_dir(path)` instead, but then other types of errors will be unnoticed (for ex. lack of permissions to read directory won't be logged, when maybe it should be).

After some thinking I think the first option may actually be better than the second option because noticing other types of errors. Imma hit accept & you let you decide (resolve this comment by yourself)

woocash2 (approved):

finality-aleph/src/party/backup.rs
woocash2: also - we should be able to match on a particular case of this error to only tolerate the `directory missing` error but this `io::Error` is somehow weird (it's not an enum) so it would require some investigation.

lesniak43: resolved it is!

PR #1500, opened 2023-11-21
A0-3517: Verify headers later in `Handler::handle_state`

lesniak43:
# Description

Headers verified only when needed, not too early.
Fixed some logs.

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/finality-aleph/src/sync/handler/mod.rs
+++ b/finality-aleph/src/sync/handler/mod.rs
@@ -766,15 +766,15 @@ where
             None => {
                 // let's start with the justification,
                 // as the header might be too far in the future
-                let mut new_info =
+                let higher_justification =
                     self.handle_justification(state.top_justification(), Some(peer.clone()))?;
                 let VerifiedHeader {
                     header,
                     maybe_equivocation_proof,
                 } = self.verify_header(state.favourite_block(), false)?;
                 maybe_proof = maybe_equivocation_proof;
-                new_info |= self.forest.update_header(&header, Some(peer), false)?;
-                HandleStateAction::maybe_extend(new_info)
+                let new_descendant = self.forest.update_header(&header, Some(peer), false)?;
+                HandleStateAction::maybe_extend(higher_justification || new_descendant)
             }
             // same session
             Some(0) => match remote_top_number >= local_top_number {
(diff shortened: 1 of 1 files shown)


finality-aleph/src/sync/handler/mod.rs
```diff
@@ -753,26 +754,39 @@ where
...
-                    || self.forest.update_header(&header, Some(peer), false)?,
-            ),
+            None => {
+                let VerifiedHeader {
+                    header,
+                    maybe_equivocation_proof,
+                } = self
+                    .verifier
+                    .verify_header(state.favourite_block(), false)
+                    .map_err(Error::HeaderVerifier)?;
```
timorleph, line 766: Technically you could get a state from an honest node in which you can interpret the justification, but the header is too new to validate, ideally you should process the justification in this case rather than erroring out here already. Not sure if this is reasonably doable.

finality-aleph/src/sync/handler/mod.rs
```diff
@@ -753,26 +754,39 @@ where
...
-                        || self.forest.update_header(&header, Some(peer), false)?,
-                ),
+                true => {
+                    let VerifiedHeader {
+                        header,
+                        maybe_equivocation_proof,
+                    } = self
+                        .verifier
+                        .verify_header(state.favourite_block(), false)
+                        .map_err(Error::HeaderVerifier)?;
```
timorleph, line 783: The above is _not_ a problem here though, so this can stay.

timorleph (approved):
One problem that is unlikely to cause trouble, but it would be great if you were able to fix it.

finality-aleph/src/sync/handler/mod.rs
```diff
@@ -753,26 +754,39 @@ where
...
-            ),
+            None => {
+                let VerifiedHeader {
+                    header,
+                    maybe_equivocation_proof,
+                } = self
+                    .verifier
+                    .verify_header(state.favourite_block(), false)
+                    .map_err(Error::HeaderVerifier)?;
+                maybe_proof = maybe_equivocation_proof;
```
kostekIV, line 767: rename in struct destructuring.

finality-aleph/src/sync/handler/mod.rs
kostekIV, line 766: nit: could we have a fn in handler `verify_header` which does exactly does -> call verify from verifier and maps an error

kostekIV (approved):

finality-aleph/src/sync/handler/mod.rs
lesniak43: how?

finality-aleph/src/sync/handler/mod.rs
lesniak43: fixed, I think

lesniak43: done

finality-aleph/src/sync/handler/mod.rs
kostekIV: maybe_equivocation_proof: maybe_proof

lesniak43: I think it shadows the original `maybe_proof`

kostekIV: oh right :P

PR #1502, opened 2023-11-21
A0-3518: Purge chain-extension: backend 1/2

pmikolajczyk41:
# Description

Great fun continued.

Since OBCE is no longer maintained, we have to rework the extension from the very ground. To avoid huge diff, I divided the work into couple of separate changes. This (second) PR:
 - brings back simplified backend part (i.e. the actual handler for extensions); simplified, because still without proper checks, weighting and tests
 - brings back the extension to the runtime

## Type of change

# Checklist:

- I have added tests


baby-liminal-extension/src/lib.rs
```diff
@@ -11,13 +11,30 @@
...
 pub type VerificationKeyIdentifier = [u8; 8];
+
+// ------ Backend stuff ----------------------------------------------------------------------------
+
+#[cfg(feature = "runtime")]
+mod backend;
+
+#[cfg(feature = "runtime")]
+pub use backend::BabyLiminalChainExtension;
+#[cfg(feature = "runtime")]
```
Marcin-Radecki, line 39: Why this feature flag is called `runtime`? As this is quite general name, and in reality it is supposed to enable some particular feature regarding to Liminal, why not call it that way for example?

baby-liminal-extension/Makefile
```diff
@@ -9,6 +9,10 @@ lint: ## Check linting rules
 	cargo +nightly fmt --all --check || exit 1
 	cargo +nightly fmt --all --check $(AT_TEST_CONTRACT) || exit 1
 
+clippy: ## Run clippy
```
Marcin-Radecki, line 12: Good idea with Makefile!

Please let us know when  added CI extends over 10 minutes, we'll then split it somehow (so invariant is to not extend a job over 10 minutes please)

baby-liminal-extension/src/lib.rs
Marcin-Radecki: Okay now I see, this is specific to this crate. All good then.

bin/runtime/Cargo.toml
```diff
@@ -157,14 +157,15 @@ runtime-benchmarks = [
 
 # Liminal-related features
 liminal = [
-    #"baby-liminal-extension",
```
Marcin-Radecki, line 160: Ideally if we we'd check if those below two places
* https://github.com/Cardinal-Cryptography/aleph-node/blob/main/.github/workflows/_build-production-node-and-runtime.yml#L30
* https://github.com/Cardinal-Cryptography/aleph-node/blob/main/.github/workflows/_build-test-node-and-runtime.yml#L31

That both runtime and `aleph-node` binary does **not** have any liminal-related symbols

pmikolajczyk41: checks added

.github/workflows/_build-production-node-and-runtime.yml
```diff
@@ -27,7 +27,10 @@ jobs:
           targets: wasm32-unknown-unknown
 
       - name: Build production binary and runtime
-        run: cargo build --profile production -p aleph-node
+        # yamllint disable-line rule:line-length
+        run: |
+          cargo build --profile production -p aleph-node
+          nm --demangle=rust target/release/aleph-node | grep -q liminal && { echo "Liminal symbols are present"; exit 1; }
```
Marcin-Radecki, line 33: I think this should be `target/production/aleph-node` instead `target/release/aleph-node`

pmikolajczyk41: oh, you're right, thank you for spotting; fixed

kostekIV (approved):
lgtm

Yithis (approved):

PR #1504, opened 2023-11-23
A0-3518: Resurrect chain extension: 2/3

pmikolajczyk41:
# Description

Yet another PR, but it is not the end!

This time, I'm bringing back backend testing. Since the chain extension mechanics are entangled with runtime and pallet contract, we have to provide mocked environment, which results in a lot of lovely code with unstable features. They were introduced because all the main calls (`store_key` and `verify`) are invoked as **static** methods of `Runtime`. Therefore, it makes testing different behaviors highly inconvenient (since we cannot change the mock behavior depending on any variable). Being able to introduce enums into generics gives us the possibility of concise defining scenarios.

To keep PRs smaller, I left charging weight and validating input for the next time.

## Type of change

# Checklist:

- I have added tests


.github/workflows/_unit-tests-and-static-checks.yml
```diff
@@ -47,6 +51,12 @@ jobs:
           command: test
           args: -- --skip clique_network
 
+      - name: Install Nightly Rust Toolchain (for unstable features in tests)
+        uses: Cardinal-Cryptography/github-actions/install-rust-toolchain@v4
```
Marcin-Radecki, line 55: One small feature request please: since `liminal` is behind the feature flag, it should be excluded from every commit to PR for now. Instead, let's add a label e.g. `liminal-CI` here https://github.com/Cardinal-Cryptography/aleph-node/blob/main/.github/workflows/on-pull-request-label.yml and spawn any liminal-related checks only if this label is added to the PR. 

On breaking liminal things e.g. by substrate updates: we should run the same liminal-checks in nightly, so then we can catch errors as a whole for every change in `aleph-node`

pmikolajczyk41: discussed offline: I'll take care of this in the next PR tomorrow (so just not to mix scopes)

Marcin-Radecki: ok so resolving this comment

deuszx (commented):
Design Q: since we use separate trait bounds for `store_key` and `verify` it is possible that we use a different `BackendExecutor` impl for each of the method - potentially leaving to some unexpected consequences. Is that intentional? We could, instead have a trait bounds on the `impl<...> for BabyLiminalChainExtension`.

deuszx (approved):

Yithis (approved):

PR #1511, opened 2023-11-27
A0-3015: Introduce AllBlockMetrics wrapper

woocash2:
# Description

Introduce `AllBlockMetrics` which is a wrapper for different types of block metrics.
Currently it only wraps `TimingBlockMetrics` but in future it will contain some other metrics.

Adjustments made to `TimingBlockMetrics`:
* from `report_block` and `report_block_if_not_present` created a single function of semantics equal to `report_block_if_not_present`, but named `report_block`.
* Changed signature of `report_block(hash, timestamp, checkpoint)` to `report_block(hash, checkpoint)` - the timestamp can be inferred inside the function
* Parameterized the struct with newly added `trait Clock` - there were some tests which checked behavior of `TimingBlockMetrics` when manipulating the times of newly appearing reports. After removing `timestamp` from `report_block` signature we needed to somehow be able to manipulate the times - thus parameterization.

## Type of change

# Checklist:

### DIFF
--- a/bin/node/src/service.rs
+++ b/bin/node/src/service.rs
@@ -7,10 +7,10 @@ use std::{
 
 use aleph_runtime::{self, opaque::Block, RuntimeApi};
 use finality_aleph::{
-    run_validator_node, AlephBlockImport, AlephConfig, BlockImporter, Justification,
-    JustificationTranslator, MillisecsPerBlock, Protocol, ProtocolNaming, RateLimiterConfig,
-    RedirectingBlockImport, SessionPeriod, SubstrateChainStatus, SyncOracle, TimingBlockMetrics,
-    TracingBlockImport, ValidatorAddressCache,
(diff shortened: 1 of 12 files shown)


bin/node/src/service.rs
```diff
@@ -136,13 +136,18 @@ pub fn new_partial(
...
+    {
+        Ok(timing_metrics) => timing_metrics,
         Err(e) => {
-            warn!("Failed to register Prometheus metrics: {:?}.", e);
-            TimingBlockMetrics::noop()
+            warn!(
+                "Failed to register Prometheus block timing metrics: {:?}.",
+                e
+            );
+            TimingBlockMetrics::<DefaultClock>::noop()
```
timorleph, line 147: Same here.

timorleph (approved):
Looks ok.

ggawryal (approved):
Gj

timorleph (approved):

PR #1512, opened 2023-11-27
Use fix & bump spec-version

kostekIV:


bin/runtime/src/lib.rs
```diff
@@ -95,7 +95,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("aleph-node"),
     impl_name: create_runtime_str!("aleph-node"),
     authoring_version: 1,
-    spec_version: 67,
+    spec_version: 69,
```
DamianStraszak, line 98: Any reason to skip 68?

You also probably want to bump versions in Cargo.toml `aleph-node` and `aleph-runtime`

kostekIV: versions bumped and spec_version set to 68

DamianStraszak (approved):

PR #1515, opened 2023-11-28
A0-3518: Move `verify` weighting to the proper place

pmikolajczyk41:
# Description

Although we don't have benchmarks for jellyfish verifier (and thus the pallet charges 0 weight for verification), we should enrich `WeightInfo` trait with `verify` method that returns for now `0` and use it in both pallet and the extension.

## Type of change

### DIFF
--- a/pallets/baby-liminal/src/weights.rs
+++ b/pallets/baby-liminal/src/weights.rs
@@ -31,6 +31,7 @@ pub trait WeightInfo {
     fn overwrite_equal_key(key_length: u32) -> Weight;
     fn overwrite_key(key_length: u32) -> Weight;
     fn delete_key(key_length: u32) -> Weight;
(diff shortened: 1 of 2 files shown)


pallets/baby-liminal/src/weights.rs
```diff
@@ -31,6 +31,7 @@ pub trait WeightInfo {
     fn overwrite_equal_key(key_length: u32) -> Weight;
     fn overwrite_key(key_length: u32) -> Weight;
     fn delete_key(key_length: u32) -> Weight;
+	fn verify() -> Weight;
```
deuszx, line 34: Cargo fmt

deuszx (approved):

kroist (approved):

PR #1516, opened 2023-11-28
A0-3554: remove pruning option for rocksdb

fixxxedpoint:
# Description

Removes `rocksdb` as the db back-end when one wants to use database pruning. Pruning when `rocksdb` is used works differently than when we use `paritydb`, i.e. some pruning state is stored in-memory instead of being db-backed. In our tests, rocksdb-pruning was pretty unstable - sometimes it works, sometimes it allocates whole available memory and crashes. In case of `paritydb` we haven't observed such problems. It also renames `--experimental-pruning` cli option to `enable-pruning`.

## Type of change

# Checklist:

### DIFF
--- a/bin/node/src/aleph_cli.rs
+++ b/bin/node/src/aleph_cli.rs
@@ -42,11 +42,10 @@ pub struct AlephCli {
     #[clap(long, default_value_t = 20)]
     max_nonfinalized_blocks: u32,
 
-    /// Experimental flag, allows pruning
-    ///
-    /// TURNING THIS FLAG ON, CAN LEAD TO MALICIOUS BEHAVIOUR AND CAN BE PUNISHED ACCORDINGLY!
+    /// Enable database pruning
+    /// Note that we only support pruning with ParityDB.
     #[clap(long, default_value_t = false)]
-    experimental_pruning: bool,
+    enable_pruning: bool,
 
     /// Maximum bit-rate per node in bytes per second of the alephbft validator network.
     #[clap(long, default_value_t = 64 * 1024)]
@@ -90,8 +89,8 @@ impl AlephCli {
         self.max_nonfinalized_blocks
     }
 
-    pub fn experimental_pruning(&self) -> bool {
-        self.experimental_pruning
+    pub fn enable_pruning(&self) -> bool {
+        self.enable_pruning
     }
 
     pub fn alephbft_bit_rate_per_connection(&self) -> u64 {
--- a/bin/node/src/aleph_cli.rs
+++ b/bin/node/src/aleph_cli.rs
@@ -42,8 +42,9 @@ pub struct AlephCli {
     #[clap(long, default_value_t = 20)]
     max_nonfinalized_blocks: u32,
 
-    /// Enable database pruning
-    /// Note that we only support pruning with ParityDB.
+    /// Enable database pruning. It removes older entries in the state-database. Pruning of blocks is not supported.
+    /// Note that we only support pruning with ParityDB database backend.
+    /// See also `--state-pruning` option for more details.
     #[clap(long, default_value_t = false)]
     enable_pruning: bool,
 
--- a/bin/node/src/pruning_config.rs
+++ b/bin/node/src/pruning_config.rs
@@ -33,7 +33,7 @@ impl PruningConfigValidator {
         state_pruning_changed || blocks_pruning_changed
     }
 
-    pub fn validate_and_fix_pruning_options(cli: &mut Cli) -> Self {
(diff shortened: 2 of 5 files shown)


bin/node/src/pruning_config.rs
```diff
@@ -0,0 +1,151 @@
...
+use primitives::DEFAULT_SESSION_PERIOD;
+use sc_cli::{Database, DatabasePruningMode, PruningParams};
+use static_assertions::const_assert;
+
+use crate::Cli;
+
+/// Anything greater than 1800, which is two normal length sessions, should be enough.
+/// We need to be able to read back so many states to retrieve the list of authorities for a session.
+const MINIMAL_STATE_PRUNING: u32 = 2048;
+const_assert!(MINIMAL_STATE_PRUNING >= 2 * DEFAULT_SESSION_PERIOD);
```
timorleph, line 11: <3

bin/node/src/pruning_config.rs
```diff
@@ -0,0 +1,151 @@
...
+                *db = DEFAULT_DATABASE_FOR_PRUNING;
+            }
+        }
+        result
+    }
+
+    pub fn report_pruning_validation_result(self) {
+        if !self.pruning_enabled {
+            if self.overwritten_pruning {
+                warn!("Pruning not enabled. Switching to keeping all block bodies and states.");
```
timorleph, line 99: Maybe add instructions how to actually enable it, i.e. "Please use --enable-pruning to...".

bin/node/src/pruning_config.rs
```diff
@@ -0,0 +1,151 @@
...
+
+        if !pruning_enabled {
+            // We need to override state pruning to our default (archive), as substrate has 256 by default.
+            // 256 does not work with our code.
+            cli.run.import_params.pruning_params.state_pruning = Some(DEFAULT_STATE_PRUNING);
+            cli.run.import_params.pruning_params.blocks_pruning = DEFAULT_BLOCKS_PRUNING;
+            return result;
+        }
+
+        match cli
```
timorleph, line 56: Nit: it would be slightly nicer if these `match`es were methods called here by name.

timorleph (approved):
Nice, only small comments.

bin/node/src/pruning_config.rs
ggawryal: And maybe panic in such case

ggawryal (approved):

bin/node/src/aleph_cli.rs
```diff
@@ -42,11 +42,10 @@ pub struct AlephCli {
     #[clap(long, default_value_t = 20)]
     max_nonfinalized_blocks: u32,
 
-    /// Experimental flag, allows pruning
-    ///
-    /// TURNING THIS FLAG ON, CAN LEAD TO MALICIOUS BEHAVIOUR AND CAN BE PUNISHED ACCORDINGLY!
+    /// Enable database pruning
+    /// Note that we only support pruning with ParityDB.
```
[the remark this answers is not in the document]
fixxxedpoint: It was tested with version that allows a node to create arbitrary many non-finalized blocks. Current version blocks this scenario. I've even prepared  a simple automated test for it...We will discuss it offline.

bin/node/src/pruning_config.rs
```diff
@@ -0,0 +1,151 @@
...
+impl PruningConfigValidator {
+    fn pruning_changed(params: &PruningParams) -> bool {
+        let state_pruning_changed =
+            params.state_pruning.unwrap_or(DEFAULT_STATE_PRUNING) != DEFAULT_STATE_PRUNING;
+        let blocks_pruning_changed = params.blocks_pruning != DEFAULT_BLOCKS_PRUNING;
+
+        state_pruning_changed || blocks_pruning_changed
+    }
+
+    pub fn validate_and_fix_pruning_options(cli: &mut Cli) -> Self {
```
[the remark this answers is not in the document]
fixxxedpoint: changed

bin/node/src/pruning_config.rs
fixxxedpoint: moved

bin/node/src/pruning_config.rs
timorleph: I'm satisfied with just the change you made now, but if you want to panic, panicking on every misconfiguration separately is also fine, there usually won't be that many of them – and this is what happens if you misspell multiple CLI arguments anyway, don't ask me how I know. ;p

ggawryal: ok let's leave it as it is

PR #1517, opened 2023-11-28
A0-3518: Proper charging in the chain extension

pmikolajczyk41:
# Description

Final part of the A0-3518. This time we bring back proper weight (gas) charging together with tests. There is also one follow-up task - benchmarking and charging argument decoding (already in Jira).

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/baby-liminal-extension/src/args.rs
+++ b/baby-liminal-extension/src/args.rs
@@ -20,6 +20,16 @@ pub struct StoreKeyArgs {
     pub key: Vec<u8>,
 }
 
+impl StoreKeyArgs {
+    /// Returns the approximate size of the encoded key. It is an upper bound, because the
+    /// encoding must also include the vector length.
+    #[cfg(feature = "runtime")]
+    pub fn approximate_key_size(encoded_len: usize) -> usize {
+        use sp_std::mem::size_of;
+        encoded_len - size_of::<AccountId>() - size_of::<VerificationKeyIdentifier>()
+    }
+}
+
 /// A struct describing layout for the `verify` chain extension.
 #[derive(Clone, Debug, PartialEq, Eq, scale::Encode, scale::Decode)]
 pub struct VerifyArgs {
--- a/baby-liminal-extension/src/backend/mod.rs
+++ b/baby-liminal-extension/src/backend/mod.rs
@@ -66,7 +66,6 @@ where
     pub fn store_key<BackendExecutor: BackendExecutorT, Environment: EnvironmentT>(
         mut env: Environment,
     ) -> ChainExtensionResult<RetVal> {
-        // todo: charge weight, validate args
         let args = env.read_as_unbounded(env.in_len())?;
         let status = match BackendExecutor::store_key(args) {
             Ok(()) => STORE_KEY_SUCCESS,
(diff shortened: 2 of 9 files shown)


baby-liminal-extension/src/args.rs
```diff
@@ -20,6 +20,16 @@ pub struct StoreKeyArgs {
     pub key: Vec<u8>,
 }
```
DamianStraszak, line 21: Do we have to even expose `StoreKey` in the extension? I don't see a case when this would be useful. It's `verify` that we want.

baby-liminal-extension/src/backend/mod.rs
```diff
@@ -78,12 +101,38 @@ where
...
+    pub fn verify<
+        BackendExecutor: BackendExecutorT,
+        Environment: EnvironmentT,
+        Weighting: WeightInfo,
+    >(
         mut env: Environment,
     ) -> ChainExtensionResult<RetVal> {
-        // todo: charge weight, validate args
+        // ------- Pre-charge optimistic weight. ---------------------------------------------------
+        let pre_charge = env.charge_weight(Weighting::verify())?;
```
DamianStraszak, line 112: `Weighting::verify()` should take some arguments, not sure which at this point, but the cost depends on something like "length of public inputs". Maybe not so crucial for the PoC though, but important to correct later.

baby-liminal-extension/src/backend/mod.rs
```diff
@@ -78,12 +101,38 @@ where
...
+        let result = BackendExecutor::verify(args);
+
+        // ------- Adjust weight if needed. --------------------------------------------------------
+        match &result {
+            // In the failure case, if pallet provides us with a post-dispatch weight, we can make
+            // an adjustment.
+            Err((_, Some(actual_weight))) => env.adjust_weight(pre_charge, *actual_weight),
+            // Otherwise (positive case, or pallet doesn't provide us with any adjustment hint), we
+            // don't need to do anything.
+            Ok(_) | Err((_, None)) => {}
```
DamianStraszak, line 131: This suggests that in the "Ok()" case we never need to adjust weight, which would suggest the `pre_charge` is super accurate. Not sure it's so easy -- but it surely depends on the details of the proof system. For a PoC we don't need to deal with these details.

DamianStraszak (approved):
This looks very good, but at some point we will need to fill in the gaps (benchmarks, more detailed computation of weight).

deuszx (approved):
Looks good. Although, I think it's the kind of code that is difficult to review by someone not fully familiar with the problem.

baby-liminal-extension/src/backend/mod.rs
pmikolajczyk41: sure, it will, but we are still missing benchmarks and thus it's no-arg

baby-liminal-extension/src/args.rs
pmikolajczyk41: discussed offline

baby-liminal-extension/src/args.rs
```diff
@@ -20,6 +20,16 @@ pub struct StoreKeyArgs {
...
 
+impl StoreKeyArgs {
+    /// Returns the approximate size of the encoded key. It is an upper bound, because the
+    /// encoding must also include the vector length.
+    #[cfg(feature = "runtime")]
+    pub fn approximate_key_size(encoded_len: usize) -> usize {
+        use sp_std::mem::size_of;
+        encoded_len - size_of::<AccountId>() - size_of::<VerificationKeyIdentifier>()
```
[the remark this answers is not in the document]
pmikolajczyk41: will be removed in the next PR

PR #1521, opened 2023-11-29
A0-3015: Finality rate metrics

woocash2:
# Description

Introduce `FinalityRateMetrics` - they measure the number of own finalized blocks and the number of own hopeless blocks (our blocks which are forking the finalized chain).

### Implementation details:
The interface consists of `{ report_own_imported, report_finalized }`. `report_own_imported` is called when own block is imported and is added to internal storage of `FinalityRateMetrics` object. `report_finalized` is called when **any** finalized block appears and:
* if it is own block, it should have been added to internal storage when reporting import - we count it as **own finalized**
* all blocks which are on the same level as the given newly finalized block are removed from storage, and if they are not the given block, they are counted as **own hopeless**.

## Type of change

# Checklist:

- I have added tests


finality-aleph/src/metrics/all_block.rs
```diff
@@ -1,21 +1,67 @@
...
+                        self.finality_rate_metrics
+                            .report_own_imported(block_hash, number);
+                    }
+                }
+                Checkpoint::Finalized => self
+                    .finality_rate_metrics
+                    .report_finalized(block_hash, number),
+                _ => {}
+            }
+        }
```
kostekIV, line 65: mv this inside fn report_block of the FinalityRateMetrics.

finality-aleph/src/metrics/finality_rate.rs
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /// Stores the imported block's hash. Assumes that the imported block is own.
+    pub fn report_own_imported(&mut self, hash: BlockHash, number: BlockNumber) {
+        let imported_cache = match self {
+            FinalityRateMetrics::Prometheus { imported_cache, .. } => imported_cache,
+            FinalityRateMetrics::Noop => return,
+        };
+
+        match imported_cache.entry(number) {
```
kostekIV, line 43: use `entry(number).or_default()`

finality-aleph/src/metrics/finality_rate.rs
```diff
@@ -0,0 +1,162 @@
...
+                            if **h == hash {
+                                own_finalized.inc();
+                            }
+                            **h != hash
+                        })
+                        .count() as u64,
+                );
+                entry.remove();
+            }
+            Entry::Vacant(_) => (),
```
kostekIV, line 80: nit: `_ -> {}`

finality-aleph/src/finalization.rs
```diff
@@ -26,7 +26,7 @@ where
     C: HeaderBackend<B> + LockImportRun<B, BE> + Finalizer<B, BE>,
 {
     client: Arc<C>,
-    metrics: AllBlockMetrics,
+    metrics: RefCell<AllBlockMetrics>,
```
kostekIV, line 29: why refcell?

finality-aleph/src/metrics/finality_rate.rs
```diff
@@ -0,0 +1,162 @@
...
+                        .iter()
+                        .filter(|h| {
+                            if **h == hash {
+                                own_finalized.inc();
+                            }
+                            **h != hash
+                        })
+                        .count() as u64,
+                );
+                entry.remove();
```
kostekIV, line 78: Not all blocks in the chain history have justifications. That means that fresh node will kepp in `imported_cache` all of them and will only get rid of them in case of restart. Lets introduce some form of pruning old entries in the cache.

finality-aleph/src/metrics/finality_rate.rs
```diff
@@ -0,0 +1,162 @@
...
+                        .get()
+                        .iter()
+                        .filter(|h| {
+                            if **h == hash {
+                                own_finalized.inc();
+                            }
+                            **h != hash
+                        })
+                        .count() as u64,
+                );
```
kostekIV, line 77: This could be more readable if done in this way:
```
let hashes = entry.get();
let new_hopeless_count = iter.filter(|h| **h != hash).count();
own_hopeless.inc_by(new_hopeless_count);
own_finalized.inc_by(hashes.len() - new_hopeless_count);
```

finality-aleph/src/metrics/finality_rate.rs
```diff
@@ -0,0 +1,162 @@
...
+
+use primitives::{BlockHash, BlockNumber};
+use substrate_prometheus_endpoint::{register, Counter, PrometheusError, Registry, U64};
+
+#[derive(Clone)]
+pub enum FinalityRateMetrics {
+    Prometheus {
+        own_finalized: Counter<U64>,
+        own_hopeless: Counter<U64>,
+        imported_cache: HashMap<BlockNumber, Vec<BlockHash>>,
```
kostekIV, line 11: nit: this vec should be set actually

kostekIV (changes requested):

finality-aleph/src/finalization.rs
woocash2: Otherwise I would have to change `finalize_block` in `trait BlockFinalizer` to accept `&mut self` instead `&self`.
This enforces changing traits in `block::substrate::finalizer` and `finalization` to accept `&mut self` instead of `&self`. I didn't want to change these traits if I didn't have to, but we can consider - what would be your opinion on that?

finality-aleph/src/metrics/all_block.rs
woocash2: can be done

finality-aleph/src/block/substrate/mod.rs
```diff
@@ -91,7 +94,8 @@ impl BlockImport<Block> for BlockImporter {
             import_existing: false,
             state: None,
         };
-        self.metrics.report_block(hash, Checkpoint::Importing);
+        self.metrics
+            .report_block(hash, Checkpoint::Importing, Some(number), Some(own));
         self.importer.import_blocks(origin, vec![incoming_block]);
```
ggawryal, line 99: This is a different behavior than we've had before - we change `origin`of imported block that is passed to the actual importer. Are we sure this is correct?

finality-aleph/src/block/substrate/mod.rs
```diff
@@ -76,9 +75,13 @@ impl BlockImporter {
 }
 
 impl BlockImport<Block> for BlockImporter {
-    fn import_block(&mut self, block: Block) {
-        let origin = BlockOrigin::NetworkBroadcast;
+    fn import_block(&mut self, block: Block, own: bool) {
+        let origin = match own {
+            true => BlockOrigin::Own,
+            false => BlockOrigin::NetworkBroadcast,
```
ggawryal, line 81: If we leave this (see other comment), maybe we should comment why we consider all other cases or `BlockOrigin` to default to `NetworkBroadcast`?

finality-aleph/src/metrics/finality_rate.rs
woocash2: Why tho? We don't ever search for any element in this collection - only increase it / iterate through its elements once.

finality-aleph/src/metrics/finality_rate.rs
ggawryal: Maybe use LruCache and bound the vector inside to avoid attacks that would cause out of memory crashes (even if they are unpractical)?

woocash2: Hmm, the blocks we store in `imported_cache` are only the blocks we produced. If we produced a block on some `level` I assume that `top_finalized < level`. So for each key in this `imported_cache` eventually there should come a finalized block which will erase it. 

That's my reasoning, but still I'm eager to implement some pruning mechanism just in case...

finality-aleph/src/finalization.rs
ggawryal: Metrics are Clone, so can't we simply clone them instead of storing them in a RefCell? All prometheus primitives (Counter, Histogram etc) are Arcs so they can be safely cloned.

finality-aleph/src/metrics/all_block.rs
```diff
@@ -1,21 +1,67 @@
...
     }
 
     /// Triggers all contained block metrics.
-    pub fn report_block(&self, hash: BlockHash, checkpoint: Checkpoint) {
-        self.timing_metrics.report_block(hash, checkpoint);
+    pub fn report_block(
+        &mut self,
+        block_hash: BlockHash,
+        checkpoint: Checkpoint,
+        block_number: Option<BlockNumber>,
```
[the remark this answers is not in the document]
woocash2: Hmm, I had a problem in legacy data provider to obtain the block number. I'll revisit and see if I can actually do it. If yes then sure, we can use `BlockId`

finality-aleph/src/finalization.rs
woocash2: The problem is the `imported_cache` which we want to mutate - cloning whole metrics won't be good. Tho I see that in `TimingBlockMetrics` the cache you implemented is under `Arc<Mutex<...>>`. I can do it the same way :+1:

finality-aleph/src/block/substrate/mod.rs
woocash2: ok

finality-aleph/src/metrics/finality_rate.rs
woocash2: In `Vec` there are our produced blocks at a particular level. Given the recent precaution measures in terms of equivocations these `Vecs` should be at most singletons :stuck_out_tongue:

I'm thinking on a different mechanism - Given that our produced blocks come in order of increasing level, `imported_cache` could just be a `Deque` - when a block comes, push it back, when a finalized block comes, erase the front until the appropriate level. Just need to verify with our main `sync` guys that block do come in such order from sync.

finality-aleph/src/metrics/finality_rate.rs
kostekIV: its a small nit. The iteration could e replaced instead with finding the element in the collection plus the hashes here should be unique. Feel free to resolve

finality-aleph/src/metrics/finality_rate.rs
```diff
@@ -0,0 +1,191 @@
...
+
+const MAX_CACHE_SIZE: usize = 1800;
+const MAX_INNER_SIZE: usize = 64;
+
+#[derive(Clone)]
+pub enum FinalityRateMetrics {
+    Prometheus {
+        own_finalized: Counter<U64>,
+        own_hopeless: Counter<U64>,
+        imported_cache: Arc<Mutex<LruCache<BlockNumber, LruCache<BlockHash, ()>>>>,
```
kostekIV, line 19: 1. add type alias
```
type ImportedHashesCache = Arc<Mutex<LruCache<BlockNumber, LruCache<BlockHash, ()>>>>;
```
2. Why lru instead of bounded vec or just set? this seems strange

finality-aleph/src/finalization.rs
kostekIV: Arc is fine, it is a solution that we use in the project.

finality-aleph/src/metrics/finality_rate.rs
kostekIV: some combination of LruCache plus bounded vec looks ok

finality-aleph/src/metrics/finality_rate.rs
woocash2: * Right, Changed to `BoundedVec`.
* Also added the suggested type alias

finality-aleph/src/metrics/finality_rate.rs
woocash2: Changed inner data type from `LruCache` to `BoundedVec`

finality-aleph/src/block/substrate/mod.rs
woocash2: I'm afraid `FinalityRateMetrics` require this change. Somehow they have to get the knowledge that the block was produced by us. Unless you have some other suggestions..

finality-aleph/src/metrics/finality_rate.rs
kostekIV: Ok, Im satisfied with this. @ggawryal if you dont have any more comments here, resolve this conversation.

finality-aleph/src/block/substrate/mod.rs
kostekIV: Ok, then lets leave it. Last change related to that is to update docs of the function import_block in src/sync/handler/mod.rs at line 476.

woocash2: ok, thx for catching the doc update - fixed it.

finality-aleph/src/metrics/finality_rate.rs
ggawryal: I forgot that there are only our blocks in the cache... In this case plain vec probably would be sufficient, giving a (theoretical) bound of the size of a block producing committee, but we can defensively leave it as it is :)

kostekIV (approved):

PR #1527, opened 2023-12-05
Explicit enumeration

kostekIV:


bin/runtime/src/lib.rs
```diff
@@ -859,59 +859,59 @@ impl pallet_proxy::Config for Runtime {
...
-        Vesting: pallet_vesting,
-        Utility: pallet_utility,
-        Multisig: pallet_multisig,
-        Sudo: pallet_sudo,
-        Contracts: pallet_contracts,
-        NominationPools: pallet_nomination_pools,
-        Identity: pallet_identity,
-        CommitteeManagement: pallet_committee_management,
-        Proxy: pallet_proxy,
+        System: frame_system = 1,
```
DamianStraszak, line 862: I think it would be great to run subwasm on this change. I suspect it's 0-based by default :P

pmikolajczyk41: @kostekIV have you run subxt for standard runtime as well? if so (no changes), then it might be 1-based indeed...

kostekIV: nope, of course its from 0. Fixed

pmikolajczyk41:
1. have you regenerate standard subxt runtime?
2. shouldn't we have some gap after removing randomness pallet?

DamianStraszak (approved):

kostekIV:
> 1. have you regenerate standard subxt runtime?
> 
>     2. shouldn't we have some gap after removing randomness pallet?

1. yep, now I have pushed it. Some minor changes most likely not related to this pr.
2. Nope, we havent yet deleted it. Its planned for after the release 13

PR #1536, opened 2023-12-11
A0-3618: move proof verification from pallet to chain extension

pmikolajczyk41:
# Description

As decided recently, we don't need verification in the pallet. It is sufficient to have directly in the chain extension.

Since we are removing jellyfish soon, I didn't migrate the verifier tests.

Technically this is a breaking change for aleph-client and cliain, but actually it's not used anywhere, hidden behind a feature flag and there will be more changes like this soon, and thus I bumped just the minor version.

## Type of change

# Checklist:


aleph-client/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aleph_client"
-version = "3.7.2"
+version = "3.8.0"
```
Yithis, line 3: As far as I understand those changes break aleph client (unused) functionalities. In this case shouldn't we update major version?

Yithis (approved):
Looks good to me

DamianStraszak (approved):

PR #1544, opened 2023-12-15
A0-2733: loop-select-ng

fixxxedpoint:
# Description

Few changes around `loop-select` statements, ensuring no channel is called after `is_terminated` returns `true`, etc.
Introduced `Bottom` type for statically checked infinite computations.

## Type of change

- mainly refactor


finality-aleph/src/sync/service.rs
```diff
@@ -730,54 +731,61 @@ where
     }
 
     /// Stay synchronized.
-    pub async fn run(mut self) {
+    pub async fn run(mut self) -> Result<Bottom, String> {
+        if self.additional_justifications_from_user.is_terminated() {
+            return Err(
+                "Channel with additional justifications from user closed before we started.",
+            )?;
```
ggawryal, line 738:
```suggestion
            );
```

finality-aleph/src/sync/service.rs
```diff
@@ -730,54 +731,61 @@ where
...
+        }
+        if self.blocks_from_creator.is_terminated() {
+            return Err("Channel with own blocks closed before we started.")?;
```
ggawryal, line 741:
```suggestion
            return Err("Channel with own blocks closed before we started.");
```

finality-aleph/src/data_io/data_provider.rs
```diff
@@ -280,25 +280,23 @@ where
...
+            best_block_in_session = self.get_best_block_in_session(best_block_in_session).await;
+            if let Some(best_block) = &best_block_in_session {
+                self.update_data(best_block);
             }
         }
     }
+
+    pub async fn run(mut self, exit: oneshot::Receiver<()>) {
+        tokio::select! {
+            _ = self.main_loop() => error!(target: "aleph-data-store", "Task for refreshing best chain finished."),
```
ggawryal, line 296: Compiler is not able to deduce that this is unreachable? If it isn't, I think `error` is indeed the best option out of `panic!` and `{}`.

ggawryal (approved):
LGTM, I like fusing notification streams in substrate network wrapper :)

finality-aleph/src/block/substrate/status_notifier.rs
```diff
@@ -64,6 +67,7 @@ impl ChainStatusNotifier<Header> for SubstrateChainStatusNotifier {
                 .map(|block| ChainStatusNotification::BlockImported(block.header))
                 .ok_or(Error::ImportStreamClosed)
             }
+            complete => Err(Error::AllStreamsTerminated)
```
timorleph, line 70: Is `complete` a keyword, or is something that I don't understand going on here?

finality-aleph/src/data_io/data_store.rs
```diff
@@ -237,26 +238,36 @@ where
         )
     }
 
-    pub async fn run(&mut self, mut exit: oneshot::Receiver<()>) {
+    pub async fn run(&mut self, mut exit: oneshot::Receiver<()>) -> Result<(), String> {
```
timorleph, line 241: Error type plx.

finality-aleph/src/network/gossip/service.rs
```diff
@@ -503,36 +504,26 @@ impl<N: RawNetwork, ES: EventStream<N::PeerId>, AD: Data, BSD: Data> Service<N,
         info!(target: LOG_TARGET, "{}", status);
     }
 
-    pub async fn run(mut self) {
+    pub async fn run(mut self) -> Result<Bottom, String> {
```
timorleph, line 507: Error type plx.

finality-aleph/src/party/manager/aggregator.rs
```diff
@@ -91,7 +93,7 @@ async fn run_aggregator<B, C, CN, LN, JS>(
     session_boundaries: &SessionBoundaries,
     mut metrics: AllBlockMetrics,
     mut exit_rx: oneshot::Receiver<()>,
-) -> Result<(), ()>
+) -> Result<(), String>
```
timorleph, line 96: Error type plx.

finality-aleph/src/party/manager/aggregator.rs
```diff
@@ -209,6 +208,13 @@ where
                 exit,
             )
             .await;
+            let result = match result {
+                Ok(_) => Ok(()),
+                Err(err) => {
```
timorleph, line 213: If you have a real error then better return it rather than masking it here.

finality-aleph/src/sync/service.rs
```diff
@@ -730,54 +731,61 @@ where
     }
 
     /// Stay synchronized.
-    pub async fn run(mut self) {
+    pub async fn run(mut self) -> Result<Bottom, String> {
```
timorleph, line 734: Error types plx, here especially important since you get wrappers much more naturally than doing a `format!` in a `map_err`.

primitives/src/lib.rs
```diff
@@ -61,6 +61,12 @@ pub type Nonce = u32;
 /// A hash of some data used by the chain.
 pub type Hash = sp_core::H256;
 
+/// Can be used to represent an infinite computation. One can use it together
+/// with [Result<Bottom, T>](Result) to indicate that a computation can finish
+/// only when some error occurs, otherwise it will spin forever
+/// (compiler/type-checker will enforce this property).
```
timorleph, line 67: This is kinda cool (⊥ is my favourite type in general), but I'm not convinced it's worth the effort if it's not a native type. I doubt everyone will start using it now, so we will only have this in a couple semi-random places, where it'll confuse any readers.

If we want to adopt this, then I would suggest making an ADR for that and actually requiring people to put ⊥s everywhere where there should be a never ending computation.

finality-aleph/src/data_io/data_provider.rs
timorleph: Hm, I don't want to keep `Bottom` as a type in general (see another comment in this review), but wouldn't an empty `match` be technically exhaustive, and thus solve this issue? That is:
```
impossible_result = self.main_loop() => match impossible_result {},
```

timorleph (changes requested):
Cool, although I'm worried that the changes are not consistent between loops – it would be great if you could make them so.

clique/src/service.rs
```diff
@@ -205,12 +208,94 @@ where
...
+                if let Some(address) = self.peer_address(&public_key) {
+                    self.spawn_new_outgoing(public_key, address, result_for_parent.clone());
+                }
+            }
+        }
+    }
+
     /// Run the service until a signal from exit.
-    pub async fn run(mut self, mut exit: oneshot::Receiver<()>) {
+    pub async fn run(mut self, mut exit: oneshot::Receiver<()>) -> Result<(), String> {
```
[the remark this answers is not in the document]
fixxxedpoint: I am going to name it `GenericStringError` ;P

primitives/src/lib.rs
DamianStraszak: Can't we use https://doc.rust-lang.org/std/primitive.never.html ?

fixxxedpoint: https://doc.rust-lang.org/reference/types/never.html

clique/src/service.rs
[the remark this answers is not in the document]
fixxxedpoint: I have mixed feelings with this. Only thing we can do with these errors is logging, we do not handle them in any way. I can introduce a new error that will be just a wrapper for `String`. Then I will have to write `Into/From` for it for `&str` conversions for the `?` operator, etc. It will be pure boilerplate. For these errors we don't even care about backtraces. Perhaps just use `Error` from anyhow (new dependency)?

finality-aleph/src/block/substrate/status_notifier.rs
fixxxedpoint: https://docs.rs/futures/latest/futures/macro.select.html

finality-aleph/src/metrics/chain_state.rs
```diff
@@ -176,21 +174,29 @@ pub async fn run_chain_state_metrics<
...
-            maybe_block = finality_notifications.next() => match maybe_block {
+            maybe_block = finality_notifications.next(), if !finality_notifications.is_terminated() => match maybe_block {
                 Some(block) => handle_block_finalized(block, &metrics),
                 None => warn!(
                     target: LOG_TARGET,
                     "Finality notification stream ended unexpectedly"
                 ),
             },
-            maybe_transaction = transaction_pool_info_provider.next_transaction() => {
+            maybe_transaction = transaction_pool_info_provider.next_transaction(), if !transaction_pool_info_provider_terminated => {
```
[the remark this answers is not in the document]
fixxxedpoint: Short version: this one doesn't `early` exit on an empty channel. Why? No idea. @ggawryal ? I will just rewrite it into early-exit version similar to clique.

primitives/src/lib.rs
DamianStraszak: Right -- the question is then if we need `Bottom` at all

finality-aleph/src/data_io/data_provider.rs
fixxxedpoint: What about using `futures::future::select` ?

finality-aleph/src/party/manager/aggregator.rs
fixxxedpoint: Do you have something more specific on your mind, cuz I can't just return some error here without refactoring Task/Subtask, etc.

primitives/src/lib.rs
fixxxedpoint: I am ok with introducing it just in few random places (why do we need to use it everywhere immediately?), is more descriptive for infinite loops/services than `Result<(), _>`, it gives some free additional compiler checks, it doesn't require any additional maintenance (no methods, converters etc.), and can be switched to `!` when it become stable (are we going to use it then or pretend it's not there?). ADR for it is an overkill, since it will be a rust's feature anyway. I can change it's name for something more descriptive. It was interesting to see that you can do it in rust. Anyway, should I remove it?

finality-aleph/src/metrics/chain_state.rs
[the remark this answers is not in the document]
ggawryal: Yes, we can rewrite this in clique style (had no particular reason why in mind, when I was writing that :))

finality-aleph/src/data_io/data_provider.rs
```diff
@@ -280,25 +280,23 @@ where
...
+            _ = exit => debug!(target: "aleph-data-store", "Task for refreshing best chain received exit signal. Terminating."),
```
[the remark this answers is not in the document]
fixxxedpoint: changed

finality-aleph/src/data_io/data_provider.rs
ggawryal: Looks good

clique/src/service.rs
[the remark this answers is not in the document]
DamianStraszak: I think defining the errors using `thiserror` could make sense. This way you have proper types implementing `Error` and in the future you can handle this errors easily. But let's wait for @timorleph

finality-aleph/src/data_io/data_store.rs
```diff
@@ -237,26 +238,36 @@ where
...
         let mut maintenance_clock = Delay::new(self.config.periodic_maintenance_interval);
         let mut import_stream = self.client.import_notification_stream();
+        if import_stream.is_terminated() {
+            return Err("Blocks import notification stream was closed.")?;
```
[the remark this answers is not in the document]
fixxxedpoint: changed

finality-aleph/src/data_io/data_store.rs
```diff
@@ -237,26 +238,36 @@ where
...
+        }
         let mut finality_stream = self.client.finality_notification_stream();
+        if finality_stream.is_terminated() {
+            return Err("Finalized-blocks notification stream was closed.")?;
```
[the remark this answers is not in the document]
fixxxedpoint: changed

finality-aleph/src/sync/service.rs
fixxxedpoint: changed

finality-aleph/src/sync/service.rs
fixxxedpoint: changed

clique/src/service.rs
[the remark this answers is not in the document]
timorleph: Just normal enum errors are much nicer, lets not add another layer of problems in case we ever manage to get to implement actual tests for these things (in which case "stringy" errors would be extremely annoying). If you are worried about the boilerplate of defining all the `Display`s and whatnot then @DamianStraszak 's suggestion of `thiserror` is reasonable – tbh we probably should've been using it all along (at the very least in node, ABFT might have some argument about minimizing deps).

finality-aleph/src/party/manager/aggregator.rs
timorleph: Ugh, they probably need a refactor then. Don't necessarily do this in this PR, but please create a task for it.

primitives/src/lib.rs
timorleph: The problem with doing it in just a couple random places is that there is additional mental load when trying to figure out why some places have it and others don't – and when the answer is "because it was done in one specific PR" it's not even something that can really be guessed. There is value in consistency. This would be slightly less of a problem if we used native Rust functionalities, but still somewhat, we don't use all the Rust functionalities that exist and would technically improve compiler checks. See for example the infamous lifetimed keychain reference in ABFT.

That being said, I see a couple reasonable approaches to this:
1. Leave this PR in its current shape, but create an ADR saying that we use `Bottom` for all functions where it's applicable and switch to `!` as soon as it's stable. This way we get temporarily confusing code, that will eventually get homogenized, which is fine. The ADR can even be very short, the argument for using this type is pretty straightforward.
2. Remove `Bottom`, but create an ADR for using some type like it eventually, probably once `!` is stabilized.
3. Remove `Bottom` and create an ADR explicitly forbidding the use of `!` to avoid confusion.
4. Remove `Bottom` and do nothing.
This is roughly in my order of preference. Note that 4 is essentially doing 3 implicitly, as there will be a discussion like this every time someone tries to re-add `Bottom`. For 2 & 3 you don't have to create any ADR immediately, it's just something that would eventually have to be done. Also, even if we go for 4 and then accept a PR with `Bottom` (probably in the form of a stabilized `!`), we would still have to at least plan an effort to homogenize the code.

tl;dr Either leave `Bottom` in and create an ADR, or remove bottom and optionally create a task for writing an ADR.

finality-aleph/src/metrics/chain_state.rs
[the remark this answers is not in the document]
fixxxedpoint: done

finality-aleph/src/network/gossip/service.rs
fixxxedpoint: done

finality-aleph/src/party/manager/aggregator.rs
fixxxedpoint: done

finality-aleph/src/party/manager/aggregator.rs
fixxxedpoint: done (added task)

finality-aleph/src/sync/service.rs
fixxxedpoint: done

clique/src/service.rs
```diff
@@ -205,12 +247,99 @@ where
...
+        }
+    }
+
     /// Run the service until a signal from exit.
-    pub async fn run(mut self, mut exit: oneshot::Receiver<()>) {
+    pub async fn run(
+        mut self,
+        mut exit: oneshot::Receiver<()>,
+    ) -> Result<(), CliqueNetworkServiceError> {
+        use CliqueNetworkServiceError as Error;
```
timorleph, line 338: Tbh I think you can just call it `Error` originally, you only have one in this module. :D

clique/src/service.rs
```diff
@@ -85,6 +92,41 @@ pub trait SpawnHandleT {
...
+            }
+            CliqueNetworkServiceError::ConnectionWorker => {
+                write!(f, "`worker_results` channel was closed.")
+            }
+        }
+    }
+}
+
+impl CliqueNetworkServiceError {
+    fn commands_stream_closed() -> Self {
```
timorleph, line 117: These constructors don't really seem necessary, no?

finality-aleph/src/data_io/data_store.rs
```diff
@@ -105,6 +107,43 @@ impl Default for DataStoreConfig {
...
+                write!(f, "Finalized block import notification stream was closed.")
+            }
+            DataStoreError::NetworkMessagesTerminated => {
+                write!(f, "Stream with network messages was closed.")
+            }
+        }
+    }
+}
+
+impl DataStoreError {
```
timorleph, line 133: Same comments as in with the other error – name could just be `Error` and constructors are not necessary.

finality-aleph/src/data_io/legacy/data_store.rs
```diff
@@ -106,6 +108,43 @@ impl Default for DataStoreConfig {
...
+                write!(f, "Finalized block import notification stream was closed.")
+            }
+            DataStoreError::NetworkMessagesTerminated => {
+                write!(f, "Stream with network messages was closed.")
+            }
+        }
+    }
+}
+
+impl DataStoreError {
```
timorleph, line 134: Same.

finality-aleph/src/metrics/chain_state.rs
```diff
@@ -135,59 +186,57 @@ pub async fn run_chain_state_metrics<
...
+    if finality_notifications.is_terminated() {
+        return Err(Error::finalized_blocks_stream_closed());
     }
 
     let metrics = match ChainStateMetrics::new(registry) {
         Ok(metrics) => metrics,
         Err(e) => {
             warn!(target: LOG_TARGET, "Failed to create metrics: {e}.");
-            return;
+            return Err(Error::unable_to_create_metrics(e));
```
timorleph, line 210: You can probably `map_err()?` here, instead of this whole `match`. The log ain't that useful and should be duplicated by error handling outside anyway.

finality-aleph/src/metrics/chain_state.rs
```diff
@@ -31,6 +31,57 @@ const TRANSACTION_CACHE_SIZE: usize = 100_000;
...
+                write!(f, "Finality notification stream ended unexpectedly.")
+            }
+            ChainStateMetricsError::TransactionStreamClosed => {
+                write!(f, "Transaction stream ended unexpectedly.")
+            }
+        }
+    }
+}
+
+impl ChainStateMetricsError {
```
timorleph, line 63: Same.

finality-aleph/src/network/gossip/service.rs
```diff
@@ -77,6 +78,49 @@ impl Display for Error {
...
+                write!(f, "Block sync user message stream ended.")
+            }
+            GossipServiceError::UnableToForwardMessageToUser => {
+                write!(f, "Cannot forward messages to user.")
+            }
+        }
+    }
+}
+
+impl GossipServiceError {
```
timorleph, line 106: Same.

finality-aleph/src/party/manager/aggregator.rs
```diff
@@ -27,6 +29,33 @@ use crate::{
...
+        match self {
+            AggregatorError::MultisignaturesStreamTerminated => {
+                write!(f, "The stream of multisigned hashes has ended.")
+            }
+            AggregatorError::UnableToProcessHash => write!(f, "Error while processing a hash."),
+        }
+    }
+}
+
+impl AggregatorError {
```
timorleph, line 49: Same.

finality-aleph/src/sync/service.rs
```diff
@@ -81,6 +81,70 @@ where
...
+                    f,
+                    "Channel with legacy internal block request from user closed."
+                )
+            }
+            SyncServiceError::CreatorChannelClosed => write!(f, "Channel with own blocks closed."),
+        }
+    }
+}
+
+impl<NetworkError, ChainEventError> SyncServiceError<NetworkError, ChainEventError> {
```
timorleph, line 122: Same.

clique/src/service.rs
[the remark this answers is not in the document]
timorleph: Well, I guess we are not addint `thiserror` then, oh well, maybe next time or when we have some time for a refactor...

timorleph (approved):
Cool, just a couple small nits.

clique/src/service.rs
fixxxedpoint: :D

timorleph: I mean, you can just use `enum` variants as constructors as we do in all other code, right? <_<"

fixxxedpoint: I thought the goal was to make it future-proof. This way we can add more details to them later more easily (e.g. back-traces). Just kidding ;)

finality-aleph/src/metrics/chain_state.rs
fixxxedpoint: done

finality-aleph/src/metrics/chain_state.rs
fixxxedpoint: done

finality-aleph/src/network/gossip/service.rs
fixxxedpoint: done

finality-aleph/src/sync/service.rs
fixxxedpoint: done

finality-aleph/src/network/gossip/service.rs
timorleph: This and the other I left unresolved were not renamed to just `Error` – is there a reason for that?

fixxxedpoint: `ServiceInterface` uses `Error`.

finality-aleph/src/sync/service.rs
fixxxedpoint: done

PR #1550, opened 2023-12-18
A0-2149: Single justification source for sync

woocash2:
# Description

Sync previously had 2 sources of justifications. Now it has only one. The channel for justification creation is done by newly introduced `ChannelProvider<T>` which can output many senders, or can be consumed to output the receiver.

## Type of change

# Checklist:

### DIFF
--- a/bin/node/src/service.rs
+++ b/bin/node/src/service.rs
@@ -8,8 +8,8 @@ use std::{
 use aleph_runtime::{self, opaque::Block, RuntimeApi};
 use finality_aleph::{
     run_validator_node, AlephBlockImport, AlephConfig, AllBlockMetrics, BlockImporter,
-    Justification, JustificationTranslator, MillisecsPerBlock, Protocol, ProtocolNaming,
-    RateLimiterConfig, RedirectingBlockImport, SessionPeriod, SubstrateChainStatus,
+    ChannelProvider, Justification, JustificationTranslator, MillisecsPerBlock, Protocol,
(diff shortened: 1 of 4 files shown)


timorleph (approved):
Nice abstraction.

lesniak43:
w8, the task specifies that this should be done the other way around - we keep the code that returns interface from `Service::new`, and we remove the bare channels passed to `new` (the idea was that we want to hide all the channels under some traits); why was this changed?

woocash2:
> w8, the task specifies that this should be done the other way around - we keep the code that returns interface from `Service::new`, and we remove the bare channels passed to `new` (the idea was that we want to hide all the channels under some traits); why was this changed?

Good point. My understanding is that non-uniform source of justification in sync is the main problem, and my changes fix that. I agree that task description says that sync should not accept any input for justifications (create it by itself and return the sender), but that would require moving the creation of sync before the creation of some substrate-based components (including RPC). Since currently we have a nice separation between substrate components creation (`bin/service`) and our components (`finality-aleph/nodes`) I didn't want to break that. I mean there are some other options - for instance, only creating justification-dependent substrate components in `bin/service` and passing them under some interface to `finality-aleph/nodes` so that their run is triggeret there after passing the justification endpoint - but it just seems a little bit too complicated... Agreeing that sync takes some input was simpler.

lesniak43:
Ok, then update the task description pls. Since @timorleph accepted this change, I think it's fine.

lesniak43 (approved):

PR #1551, opened 2023-12-18
A0-3737: Support for dynamic sudo in FE bootstrap

Marcin-Radecki:
# Description

This PR adds the possibility to spawn FE with a custom sudo account id. 

## Type of change

## Testing

1. [Run with default parameters, Alice is sudo](https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/7258857766) 

image

image

image

2. [Run with custom sudo account id, `5EyDzpeMZqrLsCoUMn6fPVYvGiScMfuA3LPAodQecvSmnLmK` is sudo](https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/7258896842)

image

image

image

## Related PRs

https://github.com/Cardinal-Cryptography/github-actions/pull/36
https://github.com/Cardinal-Cryptography/featurenet-template/pull/19

### DIFF
--- a/.github/workflows/featurenet-update.yml
+++ b/.github/workflows/featurenet-update.yml
@@ -108,7 +108,7 @@ jobs:
     # to prevent this job to be skipped when on of the parent jobs is skipped
     if: ${{ !cancelled() }}
     name: Update featurenet
-    uses: Cardinal-Cryptography/github-actions/.github/workflows/_featurenet-create.yml@A0-3737
+    uses: Cardinal-Cryptography/github-actions/.github/workflows/_featurenet-create.yml@v6
     secrets: inherit
     with:
       featurenet-name: ${{ inputs.featurenet-name }}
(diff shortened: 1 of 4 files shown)


.github/workflows/featurenet-update.yml
```diff
@@ -46,6 +46,11 @@ on:
         description: 'Use short session aleph-node binary'
         required: true
         type: boolean
+      sudo-account-id:
```
pmikolajczyk41, line 49: this MUST be the same as during creation, otherwise there will be problems, right?

Marcin-Radecki: This is a temporary solution and we do have an item which we're working right now to remove all now 5 arguments from `update FE` workflow, namely
* expiration,
* short session checkbox,
* internal checkbox,
* number of validators,
* sudo account id
so that user do not need to retype those. To answer the question - in this case, it won't be a problem since bootstrap chain is not done in `update FE` workflow, but maybe I'll remove it anyway to not cause any confusion.

pmikolajczyk41 (approved):
thank you very much!

bitsnops (approved):

PR #1553, opened 2023-12-20
A0-3735: Add featurenet test automation

Marcin-Radecki:
# Description

This PR
* changes update net test scenario (Mainnet -> `main`) to be Mainnet -> Testnet
* splits update net tests into two workflow for readability
* adds three weekly workflows that tests featurenets internals. They will be migrated to different repo at some point.

## Type of change

## Testing

Links updated after rework https://github.com/Cardinal-Cryptography/aleph-node/pull/1553/commits/34eb77c4cb0e388eb7668fec034f5e07b9eb0e71

* [Weekly test featurenets create (short session, branch)](https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/7287465811)
* [Weekly test featurenets create (short session, tag)](https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/7287465807)
* [Weekly test featurenets create (short session, full sha)](https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/7287465816)
* [Nightly update net test from Testnet to main](https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/7287465815)
* [Nightly update net test from Mainnet to Testnet](https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/7287465809)


.github/workflows/weekly-featurenets-short-session-tag.yml
```diff
@@ -0,0 +1,91 @@
...
+    secrets: inherit
+    with:
+      featurenet-name: 'ops-test-main-tag'
+      aleph-node-image: ${{ needs.get-fqdn-via-tag.outputs.fqdn-image }}
+      validators: '6'
+      expiration: '2h'
+      internal: true
+      delete-first: true
+
+  delete-tag-featurenet:
```
woocash2, line 66: Seems that these 3 workflows are very similar - they differ only by names and numbers used for expiration / validator count. Is there any chance to extracting a single job for creation / deletion of featurenets, to which these 3 workflows would only pass their custom parameters?

.github/workflows/weekly-featurenets-short-session-tag.yml
```diff
@@ -0,0 +1,91 @@
...
+    needs:
+      - get-fqdn-via-tag
+      - build-and-push-featurenet-node-image-via-tag
+    # to prevent this job to be skipped when on of the parent jobs is skipped
+    if: ${{ !cancelled() }}
+    name: Create featurenet from tag
+    uses: Cardinal-Cryptography/github-actions/.github/workflows/_featurenet-create.yml@v6
+    secrets: inherit
+    with:
+      featurenet-name: 'ops-test-main-tag'
```
woocash2, line 59: I see the name being hardcoded here and there. Any chance of extracting it to some constant at the top of the file and referring to that constant instead? This applies to all files.

woocash2 (approved):
Great, these workflows will be very useful.

I left some comments regarding duplication, although I have no experience in writing `yaml`s so treat them more like questions than suggestions - I see these workflows have a lot in common - for example all the workflows have a job named `create-featurenet-via-XXX` which differs only by names / parameters used among the workflows - is there a chance to extract some common logic to avoid duplication? :stuck_out_tongue:

.github/workflows/weekly-featurenets-short-session-tag.yml
Marcin-Radecki: Actually after refactor it will be better, but it seems there's an [issue](https://github.com/actions/runner/issues/2372) in GHA that prevents to do what [docs](https://docs.github.com/en/github-ae@latest/actions/using-workflows/workflow-syntax-for-github-actions#env) suggests, so I'll skip that one

.github/workflows/weekly-featurenets-short-session-tag.yml
Marcin-Radecki: This is very good point and actually it will lead to much better code not only in terms of less duplications, but also in terms of funcional testing - we'll call in tests the same workflow which user calls.

woocash2 (approved):
Very nice, it is way simpler than before

PR #1554, opened 2023-12-21
A0-3759: Replace `HeaderBackend`, `BlockchainEvents` and `SelectChain`

ggawryal:
# Description

## Type of change

None - refactor

# Checklist:

### DIFF
--- a/finality-aleph/src/block/mod.rs
+++ b/finality-aleph/src/block/mod.rs
@@ -233,18 +233,20 @@ where
     fn children(&self, id: BlockId) -> Result<Vec<J::Header>, Self::Error>;
 }
 
-pub trait Info {
+pub trait HeaderBackendStatus {
     fn best_id(&self) -> BlockId;
     fn genesis_hash(&self) -> BlockHash;
     fn finalized_id(&self) -> BlockId;
 }
 
 pub trait HeaderBackend<H: Header>: Send + Sync {
-    type Info: Info;
+    type Status: HeaderBackendStatus;
     type Error: Debug + Display;
     /// Get block header. Returns `None` if block is not found.
     fn header(&self, hash: BlockHash) -> Result<Option<H>, Self::Error>;
-
+    /// Get hash of a block with a given number, on the path from genesis to the currently
+    /// best block. Returned value is unspecified if queried `number` is larger than best block.
     fn hash(&self, number: BlockNumber) -> Result<Option<BlockHash>, Self::Error>;
-    fn info(&self) -> Self::Info;
+    /// Get current status of the header backend.
+    fn status(&self) -> Self::Status;
 }
--- a/finality-aleph/src/block/mod.rs
+++ b/finality-aleph/src/block/mod.rs
@@ -1,6 +1,7 @@
 use std::fmt::{Debug, Display, Error as FmtError, Formatter};
 
 use parity_scale_codec::{Codec, Decode, Encode};
+use sc_utils::mpsc::TracingUnboundedReceiver;
 
 use crate::{BlockHash, BlockNumber};
 
@@ -225,4 +226,43 @@ where
 
     /// Children of the specified block.
     fn children(&self, id: BlockId) -> Result<Vec<J::Header>, Self::Error>;
+}
+
+pub trait Info {
+    fn best_id(&self) -> BlockId;
+    fn genesis_hash(&self) -> BlockHash;
+    fn finalized_id(&self) -> BlockId;
+}
+
+pub trait HeaderBackend<H: Header>: Send + Sync {
+    type Info: Info;
+    type Error: Debug + Display;
+    /// Get block header. Returns `None` if block is not found.
+    fn header(&self, hash: BlockHash) -> Result<Option<H>, Self::Error>;
+
+    fn hash(&self, number: BlockNumber) -> Result<Option<BlockHash>, Self::Error>;
+    fn info(&self) -> Self::Info;
+}
+
+pub trait BlockImportNotification<UH: UnverifiedHeader>: Debug + Send {
+    fn hash(&self) -> BlockHash;
+    fn header(&self) -> &UH;
+    fn is_new_best(&self) -> bool;
+}
+
+pub trait FinalityNotification<UH: UnverifiedHeader>: Debug + Send {
+    fn hash(&self) -> BlockHash;
+    fn header(&self) -> &UH;
+}
+
+pub type ImportNotifications<IN> = TracingUnboundedReceiver<IN>;
+pub type FinalityNotifications<FN> = TracingUnboundedReceiver<FN>;
+
+pub trait BlockchainEvents<UH: UnverifiedHeader> {
+    type ImportNotification: BlockImportNotification<UH>;
+    type FinalityNotification: FinalityNotification<UH>;
+
+    fn import_notification_stream(&self) -> ImportNotifications<Self::ImportNotification>;
+    fn every_import_notification_stream(&self) -> ImportNotifications<Self::ImportNotification>;
+    fn finality_notification_stream(&self) -> FinalityNotifications<Self::FinalityNotification>;
 }
--- a/finality-aleph/src/block/mod.rs
+++ b/finality-aleph/src/block/mod.rs
@@ -250,3 +250,17 @@ pub trait HeaderBackend<H: Header>: Send + Sync {
     /// Get current status of the header backend.
     fn status(&self) -> Self::Status;
 }
+type SelectChainError = sp_consensus::Error;
+
+#[async_trait::async_trait]
+pub trait SelectChain<H: Header>: Sync + Send + Clone {
+    async fn leaves(&self) -> Result<Vec<BlockHash>, SelectChainError>;
+
+    async fn best_chain(&self) -> Result<H, SelectChainError>;
+
+    async fn finality_target(
+        &self,
+        base_hash: BlockHash,
+        _maybe_max_number: Option<BlockNumber>,
+    ) -> Result<BlockHash, SelectChainError>;
+}
--- a/finality-aleph/src/block/mod.rs
+++ b/finality-aleph/src/block/mod.rs
@@ -1,7 +1,6 @@
 use std::fmt::{Debug, Display, Error as FmtError, Formatter};
 
 use parity_scale_codec::{Codec, Decode, Encode};
-use sc_utils::mpsc::TracingUnboundedReceiver;
 
 use crate::{BlockHash, BlockNumber};
 
@@ -161,12 +160,18 @@ pub enum ChainStatusNotification<H: Header> {
 /// A stream of notifications about the chain status in the database changing.
 /// We assume that this will return all the events, otherwise we will end up with a broken state.
 #[async_trait::async_trait]
-pub trait ChainStatusNotifier<H: Header> {
+pub trait ChainStatusNotifier<H: Header>: Send {
     type Error: Debug + Display;
 
     /// Returns a chain status notification when it is available.
     /// This method's implementation must be cancellation safe.
     async fn next(&mut self) -> Result<ChainStatusNotification<H>, Self::Error>;
+}
+
+pub trait BlockchainEvents<H: Header>: Send {
+    type ChainStatusNotifier: ChainStatusNotifier<H>;
+
+    fn chain_status_notifier(&self) -> Self::ChainStatusNotifier;
 }
 
 /// The status of a block in the database.
@@ -242,27 +247,4 @@ pub trait HeaderBackend<H: Header>: Send + Sync {
 
     fn hash(&self, number: BlockNumber) -> Result<Option<BlockHash>, Self::Error>;
     fn info(&self) -> Self::Info;
-}
-
-pub trait BlockImportNotification<UH: UnverifiedHeader>: Debug + Send {
-    fn hash(&self) -> BlockHash;
-    fn header(&self) -> &UH;
-    fn is_new_best(&self) -> bool;
-}
-
-pub trait FinalityNotification<UH: UnverifiedHeader>: Debug + Send {
-    fn hash(&self) -> BlockHash;
-    fn header(&self) -> &UH;
-}
-
-pub type ImportNotifications<IN> = TracingUnboundedReceiver<IN>;
-pub type FinalityNotifications<FN> = TracingUnboundedReceiver<FN>;
-
-pub trait BlockchainEvents<UH: UnverifiedHeader> {
-    type ImportNotification: BlockImportNotification<UH>;
-    type FinalityNotification: FinalityNotification<UH>;
-
-    fn import_notification_stream(&self) -> ImportNotifications<Self::ImportNotification>;
-    fn every_import_notification_stream(&self) -> ImportNotifications<Self::ImportNotification>;
-    fn finality_notification_stream(&self) -> FinalityNotifications<Self::FinalityNotification>;
 }
--- a/finality-aleph/src/data_io/chain_info.rs
+++ b/finality-aleph/src/data_io/chain_info.rs
@@ -5,7 +5,7 @@ use lru::LruCache;
 
 use crate::{
     aleph_primitives::{BlockHash, BlockNumber},
-    block::{Block, Header, HeaderBackend, Info},
+    block::{Block, Header, HeaderBackend, HeaderBackendStatus},
     data_io::ChainInfoCacheConfig,
     BlockId,
 };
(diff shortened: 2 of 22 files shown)


finality-aleph/src/party/manager/mod.rs
```diff
@@ -350,11 +353,14 @@ where
...
 
         let last_block_of_previous_session = session_boundaries.first_block().saturating_sub(1);
-        let last_block_of_previous_session_hash = self
-            .client
-            .block_hash(last_block_of_previous_session)
-            .expect("Previous session ended, the block should be present")
-            .expect("Previous session ended, we should have the hash.");
+        let last_block_of_previous_session_header = self
+            .header_backend
+            .header_of_finalized_at(last_block_of_previous_session)
```
ggawryal, line 358: Double-checking - last block of previous session should be already finalized, right?

kostekIV: yep

kostekIV (commented):
Looks nice, gj. Just some nits

finality-aleph/src/party/manager/mod.rs
```diff
@@ -72,13 +70,11 @@ type CurrentNetworkType = SimpleNetwork<
...
-struct SubtasksParams<C, B, N, BE, JS>
+struct SubtasksParams<H, HB, N, JS>
 where
-    B: BlockT<Hash = BlockHash>,
-    B::Header: HeaderT<Number = BlockNumber> + UnverifiedHeader,
-    C: crate::ClientForAleph<B, BE> + Send + Sync + 'static,
-    BE: Backend<B> + 'static,
-    N: Network<VersionedNetworkData<B::Header>> + 'static,
+    H: Header,
+    HB: crate::block::HeaderBackend<H> + Send + Sync + 'static,
```
timorleph, line 76: Could you maybe just import it rather than fully qualifying here?

finality-aleph/src/party/manager/mod.rs
timorleph: Yup!

finality-aleph/src/party/manager/mod.rs
```diff
@@ -350,11 +351,14 @@ where
...
-            .expect("Previous session ended, the block should be present")
-            .expect("Previous session ended, we should have the hash.");
+        let last_block_of_previous_session_header = self
+            .header_backend
+            .header_of_finalized_at(last_block_of_previous_session)
+            .expect("Previous session ended, the block should be present.")
+            .expect("Previous session ended, the block should be finalized.");
+
+        let last_block_of_previous_session_hash =
+            UnverifiedHeader::id(&last_block_of_previous_session_header).hash();
```
timorleph, line 361: This is necessary because of the multiple traits that the header satisfies, right? If so, please add a comment here about replacing this with a normal dot call after the other traits are no longer present.

finality-aleph/src/testing/data_store.rs
```diff
@@ -204,7 +204,7 @@ async fn forks_have_different_block_hashes() {
     let genesis_hash = test_handler.genesis_hash();
     let a1 = test_handler.build_block_above(&genesis_hash).await;
     let b1 = test_handler.build_block_above(&genesis_hash).await;
-    assert_ne!(a1.hash(), b1.hash());
+    assert_ne!(a1.header().id().hash(), b1.header().id().hash());
```
timorleph, line 207: I think you can omit the `hash()` here, since just checking the non-equality of ids should be sufficient.

finality-aleph/src/testing/data_store.rs
```diff
@@ -294,7 +294,12 @@ async fn branch_not_within_session_boundaries_does_not_go_through() {
         .assert_no_message_out("Data Store let through a message not within session_boundaries")
         .await;
 
-    test_handler.finalize_block(&blocks[session_end + MAX_DATA_BRANCH_LEN].hash());
+    test_handler.finalize_block(
+        &blocks[session_end + MAX_DATA_BRANCH_LEN]
+            .header()
+            .id()
+            .hash(),
```
timorleph, line 301: Kinda related to the above comment – perhaps the interface here should be accepting ids rather than hashes, to be more conformant with our id-first approach in the main code. But this might be unnecessary work, so ehh~~, I encourage you to silently resolve this comment and pretend you didn't see it, unless you are very much in the mood for changing this.

timorleph (approved):
A couple, mostly small or even optional, comments, but in general _lovely_ PR. The places where `B` changes to `H` and it becomes clearer what the actual dependencies are especially warm my heart.

finality-aleph/src/block/mod.rs
```diff
@@ -226,3 +232,31 @@ where
...
+    type Error: Debug;
+    /// Get block header. Returns `None` if block is not found.
+    fn header(&self, id: BlockId) -> Result<Option<H>, Self::Error>;
+    /// Get hash of a finalized block with a given number. Returns Ok(None) if block exists in
+    /// the database, but is not finalized yet.
+    fn header_of_finalized_at(&self, number: BlockNumber) -> Result<Option<H>, Self::Error>;
+    /// Get currently highest finalized block.
+    fn top_finalized_id(&self) -> BlockId;
+    /// Get full id of a block with a given hash
+    fn hash_to_id(&self, hash: BlockHash) -> Result<Option<BlockId>, Self::Error>;
```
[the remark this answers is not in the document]
ggawryal: `hash_to_header: Hash -> Result<Option<Header>>` would be in a weird conflict with `header: BlockID -> Result<Option<Header>>` and I wanted to discourage users from using the former. Hoped to skip this method completely - it is used exactly once, in the aggregator.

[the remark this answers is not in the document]
timorleph: It wouldn't be _that_ much of a problem, since in most places you have easier access to a `BlockId` than `Hash`, but feel free to resolve this comment without changing that – it's kinda vibes-based anyway.

finality-aleph/src/block/mod.rs
```diff
@@ -226,3 +232,31 @@ where
     /// Children of the specified block.
     fn children(&self, id: BlockId) -> Result<Vec<J::Header>, Self::Error>;
 }
+
+pub trait HeaderBackend<H: Header>: Send + Sync {
+    type Error: Debug;
+    /// Get block header. Returns `None` if block is not found.
+    fn header(&self, id: BlockId) -> Result<Option<H>, Self::Error>;
```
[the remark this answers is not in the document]
ggawryal: Good idea (although functions in `ChainStatus` take plain `BlockId`, which is a little inconsistent)

finality-aleph/src/block/mod.rs
```diff
@@ -160,14 +160,20 @@ pub enum ChainStatusNotification<H: Header> {
...
-pub trait ChainStatusNotifier<H: Header> {
+pub trait ChainStatusNotifier<H: Header>: Send {
     type Error: Debug + Display;
 
     /// Returns a chain status notification when it is available.
     /// This method's implementation must be cancellation safe.
     async fn next(&mut self) -> Result<ChainStatusNotification<H>, Self::Error>;
 }
 
+pub trait BlockchainEvents<H: Header>: Send {
```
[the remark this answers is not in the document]
ggawryal: It's required, because we have to `Clone` `ChainStatusNotifier`s, each time when starting a new session in `NodeSessionManagerImpl`. It seems that it should be possible to avoid cloning there, but because of the `Runnable` abstraction it is not straightforward.

finality-aleph/src/testing/data_store.rs
ggawryal: Removed

finality-aleph/src/party/manager/mod.rs
ggawryal: Hmm, I could've written type equations differently, adding `H: Header` parameter and separating it from `HeaderT` (we'd need that parameter anyway in the future). Solved and now even the comment is not necessary

finality-aleph/src/party/manager/mod.rs
ggawryal: Done

finality-aleph/src/block/mod.rs
```diff
@@ -226,3 +232,31 @@ where
...
+    /// Get currently highest finalized block.
+    fn top_finalized_id(&self) -> BlockId;
+    /// Get full id of a block with a given hash
+    fn hash_to_id(&self, hash: BlockHash) -> Result<Option<BlockId>, Self::Error>;
+}
+
+type SelectChainError = sp_consensus::Error;
+
+#[async_trait::async_trait]
+pub trait SelectChain<H: Header>: Sync + Send + Clone {
```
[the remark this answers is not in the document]
ggawryal: Good finding, renamed and removed unused methods

finality-aleph/src/data_io/chain_info.rs
```diff
@@ -21,82 +19,65 @@ pub trait ChainInfoProvider: Send + Sync + 'static {
...
-        };
-
-        if let Some(header) = self.client.header(block_hash).expect("client must respond") {
-            Ok((header.hash(), num).into())
-        } else {
-            Err(())
+    fn get_finalized_at(&mut self, number: BlockNumber) -> Result<BlockId, ()> {
+        match self.client.header_of_finalized_at(number) {
+            Ok(Some(header)) => Ok(header.id()),
+            _ => Err(()),
```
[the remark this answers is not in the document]
ggawryal: Agree, but I have not easy idea how to achieve that - the problem is `CachedChainInfoProvider`, which caches calls for getting parent, and this is not a method in `HeaderBackend`. Probably removing `ChainInfoProvider` trait and separating cached and aux to a separate trait would be good enough, but I'm not sure. Created a task for that under the same epic:

finality-aleph/src/block/mod.rs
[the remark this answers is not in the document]
ggawryal: Done

finality-aleph/src/block/mod.rs
```diff
@@ -226,3 +232,31 @@ where
...
+    async fn leaves(&self) -> Result<Vec<BlockHash>, SelectChainError>;
+
+    async fn best_chain(&self) -> Result<H, SelectChainError>;
+
+    async fn finality_target(
+        &self,
+        base_hash: BlockHash,
+        _maybe_max_number: Option<BlockNumber>,
```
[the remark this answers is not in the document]
ggawryal: Removed that method

finality-aleph/src/block/mod.rs
```diff
@@ -226,3 +232,31 @@ where
...
+    fn header_of_finalized_at(&self, number: BlockNumber) -> Result<Option<H>, Self::Error>;
+    /// Get currently highest finalized block.
+    fn top_finalized_id(&self) -> BlockId;
+    /// Get full id of a block with a given hash
+    fn hash_to_id(&self, hash: BlockHash) -> Result<Option<BlockId>, Self::Error>;
+}
+
+type SelectChainError = sp_consensus::Error;
+
+#[async_trait::async_trait]
```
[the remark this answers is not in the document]
ggawryal: Done

finality-aleph/src/block/substrate/chain_status.rs
```diff
@@ -242,3 +243,42 @@ impl ChainStatus<Block, Justification> for SubstrateChainStatus {
             .collect())
     }
 }
+
+impl HeaderBackend<AlephHeader> for SubstrateChainStatus {
+    type Error = Error;
+
+    fn header(&self, id: BlockId) -> Result<Option<AlephHeader>, Self::Error> {
+        self.header(&id)
```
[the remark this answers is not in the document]
ggawryal: Changed to explicit call syntax

finality-aleph/src/data_io/chain_info.rs
```diff
@@ -21,82 +19,65 @@ pub trait ChainInfoProvider: Send + Sync + 'static {
...
     fn is_block_imported(&mut self, block: &BlockId) -> bool {
         let maybe_header = self
             .client
-            .header(block.hash())
+            .header(block.clone())
             .expect("client must answer a query");
         if let Some(header) = maybe_header {
             // If the block number is incorrect, we treat as not imported.
-            return *header.number() == block.number();
+            return header.id().number() == block.number();
```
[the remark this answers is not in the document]
ggawryal: Great catch, fixed. Also changed `get_parent_hash` for consistency, to not panic when error is found. Regarding returning errors, let's move discussion to thread your comment about it below

finality-aleph/src/testing/data_store.rs
ggawryal: [I tried :P] After a consideration, I think this'd be better done after settling block production and finalization abstractions. The whole `ClientChainBuilder` will have to be refactored, and this `TestHandler` too, as a consequence.

finality-aleph/src/block/mod.rs
```diff
@@ -226,3 +235,24 @@ where
...
+    /// Get hash of a finalized block with a given number. Returns Ok(None) if block exists in
+    /// the database, but is not finalized yet.
+    fn header_of_finalized_at(&self, number: BlockNumber) -> Result<Option<H>, Self::Error>;
+    /// Get currently highest finalized block.
+    fn top_finalized_id(&self) -> BlockId;
+    /// Get full id of a block with a given hash
+    fn hash_to_id(&self, hash: BlockHash) -> Result<Option<BlockId>, Self::Error>;
+}
+
+/// A strategy for selecting a leaf that should be considered as the chain tip.
```
timorleph, line 252: Calling it "the chain tip" introduces another name for ~the same concept. Either explain what this is about in detail ("for purposes of deciding which block to propose for finalization" or something like that) or use one of the names we already have ("best block" or "favourite block").

finality-aleph/src/data_io/chain_info.rs
```diff
@@ -21,82 +19,68 @@ pub trait ChainInfoProvider: Send + Sync + 'static {
...
-            .client
-            .header(block.hash())
-            .expect("client must answer a query");
-        if let Some(header) = maybe_header {
-            // If the block number is incorrect, we treat as not imported.
-            return *header.number() == block.number();
+        match self.client.header(block) {
+            Ok(maybe_header) => maybe_header.is_some(),
+            Err(e) => {
+                log::debug!("Error while fetching header in ChainInfoProvider: {:?}", e);
```
timorleph, line 52: Missing `target: `.

finality-aleph/src/data_io/chain_info.rs
```diff
@@ -21,82 +19,68 @@ pub trait ChainInfoProvider: Send + Sync + 'static {
...
-            .header(block.hash())
-            .expect("client must respond")
-        {
-            Ok(*header.parent_hash())
-        } else {
-            Err(())
+        match self.client.header(block) {
+            Ok(Some(header)) => Ok(header.parent_id().ok_or(())?.hash()),
+            Ok(None) => {
+                log::info!("Block not found while getting parent hash in ChainInfoProvider");
```
timorleph, line 69: Also missing `target: ` here and below.

timorleph: Oh, and import these rather than fully qualifying them plx.

timorleph (approved):
Some tiny nits left.

finality-aleph/src/data_io/chain_info.rs
ggawryal: Target `aleph-data-store` is the best match I guess

finality-aleph/src/block/mod.rs
ggawryal: Kinda agree, I was looking for more general term (`best` is somewhat associated with substrate's longest chain, and `favourite` with aleph's), but maybe let's not add 3rd name.

finality-aleph/src/data_io/legacy/data_store.rs
```diff
@@ -261,14 +257,14 @@ where
         loop {
             self.prune_pending_messages();
             self.prune_triggers();
-
             tokio::select! {
                 maybe_message = self.messages_from_network.next() => {
                     let message = maybe_message.ok_or(Error::NetworkMessagesTerminated)?;
-                    trace!(target: "aleph-data-store", "Received message at Data Store {:?}", message);
+                    trace ! (target: "aleph-data-store", "Received message at Data Store {:?}", message);
```
timorleph, line 263: ` ! `?

ggawryal: Didn't even know this is parsed correctly :facepalm: Extracted log target as a bonus.

finality-aleph/src/block/mod.rs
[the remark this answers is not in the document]
ggawryal: So I'd rather leave the `hash_to_id`. Re `top_finalized_id` I was slightly inclined to change that to return a header, but that would require some error handling - probably we can unwrap them right away, since the only errors that _should_ possibly appear there are some fatal database errors, but adding that unwrap makes me anxious.

kostekIV (approved):

PR #1561, opened 2024-01-04
A0-3835: Halo2 verifier

pmikolajczyk41:
# Description

We add halo2 SNARK verification as a runtime interface. This is the last piece in the whole pipeline that enables proof verification in smart contracts. Now, the flow is: 

```
 Smart Contract -> Chain Extension -> Runtime interface API -> Runtime interface implementation
|--- SC side ---|---------------- runtime ------------------|------------- host----------------|
```

I put runtime interface into a new crate, so that we keep `primitives` as simple as possible. It can serve for any future runtime interfaces as well. It was put outside workspace, similarly to `pallet-vk-storage` and `baby-liminal-chain-extension`, for the similar reason.

Notes:
 - to adapt to the halo2 verification API and avoid changing core code, I'm using mocked, lightweight KZG parameters
 - there is a public constant `CIRCUIT_MAX_K` which describes (lg of) the maximum number of rows in the supported circuits; for now it is ridiculously low (12, meaning that we can have up to ~4096 rows), so that testing is fast (every time we want to generate proving/verifying keys we actually have to compute `2^CIRCUIT_MAX_K` group elements); I will spend some time on it in the next week (static cache for tests, maybe some simplifications)
 - `DeserializingProofFailed`-type error no longer applies to verification process; proof deserialization happens inside `verify_proof` method and will be translated to a more general variant of `VerificationFailed`
 - companion PR: https://github.com/Cardinal-Cryptography/pse-halo2/pull/1

## Type of change

# Checklist:

- I have added tests


runtime-interfaces/src/snark_verifier/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+// `aleph-runtime-interfaces::snark_verifier::snark_verifier::<>`.
+pub use snark_verifier::verify;
+#[cfg(feature = "std")]
+pub use snark_verifier::HostFunctions;
+
+/// Log_2(max number of rows in a supported circuit).
+///
+/// Note: the same constant MUST be used in the params generation for preparing proving and
+/// verifying keys.
+pub const CIRCUIT_MAX_K: u32 = 12;
```
DamianStraszak, line 21: could we maybe add `k` to verifying_key?
And then we don't need this constant here I suppose

DamianStraszak (approved):
Brilliant work!

runtime-interfaces/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+authors = ["Cardinal", "Aleph Zero Foundation"]
+edition = "2021"
+homepage = "https://alephzero.org"
+repository = "https://github.com/Cardinal-Cryptography/aleph-node"
+
+[dependencies]
+codec = { package = "parity-scale-codec", version = "3.0", default-features = false, features = ["derive"] }
+log = { version = "0.4", default-features = false }
+
+halo2_proofs = { git = "https://github.com/Cardinal-Cryptography/pse-halo2", branch = "substrate-compatible", default-features = false, features = ["mock-kzg-params"], optional = true }
```
Yithis, line 13: As discussed I believe it may be beneficial to divide this dependency into two: curves and abstractions+verifier. But it may be added in separate pr

Yithis (approved):
Great job ^^

runtime-interfaces/src/snark_verifier/mod.rs
pmikolajczyk41: discussed offline: as it requires small changes to pallet's API, it'll be done in the next PR

PR #1092, opened 2023-03-27
Splitsplitsplit

kostekIV:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

## Type of change

Please delete options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

- I have added tests
- I have made neccessary updates to the Infrastructure
- I have made corresponding changes to the existing documentation
- I have created new documentation
- I have bumped `spec_version` and `transaction_version`
- I have bumped aleph-client version if relevant

PR #1562, opened 2024-01-05
A0-3782: Fix order of aleph authorities

woocash2:
# Description

The order of aleph authorities returned from `get_authorities_for_next_session` was not matching the order kept in `NextFinalityCommittee`. Due to conversion of `NextFinalityCommittee` into `BTree` we were losing the order initially created by shuffling the validators list in `pallets::committee_management::impls::shuffle_order_for_session`. Instead, the `next_committee_authorities` inferred its ordering according to `next_authorities` argument passed to `get_authorities_for_next_session` which had a different ordering (most likely matching the order of the **producers** list given in `pallets::committee_management::impls::shuffle_order_for_session`).

For a fix, instead converting `NextFinalityCommittee` to `BTreeSet`, convert `next_authorities` to `BTreeMap` and obtain the authority ids from the map by going in order of `NextFinalityCommittee`.

Bug most likely introduced by the `BTree` optimization [here](https://github.com/Cardinal-Cryptography/aleph-node/pull/1092/files#r1155123930). Previously there was an iteration in order of `NextFinalityCommittee` and extracting the corresponding authority ids by a linear sweep over `next_authorities`.

## Type of change

# Checklist:

- I have added tests **...**
- I have bumped `spec_version` and `transaction_version` **???**
- Checked in polkadot UI that `chain_state::aleph::nextAuthorities` is in the same order as `chain_state::aleph::nextFinalityCommittee`

### DIFF
--- a/pallets/aleph/src/lib.rs
+++ b/pallets/aleph/src/lib.rs
@@ -33,7 +33,7 @@ pub mod pallet {
     };
     use pallet_session::SessionManager;
     use primitives::SessionInfoProvider;
-    use sp_std::collections::btree_set::BTreeSet;
+    use sp_std::collections::btree_map::BTreeMap;
     #[cfg(feature = "std")]
     use sp_std::marker::PhantomData;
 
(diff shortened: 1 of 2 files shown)


pallets/aleph/src/lib.rs
```diff
@@ -132,21 +132,15 @@ pub mod pallet {
         fn get_authorities_for_next_session(
             next_authorities: Vec<(&T::AccountId, T::AuthorityId)>,
         ) -> Vec<T::AuthorityId> {
-            let next_committee_ids: BTreeSet<_> =
-                NextFinalityCommittee::<T>::get().into_iter().collect();
-
-            let next_committee_authorities: Vec<_> = next_authorities
+            let mut account_to_authority: BTreeMap<_, _> = next_authorities.into_iter().collect();
+            let stored_accounts = NextFinalityCommittee::<T>::get();
```
kostekIV, line 136: nit: I would leave the name next_committee_ids here.

kostekIV (approved):

pallets/aleph/src/lib.rs
woocash2: how about `next_committee_accounts`? To emphasize distinction between `AccountId` and `AuthorityId`

DamianStraszak (approved):
Very nice find :muscle:

pallets/aleph/src/lib.rs
kostekIV: :+1:

PR #1568, opened 2024-01-10
A0-3855: Fix charging for vk storage

pmikolajczyk41:
# Description

When a user wants to store bytes in the `vk-storage` pallet, they are charged with two fees:
 - operational one (computed based on benchmarks): `T::WeightInfo::store_key(key_len as u32)`
 - (pallet) spam protection - additional artificial fee that is purely economical: `T::StorageCharge::get().charge_for(key_len)`

Previously, I specified the total fee within `#[pallet::weight(..)]` macro. As a consequence, the latter fee was included into the block space occupation. Since the charges are pretty high, we were exhausting block limit, effectively disabling the extrinsic.

The correct way to do this is to use `PostDispatchResultWithInfo`, where we can specify additional ('acutal') charge for an extrinsic. Such fee will be dynamically adjusted, and won't be statically included in the block weight computation.

## Type of change

### DIFF
--- a/pallets/vk-storage/src/lib.rs
+++ b/pallets/vk-storage/src/lib.rs
@@ -15,6 +15,7 @@ mod tests;
 mod weights;
 
 use frame_support::{
+    dispatch::PostDispatchInfo,
     pallet_prelude::{StorageVersion, Weight},
     sp_runtime::traits::BlakeTwo256,
 };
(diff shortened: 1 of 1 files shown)


Marcin-Radecki (approved):

deuszx (approved):
I'll approve since you're saying "this is the way" but I don't understand ...

We have `T::WeightInfo::store_key` and again `T::StorageCharge` - both weights are using `key.len()` and seem to be within some kind of namespace of _storage_ (first is `store_key` and the other is `StorageCharge`. so how do they differ?

pmikolajczyk41:
`T::WeightInfo::store_key` is a function that takes the key length and returns the actual weight that it takes to store that many bytes in the storage - just a benchmarked value, i.e. operational cost.

`T::StorageCharge` is a 'functional type' that given the key length returns how much a user should additionally pay for storage - this is the economical fee, that should prevent from spamming the pallet

PR #1580, opened 2024-01-22
A0-3751: Make our `SyncOracle` compatible with `SyncingService`

lesniak43:
# Description

Our `SyncOracle` now returns an `AtomicBool` denoting if we're in major sync.

## Type of change

### DIFF
--- a/finality-aleph/src/sync_oracle.rs
+++ b/finality-aleph/src/sync_oracle.rs
@@ -26,7 +26,7 @@ pub struct SyncOracle {
 
 impl SyncOracle {
     pub fn new() -> (Self, Arc<AtomicBool>) {
-        let is_major_syncing = Arc::new(AtomicBool::new(false));
+        let is_major_syncing = Arc::new(AtomicBool::new(true));
         let oracle = SyncOracle {
             last_update: Arc::new(Mutex::new(Instant::now() - OFFLINE_THRESHOLD)),
             last_far_behind: Arc::new(Mutex::new(Instant::now())),
--- a/finality-aleph/src/sync_oracle.rs
+++ b/finality-aleph/src/sync_oracle.rs
@@ -20,6 +20,7 @@ const MAJOR_SYNC_THRESHOLD: Duration = Duration::from_secs(10);
 pub struct SyncOracle {
     last_far_behind: Arc<Mutex<Instant>>,
     last_update: Arc<Mutex<Instant>>,
+    // TODO: remove when SyncingService is no longer needed
     is_major_syncing: Arc<AtomicBool>,
 }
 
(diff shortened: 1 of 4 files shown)


timorleph (approved):
Have you made sure this lets us construct `SyncingService`?

finality-aleph/src/sync_oracle.rs
```diff
@@ -17,14 +20,18 @@ const MAJOR_SYNC_THRESHOLD: Duration = Duration::from_secs(10);
 pub struct SyncOracle {
     last_far_behind: Arc<Mutex<Instant>>,
     last_update: Arc<Mutex<Instant>>,
+    is_major_syncing: Arc<AtomicBool>,
```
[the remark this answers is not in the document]
lesniak43: done

finality-aleph/src/sync_oracle.rs
```diff
@@ -17,14 +20,19 @@ const MAJOR_SYNC_THRESHOLD: Duration = Duration::from_secs(10);
...
-        SyncOracle {
+    pub fn new() -> (Self, Arc<AtomicBool>) {
+        let is_major_syncing = Arc::new(AtomicBool::new(false));
+        let oracle = SyncOracle {
             last_update: Arc::new(Mutex::new(Instant::now() - OFFLINE_THRESHOLD)),
             last_far_behind: Arc::new(Mutex::new(Instant::now())),
-        }
+            is_major_syncing: is_major_syncing.clone(),
+        };
+        (oracle, is_major_syncing)
```
woocash2, line 35: Not confident about it so I'll put it as a Question:
Don't we need to call `self.major_sync()` here as well?

woocash2 (approved):
Accepted, tho left one question

finality-aleph/src/sync_oracle.rs
lesniak43: oh, you mean I should initialize the atomic to `true`? hmm, good question, @timorleph do we really want to start in the "major sync" state? this is how it currently works, since we set `last_far_behind` to `now`, right? I guess it makes sense IRL, but in e2e tests we need to wait 10 seconds in the beginning

timorleph: Hmm, e2e tests shouldn't be a problem, as they should start working as soon as you communicate with neighbours. Setting it to `true` should at most impact transactions then, which... meh, shouldn't be that much of a problem.

lesniak43: @woocash2 I've changed the initial value to `true`

woocash2: Aight, whatever you think is appropriate

PR #1581, opened 2024-01-24
A0-3899: Our base protocol outline

timorleph:
# Description

Adds an outline of the new base protocol, replacement for `SyncingEngine`. Does not have the protocol, but does some of the other things we will need.

I think some of the imports will have to be changed during the next substrate updates, because the definitions are now in different crates, but oh well.

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/finality-aleph/src/base_protocol/service.rs
+++ b/finality-aleph/src/base_protocol/service.rs
@@ -1,0 +1,115 @@
+use std::{
+    fmt::{Display, Formatter, Result as FmtResult},
+    sync::{
+        atomic::{AtomicBool, AtomicUsize},
+        Arc,
+    },
+};
+
+use futures::stream::StreamExt;
+use log::{debug, trace, warn};
+use sc_network_common::sync::SyncEvent;
+use sc_network_sync::{service::chain_sync::ToServiceCommand, SyncingService};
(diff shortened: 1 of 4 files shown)


finality-aleph/src/base_protocol/service.rs
```diff
@@ -0,0 +1,115 @@
...
+impl<B> Service<B>
+where
+    B: Block<Hash = BlockHash>,
+    B::Header: Header<Number = BlockNumber>,
+{
+    /// Create a new service.
+    // TODO(A0-3886): This shouldn't need to return the substrate type after replacing RPCs.
+    // In particular, it shouldn't depend on `B`. This is also the only reason why
+    // the `major_sync` argument is needed.
+    pub fn new(major_sync: Arc<AtomicBool>) -> (Self, SyncingService<B>) {
```
woocash2, line 55: Syncing Engine also returned the `NonDefaultConfig` but I understand that we are going to create it somewhere else (in `bin/service` for instance)

woocash2 (approved):
lookin tasty

finality-aleph/src/base_protocol/service.rs
timorleph: I'm not sure where we will end up creating it, but this is definitely part of the `NotificationEvent` task, since this is what interacts with it – that's what you need to get the notification events.

lesniak43 (approved):

PR #1405, opened 2023-09-18
A0-3150: bump to 1.0.0

kostekIV:
# Description
Bump substrate to 1.0.0 branch.

* also bump of subxt was needed to 0.30
* bump of nightly toolchain used in aux crates

todo: test contracts migrations

## Type of change

Please delete options that are not relevant.

- Breaking change (fix or feature that would cause existing functionality to not work as

PR #1583, opened 2024-01-25
A0-3858: Benchmark on-chain verifier (1/2)

pmikolajczyk41:
# Description

We run benchmarks for the on-chain verifier. 

Even though the actual verification happens in the `aleph-runtime-interfaces` crate, we have to put them in the chain extension library for two reasons:
1. we already have benchmarking setup ready for the CE and it is CE, that is the only consumer of these metrics
1. additionally, we need to measure verification key access

For now, we use only a single circuit. The follow-up PR will extend measurements for other layouts.

There was also a very nasty bug introduced in #1405 which turned off custom host functions from the benchmarking node. It has been fixed in this PR.

## Type of change

# Checklist:


baby-liminal-extension/src/backend/benchmarking.rs
```diff
@@ -31,12 +47,38 @@ pub struct Pallet<T> {
...
 
+/// Get the verification artifacts from the benchmark resources.
+///
+/// Since the benchmarks are run within the runtime, we don't have access to the common `std::fs` utilities.
+/// Fortunately, we can still make use of the `include_bytes` macro.
+fn get_verification_artifacts<T: pallet_vk_storage::Config>() -> (
+    VerifyArgs,
+    BoundedVec<u8, <T as pallet_vk_storage::Config>::MaximumKeyLength>,
+) {
+    // We use a macro here, because a function cannot put literal variables in the `include_bytes` macro.
```
deuszx, line 58: You could also just inline the `include_bytes!("../../benchmark-resources/5_3_proof")` and the same for `input`. 

This looks more impressive though.

pmikolajczyk41: oh yes, but I wanted to avoid specifying this long path 3 times

baby-liminal-extension/src/backend/benchmarking.rs
```diff
@@ -57,12 +99,18 @@ mod benchmarks {
         }
     }
```
deuszx, line 100: Do we also include the `.encode()` part in the benchmark?

pmikolajczyk41: encoding happens in the contract

DamianStraszak (approved):

baby-liminal-extension/build.rs
```diff
@@ -0,0 +1,247 @@
...
+
+        fn synthesize(
+            &self,
+            config: Self::Config,
+            mut layouter: impl Layouter<Fr>,
+        ) -> Result<(), Error> {
+            for instance_idx in 0..INSTANCES {
+                // For every instance, we ensure that the corresponding advice is indeed a square root of it.
+                layouter.assign_region(
+                    || "",
```
fbielejec, line 212: don't know how useful these names are, but you could call it e.g. "negative square root" here

baby-liminal-extension/build.rs
```diff
@@ -0,0 +1,247 @@
...
+        plonk::{Circuit, ConstraintSystem, Error},
+        standard_plonk::{StandardPlonk, StandardPlonkConfig},
+    };
+
+    pub struct BenchCircuit<const INSTANCES: usize, const ROW_BLOWUP: usize> {
+        /// The roots of the instances (`i`th root is a square root of `i`th instance).
+        roots: [Fr; INSTANCES],
+    }
+
+    impl<const INSTANCES: usize, const ROW_BLOWUP: usize> Default
```
fbielejec, line 155: derive macro wouldn't work?

baby-liminal-extension/build.rs
pmikolajczyk41: annotations added

baby-liminal-extension/build.rs
```diff
@@ -0,0 +1,247 @@
...
+        /// We assume here that this is the first row in the region (offset 0).
+        fn neg_root_square(
+            &self,
+            idx: usize,
+            region: &mut Region<Fr>,
+            config: &StandardPlonkConfig<Fr>,
+        ) -> Result<(), Error> {
+            region.assign_advice(|| "", config.a, 0, || Value::known(self.roots[idx]))?;
+            region.assign_advice(|| "", config.b, 0, || Value::known(self.roots[idx]))?;
+            region.assign_fixed(|| "", config.q_ab, 0, || Value::known(-Fr::one()))?;
```
fbielejec: these names seem to be usefull when debugging - call it e.g. -1

baby-liminal-extension/build.rs
```diff
@@ -0,0 +1,247 @@
...
+        /// Assign the `idx`th root to the `a` and `b` advices and `-1` to the `q_ab` fixed value.
+        ///
+        /// We assume here that this is the first row in the region (offset 0).
+        fn neg_root_square(
+            &self,
+            idx: usize,
+            region: &mut Region<Fr>,
+            config: &StandardPlonkConfig<Fr>,
+        ) -> Result<(), Error> {
+            region.assign_advice(|| "", config.a, 0, || Value::known(self.roots[idx]))?;
```
fbielejec:
```suggestion
            region.assign_advice(|| "a", config.a, 0, || Value::known(self.roots[idx]))?;
```

baby-liminal-extension/build.rs
```diff
@@ -0,0 +1,247 @@
...
+            region.assign_advice(|| "", config.b, 0, || Value::known(self.roots[idx]))?;
```
fbielejec:
```suggestion
            region.assign_advice(|| "b", config.b, 0, || Value::known(self.roots[idx]))?;
```

baby-liminal-extension/build.rs
pmikolajczyk41: already covered

baby-liminal-extension/build.rs
pmikolajczyk41: already covered

baby-liminal-extension/build.rs
pmikolajczyk41: already covered

fbielejec (approved):

PR #1591, opened 2024-01-29
A0-3957: Add proxy for foundation nomination program

ggawryal:
# Description

Adding a thin nomination proxy only capable of proxying `Staking::nominate extrinsic`. Other extrinsics (utils, batches etc.) are not needed imho. Particularly, one cannot stop nominating using this proxy.

## Type of change

# Checklist:

- I have added tests


bin/runtime/src/lib.rs
```diff
@@ -1277,6 +1278,17 @@ mod tests {
 
     use super::*;
 
+    #[test]
+    fn test_proxy_is_superset() {
```
kostekIV, line 1282: you are missing couple of asserts.

kostekIV (approved):

bin/runtime/src/lib.rs
```diff
@@ -825,18 +826,18 @@ impl InstanceFilter<RuntimeCall> for ProxyType {
                         | RuntimeCall::NominationPools(..)
                 )
             }
+            ProxyType::Nomination => {
+                matches!(
+                    c,
+                    RuntimeCall::NominationPools(pallet_nomination_pools::Call::nominate { .. })
```
DamianStraszak, line 832: At this point it doesn't seem like we need `NominationPools` here. Also I'm not sure what the `nominate` call does exactly, so I would recommend dropping it.

bin/runtime/src/lib.rs
```diff
@@ -825,18 +826,18 @@ impl InstanceFilter<RuntimeCall> for ProxyType {
...
-        match (self, o) {
-            (ProxyType::Any, _) => true,
-            (_, ProxyType::Any) => false,
-            (ProxyType::NonTransfer, ProxyType::Staking) => true,
-            (ProxyType::Staking, ProxyType::NonTransfer) => false,
-            (ProxyType::Staking, ProxyType::Staking) => true,
-            (ProxyType::NonTransfer, ProxyType::NonTransfer) => true,
-        }
+        // ProxyType::Nomination ⊆ ProxyType::Staking ⊆ ProxyType::NonTransfer ⊆ ProxyType::Any
+        (*self as u64) <= (*o as u64)
```
DamianStraszak, line 840: The reason why this was verbose is to make it safer to modify this code -- i.e., add new proxy types. What we want to avoid is someone adding a new proxy type and then the compiler not telling them `is_superset` requires attention.

I don't know what's the best way to make this right though...

DamianStraszak: Btw. this piece of code is 100% security critical. If we get this wrong, it's really bad :P

DamianStraszak (changes requested):

bin/runtime/src/lib.rs
kostekIV: Oh, right. I forgot about this when reviewing. I would suggest macro to generate arms like `is_superset!(Any < ... < Nomination)` but I dont like macros :P

ggawryal: Added exhaustive matching (and improved readability)

bin/runtime/src/lib.rs
ggawryal: Changed so that in `is_superset` I check all cases manually, and here I use the linear order trick.

DamianStraszak (approved):

PR #1592, opened 2024-01-29
A0-3900: Sync handshake logic

woocash2:
# Description

Add logic for handshake and peer slot constraints verification. The verification part checks the genesis hash contained in the handshake, and checks if we can connect the peer according to the specified constraints. Added logic for peer sets management and counting the peers of a particular role and direction (full inbound, full outbound, light).

The logic of accepting peers depending on their role and available spots has been rewritten to be more readable and should preserve the previous logic of `SyncingEngine`. The peer slot rules are:
* reserved nodes (those who have been specified during construction of `Handler` as `reserved_nodes`) can freely connect and disconnect with no constraints.
* there are separate bounds on the number of **inbound** and **outbound** full nodes.
* there is a bound on the **light** nodes.
* The above bounds **ignore** the reserved nodes.

Removed the `mod test` because we can't construct the `Handler` now without the `FullNetworkConfiguration`, and it's best to wait for the substrate update before using it.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/base_protocol/handler.rs
+++ b/finality-aleph/src/base_protocol/handler.rs
@@ -60,12 +60,11 @@ where
         // It is assumed that `default_peers_set.out_peers` only refers to full nodes, but
         // `default_peers_set.in_peers` refers to both full and light nodes.
         // Moreover, `default_peers_set_num_full` refers to the total of full nodes.
-        let max_full_out_peers =
-            net_config.network_config.default_peers_set.out_peers as usize;
-        let max_full_in_peers =
-            (net_config.network_config.default_peers_set_num_full as usize).saturating_sub(max_full_out_peers);
-        let max_light_peers =
-            (net_config.network_config.default_peers_set.in_peers as usize).saturating_sub(max_full_in_peers);
+        let max_full_out_peers = net_config.network_config.default_peers_set.out_peers as usize;
+        let max_full_in_peers = (net_config.network_config.default_peers_set_num_full as usize)
+            .saturating_sub(max_full_out_peers);
+        let max_light_peers = (net_config.network_config.default_peers_set.in_peers as usize)
+            .saturating_sub(max_full_in_peers);
 
         Handler {
             reserved_nodes,
@@ -103,7 +102,12 @@ where
             .collect()
     }
 
-    fn verify_connection(&self, peer_id: PeerId, handshake: Vec<u8>, is_inbound: bool) -> Result<Roles, ConnectError> {
+    fn verify_connection(
+        &self,
+        peer_id: PeerId,
+        handshake: Vec<u8>,
+        is_inbound: bool,
+    ) -> Result<Roles, ConnectError> {
         let handshake = BlockAnnouncesHandshake::<B>::decode_all(&mut &handshake[..])
             .map_err(|_| ConnectError::BadlyEncodedHandshake)?;
         if handshake.genesis_hash != self.genesis_hash {
@@ -116,10 +120,16 @@ where
 
         if !self.reserved_nodes.contains(&peer_id) {
             // check slot constraints for full non-reserved nodes
-            if is_inbound && handshake.roles.is_full() && self.num_full_in_peers >= self.max_full_in_peers {
+            if is_inbound
+                && handshake.roles.is_full()
+                && self.num_full_in_peers >= self.max_full_in_peers
+            {
                 return Err(ConnectError::TooManyFullInboundPeers);
             }
-            if !is_inbound && handshake.roles.is_full() && self.num_full_out_peers >= self.max_full_out_peers {
+            if !is_inbound
+                && handshake.roles.is_full()
+                && self.num_full_out_peers >= self.max_full_out_peers
+            {
                 return Err(ConnectError::TooManyFullOutboundPeers);
             }
             // check slot constraints for light nodes
@@ -140,13 +150,7 @@ where
         let role = self.verify_connection(peer_id, handshake, is_inbound)?;
 
         // update peer sets
-        self.peers.insert(
-            peer_id,
-            PeerInfo {
-                role,
-                is_inbound,
-            },
-        );
+        self.peers.insert(peer_id, PeerInfo { role, is_inbound });
 
         if !self.reserved_nodes.contains(&peer_id) {
             // update slots for full nodes
@@ -168,8 +172,11 @@ where
     }
 
     pub fn on_peer_disconnect(&mut self, peer_id: PeerId) -> Result<(), DisconnectError> {
-        let info = self.peers.remove(&peer_id).ok_or(DisconnectError::PeerWasNotConnected)?;
-        
+        let info = self
+            .peers
+            .remove(&peer_id)
+            .ok_or(DisconnectError::PeerWasNotConnected)?;
+
         if !self.reserved_nodes.contains(&peer_id) {
             // update slots for full nodes
             if info.is_inbound && info.role.is_full() {
--- a/finality-aleph/src/base_protocol/handler.rs
+++ b/finality-aleph/src/base_protocol/handler.rs
@@ -33,6 +33,7 @@ where
 {
     reserved_nodes: HashSet<PeerId>,
     peers: HashMap<PeerId, PeerInfo>,
+    // the below counters and bounds ignore the nodes which belong to `reserved_nodes`
     num_full_in_peers: usize,
     num_full_out_peers: usize,
     num_light_peers: usize,
--- a/finality-aleph/src/base_protocol/handler.rs
+++ b/finality-aleph/src/base_protocol/handler.rs
@@ -82,7 +82,6 @@ where
 
     /// Returns a list of connected peers with some additional information.
     // TODO(A0-3886): This shouldn't need to return the substrate type after replacing RPCs.
-    // In particular, it shouldn't depend on `B`.
     pub fn peers_info(&self) -> Vec<(PeerId, ExtendedPeerInfo<B>)>
     where
         B: Block<Hash = BlockHash>,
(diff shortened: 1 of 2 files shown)


finality-aleph/src/base_protocol/handler.rs
```diff
@@ -1,32 +1,89 @@
...
+            num_full_in_peers: 0,
+            num_full_out_peers: 0,
+            num_light_peers: 0,
+            genesis_hash,
         }
     }
 
     /// Returns a list of connected peers with some additional information.
     // TODO(A0-3886): This shouldn't need to return the substrate type after replacing RPCs.
     // In particular, it shouldn't depend on `B`.
```
lesniak43, line 85: does this comment now relate to the whole struct? just asking

finality-aleph/src/base_protocol/handler.rs
```diff
@@ -45,18 +102,98 @@ impl Handler {
...
+            .map_err(|_| ConnectError::BadlyEncodedHandshake)?;
+        if handshake.genesis_hash != self.genesis_hash {
+            return Err(ConnectError::BadHandshakeGenesis);
+        }
+
+        if self.peers.contains_key(&peer_id) {
+            return Err(ConnectError::PeerAlreadyConnected);
+        }
+
+        if !self.reserved_nodes.contains(&peer_id) {
```
lesniak43, line 122: too much nesting

finality-aleph/src/base_protocol/handler.rs
```diff
@@ -45,18 +102,98 @@ impl Handler {
...
+                && handshake.roles.is_full()
+                && self.num_full_out_peers >= self.max_full_out_peers
+            {
+                return Err(ConnectError::TooManyFullOutboundPeers);
+            }
+            // check slot constraints for light nodes
+            if handshake.roles.is_light() && self.num_light_peers >= self.max_light_peers {
+                return Err(ConnectError::TooManyLightPeers);
+            }
+        }
```
lesniak43, line 140: I'd suggest rewriting this into a single `match`, sth like ` match (is_inbound, is_full, is_in_saturated, is_out_saturated, is_light_saturated)`, although maybe it's too much conditions - feel free to resolve if you disagree

finality-aleph/src/base_protocol/handler.rs
```diff
@@ -45,18 +102,98 @@ impl Handler {
...
+        peer_id: PeerId,
+        handshake: Vec<u8>,
+        is_inbound: bool,
+    ) -> Result<(), ConnectError> {
+        let role = self.verify_connection(peer_id, handshake, is_inbound)?;
+
+        // update peer sets
+        self.peers.insert(peer_id, PeerInfo { role, is_inbound });
+
+        if !self.reserved_nodes.contains(&peer_id) {
```
lesniak43, line 156: too much nesting

finality-aleph/src/base_protocol/handler.rs
```diff
@@ -45,18 +102,98 @@ impl Handler {
...
-            "there should be no peers initially"
-        );
+    fn verify_connection(
+        &self,
+        peer_id: PeerId,
+        handshake: Vec<u8>,
+        is_inbound: bool,
+    ) -> Result<Roles, ConnectError> {
+        let handshake = BlockAnnouncesHandshake::<B>::decode_all(&mut &handshake[..])
+            .map_err(|_| ConnectError::BadlyEncodedHandshake)?;
```
timorleph, line 113: Having the underlying `codec` error might be informative, so please wrap it.

finality-aleph/src/base_protocol/handler.rs
```diff
@@ -45,18 +102,98 @@ impl Handler {
...
+        Ok(())
+    }
+
+    pub fn on_peer_disconnect(&mut self, peer_id: PeerId) -> Result<(), DisconnectError> {
+        let info = self
+            .peers
+            .remove(&peer_id)
+            .ok_or(DisconnectError::PeerWasNotConnected)?;
+
+        if !self.reserved_nodes.contains(&peer_id) {
```
lesniak43, line 181: too much nesting

finality-aleph/src/base_protocol/handler.rs
```diff
@@ -45,18 +102,98 @@ impl Handler {
...
+            if info.is_inbound && info.role.is_full() {
+                self.num_full_in_peers.saturating_dec();
+            }
+            if !info.is_inbound && info.role.is_full() {
+                self.num_full_out_peers.saturating_dec();
+            }
+            // update slots of light nodes
+            if info.role.is_light() {
+                self.num_light_peers.saturating_dec();
+            }
```
lesniak43, line 192: `match (is_inbound, is_full)`

finality-aleph/src/base_protocol/handler.rs
timorleph: Yeah, just do an early return with `Ok`. This also applies to the methods below.

lesniak43 (approved):
Nice job, left some comments.

finality-aleph/src/base_protocol/handler.rs
```diff
@@ -45,18 +102,98 @@ impl Handler {
...
+            if !is_inbound && role.is_full() {
+                self.num_full_out_peers += 1;
+            }
+            // update slots of light nodes
+            if role.is_light() {
+                self.num_light_peers += 1;
+            }
+        }
+
+        // TODO(A0-3901): disseminate appropriate `SyncEvent`
```
timorleph, line 170: This is IO, so it should rather happen in `Service`. And the appropriate sinks are already there. This means this method should probably return the `PeerId` on success.

Same below.

finality-aleph/src/base_protocol/handler.rs
timorleph: It's actually more annoying than that, because we cannot really remove the `B` due to the handshake type. :/ However the comment here should be changed to at least remove the dependency on `ExtendedPeerInfo`.

If we ever want to remove the `B` completely I think we will have to add handshake versioning, I don't remember whether we have a task for that, but if we do it might be a good idea to link _that_ task by the whole struct.

An alternative would be to decode the handshake _before_ putting it in here – perhaps that's even a reasonable approach? Then, for now, both the methods can be parametrized by `B` or, even better, we can put a different, unparametrized, type into the verification/peer connected method.

timorleph (approved):
Looks good, some typing/parametrization annoyances.

finality-aleph/src/base_protocol/handler.rs
woocash2: Hmm, this would have to cover (even if covering is `_ => {}`) cases which don't occur.
I'd rather stay with the `if` but let me actually convert the 2nd and 3rd branch to `else if` so it's emphasized that only one of them is possible.

finality-aleph/src/base_protocol/handler.rs
```diff
@@ -45,18 +102,98 @@ impl Handler {
...
+            if is_inbound && role.is_full() {
+                self.num_full_in_peers += 1;
+            }
+            if !is_inbound && role.is_full() {
+                self.num_full_out_peers += 1;
+            }
+            // update slots of light nodes
+            if role.is_light() {
+                self.num_light_peers += 1;
+            }
```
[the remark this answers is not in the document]
lesniak43: shouldn't we raise an error, somewhere earlier, if the role is different from full or light?

finality-aleph/src/base_protocol/handler.rs
woocash2: Right. Maybe just return `()` cuz service already has`peer_id` (and it implements `Copy` so no need to `peer_id.clone()`).

finality-aleph/src/base_protocol/handler.rs
woocash2: Removed the lowest line of the comment. We can see in further prs what we can do with removing dependencies.

finality-aleph/src/base_protocol/handler.rs
woocash2: Done

finality-aleph/src/base_protocol/handler.rs
woocash2: done

finality-aleph/src/base_protocol/handler.rs
timorleph: Uh, I actually find the `else if`s confusing – especially here, when they are paired with `return`s, below they are fine.

finality-aleph/src/base_protocol/handler.rs
timorleph: I prefer returning, as then the calls look more natural imo, but this is fine as well.

timorleph (approved):

finality-aleph/src/base_protocol/handler.rs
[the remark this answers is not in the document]
woocash2: Turns out that `is_full() || is_light()` is always true so sorry, this actually isn't an issue. (still prefer `if` tho :stuck_out_tongue:)
```rust
impl Roles {
	pub fn is_full(&self) -> bool {
		self.intersects(Self::FULL | Self::AUTHORITY)
	}
	pub fn is_light(&self) -> bool {
		!self.is_full()
	}
        ...
}
```

finality-aleph/src/base_protocol/handler.rs
woocash2: done

finality-aleph/src/base_protocol/handler.rs
woocash2: `if`

finality-aleph/src/base_protocol/handler.rs
woocash2: done

finality-aleph/src/base_protocol/handler.rs
woocash2: removed the `TODO`

finality-aleph/src/base_protocol/handler.rs
woocash2: ok, reverted to `if ... if ... if ...` in `verify_connection` only

PR #1596, opened 2024-01-29
A0-3970: Configure Call Filter for Call Runtime in Smart Contracts

DamianStraszak:
# Description

This configures the Call Runtime functionality in Pallet Contracts. From now on the contracts will be able to call the Staking and the Nomination Pools pallet as if they were signed origins.

## Type of change

- New feature

### DIFF
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -711,7 +711,7 @@ parameter_types! {
 
 // The filter for the runtime calls that are allowed to be executed by contracts.
 // Currently we allow only staking and nomination pools calls.
-pub struct ContractsCallRuntimeFilter;
+pub enum ContractsCallRuntimeFilter {}
 
 impl Contains<RuntimeCall> for ContractsCallRuntimeFilter {
     fn contains(call: &RuntimeCall) -> bool {
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -1320,6 +1320,173 @@ mod tests {
     }
 
     #[test]
+    // This test is to make sure that we don't break call-runtime.
+    fn test_staking_pallet_index() {
+        // arbitrary call that is easy to construct
+        let c = RuntimeCall::Staking(pallet_staking::Call::bond_extra { max_additional: 0 });
+        // first byte is pallet index
+        assert_eq!(c.encode()[0], 8);
+    }
+
+    #[test]
+    // This test is to make sure that we don't break call-runtime.
+    fn test_nomination_pools_pallet_index() {
+        // arbitrary call that is easy to construct
+        let c = RuntimeCall::NominationPools(pallet_nomination_pools::Call::chill { pool_id: 0 });
+        // first byte is pallet index
+        assert_eq!(c.encode()[0], 19);
+    }
+
+    fn match_staking_call(c: pallet_staking::Call<Runtime>) {
+        match c {
+            pallet_staking::Call::bond { value: _, payee: _ } => {}
+            pallet_staking::Call::bond_extra { max_additional: _ } => {}
+            pallet_staking::Call::unbond { value: _ } => {}
+            pallet_staking::Call::withdraw_unbonded {
+                num_slashing_spans: _,
+            } => {}
+            pallet_staking::Call::validate { prefs: _ } => {}
+            pallet_staking::Call::nominate { targets: _ } => {}
+            pallet_staking::Call::chill {} => {}
+            pallet_staking::Call::set_payee { payee: _ } => {}
+            pallet_staking::Call::set_controller {} => {}
+            pallet_staking::Call::set_validator_count { new: _ } => {}
+            pallet_staking::Call::increase_validator_count { additional: _ } => {}
+            pallet_staking::Call::scale_validator_count { factor: _ } => {}
+            pallet_staking::Call::force_no_eras {} => {}
+            pallet_staking::Call::force_new_era {} => {}
+            pallet_staking::Call::set_invulnerables { invulnerables: _ } => {}
+            pallet_staking::Call::force_unstake {
+                stash: _,
+                num_slashing_spans: _,
+            } => {}
+            pallet_staking::Call::force_new_era_always {} => {}
+            pallet_staking::Call::cancel_deferred_slash {
+                era: _,
+                slash_indices: _,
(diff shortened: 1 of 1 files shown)


bin/runtime/src/lib.rs
```diff
@@ -709,14 +709,27 @@ parameter_types! {
     pub CodeHashLockupDepositPercent: Perbill = Perbill::from_percent(30);
 }
 
+// The filter for the runtime calls that are allowed to be executed by contracts.
+// Currently we allow only staking and nomination pools calls.
+pub struct ContractsCallRuntimeFilter;
```
pmikolajczyk41, line 714: insignificant comment: when declaring a type that is not intended for instantiating, you can use empty enum - this way you can create a new, uninhabited type

pmikolajczyk41 (approved):

bin/runtime/src/lib.rs
deuszx: Is there any performance or memory difference between your suggestion and the struct?

pmikolajczyk41: nope

bin/runtime/src/lib.rs
```diff
@@ -709,14 +709,27 @@ parameter_types! {
...
 impl pallet_contracts::Config for Runtime {
     type Time = Timestamp;
     type Randomness = RandomnessCollectiveFlip;
     type Currency = Balances;
     type RuntimeEvent = RuntimeEvent;
     type RuntimeCall = RuntimeCall;
-    // The safest default is to allow no calls at all. This is unsafe experimental feature with no support in ink!
-    type CallFilter = Nothing;
+
+    type CallFilter = ContractsCallRuntimeFilter;
```
deuszx, line 732: What about the comment? Is it not true anymore? ie the feature now has support and is no longer experimental?

According to the [docs](https://github.com/Cardinal-Cryptography/polkadot-sdk/blob/master/substrate/frame/contracts/src/lib.rs#L269-L273) - it's now our job to never break the compatibility. This means we need some e2e test for that.

deuszx (approved):

bin/runtime/src/lib.rs
DamianStraszak: "Safest is to allow no calls at all" -- this is still true and always will be :P 
"is unsafe experimental" -- this changed, it's stable

Yes, the compatibility thing is right, although it depends mainly on pallet staking and pallet nomination pools devs in substrate. If they make breaking changes we are screwed pretty much. We need to resolve such cases in a case-by-case manner. Having automated tests here is problematic:
- they are hard to write, and they will likely yield false positives
- they can only check API, and not the semantics. If semantics changes we are screwed anyway. 

What we need to do is monitor changes in these two pallets and upon every update need to make some specific checks. We have a similar procedure for releases -- I will raise this topic today on a release meeting, and try to add something to jira.

bin/runtime/src/lib.rs
DamianStraszak: Right, I removed the pesky habitants :)

bin/runtime/src/lib.rs
DamianStraszak: Ok, I will try to add a test that will fail if the API changes. 

That being said, I don't think we should get a false feeling of security by having such a test. We need manual tests anyway.

pmikolajczyk41: `matches!(call, X::Y(_))` will still not be enough - as long as the encoding produces something 'interpretable', we will not notice breaking change :disappointed:

piotrMocz (approved):
LGTM :+1:

bin/runtime/src/lib.rs
DamianStraszak: Ok, added tests. Please say they are perfect xD

deuszx: Beautiful.

bin/runtime/src/lib.rs
```diff
@@ -1306,6 +1319,173 @@ mod tests {
...
+            pallet_nomination_pools::Call::__Ignore(..) => {}
+        }
+    }
+
+    #[test]
+    fn test_call_runtime_api_stability() {
+        // If this thing does not compile it means there are breaking changes in staking or nomination pools pallet. This affects call-runtime.
+        // Please do not fix blindly -- action required, escalate.
+        let _ = {
+            |c: RuntimeCall| match c {
```
pmikolajczyk41, line 1481: very elegant :ok_hand:

bin/runtime/src/lib.rs
```diff
@@ -1306,6 +1319,173 @@ mod tests {
...
+    fn test_nomination_pools_pallet_index() {
+        // arbitrary call that is easy to construct
+        let c = RuntimeCall::NominationPools(pallet_nomination_pools::Call::chill { pool_id: 0 });
+        // first byte is pallet index
+        assert_eq!(c.encode()[0], 19);
+    }
+
+    fn match_staking_call(c: pallet_staking::Call<Runtime>) {
+        match c {
+            pallet_staking::Call::bond { value: _, payee: _ } => {}
```
pmikolajczyk41, line 1342: I would slightly insist on specifying types explicitly... I know that it might require whole litany of imports, but this would give us one more guarantee, imo an important one

DamianStraszak: I have no idea how to do that, like in terms of labor, without spending infinite amount of time on it :P 

Hints?

h4nsu (approved):

bin/runtime/src/lib.rs
pmikolajczyk41: my IDE (I think it's actually rust-analyzer) hints with:
image

unfortunately, I cannot automatically inline it...

DamianStraszak: Ok, I tried. It's just too much manual work. Many of them require some specific imports, obviously each type has many places to import it from, there are aliases and shit like that. Also whenever I make an edit, my VSCode needs ~5sec to show the types again. It's just not worth it :/

pmikolajczyk41: okay, let's give it up now, I will do it in a follow-up PR during some long meeting xD

DamianStraszak: ok, sent on slack :)

PR #1602, opened 2024-02-05
A0-4005: Fix flooder

ggawryal:
# Description

Fixes and revamps flooder. 
Fixes:
* Fixed setting nonce for transactions
*  Added `--pool-limit` flag, so that flooder watches node's transaction pool, and tries to keep its size under the specified limit.

Changes:
* Instead of submitting a fixed number of `--transactions`, flooder will periodically send a given number of transactions at the beginning of each time interval, for a given number of time intervals (constant tps). This is similar to the previous rate-limited submissions in rounds we had, except that flooder will postpone submitting some transactions if their number would exceed the pool limit.
* Also, the main flooding part now runs for a fixed duration, and stops sending any more transactions after that period (so that the actual blockchain's processing capability in tps may be smaller than tps send by flooder).
* Fee estimation - when blocks are more than 25% full, blockchain [increases](https://research.web3.foundation/Polkadot/overview/token-economics#relay-chain-transaction-fees-and-per-block-transaction-limits) the weight fee for the next round. Flooder now tries to compute some estimation of this fee in worst case. This is only estimation, as currently getting block weight and exact weight fee is not possible afaik via `subxt` nor `RPC`s.
* If not `--skip_initialization`, flooder will also return remaining balance to main account. 

Currently, we only submit `transfer_keep_alive`. In the future, we might also want to submit contract calls, or a batch with many transfers. 

## Type of change

# Checklist:


Marcin-Radecki:
> Currently, we only submit transfer_keep_alive. In the future, we might also want to submit contract calls, or a batch with many transfers.

It's fine to send for most of the cases `transfer_allow_death` , as it is what `transfer` was - benefit is we test reap & create account scenarios.

flooder/src/main.rs
```diff
@@ -1,100 +1,258 @@
...
+            if let Err(e) = result {
+                info!("Error submitting transaction: {e:?}");
+                return Err(e);
+            }
+        }
+
+        let duration = Instant::now().saturating_duration_since(start);
+        if duration.as_secs() >= schedule.interval_duration * schedule.intervals {
+            break;
+        }
```
kostekIV, line 112: transactions now are not sent in intervals. Setting interval_duration to 1sec does not mean all planned intervals will last below 1sec. It rather means that first interval that went over 1sec ends the whole flooding.

flooder/src/main.rs
```diff
@@ -1,100 +1,258 @@
...
+        let left_duration =
+            Duration::from_secs((i + 1) * schedule.interval_duration).saturating_sub(duration);
+        sleep(left_duration).await;
+    }
+
+    info!("Flooding time has passed, left {overdue_transactions} overdue txns because of the pool limit.");
```
kostekIV, line 118: THis is not always true

flooder/src/main.rs
```diff
@@ -1,100 +1,258 @@
...
+            conn.sign_with_params(
+                transfer_all(dest.clone(), true),
+                Default::default(),
+                nonces[conn_id],
+            )?
+            .submit(status)
+            .await?;
+        }
+        debug!("Returned balance back to main account");
+    }
```
kostekIV, line 130: this should be moved out of flood fn - this is not related to this function

flooder/src/main.rs
```diff
@@ -1,100 +1,258 @@
...
+    connection: &SignedConnection,
     n: u32,
     node: F,
-    account_balance: Balance,
+    amount: Balance,
     skip: bool,
-) -> Vec<SignedConnection> {
+) -> anyhow::Result<Vec<SignedConnection>> {
+    info!(
+        "Initializing accounts, estimated total fee per account: {}TZERO",
```
kostekIV, line 143: nit: TZERO -> TOKENS or just TZERO -> ''

flooder/src/main.rs
```diff
@@ -1,100 +1,258 @@
...
+        amount as f32 / TOKEN as f32
+    );
     let mut connections = vec![];
     for i in 0..n {
         let seed = i.to_string();
-        let signer = KeyPair::new(raw_keypair_from_string(&("//".to_string() + &seed)));
+        let signer = KeyPair::new(raw_keypair_from_string(
+            &(ACCOUNTS_SEED_PREFIX.to_string() + &seed),
```
kostekIV, line 150: use format

flooder/src/main.rs
```diff
@@ -1,100 +1,258 @@
...
+    ext::sp_core::{sr25519, Pair},
+    tx::TxPayload,
+    utils::{MultiAddress, Static},
+};
+use tokio::time::{sleep, Duration, Instant};
 
 mod config;
 
+/// Accounts used for the experiment (except for the main account) will have seed //1 etc.
+const ACCOUNTS_SEED_PREFIX: &str = "//";
```
kostekIV, line 26: it could be local to function const or even this could be inlined

flooder/src/main.rs
```diff
@@ -1,100 +1,258 @@
...
 
     if skip {
-        return connections;
+        return Ok(connections);
     }
-    for conn in connections.iter() {
+
+    let nonce = connection.account_nonce(connection.account_id()).await?;
+    for (i, conn) in connections.iter().enumerate() {
         connection
```
kostekIV, line 161: renmae connection to something else as I was reading this 10 times thinking connection sends funds to itself.

flooder/src/main.rs
```diff
@@ -1,100 +1,258 @@
...
+            .sign_with_params(
+                transfer_keep_alive(conn.account_id().clone(), amount),
+                Default::default(),
+                nonce + i as Nonce,
+            )?
+            .submit(TxStatus::Submitted)
+            .await?;
     }
 
     connection
```
kostekIV, line 171: What is for this transfer?

kostekIV (commented):
Nice usage of nonces :) Great improvement over the dead flooder

ggawryal:
>It's fine to send for most of the cases transfer_allow_death , as it is what transfer was - benefit is we test reap & create account scenarios.

Probably transfer_allow_death transactions are a little lighter, I don't think it matters much. Reaping account is hard to test there, as the total sum of fees per node is impossible to compute exactly. At the end we're returning the balance to main account with `transfer_all`, so maybe there we can force reaping these accounts. 

> What is the pool limit here? Pool limit of the node or the flooder?

I meant the flooder `--pool-limit` flag. It should be usually set slightly below the node's pool limit, to allow also some external (non-flooder) transactions to pass. Note, that flooder itself doesn't store any transactions - it signs them on the fly, as it's relatively cheap - on my laptop, signing 1000 transactions is ~50ms in one thread.

flooder/src/main.rs
ggawryal: The timer runs from the start, so this checks the total flooding duration. Intervals try to last 1sec, but if previous ones ended too late (can happen mostly because rpc node can't confirm adding that many of them to the pool in that short time), those intervals will "try to catch up". So finally, there could be less intervals than `schedule.intervals` in such situations.

aleph-client/src/connections.rs
```diff
@@ -171,6 +183,19 @@ pub trait SignedConnectionApi: ConnectionApi {
...
+    /// * `tx` - encoded transaction payload
+    /// * `params` - optional tx params e.g. tip
+    /// * `nonce` - tx nonce. If None, current signer's on chain account nonce will be used
+    /// # Returns
+    /// A signed transaction ready to be submitted via this connection.
+    fn sign_with_params<Call: TxPayload + Send + Sync>(
+        &self,
+        tx: Call,
+        params: ParamsBuilder,
+        nonce: Nonce,
```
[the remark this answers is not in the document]
ggawryal: Updated doc. The problem with getting nonce from chain is that it is an async operation, so it's better to get this, if needed, on higher level

aleph-client/src/pallets/author.rs
```diff
@@ -15,4 +19,15 @@ impl<C: AsConnection + Sync> AuthorRpc for C {
         let bytes = self.as_connection().as_client().rpc().rotate_keys().await?;
         SessionKeys::decode(&mut bytes.0.as_slice()).map_err(|e| e.into())
     }
+
+    async fn pending_extrinsics_len(&self) -> anyhow::Result<u64> {
+        Ok(self
```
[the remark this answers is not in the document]
ggawryal: Done

flooder/src/main.rs
ggawryal: Isn't it?

flooder/src/main.rs
ggawryal: Moved, required returning nonces. Alternatively we could submit a tx with current nonce, but higher tip, but I guess some race conditions could occur that way.

flooder/src/main.rs
ggawryal: Removed

flooder/src/main.rs
ggawryal: Done

flooder/src/main.rs
ggawryal: Changed to fn const

flooder/src/main.rs
ggawryal: Renamed to `main_connection`

flooder/src/main.rs
ggawryal: It was a trick (a transfer to itself indeed), so that we can wait on finalization of this transaction instead of awaiting on all the previous ones :P . Removed and added waiting for last transaction finalization in the loop.

flooder/src/main.rs
```diff
@@ -1,100 +1,277 @@
...
+                        start_finalized_number,
+                    ),
+                    start_finalized_hash,
+                );
 
-                    info!("sleeping for {}ms", left_duration.as_millis());
-                    sleep(left_duration).await;
+                async move {
+                    for tx_id in 0..per_task as u32 {
+                        conn.sign_with_params(
```
fixxxedpoint, line 103: I wonder how much time do we spend on generating and signing txs and how much on submitting them.

aleph-client/src/pallets/author.rs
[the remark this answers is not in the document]
kostekIV: is it?

flooder/src/main.rs
kostekIV: Yep, this is confusing, lest fix it. Duration interval should be enforced and then the number of intervals also should be enforced. what about timeout set to duration interval?

flooder/src/main.rs
```diff
@@ -1,100 +1,277 @@
...
+        .unwrap() as u64;
+
+    let mut nonces: Vec<u32> = vec![];
+    for conn in connections {
+        nonces.push(conn.account_nonce(conn.account_id()).await?);
+    }
+
+    let mut overdue_transactions = 0;
+    for i in 0..schedule.intervals {
+        let pending_in_pool = connections[0].pending_extrinsics_len().await?;
```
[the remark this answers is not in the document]
ggawryal: I'd rather stay with only option 2, as unfortunately for 1) it seems, that:
* when submitting a transaction that exceeds tx-pool's pool limit, it will be accepted, but some other transaction (with lower priority, or the oldest one in case of draws) will be dropped instead, with all transactions depending on it, i.e. with the same sender, but higher nonce. This will make deciding which transaction are dropped kida unpredictable, which complicates nonce computation.
* When node's txpool is almost full, it starts behaving strangely when I test it - maybe rpc node doesn't propagate all transactions to other nodes in the network, idk and haven't had time to debug.

flooder/src/main.rs
```diff
@@ -1,100 +1,277 @@
...
     }
 
-    connection
-        .transfer_keep_alive(connection.account_id().clone(), 1, TxStatus::Finalized)
-        .await
-        .unwrap();
+    Ok(connections)
+}
+
+/// Only a rough estimation, for the worst case where blocks are 75% full
```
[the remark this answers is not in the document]
ggawryal: Added

flooder/src/main.rs
```diff
@@ -1,100 +1,277 @@
...
+                        .await?;
+                    }
+                    anyhow::Ok(())
                 }
             })
-        })
-        .collect();
+            .collect::<Vec<_>>();
+
+        for result in join_all(tasks).await {
```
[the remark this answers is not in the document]
ggawryal: I think this is more or less equivalent to `FuturesUnordered`. I modified spawning separate tasks per connection to spawning them in each interval and joining after its end, as this way it's much easier to deal with `pool-limit` and overdue transactions - I don't see any way of doing it not using sth like a barrier to synchronize and atomic `overdue_transactions`. Finally, I went with spawning tasks each interval, as I guess this should be only relatively slightly slower and easier to code.

flooder/src/main.rs
```diff
@@ -1,100 +1,277 @@
...
+                        )?
+                        .submit(status)
+                        .await?;
+                    }
+                    anyhow::Ok(())
                 }
             })
-        })
-        .collect();
+            .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
ggawryal: Indeed, thanks.

flooder/src/main.rs
ggawryal: Kinda fixed, is this good enough?

flooder/src/main.rs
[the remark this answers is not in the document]
ggawryal: Never mind, I was wrong. This was concurrent, but not multithreaded.  Reverted previous style with spawning separate threads (at cost of slightly worse handling of overdue transactions).

flooder/src/main.rs
ggawryal: Actually, I reverted this and refactored to use `time::interval` await with `MissedTickBehavior::Burst`. With this refactor, I see no point in artificially enforcing constant interval lengths, especially since changes caused by spawning tasks made distributing overdue transactions less balanced. Is it ok for you?

ggawryal:
I rolled back spawning tasks per connection so that the code runs in multiple threads. 
There is also a problem when we spam nodes with too many transactions - sometimes nodes receive transactions signed by the same account in incorrect order, causing them to go to future pool and be waiting until pool maintenance to be brought back to ready pool. This causes a discrepancy between flooder's set tps and actual network tps, even though all nodes behave normal-ish in terms of slo.
This can be fixed by having multiple accounts per connection, but I'd rather add this in next PR.

aleph-client/src/pallets/author.rs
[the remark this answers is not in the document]
kostekIV: hmm, ok

flooder/src/main.rs
kostekIV: ok

kostekIV (approved):

fixxxedpoint (approved):
Great work!

PR #1614, opened 2024-02-12
A0-4036: Incorporate on-chain verifier to the node (remove `liminal` feature)

pmikolajczyk41:
# Description

1. include `pallet-vk-storage`, `baby-liminal-extension`, `runtime-interfaces` in the Rust workspace
2. clean-up CI (remove liminal-specific jobs and workflows)
3. remove `liminal` feature from all crates (inluding `aleph-runtime`, `aleph-node`, `aleph-client`, `runtime-interfaces`, scripts)

Direct follow-up: _put on-chain verifier behind runtime-checks using `feature-control` pallet_ - I didn't include that here so that this PR is purely logistic, without additional runtime logic to check.

Note: I removed `clippy` check for the `ink(-std)` feature for `baby-liminal-extension`, because:
 - the crate has `aleph-runtime` in its `dev-dependencies` - it is used for testing `runtime-std` feature (mutually exclusive with `ink-std`)
 - `aleph-runtime` depends on `baby-liminal-extension` with the `runtime` feature
 - and thus `baby-liminal-extension` is eventually built with both `ink-std` and `runtime` features, which leads to compilation failure

## Type of change

# Checklist:


Cargo.toml
```diff
@@ -11,21 +12,20 @@ members = [
     "pallets/feature-control",
     "pallets/committee-management",
     "pallets/support",
+    "pallets/vk-storage",
     "primitives",
     "rate-limiter",
+    "runtime-interfaces",
 ]
 
 exclude = [
```
Marcin-Radecki, line 21: Nice :) So now `exclude` resembles crates that should be moved to the separate repository (in the future).

CC @kostekIV

kostekIV (approved):

.github/workflows/_unit-tests-and-static-checks.yml
```diff
@@ -45,4 +50,16 @@ jobs:
...
+
+      # Run unit tests for the chain extension: they require a different setup.
+
+      - name: Install Nightly Rust Toolchain
+        uses: Cardinal-Cryptography/github-actions/install-rust-toolchain@v6
+        with:
+          channel: nightly-2023-05-22-unknown-linux-gnu
+          targets: wasm32-unknown-unknown
+
+      - name: Run unit tests for chain extension
```
Marcin-Radecki, line 63: Let's add this step as a parallel job - we have a bunch of large runners to use and adding this as a step here increases the whole PR pipeline by 3-4 minutes (https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/7912430437?pr=1614), where's baseline is that whole `.github/workflows/_unit-tests-and-static-checks.yml` runs now around 8.5 minutes https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/7917287724

bin/runtime/src/lib.rs
```diff
@@ -905,37 +899,6 @@ impl pallet_feature_control::Config for Runtime {
...
-        Identity: pallet_identity = 20,
-        CommitteeManagement: pallet_committee_management = 21,
-        Proxy: pallet_proxy = 22,
-        FeatureControl: pallet_feature_control = 23,
-    }
-);
-
-#[cfg(feature = "liminal")]
 construct_runtime!(
     pub struct Runtime {
```
Marcin-Radecki, line 903:
```
VkStorage: pallet_vk_storage = 41,
```
do we leave it as it is? we still have option to give it `24`, ie next consecutive number, since this change is not released anywhere yet. `VkStorage` is now standard pallet of our runtime.

bin/runtime/src/lib.rs
```diff
@@ -356,17 +356,14 @@ impl pallet_aleph::Config for Runtime {
     type NextSessionAuthorityProvider = Session;
 }
 
-#[cfg(feature = "liminal")]
```
Marcin-Radecki, line 359: There are still two leftoverts in this file

image

Marcin-Radecki (approved):

bin/runtime/src/lib.rs
pmikolajczyk41: removed

bin/runtime/src/lib.rs
pmikolajczyk41: decreased to 41, aleph-client regenerated

.github/workflows/_unit-tests-and-static-checks.yml
pmikolajczyk41: done

PR #1615, opened 2024-02-12
A0-4048: substrate 1.3.0

lesniak43:
# Description

Updating Substrate to v1.3.0

Bumping rust versioin.

## Type of change

# Checklist:

- I have bumped aleph-client version


bin/runtime/src/lib.rs
```diff
@@ -1478,6 +1491,7 @@ mod tests {
                 change_rate: _,
             } => {}
             pallet_nomination_pools::Call::claim_commission { pool_id: _ } => {}
+            pallet_nomination_pools::Call::adjust_pool_deposit { .. } => {}
```
DamianStraszak, line 1494: This is probably easy to miss, so in case you missed it -- see the comment in `test_call_runtime_api_stability`. 
Also I will accept any suggestions to make this hard to miss, or make this test better :P

baby-liminal-extension/test_contract/Cargo.toml
kostekIV, line 1: nit: mode changed

finality-aleph/src/metrics/chain_state.rs
```diff
@@ -643,7 +643,7 @@ mod test {
             .extrinsic(AccountKeyring::Alice, AccountKeyring::Bob, 0);
         let xt2 = setup
             .pool
-            .extrinsic(AccountKeyring::Dave, AccountKeyring::Eve, 0);
```
kostekIV, line 646: looks suspicious

kostekIV (commented):
Looks good

baby-liminal-extension/test_contract/Cargo.toml
lesniak43: should it be executable?

kostekIV: nope, so this changed this file from executable to not executable?

lesniak43: yup

bin/runtime/src/lib.rs
kostekIV: Tbh I think this test is not that needed. Any breaking change should be catched up by one of the 2 tools we have to inspect runtime changes and/or should also be described in polkadot-sdk releases, since well they are breaking. We could add checks to the ci if the given commit introduces such changes using one of the tools we have but I think there would be a lot of false positives

kostekIV: If we are worried about, allowing more calls in the proxy accounts that intended we should add action point to the releases to check it before relaesing

DamianStraszak: The point of this test is to catch such breaking changes very early, and not as the last check before the release, because then it's kind of too late. The reason why we want to catch it early is because `call_runtime` on contracts makes this extra sensitive, to the extent that some such changes we will not be able to merge at all. 

I'm not sure what the best approach here should be TBH, but my concern is that if we wait till release checks, then such changes could be already buried under dozens of other commits and it will be harder to react.

bin/runtime/src/lib.rs
```diff
@@ -698,6 +703,12 @@ impl pallet_treasury::Config for Runtime {
     type SpendOrigin = frame_support::traits::NeverEnsureOrigin<u128>;
     type SpendPeriod = SpendPeriod;
     type WeightInfo = pallet_treasury::weights::SubstrateWeight<Runtime>;
+    type AssetKind = ();
+    type Beneficiary = Self::AccountId;
+    type BeneficiaryLookup = IdentityLookup<Self::AccountId>;
+    type Paymaster = PayFromAccount<Balances, TreasuryAccount>;
+    type BalanceConverter = UnityAssetBalanceConversion;
+    type PayoutPeriod = ConstU32<0>;
```
DamianStraszak, line 711: I'm afraid I have no idea what this configuration does. At this point we don't "actively" use pallet_treasury, it's only accumulating some fees etc., but we don't do anything with them. Hopefully the new configuration is not doing anything crazy :P If nobody knows, then I guess this requires creating a jira task to research this, and make sure it does not affect us negatively. But I wouldn't consider this a blocker for merging.

bin/runtime/src/lib.rs
```diff
@@ -236,7 +238,7 @@ parameter_types! {
     pub const MaxLocks: u32 = 50;
     // We have only 2 reasons for holds - CodeUploadDeposit and StorageDeposit.
     pub const MaxHolds: u32 = 2;
-    pub const MaxFreezes: u32 = 0;
+    pub const MaxFreezes: u32 = 1;
```
DamianStraszak, line 241: Maybe we want here 50 too, I think there is no downside in setting it high. Do you have a reference point maybe on how Polkadot runtime has it, or some parachain? I think we will set `MaxHolds` to 50 in one of hotfixes too.

DamianStraszak (approved):
Very nice!

bin/runtime/src/lib.rs
kostekIV: Here - https://github.com/polkadot-fellows/runtimes/blob/main/relay/polkadot/src/lib.rs#L323
They use value 8 here :)

aleph-client/src/contract/convertible_value.rs
```diff
@@ -200,6 +200,86 @@ where
     }
 }
 
+impl TryFrom<ConvertibleValue> for Option<String> {
```
kostekIV, line 203: We should ask around if anyone is using Option<T> where T is not a string and accountid

kostekIV (approved):
Lgtm, moslty just comments that we need to create jira tasks

bin/runtime/src/lib.rs
lesniak43: I'll create a task.

bin/runtime/src/lib.rs
lesniak43: Set to 50.

aleph-client/src/contract/convertible_value.rs
lesniak43: I'll ask, and remove the unnecessary code.

PR #1616, opened 2024-02-13
A0-4055: hide on-chain verifier also behind runtime checks

pmikolajczyk41:
# Description

We are using `feature-control` pallet to ensure that nobody can access on-chain verifier related code, unless this feature is enabled.

E2E tests will be added after this and #1614 are merged, so that we don't have to tackle with `liminal` compilation feature in e2e-test launching setup.

## Type of change

### DIFF
--- a/baby-liminal-extension/src/backend/tests.rs
+++ b/baby-liminal-extension/src/backend/tests.rs
@@ -6,7 +6,7 @@ mod weighting;
 use aleph_runtime::Runtime as AlephRuntime;
 use aleph_runtime_interfaces::snark_verifier::VerifierError::*;
 use frame_support::pallet_prelude::Weight;
-use pallet_contracts::chain_extension::{ChainExtension, RetVal};
+use pallet_contracts::chain_extension::RetVal;
 
 use crate::{
     backend::{
@@ -34,11 +34,6 @@ fn simulate_verify<Exc: BackendExecutor>(expected_ret_val: u32) {
     let expected_charged = <TestWeight as WeightInfo>::verify()
         + TestWeight::verify_read_args(verify_args().len() as u32);
     assert_eq!(charged, expected_charged);
-}
-
-#[test]
-fn extension_is_enabled() {
-    assert!(BabyLiminalChainExtension::<AlephRuntime>::enabled())
 }
 
 #[test]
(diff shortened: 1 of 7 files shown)


Marcin-Radecki (approved):

bin/runtime/src/lib.rs
```diff
@@ -143,11 +143,26 @@ parameter_types! {
     pub const SS58Prefix: u8 = ADDRESSES_ENCODING;
 }
 
+pub enum CallFilter {}
```
kostekIV, line 146: nit:
```suggestion
pub enum CallFilter;
```

baby-liminal-extension/src/backend/tests.rs
```diff
@@ -36,11 +36,6 @@ fn simulate_verify<Exc: BackendExecutor>(expected_ret_val: u32) {
     assert_eq!(charged, expected_charged);
 }
 
-#[test]
-fn extension_is_enabled() {
-    assert!(BabyLiminalChainExtension::<AlephRuntime>::enabled())
-}
-
```
kostekIV, line 43: why not negate the assertion?

bin/runtime/src/lib.rs
```diff
@@ -417,6 +432,8 @@ parameter_types! {
     pub const MaxPointsToBalance: u8 = 10;
 }
 
+#[cfg(feature = "liminal")]
+use pallet_feature_control::Feature;
```
kostekIV, line 436: could we move those uses up? :D

kostekIV (approved):

bin/runtime/src/lib.rs
pmikolajczyk41: this is invalid syntax for enum (would work for struct though)

bin/runtime/src/lib.rs
pmikolajczyk41: oh, I believed that my built-in formatter will do it automatically... anyway moved up

baby-liminal-extension/src/backend/tests.rs
pmikolajczyk41: because soon it will be by default enabled again... but okay, reverted change to show current state

bin/runtime/src/lib.rs
kostekIV: oh, why is this enum >;

pmikolajczyk41: https://github.com/Cardinal-Cryptography/aleph-node/pull/1596#discussion_r1470307989 :P

baby-liminal-extension/src/backend/tests.rs
pmikolajczyk41: oh, sorry, this is why: https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/7902869902/job/21569473967?pr=1616 

I forgot

PR #1618, opened 2024-02-13
A0-3728: script for fixing double providers counter

Marcin-Radecki:
# Description

Script attaches to given WS URL and 
* queries all `System.Account` keys which have `providers` counter set to 2 (from single block state),
* transforms above `AccountInfo` data in a way only `providers` counter is modifies and set to 1
* performs `System.setStorage` with list of key-value tuples (storage_key, storage_value)
* checks that from the original state there were no events emitted that concerns double providers accounts,
* at the end makes sure there are no double providers account anymore.

## Type of change

### DIFF
--- a/scripts/accounts-invariants/pallet-balances-maintenance.py
+++ b/scripts/accounts-invariants/pallet-balances-maintenance.py
@@ -1,4 +1,5 @@
 #!/bin/python3
+import copy
 import datetime
 import json
 import pathlib
@@ -59,11 +60,11 @@ Accounts that do not satisfy those checks are written to accounts-with-failed-in
                         type=int,
                         default=128,
                         help='How many accounts to upgrade in a single transaction. Default: 128')
-    parser.add_argument('--set-storage-in-batch',
+    parser.add_argument('--double-providers-accounts-to-fix',
                         type=int,
-                        default=16384,
-                        help='How many storage value pairs to update in a single System.setStorage transaction.'
-                             ' Default: 16384')
+                        default=10,
+                        help='How many accounts to fix in one script run session.'
+                             ' Default: 10')
     group = parser.add_mutually_exclusive_group()
     group.add_argument('--fix-free-balance',
                        action='store_true',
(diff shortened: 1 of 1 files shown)


scripts/accounts-invariants/pallet-balances-maintenance.py
```diff
@@ -333,6 +356,245 @@ def upgrade_accounts(chain_connection,
...
+                                                    block_hash=set_storage_block_hash).value
+    log.debug(f"Account state in the setStorage block: {set_storage_info_state}")
+    assert parent_account_info_state['providers'] == 2, f"Providers before fix for account {account_id} is not 2!"
+    assert set_storage_info_state['providers'] == 1, f"Providers after fix for account {account_id} is not 1!"
+    changed_state_for_sake_of_assert = copy.deepcopy(parent_account_info_state)
+    changed_state_for_sake_of_assert['providers'] = 1
+    assert changed_state_for_sake_of_assert == set_storage_info_state, \
+        f"Parent account info state is different from set storage state with more than providers counter! " \
+        f"Parent state: {parent_account_info_state} " \
+        f"Set storage state: {set_storage_info_state}"
```
DamianStraszak, line 410: You can also print `set_storage_block_number` so that it's easier to find, in case it fires

scripts/accounts-invariants/pallet-balances-maintenance.py
```diff
@@ -333,6 +356,245 @@ def upgrade_accounts(chain_connection,
...
+        call_params={
+            'call': set_storage_call,
+            'weight': {
+                'proof_size': 0,
+                'ref_time': max_weight,
+            },
+        }
+    )
+
+    extrinsic = chain_connection.create_signed_extrinsic(call=sudo_unchecked_weight_call,
```
DamianStraszak, line 471: can you add a small tip to the extrinsic, like `0.001` TZERO, to make sure it's first in block?

DamianStraszak (approved):

scripts/accounts-invariants/pallet-balances-maintenance.py
Marcin-Radecki: done

scripts/accounts-invariants/pallet-balances-maintenance.py
Marcin-Radecki: done

PR #1619, opened 2024-02-13
A0-3589: Direct storage read functions

woocash2:
# Description

An implementation of direct storage access to support backwards-compatible runtime API.
* Added `read_storage_map` which reads the element of a runtime storage map directly
* Improved error handling
* Added test which checks that `read_storage_map` and `read_storage_value` work assuming appropriate storage key naming convention (map element is under appropriately digested version of `pallet + map_name + key`, storage value similarly at `pallet + storage_value_name`).

## Type of change

## Checklist:
- I have added tests

### DIFF
--- a/finality-aleph/src/runtime_api.rs
+++ b/finality-aleph/src/runtime_api.rs
@@ -5,7 +5,8 @@ use std::{
 };
 
 use aleph_runtime::SessionKeys;
-use parity_scale_codec::{Decode, DecodeAll, Error as DecodeError};
+use frame_support::StorageHasher;
+use parity_scale_codec::{Decode, DecodeAll, Encode, Error as DecodeError};
 use sc_client_api::Backend;
 use sp_application_crypto::key_types::AURA;
 use sp_core::twox_128;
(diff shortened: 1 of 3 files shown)


finality-aleph/src/runtime_api.rs
```diff
@@ -63,36 +64,74 @@ where
...
-            .client
-            .storage(at_block, &sc_client_api::StorageKey(storage_key))
-        {
-            Ok(Some(e)) => e,
-            _ => return Err(ApiError::NoStorage(pallet.to_string(), item.to_string())),
-        };
-
-        D::decode_all(&mut encoded.0.as_ref()).map_err(ApiError::DecodeError)
+    #[allow(dead_code)]
+    fn read_storage_map<H: StorageHasher, D: Decode, E: Encode>(
```
Marcin-Radecki, line 94: Is this method supposed to read key value from only from `StorageMap` which keys are hashed using `Blake2_128Concat`? As there are `StorageMap`s like `Staking.bonded` that use `Twox64Concat`

Marcin-Radecki: Ah no, all good - it's first template parameter!

finality-aleph/src/runtime_api.rs
```diff
@@ -115,10 +154,80 @@ where
...
+    use frame_support::Twox128;
+    use parity_scale_codec::Encode;
+    use sp_runtime::Storage;
+    use substrate_test_client::ClientExt;
+
+    use super::*;
+    use crate::testing::mocks::{TestClientBuilder, TestClientBuilderExt};
+
+    #[tokio::test]
+    async fn test_storage_reads() {
```
Marcin-Radecki, line 181: Maybe we can add a test which uses `Twox64Concat`, as there are `StorageMaps` on our chain that uses those

finality-aleph/src/runtime_api.rs
```diff
@@ -63,36 +64,74 @@ where
...
+    #[allow(dead_code)]
+    fn read_storage_map<H: StorageHasher, D: Decode, E: Encode>(
+        &self,
+        pallet: &str,
+        item: &str,
+        key: E,
+        at_block: BlockHash,
+    ) -> Result<D, ApiError> {
+        let mut storage_key = [twox_128(pallet.as_bytes()), twox_128(item.as_bytes())].concat();
+        let p = key.using_encoded(H::hash);
```
Marcin-Radecki, line 102: Does `using_encoded` take into account that this should be a hash storage keys with concat(twox64(key), key)? From the test I see it's not, so can one use here something like `Blake2_128Concat` as a `H` when calling this function to mimic real-world example? I know this works as it is, but in reall chain `Blake2_128Concat` is used not just `Twox128` - maybe latter is used as well, I did not check all metadata

finality-aleph/src/runtime_api.rs
ggawryal: Also `at_block` and some errors are not tested, maybe it's worth having such tests, as storage reads are quite an important functionality?

ggawryal (commented):
Overall looks good. The task desription in jira also includes "_reading value from double map_", so I assume this drops in next pr?

kostekIV (approved):

finality-aleph/src/runtime_api.rs
woocash2: Good point, I'll add another tests

finality-aleph/src/runtime_api.rs
woocash2: Not sure if I understand the question. Anything that implements `StorageHasher` can be used here as `H` which includes `Blake2_128Concat`. I'll have `Twox128` changed in tests to something which resembles more the real scenario.

finality-aleph/src/runtime_api.rs
```diff
@@ -63,36 +64,74 @@ where
...
+            other => other,
+        }
     }
 }
 
 #[derive(Clone, Debug)]
 pub enum ApiError {
-    NoStorage(String, String),
+    StorageAccessFailure,
+    NoStorage,
```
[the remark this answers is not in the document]
woocash2: Yeah, so in `access_storage` we don't have the `String` to pass to the `NoStorage` error, we only have the byte array which shouldn't be logged in `Display`. Thus I have introduced `NoStorage` only to be propagated from `access_storage` to `read_storage_map` and `read_storage_value`, so that they can convert it into a more specific `NoStorage...` error which can be given the `String` which represents the "unencoded key" to the storage. It's not the most beautiful way, I know, but the other options I considered (returning `Result<Option, ApiError>` from `access_storage` and converting `None` to `NoStorage` on the upper level, or introducing a new error type for `access_storage`) seemed like just too complicated.

woocash2:
@ggawryal no, the double map most likely won't be needed. In `AlephSessionApi` we don't use any storage of this type.

finality-aleph/src/runtime_api.rs
Marcin-Radecki: Okay, that's what I thought for. Thanks

finality-aleph/src/runtime_api.rs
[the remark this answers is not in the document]
ggawryal: Ok, we can leave it then

finality-aleph/src/runtime_api.rs
woocash2: done

finality-aleph/src/runtime_api.rs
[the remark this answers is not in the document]
woocash2: done

finality-aleph/src/runtime_api.rs
```diff
@@ -115,10 +154,80 @@ where
...
+        let map_value2 = runtime_api.read_storage_map::<Twox128, u32, &str>(
+            "Pallet",
+            "Map",
+            "Key2",
+            genesis_hash,
+        );
+        let storage_value =
+            runtime_api.read_storage_value::<u32>("Pallet", "StorageValue", genesis_hash);
+
+        assert!(map_value1.is_ok());
```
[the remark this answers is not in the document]
woocash2: done

Marcin-Radecki (approved):

ggawryal (approved):
LGTM

PR #1621, opened 2024-02-15
A0-4060: Fix spurious fails of data provider tests

ggawryal:
# Description
Sleep with retires when expecting Some data (still theoretically not ideal, but practically it sleeps 15s in the worst case before failing, when practically 5ms should be sufficient) .

## Type of change

Change in tests pipeline

# Checklist:

### DIFF
--- a/finality-aleph/src/data_io/data_provider.rs
+++ b/finality-aleph/src/data_io/data_provider.rs
@@ -353,7 +353,7 @@ mod tests {
     use crate::{
         data_io::{
             data_provider::{ChainTracker, ChainTrackerConfig},
-            DataProvider, MAX_DATA_BRANCH_LEN,
+            AlephData, DataProvider, MAX_DATA_BRANCH_LEN,
         },
         metrics::AllBlockMetrics,
         party::manager::Runnable,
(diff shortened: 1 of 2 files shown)


lesniak43 (approved):

finality-aleph/src/data_io/legacy/data_provider.rs
```diff
@@ -400,11 +400,29 @@ mod tests {
...
 
-    // Sleep enough time so that the internal refreshing in ChainTracker has time to finish.
+    // Even though sleep time should be more enough, this may return None instead of Some very occasionally.
     async fn sleep_enough() {
         sleep(REFRESH_INTERVAL + REFRESH_INTERVAL + REFRESH_INTERVAL).await;
     }
 
+    // Retries `sleep_enough` until data in provider is available.
+    // This theoretically might fail, but practically shouldn't.
+    async fn sleep_until_data_available(data_provider: &mut DataProvider) -> AlephData {
```
woocash2, line 410: I don't like that we have to repeat exactly the same code for the legacy version. There's nothing we can do :fr: with it in this PR I guess, but maybe (the team) should consider at least having some `common` files for legacy and current in the future. (just a comment so resolve it)

woocash2 (approved):

finality-aleph/src/data_io/data_provider.rs
```diff
@@ -407,10 +407,31 @@ mod tests {
     }
 
     // Sleep enough time so that the internal refreshing in ChainTracker has time to finish.
+    // Even though the sleep time should be enough, this may return None instead of Some very occasionally.
     async fn sleep_enough() {
```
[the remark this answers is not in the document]
ggawryal: Removed

finality-aleph/src/data_io/legacy/data_provider.rs
ggawryal: Even better, we'll be removing whole legacy in data_io after release 13 - A0-3461

PR #1623, opened 2024-02-16
A0-2674: Refactor service.rs

kostekIV:
part 2/x of refactoring

### DIFF
--- a/bin/node/src/service.rs
+++ b/bin/node/src/service.rs
@@ -217,7 +217,7 @@ fn get_aleph_runtime_vars(client: &Arc<FullClient>) -> AlephRuntimeVars {
 fn get_validator_address_cache(aleph_config: &AlephCli) -> Option<ValidatorAddressCache> {
     aleph_config
         .no_collection_of_extra_debugging_data()
-        .then(|| ValidatorAddressCache::new())
+        .then(ValidatorAddressCache::new)
 }
 
 fn get_net_config(
--- a/bin/node/src/service.rs
+++ b/bin/node/src/service.rs
@@ -38,6 +38,18 @@ use crate::{
 type FullClient = sc_service::TFullClient<Block, RuntimeApi, AlephExecutor>;
 type FullBackend = sc_service::TFullBackend<Block>;
 type FullSelectChain = sc_consensus::LongestChain<FullBackend, Block>;
+type ServiceComponents = sc_service::PartialComponents<
+    FullClient,
+    FullBackend,
+    FullSelectChain,
+    sc_consensus::DefaultImportQueue<Block>,
+    sc_transaction_pool::FullPool<Block, FullClient>,
+    (
+        ChannelProvider<Justification>,
+        Option<Telemetry>,
+        AllBlockMetrics,
+    ),
+>;
 
 struct LimitNonfinalized(u32);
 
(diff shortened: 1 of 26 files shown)


bin/node/src/service.rs
```diff
@@ -183,30 +186,36 @@ pub fn new_partial(config: &Configuration) -> Result<ServiceComponents, ServiceE
...
+    pub millisecs_per_block: MillisecsPerBlock,
+}
+
+fn get_aleph_runtime_vars(client: &Arc<FullClient>) -> AlephRuntimeVars {
+    let finalized = client.info().finalized_hash;
+
+    let session_period = SessionPeriod(client.runtime_api().session_period(finalized).unwrap());
+
+    let millisecs_per_block =
+        MillisecsPerBlock(client.runtime_api().millisecs_per_block(finalized).unwrap());
```
timorleph, line 200: Here and above plx use `expect` with an explanation as to why (i.e. that we always have these, right?).

bin/node/src/service.rs
```diff
@@ -183,30 +186,36 @@ pub fn new_partial(config: &Configuration) -> Result<ServiceComponents, ServiceE
...
+
+    AlephRuntimeVars {
+        session_period,
+        millisecs_per_block,
+    }
+}
+
+fn get_validator_address_cache(aleph_config: &AlephCli) -> Option<ValidatorAddressCache> {
+    match aleph_config.no_collection_of_extra_debugging_data() {
+        false => Some(ValidatorAddressCache::new()),
```
timorleph, line 210: Maybe you want to use [`then_some`](https://doc.rust-lang.org/std/primitive.bool.html#method.then_some)? If you don't like this feel free to just resolve.

bin/node/src/service.rs
```diff
@@ -375,51 +302,104 @@ pub fn new_authority(
...
+            force_authoring: config.force_authoring,
             backoff_authoring_blocks,
-            keystore: keystore_container.local_keystore(),
+            keystore: service_components.keystore_container.local_keystore(),
             sync_oracle: sync_oracle.clone(),
             justification_sync_link: (),
             block_proposal_slot_portion: SlotProportion::new(2f32 / 3f32),
             max_block_proposal_slot_portion: None,
-            telemetry: telemetry.as_ref().map(|x| x.handle()),
+            telemetry: service_components.other.1.as_ref().map(|x| x.handle()),
```
timorleph, line 312: Stuff like `service_components.other.1` is quite ugly. As far as I can tell there is no real reason to use `PartialComponents` and we could just write our own config/components struct that has more descriptive names for everything we need – at least I think so, there seem to be no uses of this struct from functions that are not ours. Are you planning to do something like this as a continuation of this refactor?

timorleph (approved):
Definitely better than before, left some minor comments.

bin/node/src/service.rs
kostekIV: YOu are 100% right about PartialComponents. I wanted to do it in this part but I forgot to do it. Now, I think its best to have it in the continuation :P

bin/node/src/service.rs
kostekIV: done

bin/node/src/service.rs
kostekIV: done

timorleph: Heh, good idea with using `then` instead of `then_some`, but I think the execution is wrong – wouldn't `ValidatorAddressCache::new` work as an argument, without the `|| ` and `()`?

kostekIV: Yep, you are like a clippy on this one :)

PR #1629, opened 2024-02-23
A0-4099: New pallets API in aleph-client

pmikolajczyk41:
# Description

Recently we added the feature control and vk storage pallet - in order to test them in e2e tests, we need its API covered by aleph-client.

## Type of change

# Checklist:

### DIFF
--- a/aleph-client/Cargo.toml
+++ b/aleph-client/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aleph_client"
-version = "3.9.0"
+version = "3.9.1"
 edition = "2021"
 authors = ["Cardinal"]
 documentation = "https://docs.rs/aleph_client"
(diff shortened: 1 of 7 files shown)


aleph-client/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aleph_client"
-version = "3.9.0"
+version = "3.9.1"
```
Marcin-Radecki, line 3: Nitpick: since we're changing this line, let's bump to `3.10.0`. I know it does not matter since we don't publish it anywhere but for sake being compliant with semver we should bump minor version, as we're adding new methods to API of `aleph-client`.

Marcin-Radecki (approved):

aleph-client/Cargo.toml
pmikolajczyk41: oh, yes, totally true - I'm sorry, a moment earlier I was working with drink, where I'm bumping always patch component (it's still at `0.x.x`) :see_no_evil:

kostekIV (approved):

PR #1637, opened 2024-03-01
A0-4020: Pallet operations. 

Marcin-Radecki:
# Description

An account can have an underflow of a `consumers` counter. 
Account categories that are impacted by this issue depend on a chain runtime,
but specifically for AlephNode runtime are as follows:
* `consumers`  == 0, `reserved`  > 0
* `consumers`  == 1, `balances.Locks` contain an entry with `id`  == `vesting`
* `consumers`  == 2, `balances.Locks` contain an entry with `id`  == `staking`
* `consumers`  == 3, `balances.Locks` contain entries with `id`  == `staking`
   and account id is in `session.nextKeys`

This PR adds `pallet-operations` with a single extrinsic `fix_accounts_consumers_underflow` that checks if the account falls into one of the above categories, and increases its `consumers` counter in such scenario.

## Type of change

## Testing

There are 4 cases to test, and all of them are tested in the unit tests. 3 out of 4 cases were also tested on local chain with cloned state, except 4th scenario (validator).

## Companion PRs

https://github.com/Cardinal-Cryptography/aleph-node/pull/1638

# TODO

* weights adjustment for `fix_accounts_consumers_underflow`

### DIFF
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -919,7 +919,7 @@ construct_runtime!(
         Identity: pallet_identity = 20,
         CommitteeManagement: pallet_committee_management = 21,
         Proxy: pallet_proxy = 22,
-        Operations: pallet_operations = 23,
+        Operations: pallet_operations = 255,
     }
 );
 
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -376,6 +376,13 @@ impl pallet_elections::Config for Runtime {
     type BannedValidators = CommitteeManagement;
 }
 
+impl pallet_operations::Config for Runtime {
+    type RuntimeEvent = RuntimeEvent;
+    type AccountInfoProvider = System;
+    type BalancesProvider = Balances;
+    type NextKeysSessionProvider = Session;
+}
+
 impl pallet_committee_management::Config for Runtime {
     type RuntimeEvent = RuntimeEvent;
     type BanHandler = Elections;
@@ -912,6 +919,7 @@ construct_runtime!(
         Identity: pallet_identity = 20,
         CommitteeManagement: pallet_committee_management = 21,
         Proxy: pallet_proxy = 22,
+        Operations: pallet_operations = 23,
     }
 );
 
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -97,7 +97,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("aleph-node"),
     impl_name: create_runtime_str!("aleph-node"),
     authoring_version: 1,
-    spec_version: 69,
+    spec_version: 70,
     impl_version: 1,
     apis: RUNTIME_API_VERSIONS,
     transaction_version: 18,
--- a/pallets/operations/src/lib.rs
+++ b/pallets/operations/src/lib.rs
@@ -23,6 +23,7 @@ pub use pallet::*;
 #[pallet_doc("../README.md")]
 pub mod pallet {
     use frame_support::pallet_prelude::*;
+    use frame_support::weights::constants::WEIGHT_REF_TIME_PER_MILLIS;
     use frame_system::{ensure_signed, pallet_prelude::OriginFor};
 
     use crate::{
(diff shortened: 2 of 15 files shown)


pallets/operations/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+        ///	`fix_accounts_consumers_underflow` checks if the account falls into one of above
+        /// categories, and increase its `consumers` counter.
+        ///
+        /// - `origin`: Must be `Signed`.
+        /// - `who`: An account to be fixed
+        ///
+        #[pallet::call_index(0)]
+        // TODO: below weight is a maximum of DbReads ext can done, but it was not benchmarked
+        #[pallet::weight(
+        Weight::from_parts(0, 0)
```
DamianStraszak, line 74: For the weight, let's just put here 2% of the whole block weight -- this is more than enough, so it's safe.

AFAIK a whole block is worth 1 AZERO, so this will cost 0.02 AZERO, which is reasonable.

DamianStraszak (approved):

pallets/operations/src/lib.rs
Marcin-Radecki: Done: 

image

DamianStraszak: Just to make sure, is the fee ~0.02 :)?

Marcin-Radecki: Here's an example of batching this call in a single block (25 calls). 50% block capacity, fee 0.2 AZERO.

image

Marcin-Radecki: ie so one transaction will cost 0.008 AZERO, is it okay?

DamianStraszak: Ok, It works -- actually:
- 10^12 weight = 1 AZERO
- 10^12 weight = 1 sec, reference time
- we allow a block to have 0.4 sec of weight measured in ref_time. 

So 2% of block = 0.02*0.4 = 0.008 AZERO.

bin/runtime/src/lib.rs
```diff
@@ -912,6 +919,7 @@ construct_runtime!(
         Identity: pallet_identity = 20,
         CommitteeManagement: pallet_committee_management = 21,
         Proxy: pallet_proxy = 22,
+        Operations: pallet_operations = 23,
```
kostekIV, line 922: can we put some large number here? We will most likely remove this pallet in the future, right?

pallets/operations/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+#[cfg(test)]
+mod tests;
+
+use frame_support::traits::{LockIdentifier, StorageVersion};
+
+const STORAGE_VERSION: StorageVersion = StorageVersion::new(0);
+pub(crate) const LOG_TARGET: &str = "pallet-operations";
+// harcoding as those consts are not public in substrate
+pub(crate) const STAKING_ID: LockIdentifier = *b"staking ";
+pub(crate) const VESTING_ID: LockIdentifier = *b"vesting ";
```
kostekIV, line 18: just a nit nothing important, use either pub or const here (without (crate) so timorl will be happy)

pallets/operations/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+        /// - `origin`: Must be `Signed`.
+        /// - `who`: An account to be fixed
+        ///
+        #[pallet::call_index(0)]
+        #[pallet::weight(
+        Weight::from_parts(WEIGHT_REF_TIME_PER_MILLIS.saturating_mul(8), 0)
+        )]
+        pub fn fix_accounts_consumers_underflow(
+            origin: OriginFor<T>,
+            who: T::AccountId,
```
kostekIV, line 77: was there a reason we did not want a batched version?

kostekIV (approved):
LGTM

bin/runtime/src/lib.rs
DamianStraszak: Whether we remove or not I don't know, but I think putting a large index makes sense.

pallets/operations/src/lib.rs
Marcin-Radecki: No hard reason, I implemented that since in the first place I though it would be easier to adjust the weight for a call that has single accounts as an argument.

DamianStraszak: Yeah, let's keep it simple. This is good now.

pallets/operations/src/lib.rs
Marcin-Radecki: Okay

PR #1643, opened 2024-03-07
A0-4020: fix for fix accounts consumers underflow

Marcin-Radecki:
# Description

`operations.fix_accounts_consumers_underflow` does not take into account controller != stash scenario, in which case `consumers` should not be incremented on stash account, but on the controller account.

## Type of change

### DIFF
--- a/pallets/operations/src/impls.rs
+++ b/pallets/operations/src/impls.rs
@@ -11,7 +11,7 @@ use sp_runtime::DispatchError;
 
 use crate::{
     pallet::{Config, Event, Pallet},
-    traits::{AccountInfoProvider, BalancesProvider, NextKeysSessionProvider},
+    traits::{AccountInfoProvider, BalancesProvider, BondedStashProvider, NextKeysSessionProvider},
     LOG_TARGET, STAKING_ID, VESTING_ID,
 };
 
@@ -51,8 +51,21 @@ impl<T: Config> Pallet<T> {
         let has_staking_lock = Self::has_lock(&locks, STAKING_ID);
         let nominator_has_consumers_underflow = consumers == 2 && has_staking_lock;
         let has_next_session_keys = T::NextKeysSessionProvider::has_next_session_keys(who);
-        let validator_has_consumers_underflow =
-            consumers == 3 && has_staking_lock && has_next_session_keys;
+        let stash_different_than_controller = match T::BondedStashProvider::get_controller(who) {
+            Some(controller) => *who != controller,
+            None => {
+                log::warn!(
+                    target: LOG_TARGET,
+                    "Account {:?} has staking lock, but its Bonded is empty!",
+                    HexDisplay::from(&who.encode())
+                );
+                false
+            }
+        };
+        let validator_has_consumers_underflow = consumers == 3
+            && has_staking_lock
+            && has_next_session_keys
+            && stash_different_than_controller;
         vester_has_consumers_underflow
             || nominator_has_consumers_underflow
             || validator_has_consumers_underflow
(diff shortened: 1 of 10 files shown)


kostekIV (approved):

pallets/operations/src/impls.rs
```diff
@@ -1,67 +1,85 @@
...
+            expected_consumers += 1;
+        }
+        let has_vesting_lock = Self::has_vesting_lock(&who);
+        let has_staking_lock = Self::has_staking_lock(&who);
+        if has_staking_lock || has_vesting_lock {
+            expected_consumers += 1;
+            if has_staking_lock {
+                expected_consumers += 1;
+            }
+        }
```
DamianStraszak, line 32: Why is the above like that, and not:
```
if has_staking_lock { expected_consumers += 1;}
if has_vesting_lock { expected_consumers += 1;}
```

pallets/operations/src/impls.rs
```diff
@@ -1,67 +1,85 @@
...
-    fn no_consumers_some_reserved(who: &T::AccountId, consumers: u32) -> bool {
-        let is_reserved_not_zero = T::BalancesProvider::is_reserved_not_zero(who);
+    fn has_staking_lock(who: &T::AccountId) -> bool {
+        let locks = T::BalancesProvider::locks(who);
+        Self::has_lock(&locks, STAKING_ID)
+    }
 
-        consumers == 0 && is_reserved_not_zero
+    fn has_next_session_keys_and_account_is_controller(who: &T::AccountId) -> bool {
+        let has_next_session_keys = T::NextKeysSessionProvider::has_next_session_keys(who);
```
DamianStraszak, line 69: Why do we need to check if account is controller, instead of only if it has `session_keys`?

pallets/operations/src/impls.rs
Marcin-Radecki: 1. The `consumers` counter logic regarding locks is as [follows](https://github.com/Cardinal-Cryptography/polkadot-sdk/blob/aleph-v1.2.0/substrate/frame/balances/src/lib.rs#L1004-L1023) (in version 13):
* when there are no locks and one is added, `consumers` is increased
* when the last lock is removed, `consumers` is decreased.
Vesting pallet as itself does not increase `consumers` counter, only [sets a vesting lock](https://github.com/Cardinal-Cryptography/polkadot-sdk/blob/aleph-v1.2.0/substrate/frame/vesting/src/lib.rs#L598).
Note: this locking `consumers` logic is removed in `polkadot-sdk` 1.4.0 by this change https://github.com/paritytech/polkadot-sdk/pull/1976

2. In addition to 1., `staking.bond` [creates additional lock, always on stash](https://github.com/paritytech/polkadot-sdk/pull/1976). It also [sets staking lock, always on stash account](https://github.com/Cardinal-Cryptography/polkadot-sdk/blob/aleph-v1.2.0/substrate/frame/staking/src/pallet/impls.rs#L285), but it falls into increment described in 1.

pallets/operations/src/impls.rs
Marcin-Radecki: From version 12 onwards, we would not have to check, yet in version 11, where it was still possible to set controller != stash, `session.set_keys`:
* retrieved **controller** account id via [this line](https://github.com/Cardinal-Cryptography/substrate/blob/aleph-v0.9.38/frame/session/src/lib.rs#L804), using [`ValidatorIdOf` trait](https://github.com/Cardinal-Cryptography/aleph-node/blob/r-11.4/bin/runtime/src/lib.rs#L392), that is a [`pallet_staking::StashOf` trait](https://github.com/Cardinal-Cryptography/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/frame/staking/src/lib.rs#L924), which queries `staking.ledger` storage map. This means that input argument to `session.set_keys` is a controller account, which implies that [`consumers` counter is increased on the controller account](https://github.com/Cardinal-Cryptography/substrate/blob/aleph-v0.9.38/frame/session/src/lib.rs#L810), even though keys are stash always on a stash account.

DamianStraszak (approved):

PR #1648, opened 2024-03-11
A0-4089: E2E tests for disabling on-chain verifier

pmikolajczyk41:
# Description

Finally we add E2E tests ensuring that feature control is working as intended, i.e. we can easily turn on and off on-chain verifier feature.

Additionally, this PR introduces two small fixes:
 - in `aleph-client`, `FeatureControlSudoApi` has now a correct implementation that uses `sudo_unchecked` instead of simply `send_tx` (would work if `runtime` was configured with explicit `sudo` account instead of `root` origin)
 - `ink` dependency doesn't have default features on (this actually was breaking any ink! contract that was depending on our extension)

E2E run: https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/8235057473

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,93 +79,93 @@ tiny-bip39 = { version = "1.0" }
 tokio = { version = "1.36" }
 rand_pcg = { version = "0.3.1", default-features = false }
 
-frame-benchmarking = { git = "https://github.com/Cardinal-Cryptography/polkadot-sdk.git", branch = "aleph-v1.3.0", default-features = false }
-frame-benchmarking-cli = { git = "https://github.com/Cardinal-Cryptography/polkadot-sdk.git", branch = "aleph-v1.3.0", default-features = false }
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -59,7 +59,7 @@ futures-timer = { version = "3.0" }
 hash-db = { version = "0.16", default-features = false }
 hex = { version = "0.4" }
 hex-literal = { version = "0.3" }
-ink = { version = "5.0.0-rc" }
+ink = { version = "5.0.0-rc", default-features = false }
```
deuszx, line 62: There's already 5.0.0-rc3 ([crates](https://crates.io/crates/ink/versions)) but I'd be wary of using RC at all.

pmikolajczyk41: not related here, but +1

I will bump this once release of ink5 is concluded

kostekIV (approved):

Marcin-Radecki (approved):

PR #1653, opened 2024-03-15
A0-4172: Fix for FE e2e nightly

Marcin-Radecki:
# Description

When an ECR image exists, this workflow is a no-op and silently skips e2e test run. This is because GHA skips a job when any parent job is skipped.

Unfortunately, there is no way to test this as the scenario in which bug occurs is when there's already an image on ECR, ie when e.g. running this workflow from `main`. So I've just run workflow from this PR https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/8292896136, and if its okay then we can merge this PR and re-run from main to see if it helped.

## Type of change

### DIFF
--- a/.github/workflows/nightly-fe-e2e-tests.yml
+++ b/.github/workflows/nightly-fe-e2e-tests.yml
@@ -57,7 +57,7 @@ jobs:
     needs:
       - get-full-docker-image-path
       - build-and-push-featurenet-node-image
-    # to prevent this job to be skipped when on of the parent jobs is skipped
+    # to prevent this job to be skipped when one of the parent jobs is skipped
     if: ${{ !cancelled() }}
     name: Create featurenet from ${{ github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-fe-e2e-tests.yml
```diff
@@ -57,7 +57,7 @@ jobs:
     needs:
       - get-full-docker-image-path
       - build-and-push-featurenet-node-image
-    # to prevent this job to be skipped when on of the parent jobs is skipped
+    # to prevent this job to be skipped when one of the parent jobs is skipped
```
timorleph, line 60: "to prevent this job from being skipped when one...", right? same below

timorleph (approved):

.github/workflows/nightly-fe-e2e-tests.yml
Marcin-Radecki: Right, comment was removed afterall

pmikolajczyk41 (approved):

PR #1650, opened 2024-03-12
A0-4159: Substrate v1.4.0

lesniak43:
# Description

Updating Substrate to v1.4.0

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

- I have bumped aleph-client version

PR #1668, opened 2024-03-29
A0-4216: Does not depend on actual runtime API in aleph-node

Marcin-Radecki:
tl;dr 
Goal of this PR is to have only chainspec generation as only `aleph-runtime` dependency in `aleph-node`.

In aleph-node, we depend on aleph-runtimes RuntimeAPI type. This is not required, as per https://github.com/Cardinal-Cryptography/polkadot-sdk/commit/e53d15aa20e77eb35ccb590bb7ef940d3eca7dbb :

> These runtime api implementations are only used to make the compiler
think that we have implemented all required runtime apis. They will notbe called as we switch the executor to WasmExecutor. In the nearfuture we will not require these fake implementations anymore afterSubstrate has shifted away from this compile time requirement.This brings us the advantage that the polkadot-service doesn't need todepend on the runtimes for getting the RuntimeApi   also removes around 1min of build time on my machine ;)

1. To achieve so we need to implement similar idea like https://github.com/Cardinal-Cryptography/polkadot-sdk/blob/aleph-v1.4.0/polkadot/node/service/src/fake_runtime_api.rs into our `aleph-node` binary
2. There’s an important remark on how this fake runtime must be implemented - it does not need to have all the same entries like [`impl_runtime_apis!`](https://github.com/Cardinal-Cryptography/aleph-node/blob/main/bin/runtime/src/lib.rs#L1001-L1259) has - in particular, it does not need an implementation for 
* `pallet_nomination_pools_runtime_api::NominationPoolsApi`
* `pallet_staking_runtime_api::StakingApi`
* `pallet_contracts::ContractsApi`

ie, code compiles without them, even though real runtime has those. Why? Because this fake runtime API is only used only for sake of compilation:
* for explicit calls in aleph-node to runtime API, e.g. [this requires primitives::AlephSessionApi to be implemented for fake_runtime_api](https://github.com/Cardinal-Cryptography/aleph-node/blob/main/bin/node/src/service.rs#L200) 
* [because of trait bounds when creating RPC client in node](https://github.com/Cardinal-Cryptography/aleph-node/blob/main/bin/node/src/rpc.rs#L49-L51)
* [because of RPC modules to include in node](https://github.com/Cardinal-Cryptography/aleph-node/blob/main/bin/node/src/rpc.rs#L69-L83)

3. We use runtime’s SessionKeys, not opaque::SessionKeys when impl sp_session::SessionKeys<Block> for Runtime  - this is because Polkadot and other parachains do the same, and code compiles after the change.
4. Unifications of imports of primitives in aleph-node. 
5. Use fake_runtime in `finality-aleph`'s tests - no more aleph-runtime dependency!
6. Bumped `version` of `primitives` to be 0.14.0+dev - primitives needs to be versioned the same as runtime

### DIFF
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -1070,10 +1070,6 @@ impl_runtime_apis! {
         ) -> pallet_transaction_payment_rpc_runtime_api::RuntimeDispatchInfo<Balance> {
             TransactionPayment::query_info(uxt, len)
         }
-        // TODO check if it's needed, it exists in polkadot-sdk 1.4.0 fake_runtime.rs
-        // fn query_info(_: <Block as BlockT>::Extrinsic, _: u32) -> RuntimeDispatchInfo<Balance> {
-        //     unimplemented!()
-        // }
         fn query_fee_details(
             uxt: <Block as BlockT>::Extrinsic,
             len: u32,
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -44,10 +44,9 @@ use parity_scale_codec::{Decode, Encode, MaxEncodedLen};
 use primitives::{
     staking::MAX_NOMINATORS_REWARDED_PER_VALIDATOR, wrap_methods, Address,
     AlephNodeSessionKeys as SessionKeys, ApiError as AlephApiError, AuraId, AuthorityId as AlephId,
-    Block as AlephBlock, BlockId as AlephBlockId, BlockNumber as AlephBlockNumber,
-    Header as AlephHeader, SessionAuthorityData, SessionCommittee, SessionIndex,
-    SessionInfoProvider, SessionValidatorError, Version as FinalityVersion, ADDRESSES_ENCODING,
-    DEFAULT_BAN_REASON_LENGTH, DEFAULT_MAX_WINNERS, DEFAULT_SESSIONS_PER_ERA,
+    BlockNumber as AlephBlockNumber, Header as AlephHeader, SessionAuthorityData, SessionCommittee,
+    SessionIndex, SessionInfoProvider, SessionValidatorError, Version as FinalityVersion,
+    ADDRESSES_ENCODING, DEFAULT_BAN_REASON_LENGTH, DEFAULT_MAX_WINNERS, DEFAULT_SESSIONS_PER_ERA,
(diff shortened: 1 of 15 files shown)


bin/runtime/src/lib.rs
```diff
@@ -1095,6 +1070,10 @@ impl_runtime_apis! {
         ) -> pallet_transaction_payment_rpc_runtime_api::RuntimeDispatchInfo<Balance> {
             TransactionPayment::query_info(uxt, len)
         }
+        // TODO check if it's needed, it exists in polkadot-sdk 1.4.0 fake_runtime.rs
```
pmikolajczyk41, line 1073: should we have a Jira for that? or let know @lesniak43 who is working on https://github.com/Cardinal-Cryptography/aleph-node/pull/1650?

finality-aleph/Cargo.toml
```diff
@@ -79,7 +79,6 @@ substrate-test-runtime = { workspace = true }
 substrate-test-client = { workspace = true }
 sc-block-builder = { workspace = true }
 sc-basic-authorship = { workspace = true }
-aleph-runtime = { workspace = true }
```
pmikolajczyk41, line 82: :heart:

primitives/src/lib.rs
```diff
@@ -319,6 +321,7 @@ pub struct VersionChange {
     pub session: SessionIndex,
 }
 
+// TODO this should be moved to separate crate pallet-aleph-session-api
```
pmikolajczyk41, line 324: jira?

primitives/src/fake_runtime_api.rs
```diff
@@ -0,0 +1,250 @@
+// Copyright (C) Parity Technologies (UK) Ltd.
```
pmikolajczyk41, line 1: do we need/want it?

bin/runtime/src/lib.rs
Marcin-Radecki: I've created a JIRA

primitives/src/lib.rs
Marcin-Radecki: Created

primitives/src/fake_runtime_api.rs
Marcin-Radecki: Good point, now we don't

woocash2 (approved):
looks clean

PR #1670, opened 2024-04-04
More space

kostekIV:


bin/runtime/src/lib.rs
```diff
@@ -97,7 +97,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("aleph-node"),
     impl_name: create_runtime_str!("aleph-node"),
     authoring_version: 1,
-    spec_version: 71,
```
Marcin-Radecki, line 100: Please use at least 73, as current Mainent has 72

Marcin-Radecki (approved):

bin/runtime/src/lib.rs
Marcin-Radecki: Or you can go straight ahead and use `13_300_000` here.

DamianStraszak (approved):

bin/runtime/src/lib.rs
kostekIV: done

PR #1675, opened 2024-04-09
A0-4208: Hide NativeExecutionDispatch behind feature flags

Marcin-Radecki:
# Description

Hide implementation of `NativeExecutionDispatch` for the `ExecutorDispatch` behind `local-debugging`. `NativeExecutionDispatch` is needed for debugging locally, as well as for `runtime-benchmarks`. Pehaps it's not needed for `try-runtime`, but it's subject to the next PRs.

## Type of change

# Checklist:

https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/8628234269
https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/8615033603

### DIFF
--- a/bin/node/src/executor.rs
+++ b/bin/node/src/executor.rs
@@ -9,7 +9,7 @@ use sc_service::Configuration;
     feature = "local-debugging",
     feature = "try-runtime"
 )))]
-pub mod executor {
+pub mod aleph_executor {
     use super::Configuration;
     use sc_executor::WasmExecutor;
 
@@ -29,7 +29,7 @@ pub mod executor {
     feature = "local-debugging",
     feature = "try-runtime"
 ))]
-pub mod executor {
+pub mod aleph_executor {
     use super::Configuration;
     use sc_executor::NativeElseWasmExecutor;
 
--- a/bin/node/src/executor.rs
+++ b/bin/node/src/executor.rs
@@ -10,8 +10,9 @@ use sc_service::Configuration;
     feature = "try-runtime"
 )))]
 pub mod aleph_executor {
-    use super::Configuration;
     use sc_executor::WasmExecutor;
+
+    use super::Configuration;
 
     type ExtendHostFunctions = (
         sp_io::SubstrateHostFunctions,
@@ -30,8 +31,9 @@ pub mod aleph_executor {
     feature = "try-runtime"
 ))]
 pub mod aleph_executor {
-    use super::Configuration;
     use sc_executor::NativeElseWasmExecutor;
+
+    use super::Configuration;
 
     pub struct ExecutorDispatch;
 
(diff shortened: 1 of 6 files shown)


bin/node/src/executor.rs
```diff
@@ -1,27 +1,64 @@
...
+    use sc_executor::NativeElseWasmExecutor;
+
+    use super::Configuration;
 
-    fn dispatch(method: &str, data: &[u8]) -> Option<Vec<u8>> {
-        aleph_runtime::api::dispatch(method, data)
+    pub struct ExecutorDispatch;
+
+    impl sc_executor::NativeExecutionDispatch for ExecutorDispatch {
+        #[cfg(feature = "runtime-benchmarks")]
```
ggawryal, line 41: Just to make sure - we don't need  `sp_io::SubstrateHostFunctions` for `NativeElseWasmExecutor`, right?

ggawryal (approved):
LGTM

Marcin-Radecki:
> Cool! The only thing that slightly worries me is that now many tests run by default using different code than the actual node (i.e. native), but there are still some that use wasm (e.g. the runtime update tests), so this should mostly be fine, right?

Actually, `NativeExecutionDispatch` does not influence whether e2e or dev/FE chains run with native or wasm execution - they all run with wasm execution anyway. What `NativeExecutionDispatch` influences though is local debugging - [substrate executor might take request to use native into account](https://github.com/Cardinal-Cryptography/polkadot-sdk/blob/aleph-v1.3.0/substrate/client/executor/src/executor.rs#L683-L693) when `NativeExecutionDispatch` is implemented, for `WasmExecutor` it always ignores [such request](https://github.com/Cardinal-Cryptography/polkadot-sdk/blob/aleph-v1.3.0/substrate/client/executor/src/executor.rs#L495). 

I'm not sure that `NativeExecutionDispatch` is required for `try-runtime` or `runtime-benchmarks`, yet I don't want to change the current behavior for those builds, as this PR only aims to make explicit wasm execution which is used by default (same as for [Polkadot](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/node/service/src/lib.rs#L436))

bin/node/src/service.rs
```diff
@@ -31,10 +31,11 @@ use sp_consensus_aura::{sr25519::AuthorityPair as AuraPair, Slot};
 use crate::{
     aleph_cli::AlephCli,
     chain_spec::DEFAULT_BACKUP_FOLDER,
-    executor::AlephExecutor,
+    executor as aleph_executor,
```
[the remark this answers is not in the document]
Marcin-Radecki: Good point

bin/node/src/executor.rs
Marcin-Radecki: Correct, `NativeElseWasmExecutor` constructor [takes care of it](https://github.com/paritytech/polkadot-sdk/blob/release-polkadot-v1.4.0/substrate/client/executor/src/executor.rs#L610).

PR #1683, opened 2024-04-12
A0-4231: Use json chainspec builder. Move chainspec generation to separate module.

Marcin-Radecki:
# Description

This is 1 out of N (2 or 3) PRs which goal is to remove `aleph-runtime` dependency from `aleph-node`. 

This PR:
* uses `json` chainspec builder (introduced in `polkadot 1.4.0`)
* moves chainspec generation to a separate module. Idea here is that this module eventually becomes new crate.

Latter has bigger design goal, which is to move below commands to separate bin crate called `generate-chain-spec`, ie moving `bootstrap-chain` command out of `aleph-node` binary. This will be a breaking change, but since it will be out in 14 release, with proper communication should be fine, as it mostly affects new chains - ie dev and testing ones. 

The architecture of `Command` and new module files might not be ideal, but will be improved in the subsequent PRs.

# Testing

- run e2e nightly
- compare manually generated non-raw chainspec with `main` version


bin/node/src/chain_spec/builder.rs
```diff
@@ -0,0 +1,164 @@
...
+fn calculate_initial_endowment(accounts: &[AccountId]) -> u128 {
+    let total_issuance = 300_000_000u128 * 10u128.pow(TOKEN_DECIMALS);
+    // due to known issue https://github.com/paritytech/polkadot-sdk/pull/2987/files,
+    // we need to make sure returned number is un u64 range, otherwise serde_json::json macro fails
+    // this is fixed in polkadot-sdk 1.6.0
+    total_issuance / (accounts.len() as u128) / 10
+}
+
+/// Configure initial storage state for FRAME modules.
+#[allow(clippy::too_many_arguments)]
```
timorleph, line 77: Maybe you could fix this since you are moving the code – or is there a reason why the arguments cannot be consolidated into a struct (or several)?

Also, are you sure this is even necessary? 5 arguments seems a bit low to trigger this check.

lesniak43 (approved):

timorleph (approved):
Looks good, only some nitpicking.

bin/node/src/chain_spec/builder.rs
Marcin-Radecki: Yes, I forgot to remove it, good point

bin/node/src/chain_spec/builder.rs
```diff
@@ -0,0 +1,164 @@
...
+    .with_properties(system_properties(token_symbol))
+    .build())
+}
+
+/// Calculate initial endowments such that total issuance is kept approximately constant.
+fn calculate_initial_endowment(accounts: &[AccountId]) -> u128 {
+    let total_issuance = 300_000_000u128 * 10u128.pow(TOKEN_DECIMALS);
+    // due to known issue https://github.com/paritytech/polkadot-sdk/pull/2987/files,
+    // we need to make sure returned number is un u64 range, otherwise serde_json::json macro fails
+    // this is fixed in polkadot-sdk 1.6.0
```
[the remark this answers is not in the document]
Marcin-Radecki: Sure

Marcin-Radecki:
Thanks for review, I "dared" to resolve conversations myself as they are straightforward.

PR #1687, opened 2024-04-16
A0-4231: Refactor all keystore and chainspec generation to seprate module

Marcin-Radecki:
# Description

Folow up after https://github.com/Cardinal-Cryptography/aleph-node/commit/5a8b1c380fe3ebae863ba8c89ef4aa73b01d0d16

This is 2nd out of 3 PRs to remove `aleph-runtime` dependency in `aleph-node`. In this PR:
* The whole keystore and chainspec generation logic is moved to `chain_spec` module. This module will be extracted as a whole, with small adjustment, to a separate crate
* `bootstrap-chain` command no longer supports `--faucet-account` switch. This is because only reason for it to exists was to set an initial endowment, exactly what `--rich-accounts` does. So it's suggested to migrate all `--faucet-accounts <account-id>` to `--rich-accounts <account-id>`. This is a breaking change, yet no impactful it's not used frequently.
* `bootstrap-node` command is removed. This is because it was doing the same what `bootstrap-chain` does, which now have `--authorities-account-ids` switch to differentiate between RPC and validator nodes

## Type of change

# Checklist:

* e2e tests: https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/8706331881


bin/node/src/chain_spec/cli.rs
```diff
@@ -36,28 +24,29 @@ pub struct ChainParams {
     #[arg(long, default_value = "DZERO")]
     token_symbol: String,
 
-    /// AccountIds of authorities forming the committee at the genesis (comma delimited)
+    /// all account ids that needs to have session keys generated when bootstraping chain (comma delimeted)
```
pmikolajczyk41, line 27: typo in 'delimeted'

bin/node/src/chain_spec/cli.rs
```diff
@@ -36,28 +24,29 @@ pub struct ChainParams {
...
     #[arg(long, value_delimiter = ',', value_parser = parse_account_id, num_args = 1..)]
     account_ids: Vec<AccountId>,
 
+    /// AccountIds of authorities forming the committee at the genesis (comma delimited)
+    /// If empty, then `--account-ids` are used as authorities
```
pmikolajczyk41, line 32: we should probably add a note that it should be a subset of `account_ids`

bin/node/src/chain_spec/commands.rs
```diff
@@ -0,0 +1,111 @@
...
+    keystore::{
+        create_abft_backup_dir, create_account_session_keys, create_p2p_key, open_keystore,
+    },
+    AlephNodeChainSpec, DEFAULT_BACKUP_FOLDER,
+};
+
+/// This command generates session keys and libp2p key for all input accounts.
+/// Those keys are written to the keystore in a given base path. It also generates a chainspec.
+#[derive(Debug, Parser)]
+pub struct BootstrapChainCmd {
```
pmikolajczyk41, line 21: this makes me just slightly uncomfortable, that bootstrapping chain (in particular session keys generation, keystore preparation) lives in the `chain_spec` module, which should be somehow independent... but as far as I understand, we would have to artificially split the `BootstrapChainCmd::run` body into two distinct commands, losing a convenient way to prepare local environment with a single instruction :[

I would opt for a better name for the `chain_spec` module (soon-to-be-a-crate), but feel free to ignore this comment.

pmikolajczyk41 (approved):
just one minor naming doubt, but apart from this: I love the changes!

bin/node/src/chain_spec/commands.rs
```diff
@@ -0,0 +1,111 @@
...
+                create_abft_backup_dir(account_base_path.path(), backup_dir);
+                let keystore = open_keystore(&self.keystore_params, chain_id, &account_base_path);
+                let session_keys = create_account_session_keys(&keystore, account_id);
+                let node_key_path = account_base_path.path().join(node_key_file);
+                create_p2p_key(node_key_path.as_path());
+                session_keys
+            });
+        let only_authorities = account_session_keys
+            .filter(|account_session_key| authorities.contains(&account_session_key.account_id))
+            .collect();
```
pmikolajczyk41, line 76: we may want to log those authorities that were ignored, because they haven't been included in the `account_ids`

aleph-client/src/connections.rs
```diff
@@ -419,7 +419,7 @@ impl<S: AsSigned + Sync> SignedConnectionApi for S {
 
 impl Connection {
     const DEFAULT_RETRIES: u32 = 10;
-    const RETRY_WAIT_SECS: u64 = 3;
+    const RETRY_WAIT_SECS: u64 = 6;
```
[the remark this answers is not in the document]
Marcin-Radecki: This is a fix for https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/8712537135/job/23899625772  - it takes more than 30 seconds for 7 validators on the single (free) GH machine to start RPC server

bin/node/src/chain_spec/cli.rs
Marcin-Radecki: Yes, and let's throw an error if it is not

bin/node/src/chain_spec/commands.rs
Marcin-Radecki: Excellent point, this should be named differently.  tried to make two commands, but code we have now is convenient  to do as as single command, as we need to read session keys from the storage to produce chainspec.

bin/node/src/chain_spec/commands.rs
Marcin-Radecki: When authorities will be a subset of acount_ids (https://github.com/Cardinal-Cryptography/aleph-node/pull/1687#discussion_r1568409247), then this is no issue

bin/node/src/cli.rs
```diff
@@ -69,14 +70,10 @@ pub enum Subcommand {
     #[command(subcommand)]
     Key(sc_cli::KeySubcommand),
 
-    /// Populate authorities keystore and generate chainspec in JSON format (printed to stdout)
-    /// Use `--raw` to produce the so called raw chainspec
+    /// Generates keystore (libp2p key and session keys), and generates chainspec to stdout
     BootstrapChain(BootstrapChainCmd),
 
-    /// Generate and print to stdout keys for a single node
-    BootstrapNode(BootstrapNodeCmd),
```
kostekIV, line 77: This is not good imho, incorporating boostrapnode into bootsrapchain is a no go for me.

kostekIV (changes requested):

kostekIV (approved):

PR #1689, opened 2024-04-17
A0-3848: Update weekly nightly start FE tests to more recent aleph-node

Marcin-Radecki:
We should be using more recent `aleph-node` images on our CI.

### DIFF
--- a/.github/workflows/weekly-featurenets-short-session-full-sha.yml
+++ b/.github/workflows/weekly-featurenets-short-session-full-sha.yml
@@ -24,7 +24,7 @@ jobs:
     secrets: inherit
     with:
       featurenet-name: 'ops-test-full-hash'
-      ref: '787c39186884aaf86ed48f46f575a8cd8cfd13e8'
+      ref: '87ebe7c4856b4f3365ddba55b55089713a96631a'
       expiration: '2h'
       validators: '5'
       internal: true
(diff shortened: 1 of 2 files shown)


timorleph (approved):
Oh, definitely! Additionally, updating this should probably be part of the release procedure, no?

Marcin-Radecki:
> Oh, definitely! Additionally, updating this should probably be part of the release procedure, no?

Could be yes, but I'd rather automate it, so e.g. make a step ni GHA that returns SHA/tag of current Testnet etc

.github/workflows/weekly-featurenets-short-session-full-sha.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     secrets: inherit
     with:
       featurenet-name: 'ops-test-full-hash'
-      ref: '787c39186884aaf86ed48f46f575a8cd8cfd13e8'
+      ref: '87ebe7c4856b4f3365ddba55b55089713a96631a'
```
pmikolajczyk41, line 27: these sha are very opaque :{ how do you choose one?

Marcin-Radecki: This is just a sha of `r-13-3`. Agree is not intuitive what should be there, but we'll do a follow-up step that returns a tag/SHA from the current Testnet - in the next PR.

pmikolajczyk41 (approved):

PR #1692, opened 2024-04-17
A0-3461: Update ABFT and remove legacy data IO

timorleph:
# Description

Updates AlephBFT, letting us get rid of the ugly `data_io::legacy` hack. Note that this updates `aleph_bft_crypto` which, in principle could be a compatibility issue. Fortunately, the change from 0.8 to 0.9 is only in some minor interfaces, the internal representation and, more importantly, encoding of all the types remains unchanged.

## Type of change

- ABFT update

# Checklist:


finality-aleph/src/abft/legacy/mod.rs
```diff
@@ -80,8 +86,10 @@ pub fn create_aleph_config(
...
 ) -> Config {
-    let mut config = default_config(n_members.into(), node_id.into(), session_id.0 as u64);
-    config.delay_config.unit_creation_delay = unit_creation_delay_fn(unit_creation_delay);
-    config.max_round = MAX_ROUNDS;
-    config
+    let mut delay_config = default_delay_config();
+    delay_config.unit_creation_delay = unit_creation_delay_fn(unit_creation_delay);
+    match create_config(n_members.into(), node_id.into(), session_id.0 as u64, MAX_ROUNDS, delay_config, Duration::from_millis(SESSION_LEN_LOWER_BOUND_MS as u64)) {
+        Ok(config) => config,
+        Err(_) => panic!("Incorrect setting of delays. Make sure the total AlephBFT session time is at least {} ms.", SESSION_LEN_LOWER_BOUND_MS),
```
kostekIV, line 93: could we include err in the panic msg?

finality-aleph/src/party/backup.rs
```diff
@@ -41,8 +43,14 @@ impl From<io::Error> for BackupLoadError {
 
 impl std::error::Error for BackupLoadError {}
 
-pub type Saver = Box<dyn Write + Send + Sync>;
-pub type Loader = Box<dyn Read + Send + Sync>;
+pub trait BothRead: Read + AsyncRead {}
+impl<T: Read + AsyncRead> BothRead for T {}
+
+pub trait BothWrite: Write + AsyncWrite {}
```
kostekIV, line 49: hmm naming is strange :P

kostekIV (approved):
Niceone

finality-aleph/src/abft/legacy/mod.rs
timorleph: The error is essentially a unit, so that doesn't make sense really. :/ It should probably be fixed on ABFT side.

finality-aleph/src/party/backup.rs
timorleph: Yeah, these are here only because you cannot have something `dyn` with more than one real trait. I still cannot think of a better name. :<

kostekIV: yrsh, me too :P

woocash2 (commented):
For such PR, In my opinion e2e testing is very important. As a must-have I suggest running the local-test
```
python local-tests/test_recover_after_abft_update.py
```
which tests finalization + recovery behavior.

Ideally we should repeat the following manual test which I performed ([link](https://www.notion.so/cardinalcryptography/2023-12-ABFT-update-test-1ee747e3040c47a4a1bfa490ca722481)), or a simplified version of it, which just checks finality (doesn't care about recovery, although, it's of course desired to run the full test). I could perform this test for you when I find some time.

timorleph:
The python test passes (I can send you the logs on request), I would wait with the more thorough ones until we have an actual release candidate (or at least branch).

woocash2 (approved):

finality-aleph/src/party/backup.rs
kostekIV: meh

PR #1694, opened 2024-04-18
A0-4264:  resolve cyclic polkadot-sdk dependency

Marcin-Radecki:
# Description

See https://github.com/Cardinal-Cryptography/polkadot-sdk/pull/6

I run `cargo update` on `main` to get the above incorporated`.

It turns out this saves almost 2x build time - from 18 minutes to 10 minutes.

## Type of change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -381,7 +381,7 @@ dependencies = [
  "sp-runtime",
  "sp-session",
  "sp-staking",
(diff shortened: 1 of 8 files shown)


timorleph (approved):
cc @lesniak43 the linked commit on top of 1.4 should also be cpd into your update

pmikolajczyk41 (approved):

PR #1700, opened 2024-04-23
A0-4251: Move `sign_with_params` to a new trait in aleph client

ggawryal:
# Description

In #1602, we've added `sign_with_params` to `SignedConnectionApi` in aleph-client to be able to specify nonce in lower level api for signing transactions, without submitting. Normally, this could be done in synchronous manner.
However, in Most bridge we hide signing keys in a separate enclave, and implement `SignedConnectionApi` by a client talking to that enclave, to be a proxy between that enclave and a user. This:
1. Complicates creating synchronous signing implementation for that proxy, as a channel to the enclave is async,
2. Forces implementing that lower level method, when it is in fact not needed in Most use case.

Therefore, I'm splitting that trait.
Also, in that PR I forgot to make `SubmittableExtrinsic` public, so I'm doing it now.

## Type of change

# Checklist:

- I have bumped `aleph-client` version


obrok (approved):
Code itself looks good to me. I guess, the question is. if we want to do _any_ changes directed at making `aleph_client` more usable when in fact it should not be used (see discussion on slack).

aleph-client/src/connections.rs
```diff
@@ -183,6 +183,18 @@ pub trait SignedConnectionApi: ConnectionApi {
...
     ) -> anyhow::Result<TxInfo>;
 
+    /// Returns account id which signs this connection
+    fn account_id(&self) -> &AccountId;
+
+    /// Returns a [`KeyPair`] which signs this connection
+    fn signer(&self) -> &KeyPair;
+
+    /// Tries to convert [`SignedConnection`] as [`RootConnection`]
+    async fn try_as_root(&self) -> anyhow::Result<RootConnection>;
```
kostekIV, line 193: actually lets add a default impl where we return an error like `not upgradable to root` or smth

kostekIV (approved):
lgtm

aleph-client/src/connections.rs
ggawryal: done

PR #1701, opened 2024-04-24
A0-4278: Change default finality version to current (for new chains)

Marcin-Radecki:
# Description

As per title

## Type of change

# Checklist


Marcin-Radecki:
It does not make sense to change default to current. Better to start new chain with legacy, and send `set_finality_version` instead.

PR #1705, opened 2024-04-26
A0-4289: Rewrite Gossip

lesniak43:
# Description

Replacing gossip service with `SyncNetworkService` and `NotificationService` which are responsible for managing new connections and sending data respectively. Now every protocol will be handled by a separate `NotificationService`, thus making our gossip code obsolete.

## Type of change

- Refactor


finality-aleph/src/network/substrate.rs
```diff
@@ -60,163 +83,230 @@ impl ProtocolNaming {
...
+        trace!(target: LOG_TARGET, "Disconnected event for peer {:?}", remote);
+        let addresses: Vec<_> = iter::once(remote).collect();
+
+        for protocol in self.protocols() {
+            let name = self.naming.protocol_name(&protocol);
+            if let Err(e) = self
+                .network
+                .remove_peers_from_reserved_set(name.clone(), addresses.clone())
+            {
+                warn!(target: LOG_TARGET, "Error while removing peer from {} reserved set: {}", name, e)
```
kostekIV, line 149: in peer connected this is error, here its warn - any reason for tht?

finality-aleph/src/network/substrate.rs
```diff
@@ -1,8 +1,16 @@
-use std::{collections::HashMap, iter, pin::Pin, sync::Arc};
+use core::fmt;
```
timorleph, line 1: uh, rather import the types, renamed if needed, and better from `std` rather than `core` for at least some consistency.

Unfortunately we are very inconsistent in how refer to all the types when implementing `Display`. >x

kostekIV (commented):
is this mostly deletion with some ctr+c +ctrl+v? Where is an interesting part to look at?

finality-aleph/src/network/substrate.rs
```diff
@@ -60,163 +83,230 @@ impl ProtocolNaming {
...
         }
     }
-}
 
-#[async_trait]
-impl<B: Block, H: ExHashT> EventStream<PeerId> for NetworkEventStream<B, H> {
-    async fn next_event(&mut self) -> Option<Event<PeerId>> {
-        use Event::*;
+    fn protocols(&self) -> [Protocol; 2] {
+        [Protocol::Authentication, Protocol::BlockSync]
```
timorleph, line 120: This can just be a constant, `Protocol::ALL` or somehting.

finality-aleph/src/network/substrate.rs
```diff
@@ -60,163 +83,230 @@ impl ProtocolNaming {
...
+    }
+
+    fn peer_connected(&mut self, remote: PeerId) {
+        let multiaddress: Multiaddr =
+            iter::once(MultiaddressProtocol::P2p(remote.into())).collect();
+        trace!(target: LOG_TARGET, "Connected event from address {:?}", multiaddress);
+
+        for protocol in self.protocols() {
+            let name = self.naming.protocol_name(&protocol);
```
timorleph, line 129: Alternatively (or in addition to?) to the above, add a method to `ProtocolNaming` and iterate over it here.

finality-aleph/src/network/substrate.rs
```diff
@@ -60,163 +83,230 @@ impl ProtocolNaming {
...
-                                peer,
-                                vec![(Protocol::Authentication, notification.into())],
-                            ));
-                        },
-                    }
-                },
+/// A thin wrapper around sc_network::config::NotificationService that stores a list
+/// of all currently connected peers, and introduces a few convenience methods to
+/// allow broadcasting messages and sending data to random peers.
+pub struct NotificationService {
```
timorleph, line 171: Terrible name, we use `Service` to mean something that should be `run` in this codebase. It should probably be called `*Network`, since that is the main purpose of it. Maybe `ProtocolNetwork`?

finality-aleph/src/network/substrate.rs
```diff
@@ -60,163 +83,230 @@ impl ProtocolNaming {
...
+        peer_ids
+            .intersection(&self.connected_peers)
+            .choose(&mut thread_rng())
+            .or_else(|| self.connected_peers.iter().choose(&mut thread_rng()))
+    }
+
+    fn handle_network_event(&mut self, event: SubstrateEvent) -> Option<(Vec<u8>, PeerId)> {
+        use SubstrateEvent::*;
+        match event {
+            ValidateInboundSubstream {
```
timorleph, line 196: Darn, it just occurred to me that we could simplify compatibility changes in the network massively by using these handshakes. Very out of scope for this PR, but perhaps worth creating a task.

finality-aleph/src/network/substrate.rs
```diff
@@ -60,163 +83,230 @@ impl ProtocolNaming {
...
+                );
+                return Ok(());
+            }
+        };
+        self.send_to(data, peer_id)
+    }
+
+    fn broadcast(&mut self, data: D) -> Result<(), Self::Error> {
+        for peer in self.connected_peers.clone() {
+            self.send_to(data.clone(), peer)?;
```
timorleph, line 278: I don't think you want to stop at the first fail, no?

finality-aleph/src/network/substrate.rs
```diff
@@ -60,163 +83,230 @@ impl ProtocolNaming {
...
-                    }
-                },
+/// A thin wrapper around sc_network::config::NotificationService that stores a list
+/// of all currently connected peers, and introduces a few convenience methods to
+/// allow broadcasting messages and sending data to random peers.
+pub struct NotificationService {
+    service: Box<dyn sc_network::config::NotificationService>,
+    connected_peers: HashSet<PeerId>,
+    last_status_report: time::Instant,
+}
```
timorleph, line 175: Hmm~, you deleted the metrics from `gossip` and as far as I can tell there is completely no equivalent in the logic of this new object, so nothing can be re-added, right?

That's kinda fine, but you gotta create a follow-up task to clean up Grafana.

finality-aleph/src/network/substrate.rs
```diff
@@ -22,6 +36,15 @@ const AUTHENTICATION_PROTOCOL_NAME: &str = "/auth/0";
 /// Name of the network protocol used by Aleph Zero to synchronize the block state.
 const BLOCK_SYNC_PROTOCOL_NAME: &str = "/sync/0";
 
+/// Protocols used by the network.
+#[derive(Debug, PartialEq, Eq, Copy, Clone, Hash)]
+pub enum Protocol {
```
timorleph, line 41: This is technically not strictly substrate and perhaps would be better in `mod.rs`. Not insisting though.

Hmm~, alternatively... is this even needed now? I.e. does any other component now use bare `Protocol`s rather than just `ProtocolNaming`, in which case it would be sufficient to only have it behave reasonably with constant strings it contains? On a quick glance this might be the case, but I'm not completely sure.

timorleph (changes requested):
Very cool, but still found some ways to complain. ;>

finality-aleph/src/network/substrate.rs
lesniak43: `send_to` always returns `Ok`, at some point we might want to have a custom broadcast error, but now it's pointless to add the logic supporting it

finality-aleph/src/network/substrate.rs
lesniak43: yes

bin/node/src/service.rs
```diff
@@ -228,7 +227,8 @@ fn get_net_config(
 ) -> (
     FullNetworkConfiguration,
     ProtocolNaming,
-    NotificationServices,
+    Box<dyn sc_network::config::NotificationService>,
+    Box<dyn sc_network::config::NotificationService>,
```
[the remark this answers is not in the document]
lesniak43: made it so that `get_net_config` returns our type instead, looks better even without a type alias now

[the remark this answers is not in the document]
lesniak43: fixed

finality-aleph/src/network/substrate.rs
lesniak43: fixed

finality-aleph/src/network/substrate.rs
lesniak43: you tell me...

finality-aleph/src/network/substrate.rs
timorleph: Added A0-4292 then.

finality-aleph/src/network/substrate.rs
timorleph: Then an `expect` explaining this would be appropriate here.

finality-aleph/src/nodes.rs
```diff
@@ -131,14 +132,12 @@ where
         }
     });
 
-    let (gossip_network_service, authentication_network, block_sync_network) =
-        GossipService::new(network_event_stream, spawn_handle.clone(), registry.clone());
     let gossip_network_task = async move {
```
kostekIV, line 135: you probablt want to rename gossip_network_task

finality-aleph/src/network/substrate.rs
lesniak43: fixed

finality-aleph/src/network/substrate.rs
kostekIV: lets make it error for consistency :P

finality-aleph/src/network/substrate.rs
```diff
@@ -60,163 +82,230 @@ impl ProtocolNaming {
...
+            self.service.protocol(),
+            self.connected_peers.len()
+        ));
+        info!(target: LOG_TARGET, "{}", status);
+    }
+}
+
+#[derive(Debug)]
+pub enum ProtocolNetworkError {
+    NetworkStreamTerminated,
```
kostekIV, line 228: Isnt it the same error as before? Cant we use one enum `NetworkError` for example?

finality-aleph/src/network/substrate.rs
```diff
@@ -60,163 +82,230 @@ impl ProtocolNaming {
...
+                .checked_add(STATUS_REPORT_INTERVAL)
+                .unwrap_or(time::Instant::now()),
+            STATUS_REPORT_INTERVAL,
+        );
+        loop {
+            tokio::select! {
+                maybe_event = self.service.next_event() => {
+                    let event = maybe_event.ok_or(Self::Error::NetworkStreamTerminated)?;
+                    if let Some((message, peer_id)) = self.handle_network_event(event) {
+                        match D::decode(&mut &message[..]) {
```
kostekIV, line 294: nit: decode -> decode_all

finality-aleph/src/network/substrate.rs
lesniak43: done even better

finality-aleph/src/network/substrate.rs
lesniak43: done even better

finality-aleph/src/network/substrate.rs
lesniak43: annihilated

finality-aleph/src/lib.rs
```diff
@@ -84,29 +86,62 @@ pub use crate::{
...
-        Protocol::BlockSync => MAX_BLOCK_SYNC_MESSAGE_SIZE,
-    }
+/// Name of the network protocol used by Aleph Zero to disseminate validator
+/// authentications.
+const AUTHENTICATION_PROTOCOL_NAME: &str = "/auth/0";
+
+/// Name of the network protocol used by Aleph Zero to synchronize the block state.
+const BLOCK_SYNC_PROTOCOL_NAME: &str = "/sync/0";
+
+pub struct NetConfig {
```
timorleph, line 96: All this would probably be better in `network/mod.rs` and only reexported here. Also, plx add doc.

timorleph (approved):
Nice, only very minor comments left.

finality-aleph/src/network/substrate.rs
lesniak43: you mean `SyncNetworkServiceError`?
`SyncNetworkService` will be removed soon + these are 2 different kinds of networks

finality-aleph/src/network/substrate.rs
lesniak43: done

finality-aleph/src/network/substrate.rs
lesniak43: added `let _ = ...` to ignore all the potential errors for now

finality-aleph/src/nodes.rs
lesniak43: done

finality-aleph/src/network/substrate.rs
lesniak43: done

finality-aleph/src/network/substrate.rs
kostekIV: ok, but the error is kinda the same? ANyway, feel free to resolve or go with my suggestion since one of them will be removed anyway

PR #1707, opened 2024-04-29
A0-4277: Unit creation delay

woocash2:
# Description

Change `DEFAULT_UNIT_CREATION_DELAY` from `300` ms to `200` ms. Change `MAX_ROUNDS` and exponential slowdown params accordingly so that the time to reach it after updating unit creation delay is similar to the time before the update.

Change the starting round of exponential slowdown from proportionally, i.e. from 5000 rounds to 7500 rounds, so that the no-slowdown time is 25 min which remains significantly longer than the expected session time (15 min).

## Testing
The following dummy test-case in AlephBFT:
```rust
let t1: Duration = (0..7000).map(|t| exponential_slowdown(t, 300.0, 5000, 1.005)).sum();
let t2: Duration = (0..10000).map(|t| exponential_slowdown(t, 200.0, 7500, 1.004)).sum();
println!("T1: {:?}", t1);
println!("T2: {:?}", t2);
```
gave output:
```
T1: 1290504.822s
T2: 1081023.038s
```
which is ~14.9 and ~12.5 days correspondingly. This means that the new time to reach max round will be 12.5 days which is not much smaller than 14.9 days, and at the same time, noticeably longer than the required 7 days.

**Reference**: [exponential slowdown](https://github.com/Cardinal-Cryptography/AlephBFT/blob/main/consensus/src/config.rs#L97)

## Type of change


ggawryal (approved):
LGTM, please don't forget to update also docker entry points at the appropriate time:
* https://github.com/Cardinal-Cryptography/aleph-node/blob/071be6c6595fc5b376d3e20f06d0df0269992506/docker/docker_entrypoint.sh#L31
*https://github.com/Cardinal-Cryptography/aleph-node/blob/071be6c6595fc5b376d3e20f06d0df0269992506/docker/common.yml#L11
*https://github.com/Cardinal-Cryptography/aleph-node/blob/071be6c6595fc5b376d3e20f06d0df0269992506/docker/smartnet-compose.yml#L19

PR #1708, opened 2024-04-30
A0-3901: Handle events in the base protocol

timorleph:
# Description

Make the base protocol service handle network events. In particular, it internally adds all the new peers to the other protocols, so it's not necessary to produce events for any outside actors.

Also some minor improvements for things I found on the way.

## Type of change

# Checklist:


finality-aleph/src/base_protocol/handler.rs
```diff
@@ -1,45 +1,147 @@
...
+            AlreadyConnected(peer_id) => write!(f, "peer {} already connected", peer_id),
+            TooManyPeers(role, direction) => {
+                write!(f, "too many {:?} nodes connected {:?}", role, direction)
+            }
+        }
+    }
 }
 
-/// Handler for the base protocol.
+struct ConnectionLimiter {
```
woocash2, line 81:
```suggestion
struct ConnectionLimits {
```

finality-aleph/src/base_protocol/handler.rs
```diff
@@ -1,45 +1,147 @@
...
     peers: HashMap<PeerId, PeerInfo>,
-    // the below counters and bounds ignore the nodes which belong to `reserved_nodes`
-    num_full_in_peers: usize,
-    num_full_out_peers: usize,
-    num_light_peers: usize,
-    max_full_in_peers: usize,
-    max_full_out_peers: usize,
-    max_light_peers: usize,
+    // the limiter ignores the nodes which belong to `reserved_nodes`
+    limiter: ConnectionLimiter,
```
woocash2, line 144:
```suggestion
    connection_limits: ConnectionLimits,
```

finality-aleph/src/base_protocol/handler.rs
```diff
@@ -1,45 +1,147 @@
...
-    PeerAlreadyConnected,
-    TooManyFullInboundPeers,
-    TooManyFullOutboundPeers,
-    TooManyLightPeers,
+    /// The peer is running on a different chain.
+    BadHandshakeGenesis(BlockHash),
+    /// The peer is already connected.
+    AlreadyConnected(PeerId),
+    /// There are too many peer of this type already connected.
+    TooManyPeers(Role, Direction),
```
woocash2, line 58: In theory it allows creation of pair `Light, Outbound` which doesn't make sense, although reducing the number of variants is cool. Consider a meet-in-the-middle alternative with 2 variants instead of 3:
```rust
TooManyFullPeers(Direction),
TooManyLightPeers,
```
This is minor stuff tho, feel free to ignore & resolve according to your preference

finality-aleph/src/base_protocol/service.rs
```diff
@@ -102,16 +122,75 @@ where
...
+            NotificationStreamOpened {
+                peer,
+                handshake,
+                direction,
+                negotiated_fallback: _,
+            } => match self.handler.on_peer_connect(peer, handshake, direction) {
+                Ok(()) => {
+                    let multiaddress: Multiaddr =
+                        iter::once(MultiaddressProtocol::P2p(peer.into())).collect();
+                    trace!(target: LOG_TARGET, "Connected event from address {:?}.", multiaddress);
```
woocash2, line 156:
```suggestion
                    trace!(target: LOG_TARGET, "Connect event from address {:?}.", multiaddress);
```
nit

finality-aleph/src/base_protocol/service.rs
```diff
@@ -102,16 +122,75 @@ where
...
+                            iter::once(multiaddress.clone()).collect(),
+                        ) {
+                            error!(target: LOG_TARGET, "Adding peer to the {} reserved set failed: {}.", name, e);
+                        }
+                    }
+                }
+                Err(e) => debug!(target:LOG_TARGET, "Failed to accept connection: {}.", e),
+            },
+            NotificationStreamClosed { peer } => {
+                trace!(target: LOG_TARGET, "Disconnected event for peer {:?}", peer);
```
woocash2, line 169:
```suggestion
                trace!(target: LOG_TARGET, "Disconnect event for peer {:?}", peer);
```
nit

finality-aleph/src/base_protocol/service.rs
```diff
@@ -102,16 +122,75 @@ where
...
         loop {
-            let command = self
-                .commands_from_user
-                .next()
-                .await
-                .ok_or(NoIncomingCommands)?;
-            self.handle_command(command);
+            tokio::select! {
+                command = self.commands_from_user.next() => self.handle_command(command.ok_or(NoIncomingCommands)?),
+                event = self.events_from_network.next_event() => self.handle_network_event(event.ok_or(NoNetworkEvents)?),
```
woocash2, line 192: Likely the answer is "YES", but let me ask to make sure: is breaking the select loop in case of an error in one of the branches correct?

finality-aleph/src/base_protocol/handler.rs
```diff
@@ -49,122 +151,90 @@ where
...
-            self.num_light_peers.saturating_dec();
+        if !self.is_reserved(&peer_id) {
+            self.limiter.remove(&peer)
         }
 
         Ok(())
     }
+
+    /// Checks whether an inbound peer would be accepted.
+    pub fn accept_inbound(
```
woocash2, line 232: As for this function:
* Seems we could live without it, it's only used to handle `ValidateInboundSubstream`
* But if you want to keep it, maybe it should be renamed to `verify_inbound_connection`
* Moreover it's kinda tedious how it converts the inner `Ok` value to `()`. It could have the same return type as `verify_connection`, and we could handle it in `ValidateInboundSubstream` with `Ok(_) => ...`.

woocash2 (approved):
LGTM, left a few nits / gentle suggestions / non-alarming questions

finality-aleph/src/base_protocol/handler.rs
timorleph: Done.

finality-aleph/src/base_protocol/handler.rs
timorleph: Done, although only "limits" as it should be enough.

finality-aleph/src/base_protocol/handler.rs
timorleph: Yeah, good point, it bothered me a bit as well. Fixed as suggested.

finality-aleph/src/base_protocol/service.rs
timorleph: Done.

finality-aleph/src/base_protocol/service.rs
timorleph: If any of the streams returns `None` we are definitely done, so yes.

PR #1710, opened 2024-05-02
A0-3010: Fix reorg metrics buckets

ggawryal:
# Description

Changing buckets, because I set them incorectly previously. For slo, we want to measure reorgs of length >= 1, >= 2, >= 3, >= 5 and >= 10. In Prometheus, buckets are [0; x] for some values of x, so to count e.g. reorgs of length >= 5 we have to subtract those <= 4 from all the reorgs.
Hence, I decreased all bucket values by 1.

## Type of change

# Checklist:


woocash2 (approved):
Makes sense

fixxxedpoint (approved):

PR #1711, opened 2024-05-03
A0-4295: Base protocol config creation

timorleph:
# Description

Config creation for the base protocol, mostly dummy values, but this is what we want in our case. Not yet used, will have to be used in tandem with the network building.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/base_protocol/config.rs
+++ b/finality-aleph/src/base_protocol/config.rs
@@ -11,10 +11,17 @@ use sp_runtime::traits::{Block, Header};
 
 use crate::{BlockHash, BlockNumber};
 
+// NOTE: `set_config` will be ignored by `protocol.rs` as the base
+// protocol is still hardcoded into the peerset.
+const DUMMY_SET_CONFIG: SetConfig = SetConfig {
+    in_peers: 0,
+    out_peers: 0,
+    reserved_nodes: Vec::new(),
+    non_reserved_mode: NonReservedPeerMode::Deny,
+};
+
 /// Generate a config for the base protocol and the notification service that should be passed to its service.
-pub fn generate_config<B>(
-    genesis_hash: B::Hash,
-) -> (NonDefaultSetConfig, Box<dyn NotificationService>)
+pub fn setup<B>(genesis_hash: B::Hash) -> (NonDefaultSetConfig, Box<dyn NotificationService>)
 where
     B: Block<Hash = BlockHash>,
     B::Header: Header<Number = BlockNumber>,
@@ -42,13 +49,6 @@ where
                 genesis_hash,
             ),
         )),
-        // NOTE: `set_config` will be ignored by `protocol.rs` as the base
-        // protocol is still hardcoded into the peerset.
-        SetConfig {
-            in_peers: 0,
-            out_peers: 0,
-            reserved_nodes: Vec::new(),
-            non_reserved_mode: NonReservedPeerMode::Deny,
-        },
+        DUMMY_SET_CONFIG,
     )
 }
--- a/finality-aleph/src/base_protocol/config.rs
+++ b/finality-aleph/src/base_protocol/config.rs
@@ -1,0 +1,54 @@
+use std::iter;
+
+use array_bytes::bytes2hex;
+use sc_network::{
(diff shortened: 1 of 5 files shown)


finality-aleph/src/base_protocol/config.rs
```diff
@@ -0,0 +1,54 @@
...
+    NotificationService,
+};
+use sc_network_common::sync::message::BlockAnnouncesHandshake;
+use sp_core::H256;
+use sp_runtime::traits::{Block, Header};
+
+use crate::{BlockHash, BlockNumber};
+
+/// Generate a config for the base protocol and the notification service that should be passed to its service.
+pub fn generate_config<B>(
```
kostekIV, line 15: Find a better name for this function - this does not only create config

finality-aleph/src/base_protocol/config.rs
```diff
@@ -0,0 +1,54 @@
...
+    genesis_hash: B::Hash,
+) -> (NonDefaultSetConfig, Box<dyn NotificationService>)
```
kostekIV, line 17: Consider using struct here instead of tuple + add some alias for `Box<....>` pls

lesniak43: When removing gossip I've solved this by returning our type that consumes the `Box<dyn ...>` - don't you want to do something similar here?

lesniak43 (approved):

finality-aleph/src/base_protocol/config.rs
timorleph: Our `base_protocol`'s `Service` takes a raw `Box` of this form, and a lot of other arguments that it would be inconvenient and confusing to pass in here.

finality-aleph/src/base_protocol/config.rs
kostekIV: Ok, why I have problem with that: 
- normally when Im reading function `generate_smth` I expect it to create `smth`. Extra stuff from such function is misleading, especially when its Boxed trait ;d
- I feel like this funciton is acually more about protocol service rather than just config, even though the service is a side effect of the ::new.

My propositions:
* `setup_base_protocol_service`
* `setup_protocol_service`
* `configure[_base]_protocol_service`

finality-aleph/src/base_protocol/config.rs
```diff
@@ -0,0 +1,54 @@
...
+            ),
+        )),
+        // NOTE: `set_config` will be ignored by `protocol.rs` as the base
+        // protocol is still hardcoded into the peerset.
+        SetConfig {
+            in_peers: 0,
+            out_peers: 0,
+            reserved_nodes: Vec::new(),
+            non_reserved_mode: NonReservedPeerMode::Deny,
+        },
```
kostekIV, line 52: nit: Add fn `dummy_set_config` or if possible `const DummyConfig: SetConfig = {...}`

finality-aleph/src/base_protocol/config.rs
kostekIV: Feel free to resolve but without different function name I would get confused what are those things returned from here.

finality-aleph/src/base_protocol/config.rs
timorleph: I strongly oppose calling it `service`, since this has a completely different meaning in our codebase, unfortunately. x< How about `setup_base_protocol`? Or perhaps just `setup` or `configure` (with a slight preference for the latter from me), since they are already in `base_protocol`? If we want to stress that we are also returning that interface maybe we could incorporate that word somehow, although I don't see how.

kostekIV: just setup is enough - since it will be imported form the `base_protocol`

finality-aleph/src/base_protocol/config.rs
timorleph: Oooh, cool, the `const` thingy works, nice. Changed as requested.

finality-aleph/src/base_protocol/config.rs
timorleph: Renamed then.

kostekIV (approved):

PR #1716, opened 2024-05-08
A0-2980: Wide fork scenario test

woocash2:
# Description

Wide fork scenario test. It is described pretty in detail in the comments.

## Type of change

- Unit test

# Checklist:
- I have added tests

PR #1719, opened 2024-05-10
A0-2980: Fix sync handler rejecting proper blocks.

woocash2:
# Description

When a block comes as a response to a request, we first process its header, similarly as we would get this header in a response. 

Additionally, next to making the `forest.importable(header.id)` check in `handle_header` which checks if forest would accept the corresponding block, we also allow for imports in a sequence. This means that if in the same response processing we imported the parent of the block, then we can also import the block, no matter what `forest.importable(...)` has returned.

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/finality-aleph/src/sync/handler/mod.rs
+++ b/finality-aleph/src/sync/handler/mod.rs
@@ -643,6 +643,11 @@ where
         (new_highest, None)
     }
 
+    /// Processes the given `header` passed by `peer`. May return an error which indicates that the header couldn't
+    /// have been verified correctly, forest couldn't have been updated with the header, or the corresponding block
+    /// is not needed to be imported. If the block is not importable from the perspective of the forest, we can
+    /// ignore it if we detect that the handler processes a sequence of block imports which is indicated by
+    /// `last_imported` being the id of the parent of the block of the `header`.
     fn handle_header(
         &mut self,
         header: UnverifiedHeaderFor<J>,
@@ -655,7 +660,9 @@ where
                 header: h,
                 maybe_equivocation_proof,
             }) => {
-                maybe_equivocation_proof.map(|proof| equivocation_proofs.push(proof));
+                if let Some(proof) = maybe_equivocation_proof {
+                    equivocation_proofs.push(proof);
+                }
                 h
             }
             Err(e) => return Err(e),
@@ -681,9 +688,9 @@ where
     /// dropped. Equivocated headers are processed in the same way as the ordinary ones.
     /// In any case, the Handler finishes in a sane state.
     ///
-    /// Note that this method does not verify nor import blocks. The received blocks
-    /// are stored in a buffer, and might be silently discarded in the future
-    /// if the import fails.
+    /// Note that this method does not import blocks. The received blocks are stored in a buffer,
+    /// and might be silently discarded in the future if the import fails. Headers of the blocks
+    /// are processed as if they came in the response themselves.
     pub fn handle_request_response(
         &mut self,
         response_items: ResponseItems<B, J>,
@@ -2931,8 +2938,8 @@ mod tests {
 
         // the result should be handled correctly, blocks should get imported
         let (new_justified, equivocations, maybe_error) = h2.handle_request_response(items, 1);
-        assert_eq!(new_justified, false);
+        assert!(!new_justified);
         assert!(equivocations.is_empty());
-        assert!(matches!(maybe_error, None));
+        assert!(maybe_error.is_none());
     }
 }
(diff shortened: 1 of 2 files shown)


woocash2:
Some weird stuff happened when I was rebasing my local branches so this PR includes some commits which inherited names from [the previous PR](https://github.com/Cardinal-Cryptography/aleph-node/pull/1716). When reviewing it may be more convenient to inspect **files changed** instead going through commits 1 by 1.

finality-aleph/src/sync/handler/mod.rs
```diff
@@ -643,6 +643,43 @@ where
...
+                h
+            }
+            Err(e) => return Err(e),
+        };
+        if let Err(e) = self.forest.update_header(&h, Some(peer), false) {
+            return Err(Error::Forest(e));
+        }
+        if !self.forest.importable(&h.id())
+            && !last_imported
+                .map(|id| h.parent_id().map(|p_id| id == p_id).unwrap_or(false))
```
timorleph, line 675: `flatten` instead of the double `unwrap_or`? Or rather `flat_map`.

finality-aleph/src/sync/handler/mod.rs
```diff
@@ -679,56 +716,35 @@ where
...
+                                _ => e,
+                            }),
                         );
                     }
-                }
-                ResponseItem::Block(b) => {
-                    if self.forest.skippable(&b.header().id()) {
-                        continue;
+                    last_imported = Some(b.header().id());
+                    if let Err(e) = self.import_block(b, false) {
```
timorleph, line 745: `import_block` here now completely unnecessarily verifies the header and checks for equivocations, please fix that – keeping in mind that we still have to do that for own blocks. Perhaps rename to `import_own_block` and here use `block_importer.import_block` directly?

finality-aleph/src/sync/handler/mod.rs
```diff
@@ -2879,4 +2895,53 @@ mod tests {
...
+            match n1.next().await {
+                Ok(BlockImported(header)) => {
+                    h1.block_imported(header)
+                        .expect("block imported should succeed");
+                }
+                _ => panic!("should notify about imported block"),
+            }
+        }
+
+        println!("kurcze 1");
```
timorleph, line 2920: Little chickens are unfortunately not welcome in our code, as that would make it non-vegan.

finality-aleph/src/sync/handler/mod.rs
```diff
@@ -2879,4 +2895,53 @@ mod tests {
...
+                assert_eq!(
+                    SimplifiedItem::from_response_items(items.clone()),
+                    vec![B(1), B(2)],
+                );
+                items
+            }
+            _ => panic!("should be response"),
+        };
+
+        println!("kurcze 2");
```
timorleph, line 2939: Here as well.

finality-aleph/src/sync/handler/mod.rs
```diff
@@ -2879,4 +2895,53 @@ mod tests {
...
+
+        // The result should be handled correctly, blocks should get imported.
```
timorleph, line 2941: Wasn't there some way to check whether the import calls happen? Just for completeness. Ignore if it would be annoying, the handler returning no errors should be indication enough in practice.

timorleph (changes requested):
One function is now doing redundant work, plx fix.

timorleph:
Oh, almost forgot – also remove the header from the initial block broadcast while you are at it. You might do that in a separate PR if you want, but technically this is the PR that makes it redundant.

finality-aleph/src/sync/handler/mod.rs
woocash2: Apologies, no more chickens :pray:

finality-aleph/src/sync/handler/mod.rs
woocash2: Good point. I noticed that but wanted to avoid changing logic too much. Changed as suggested.

finality-aleph/src/sync/handler/mod.rs
woocash2: Good point, done

finality-aleph/src/sync/handler/mod.rs
woocash2: Sorry, I don't see how to use `flat_map` to simplify this... :thinking:

timorleph: Change the `map` at the beginning of this line into `flat_map` and remove the `unwrap_or` inside the closure.

woocash2: `Option` doesn't support `flat_map`.

woocash2: I think `and_then` has similar semantics, let me try this

timorleph (approved):
Nice.

lesniak43 (approved):

PR #1723, opened 2024-05-14
L1-220: Request chain extension more liberally

timorleph:
# Description

Now any header with a number higher than the local favourite block can trigger a chain extension request. This makes the new sync keep up with blocks much better.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/sync/forest/mod.rs
+++ b/finality-aleph/src/sync/forest/mod.rs
@@ -359,7 +359,7 @@ where
 
     /// Updates the provided header, returns:
     /// 1. If required is set whether it became a new explicitly required.
-    /// 2. Otherwise whether it's a new header above the favourite block.
+    /// 2. Otherwise whether it's a new header with a number greater than the favourite block.
     pub fn update_header(
         &mut self,
         header: &J::Header,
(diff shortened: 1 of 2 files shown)


woocash2 (approved):
looks good :+1:

finality-aleph/src/sync/forest/mod.rs
```diff
@@ -359,25 +359,25 @@ where
 
     /// Updates the provided header, returns:
     /// 1. If required is set whether it became a new explicitly required.
-    /// 2. Otherwise whether it's a new descendant of the highest justified.
+    /// 2. Otherwise whether it's a new header above the favourite block.
```
ggawryal, line 362: The definition of `above` is not clear to me. How about:
```suggestion
    /// 2. Otherwise whether it's a new header above the favourite block (in terms of block number only).
```

ggawryal (approved):
LGTM

finality-aleph/src/sync/forest/mod.rs
timorleph: Good point, reworded to be more explicit.

PR #1724, opened 2024-05-15
L1-221: Respond to peer info requests

timorleph:
# Description

These are needed for the node to successfully respond to "health" RPCs used by a lot of our infra.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/network/base_protocol/handler.rs
+++ b/finality-aleph/src/network/base_protocol/handler.rs
@@ -6,6 +6,8 @@ use std::{
 use parity_scale_codec::{DecodeAll, Error as CodecError};
 use sc_network::{config::NetworkConfiguration, service::traits::Direction, PeerId};
 use sc_network_common::{role::Roles, sync::message::BlockAnnouncesHandshake};
+use sc_network_sync::types::ExtendedPeerInfo;
+use sp_core::H256;
 use sp_runtime::traits::{Block, Header, Saturating};
 
 use crate::{BlockHash, BlockNumber};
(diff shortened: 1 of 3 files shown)


Marcin-Radecki (approved):

finality-aleph/src/network/base_protocol/handler.rs
```diff
@@ -40,6 +51,21 @@ impl PeerInfo {
...
+where
+    B: Block<Hash = BlockHash>,
+    B::Header: Header<Number = BlockNumber>,
+{
+    fn from(peer_info: PeerInfo) -> Self {
+        ExtendedPeerInfo {
+            roles: peer_info.role.into(),
+            // this is never actually inspected, so return dummy values
+            best_hash: H256([0; 32]),
+            best_number: 0,
```
kostekIV, line 64: use const with names `DUMMY_*`

kostekIV: Actually, how hard would it be to get those values

kostekIV (approved):

finality-aleph/src/network/base_protocol/handler.rs
timorleph: Added `const`s.

Annoyingly hard. It would require:
1. Connecting the base protocol with sync. They are created quite far apart, so that complexity would spread all over the code.
2. Making sync actually track this – we have that info there, but only implicitly at this point.
Imo completely not worth it, if we wanted to return similar info it would be better to create our own RPC. Also, as mentioned in the comment, this _never_ actually gets inspected, the only call checks only the length of the returned vector, even having vaguely correct `Roles` is a bit of overkill.

PR #1728, opened 2024-05-16
A0-2963: select-loop imprv

fixxxedpoint:
# Description

Few small improvements around async/select-loops.

## Type of change

### DIFF
--- a/clique/src/service.rs
+++ b/clique/src/service.rs
@@ -105,7 +105,12 @@ impl<SH: SpawnHandleT> SpawnHandleExt for SH {
             task.await;
             let _ = tx.send(());
         });
-        Box::pin(rx.map_err(|_| ()))
+        Box::pin(rx.map_err(move |_| {
+            warn!(
+                target: LOG_TARGET,
+                "Task '{name}' exited early."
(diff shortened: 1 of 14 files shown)


clique/src/service.rs
```diff
@@ -92,6 +94,21 @@ pub trait SpawnHandleT {
...
+        &self,
+        name: &'static str,
+        task: impl Future<Output = ()> + Send + 'static,
+    ) -> Pin<Box<dyn Future<Output = Result<(), ()>> + Send>> {
+        let (tx, rx) = oneshot::channel();
+        self.spawn(name, async move {
+            task.await;
+            let _ = tx.send(());
+        });
+        Box::pin(rx.map_err(|_| ()))
```
kostekIV, line 108: nit: I dont like we are loosing an error here - though I know it would probably never be a problem hence nit

kostekIV (approved):
LGTM

clique/src/service.rs
fixxxedpoint: changed

ggawryal (approved):

PR #1734, opened 2024-05-17
A0-4224: Extracted pallet alleph runtime API to a separate crate

Marcin-Radecki:
# Description

This is a refactor which brings two new crates to `aleph-node` workspace:
* `fake-runtime-api`, that is used both by `finality-aleph` and `aleph-node` (was module in `primitives`)
* `pallet-aleph-runtime-api`, that is used by `aleph-runtime` and `finality-aleph` (latter uses just one trait)
The motivation is that we want `primitives` to be just bags of costs and types). Also, it follows an upstream approach, e.g., https://github.com/Cardinal-Cryptography/polkadot-sdk/blob/master/substrate/frame/nomination-pools/runtime-api/src/lib.rs


bin/node/src/service.rs
```diff
@@ -5,17 +5,16 @@ use std::{
...
     BlockImporter, BuildNetworkOutput, ChannelProvider, Justification, JustificationTranslator,
     MillisecsPerBlock, RateLimiterConfig, RedirectingBlockImport, SessionPeriod,
     SubstrateChainStatus, SyncOracle, TracingBlockImport, ValidatorAddressCache,
 };
 use log::warn;
-use primitives::{
-    fake_runtime_api::fake_runtime::RuntimeApi, AlephSessionApi, Block, DEFAULT_BACKUP_FOLDER,
-    MAX_BLOCK_SIZE,
-};
+use pallet_aleph_runtime_api::AlephSessionApi;
```
ggawryal, line 16: Do we need that `AlephSessionApi` import there?

ggawryal (approved):

bin/node/src/service.rs
Marcin-Radecki: Actually it is needed, there's below compilation error if it's not there:
```
error[E0599]: no method named `session_period` found for struct `ApiRef<'_, RuntimeApiImpl<Block<Header<u32, BlakeTwo256>, UncheckedExtrinsic>, ...>>` in the current scope
   --> bin/node/src/service.rs:198:14
    |
196 | /         client
197 | |             .runtime_api()
198 | |             .session_period(finalized)
    | |             -^^^^^^^^^^^^^^ method not found in `ApiRef<'_, RuntimeApiImpl<Block<Header<u32, BlakeTwo256>, UncheckedExtrinsic>, ...>>`
    | |_____________|
    | 
    |
    = note: the full type name has been written to '/home/ec2-user/remote-builds/17432567434322764125/target/release/deps/aleph_node-d7a11b894232f461.long-type-17325937729136008227.txt'
    = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
    |
3   + use pallet_aleph_runtime_api::AlephSessionApi;
```

woocash2 (approved):

PR #1735, opened 2024-05-20
L1-55: Bump spec_version to match new naming convention

Marcin-Radecki:
# Description

* change aleph-node and aleph-runtime version in their Cargo.toml from 0.14.0+dev to 14.0.0+dev
* on main, change aleph-runtime spec_version to 14_000_000

## Type of change


bin/runtime/src/lib.rs
```diff
@@ -79,7 +79,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("aleph-node"),
     impl_name: create_runtime_str!("aleph-node"),
     authoring_version: 1,
-    spec_version: 73,
+    spec_version: 14_000_000,
```
DamianStraszak, line 82: If we ever reach `>0` major version, then I think we might get into trouble with running beyond u32

lesniak43 (approved):

PR #1739, opened 2024-05-20
L1-121: SelectChain based on the favourite block

kostekIV:
Implementation of the select-chain that follows favourite block tracked by forest.

Only best_chain function is important to implement - as others are actually never used but introduces quite a complexity take a look at the commit before `leaves are not needed` :P


finality-aleph/src/sync/mod.rs
```diff
@@ -19,8 +19,11 @@ mod task_queue;
 mod tasks;
 mod ticker;
 
+mod select_chain;
```
timorleph, line 22: No reason to separate it from the above I think.

finality-aleph/src/sync/select_chain.rs
```diff
@@ -0,0 +1,112 @@
...
+#[async_trait::async_trait]
+impl<B, H> SelectChain<B> for FavouriteSelectChain<B>
+where
+    B: BlockT<Header = H, Hash = BlockHash>,
+    B: Block<Header = H, Hash = BlockHash>,
+    H: Sync + Send + Clone,
+{
+    async fn leaves(&self) -> Result<Vec<<B as BlockT>::Hash>, Error> {
+        // this is never used in the current version
+        Ok(vec![])
```
timorleph, line 104: As a slightly more informative alternative you could put at least the favourite in this vector.

You could also put leaves in there, as we actually keep that info in sync already, but if this is not used then it's probably not worth it.

If you like neither of these options at least change this to `Vec::new()`, since this is how we create empty `Vec`s.

finality-aleph/src/sync/select_chain.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub fn update_favourite(&self, new_favourite: Header) {
+        if let Err(TrySendError::Full(_)) = self.favourite_sender.try_send(new_favourite) {
+            warn!(target: LOG_TARGET, "SelectChainStateHandler channel full, skipping one notification")
+        }
+    }
+}
+
+pub struct StateWriter<B: Block> {
+    events: Receiver<<B as Block>::Header>,
+    favourite: Arc<RwLock<<B as Block>::Header>>,
```
timorleph, line 62: Welp, all of this is incredibly convoluted, channel to separate task which updates a rwlocked value. This also made you add a `_tx` to all tests and ignore it, which kinda sucks.

Alternative proposal – use a (probably async?) `RwLock` directly, hold and update it in `Service` and have `Forest` and `Handler` return the updated favourite header whenever needed. This should cut down on components needed and make testing of this feature in `Forest` and `Header` more natural.

timorleph (changes requested):
I don't like the design here, explanation in one comment.

finality-aleph/src/sync/select_chain.rs
kostekIV: I had implementation with the leaves but I dont think it would be worth it tbh (you can see it in the commit history). I will change it to `Vec::new()`

finality-aleph/src/sync/select_chain.rs
timorleph: `Handler` in general is not supposed to be doing any IO operations, whether sync or async, to make testing simpler. In particular, if you were to test the `_tx` you would have to read from it rather than just examine the return values.

Considering that the `SelectChain` trait has async methods it's users might expect to be able to give away control whenever blocking on these, so using an async lock would be advisable anyway.

kostekIV: After offline discussion changed the design to use oneshot channel. I think it looks quite nice now :)

finality-aleph/src/sync/select_chain.rs
kostekIV: done

finality-aleph/src/sync/mod.rs
kostekIV: fixed

finality-aleph/src/sync/select_chain.rs
```diff
@@ -1,111 +1,55 @@
...
     async fn best_chain(&self) -> Result<<B as BlockT>::Header, Error> {
-        let best = self.favourite.read();
+        let (rx, tx) = oneshot::channel();
+
+        self.favourite_block_request
+            .unbounded_send(rx)
+            .expect("The second end of the channel should be operational");
+        let best = tx
+            .await
+            .expect("The second end of the channel should be operational");
```
timorleph, line 50: The two `expect`s here should just be errors.

finality-aleph/src/sync/select_chain.rs
```diff
@@ -1,111 +1,55 @@
...
+
+        debug!(target: LOG_TARGET, "Best chain: {:?}", best);
```
timorleph, line 52: The `clone` below shouldn't be necessary, no?

finality-aleph/src/sync/select_chain.rs
```diff
@@ -1,111 +1,55 @@
...
+#[derive(Clone)]
+pub struct FavouriteSelectChain<B: Block> {
+    favourite_block_request: mpsc::UnboundedSender<oneshot::Sender<B::Header>>,
 }
 
-impl<Header> SelectChainStateHandler<Header> {
-    pub fn new() -> (Receiver<Header>, Self) {
-        let (rx, tx) = channel(CHANNEL_SIZE as usize);
+impl<B: Block> FavouriteSelectChain<B> {
+    pub fn new() -> (Self, mpsc::UnboundedReceiver<oneshot::Sender<B::Header>>) {
```
timorleph, line 18: Minor interface gripes –  how about having this return a single object that:

1. can return `impl SelectChain`s (by `clone`ing a sender it keeps inside),
2. can be deconstructed into a channel?

This would make the usage of it in `service.rs` and `nodes.rs` clearer (no weird pairs).

finality-aleph/src/sync/service.rs
```diff
@@ -757,6 +760,17 @@ where
                     self.handle_own_block(block);
                 },
 
+                maybe_favourite_block_sender = self.favourite_block_request.next() => {
+                    let favourite_block_sender = maybe_favourite_block_sender.ok_or(Error::FavouriteRequestChannelClosed)?;
+                    let favourite = self.handler.favourite_block();
+                    if let Err(_) = favourite_block_sender.send(favourite) {
+                        warn!(
+                            target: LOG_TARGET,
+                            "Error sending favourite block on request."
```
timorleph: This sounds confusingly as if it might be about sync requests over the network. Perhaps "Failed to respond with favourite header to user." or something like that. Also eliminates the "Error" in a `warn`.

timorleph (approved):
Lovely, only minor comments.

finality-aleph/src/sync/service.rs
kostekIV: ok, done

finality-aleph/src/sync/select_chain.rs
kostekIV: yep, removed

finality-aleph/src/sync/select_chain.rs
kostekIV: done

finality-aleph/src/sync/select_chain.rs
kostekIV: I think done? with some minor changes

timorleph (approved):
Ohh, making proper `ServiceComponents` is great, thanks!

lesniak43 (approved):

PR #1740, opened 2024-05-21
L1-230: Report sync mode

fixxxedpoint:
# Description

Outputs with `info` level status of the sync service - major/normal mode.

## Type of change


finality-aleph/src/sync/service.rs
```diff
@@ -706,6 +708,17 @@ where
         };
     }
 
+    fn report_sync_status(&mut self) {
+        let prev_status = self.major_sync_last_status;
+        let new_status = self.handler.major_sync();
+        match (prev_status, new_status) {
+            (false, true) => info!(target: LOG_TARGET, "Switched to major-sync mode."),
+            (true, false) => info!(target: LOG_TARGET, "Disabled major-sync mode."),
```
timorleph, line 716: It's not really a "mode", since the behaviour of the protocol is identical, perhaps "Switched to major sync state." and "No longer in major sync state." would be better?

lesniak43: I'd also suggest renaming the method to `report_sync_status_change`, since it does nothing most of the time.

lesniak43 (approved):

woocash2 (approved):
If we don't consider major / not major sync as "modes" then you can also consider changing the PR name -> "Report sync state" or something like that.

PR #1749, opened 2024-05-28
L1-191: Slo metrics refactor

ggawryal:
# Description

## Type of change

# Checklist:

- I have added tests


finality-aleph/src/metrics/best_block_related.rs
```diff
@@ -0,0 +1,177 @@
+use substrate_prometheus_endpoint::{
+    register, Gauge, Histogram, HistogramOpts, PrometheusError, Registry, U64,
+};
+
+use crate::{block::TreePathAnalyzer, BlockId};
+
+pub enum BestBlockRelatedMetrics<TPA> {
```
timorleph, line 7: Nit: The "related" in this name is kinda unnecessary, especially with "metrics" being plural.

finality-aleph/src/metrics/mod.rs
```diff
@@ -1,15 +1,15 @@
...
-pub use chain_state::run_chain_state_metrics;
-pub use finality_rate::FinalityRateMetrics;
+pub use slo::{run_metrics_service, SloMetrics};
+pub use timing::Checkpoint;
+pub type TimingBlockMetrics = timing::TimingBlockMetrics<DefaultClock>;
 use substrate_prometheus_endpoint::{exponential_buckets, prometheus};
-pub use timing::{Checkpoint, DefaultClock, TimingBlockMetrics};
-pub use transaction_pool::TransactionPoolInfoProvider;
+
+use crate::metrics::timing::DefaultClock;
```
timorleph, line 12: Nit: inconsistent with other uses above, should probably be `use timing::DefaultClock;`?

finality-aleph/src/metrics/slo.rs
```diff
@@ -0,0 +1,155 @@
...
+    finality_notifications: &mut FinalityNotifications<Block>,
+    transaction_pool_stream: &mut TS,
+) -> Result<(), Error> {
+    if metrics.is_noop() {
+        info!(target: LOG_TARGET, "Stopping metrics service: metrics were disabled.");
+        return Ok(());
+    }
+    loop {
+        tokio::select! {
+            maybe_block = import_notifications.next() => {
```
timorleph, line 42: If you are already using import/finality notifications then imo it would be better to have this happen in `sync`. In particular the whole business with figuring out the length of the reorg should become much simpler (no need for the trait and whatnot), but also it would better reflect the nodes internal notion of best (or rather favourite) block.

The runner here would still be needed for transactions though.

finality-aleph/src/nodes.rs
```diff
@@ -57,7 +57,7 @@ where
     C: crate::ClientForAleph<Block, BE> + Send + Sync + 'static,
     C::Api: AlephSessionApi<Block> + AuraApi<Block, AuraId>,
     BE: Backend<Block> + 'static,
-    TP: TransactionPool<Block = Block> + 'static,
+    TP: TransactionPool<Block = Block, Hash = primitives::Hash> + 'static,
```
timorleph, line 60: Nit: you probably want `BlockHash` here, although I think it refers to the same thing underneath? Or maybe I'm very wrong about this, but then I would like to know. :P

finality-aleph/src/metrics/transaction_pool.rs
```diff
@@ -1,63 +1,137 @@
...
+                        2000.0,
+                        BUCKETS_FACTOR,
+                        2,
+                        8,
+                    )?),
+                )?,
+                registry,
+            )?,
+            transactions_not_seen_in_the_pool: register(
+                Counter::new("aleph_transactions_not_seen_in_the_pool", "no help")?,
```
timorleph, line 54: Consider writing proper help messages after all maybe? Not a strict requirement, as I wouldn't quite know where to look for these other than in the code, but it's still nice to have stuff done properly.

There are some other places where it's set as "no help", this one just made me decide to write this comment after all. :P

finality-aleph/src/block/mod.rs
```diff
@@ -256,3 +256,18 @@ pub trait BestBlockSelector<H: Header>: Sync + Send + Clone {
     /// Return header of the leaf that should be considered the best block.
     async fn select_best(&self) -> Result<H, Self::Error>;
 }
+
+pub trait TreePathAnalyzer: Sync {
+    type Error: Display + Debug;
+
+    fn lowest_common_ancestor(&self, a: &BlockId, b: &BlockId) -> Result<BlockId, Self::Error>;
```
timorleph, line 263: I'm not quite clear as to why this is the trait in the first place.

Anyway, there is a comment in another file that would result in this being unnecessary, but otherwise I would like to know why this is a trait. Probably in some tests, but then maybe it's better to directly define the below function only?

timorleph (approved):
Cool stuff, one pretty major (although possibly negotiable?) comment and some nits.

finality-aleph/src/metrics/best_block_related.rs
woocash2: maybe we can change the filename too

woocash2 (approved):
LGTM

finality-aleph/src/nodes.rs
ggawryal: Hash here is the transaction hash, so I wanted to stress that. Anyway, `BlockHash = primitives::Hash = H256`, so this is only an aesthetic preference.

finality-aleph/src/metrics/slo.rs
timorleph: The source-of-truth for what "favourite" is is actually in the forest, so that happens on receiving an import notification by definition. The info you use now is afaik less accurate (I think it might even treat all reaorgs as reorgs to the top finalized actually, although I haven't checked).

After a quick glance into the forest I'm a tad more worried about getting the reorg length from there though, it might not be as trivial as I assumed. :/

ggawryal: Ah, ok. And reorgs could be detected by looking into db for the lca when updating favorite block, right?

timorleph: Yeah, although I'm slightly worried about this call in general, DB operations like this turned out to be particularly costly in the past, so it would be nice to not have them done here. Looking for lca in the forest should be simple to implement and much faster. On the other hand while there _used_ to be problems with the DB, I doubt there will be now, since we shouldn't be getting reorgs of length >20 ever, and that _cannot_ take long, so maybe it's not worth it.

tl;dr do as you with with reorg length calculation :P

finality-aleph/src/metrics/best_block_related.rs
ggawryal: Renamed to `best_block`

finality-aleph/src/metrics/mod.rs
ggawryal: You're right

finality-aleph/src/metrics/transaction_pool.rs
ggawryal: Added something

finality-aleph/src/block/mod.rs
ggawryal: Removed trait and moved impl to `SubstrateChainState`

finality-aleph/src/metrics/slo.rs
ggawryal: > The info you use now is afaik less accurate (I think it might even treat all reaorgs as reorgs to the top finalized actually, although I haven't checked).

If I'm not mistaken, I think I might be missing some temporary reorgs (succeeded by another reorg shortly after), but their length should be fine. Also, info from import notification streams is what frontend sees, so it is more important than what happens locally in the forest. 
Also, a slight improvement to this would be to detect reorgs also on every favorite block request by select chain - this way we won't miss any reorg visible from outside of sync.

finality-aleph/src/metrics/best_block_related.rs
timorleph: Even nitter: Missed the names in `metrics/slo.rs`.

finality-aleph/src/nodes.rs
timorleph: If you want to stress the distinction then there should be a separate constant in `primitives` called `TransactionHash` – I think that's actually a nice idea, even though I believe there are some uses within the codebase that would strictly speaking be wrong because of that (no need to look for these in this PR, we will eventually find them or they don't matter enough).

finality-aleph/src/sync/service.rs
```diff
@@ -585,20 +586,24 @@ where
...
-                            .broadcast(NetworkData::RequestResponse(broadcast))
-                        {
-                            warn!(
-                                target: LOG_TARGET,
-                                "Error broadcasting newly created block: {}.", e
-                            )
-                        };
+                    Ok(block_to_broadcast) => {
+                        if let Some(broadcast) = block_to_broadcast {
+                            own_block = true;
```
timorleph, line 595: Why the change from a match arm to an internal if? Not much wrong with that, question mostly out of curiosity.

finality-aleph/src/metrics/slo.rs
timorleph: With the changes that @kostekIV made to best block detection recently I think the approach you implemented (computing the length of the reorg exactly when the DB decides a newly imported block is the new best block) will slightly(?) overestimate the length of reorgs. But my intuition here is far from perfect, what do you think Jan?

I also think that changing the return values just requires writing proper types for them, and then it shouldn't be that bad? But no pressure on that point, if you didn't like what this looked like, it can stay as it is currently.

kostekIV: Yep they will be overestimated. Its because the best block can change during finalization and we will compute the lenght of the reorg for imported blocks after real reorg happened

ggawryal: Hmm, but the length of the reorg is the length of the retracted branch. The lowest common ancestor of the last imported block pre reorg caused by finalization, and any next imported block is the same, so the retracted part length also is, no?

finality-aleph/src/nodes.rs
ggawryal: Done

finality-aleph/src/metrics/best_block_related.rs
ggawryal: Good spotting

timorleph: Ahh, you really like the "related", don't you? xd You still left one variable with that in the constructor and then you have to use the `foo: bar` notation. I _hope_ this was the result of you using some automated renaming, not actually trying to keep it. :P

finality-aleph/src/metrics/slo.rs
timorleph: Oh, yeah, that should work, fair enough.

finality-aleph/src/metrics/best_block_related.rs
ggawryal: It seems that my IDE liked the previous name more. Fixed and hope my IDE won't eat my whole ram next time as a revenge.

PR #1750, opened 2024-05-28
L1-215: Add TxPause and SafeMode to runtime

kostekIV:


lesniak43 (approved):

bin/runtime/src/lib.rs
```diff
@@ -895,6 +896,68 @@ impl pallet_feature_control::Config for Runtime {
...
+    pub const DisallowPermissionlessRelease: Option<AlephBlockNumber> = None;
+}
+
+/// Calls that can bypass the safe-mode pallet.
+pub struct SafeModeWhitelistedCalls;
+impl Contains<RuntimeCall> for SafeModeWhitelistedCalls {
+    fn contains(call: &RuntimeCall) -> bool {
+        matches!(
+            call,
+            RuntimeCall::System(_) | RuntimeCall::SafeMode(_) | RuntimeCall::Timestamp(_)
```
ggawryal, line 919: How about `RuntimeCall::Sudo`? Is it allowed by deafult, or we don't want it?

ggawryal (approved):

bin/runtime/src/lib.rs
kostekIV: added

PR #1751, opened 2024-05-28
CW-53 Move baby-liminal-extenstion, vk-storage, feature-control outside workspace

pmikolajczyk41:
# Description

The following packages have been moved outside the workspace:
* `pallets/vk-storage`
* `pallets/feature-control`
* `baby-liminal-extension`

Previously, those were `cargo clipp`ed and `cargo test`ed using `cargo test—-workspace`; now, those checks are made by `check excluded packages`. This, unfortunately, adds 4 minutes to this workflow, which is not ideal. 

Not in `baby-liminal-extension`'s Makefile, I commented out the `test-backend` target. This target depends on aleph-runtime having `pallet-vk-storage` and `pallet-feature-control` enabled. I'll leave it for the next PRs to fix.

## Type of change


baby-liminal-extension/Makefile
```diff
@@ -3,7 +3,9 @@
 AT_TEST_CONTRACT = --manifest-path test_contract/Cargo.toml
 TEST_TOOLCHAIN = nightly-2024-02-14 # needed for unstable features regarding const generics
 
-test: test-frontend test-backend ## Run all tests
+# test-backend is disabled as whe this package was moving out of the workspace, it was removed from the aleph-runtime
+# aleongisde with vk-storage and feature-control pallets; test-backend depends on aleph-runtime having those two pallets
```
kostekIV, line 7:
```suggestion
# test-backend is disabled as when this package was moving out of the workspace, it was removed from the aleph-runtime
# alongisde with vk-storage and feature-control pallets; test-backend depends on aleph-runtime having those two pallets
```

bin/runtime/src/lib.rs
```diff
@@ -1000,10 +964,8 @@ construct_runtime!(
         Identity: pallet_identity = 20,
         CommitteeManagement: pallet_committee_management = 21,
         Proxy: pallet_proxy = 22,
-        FeatureControl: pallet_feature_control = 23,
-        VkStorage: pallet_vk_storage = 24,
-        SafeMode: pallet_safe_mode = 25,
-        TxPause: pallet_tx_pause = 26,
+        SafeMode: pallet_safe_mode = 23,
+        TxPause: pallet_tx_pause = 24,
```
kostekIV, line 968: Im not sure about this - we should probably keep a hole here and start from 25

kostekIV (approved):

bin/runtime/src/lib.rs
DamianStraszak, line 968: I know @kostekIV already asked this, but what's the situation here? Do we have any pallet number >=23 on testet/mainnet?

DamianStraszak (approved):

bin/runtime/src/lib.rs
Marcin-Radecki: We do have https://github.com/Cardinal-Cryptography/aleph-node/blob/r-13.3/bin/runtime/src/lib.rs#L940-L966 
```
        BabyLiminal: pallet_baby_liminal = 41,
        Operations: pallet_operations = 255,
```
Although it does not matter since those are not 23 or 24? Also, those pallets do not keep any storages

PR #1756, opened 2024-06-04
L1-258: Make all e2e tests run on a chain with up to speed finalization

Marcin-Radecki:
# Description

Recently, on main, there have been spurious nightly e2e logic test failures. While the exact root cause is unknown, they happen when the brand new chain is not yet up to speed with finalization. The scope of this task is to wait for finalization to spin off before starting an e2e case.

Currently, there’s an optional check called follow-up-finalization-check that is run after some e2e tests. It’s very cheap regarding time consumed, so we can always run before and after each e2e test. That fulfills the requirements of this task. Note that some e2e expects finalization to break after they run. 

When testing this PR, a bug in the previous change was accidentally found, causing 2 e2e cases to not run. A safeguard was added to prevent such situations in the future.

## Type of change

# Checklist:

[Nightly e2e logic tests](https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/9365029928) - OK
[Safeguard against empty e2e CI run](https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/9364653772/job/25778573449) - OK

### DIFF
--- a/.github/actions/run-e2e-test/action.yml
+++ b/.github/actions/run-e2e-test/action.yml
@@ -25,9 +25,6 @@ inputs:
   non-reserved-seats:
     description: 'Number of non-reserved seats available to validators.'
     required: false
-  follow-up-finalization-check:
-    description: 'Whether to run a follow-up finalization check.'
-    required: false
   deploy-adder:
     description: 'Whether to deploy the adder sample contract to the node.'
     required: false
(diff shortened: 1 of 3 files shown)


.github/actions/run-e2e-test/action.yml
```diff
@@ -89,6 +90,13 @@ runs:
       shell: bash
       run: docker load -i aleph-e2e-client.tar
 
+    - name: Check finalization before e2e test
```
ggawryal, line 93:
```suggestion
    - name: Wait for the finalization before e2e test
```

ggawryal (approved):

timorleph (approved):

PR #1758, opened 2024-06-06
L1-66: Pallet operations latest changes

Marcin-Radecki:
# Description

This PR extends and makes final logic changes for pallet operations. Now, it works in general mode, which calculates the expected `consumers` counter, compares it to the current one, and decreases or increases it when necessary. There are two main additions to pallet logic:
* It is now aware of whether an account is a contract one - in such cases it needs +1 `consumers`
* from `polkadot-sdk 1.4.0`, balance locks do not contribute to the `consumers` counter anymore. Hence, that logic was removed from pallet operations. 

## Type of change

## Companion PRs

https://github.com/Cardinal-Cryptography/aleph-node/pull/1760

### DIFF
--- a/pallets/operations/src/impls.rs
+++ b/pallets/operations/src/impls.rs
@@ -1,45 +1,56 @@
 #![allow(clippy::nonminimal_bool)]
 
-use frame_support::{dispatch::DispatchResult, traits::LockIdentifier, WeakBoundedVec};
-use pallet_balances::BalanceLock;
+use frame_support::dispatch::DispatchResult;
 use parity_scale_codec::Encode;
 use sp_core::hexdisplay::HexDisplay;
 use sp_runtime::DispatchError;
 
 use crate::{
     pallet::{Config, Event, Pallet},
-    traits::{AccountInfoProvider, BalancesProvider, BondedStashProvider, NextKeysSessionProvider},
-    LOG_TARGET, STAKING_ID,
+    traits::{
+        AccountInfoProvider, BalancesProvider, BondedStashProvider, ContractInfoProvider,
+        NextKeysSessionProvider,
+    },
+    LOG_TARGET,
 };
 
 impl<T: Config> Pallet<T> {
-    /// Checks if account has an underflow of `consumers` counter. In such case, it increments
-    /// it by one.
-    pub fn fix_underflow_consumer_counter(who: T::AccountId) -> DispatchResult {
+    /// Calculate expected consumers counter for a `who` account, and if actual
+    /// counter is not as expected, increment or decrement current counter
+    pub fn fix_consumer_counter(who: T::AccountId) -> DispatchResult {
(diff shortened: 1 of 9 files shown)


pallets/operations/src/impls.rs
```diff
@@ -1,45 +1,56 @@
...
+            expected_consumers += 1;
+        }
+        if Self::is_bonded(&who) {
             expected_consumers += 1;
         }
         if Self::has_next_session_keys_and_account_is_controller(&who) {
             expected_consumers += 1;
         }
 
+        #[allow(clippy::comparison_chain)]
```
kostekIV, line 37: Why? with match curretnt_consumers.cmp(expected_consumers) it would look better :P

pallets/operations/src/lib.rs
```diff
@@ -53,22 +56,29 @@ pub mod pallet {
...
+        /// +1 consumers for a controller account that is in session.next_keys
         /// +1 consumers if account bonded
         ///
-        ///	`fix_accounts_consumers_underflow` calculates expected consumers counter and comperes
+        ///	`fix_accounts_consumers_counter` calculates expected consumers counter and compares
         /// it with current consumers counter, incrementing by one in case of an underflow
+        /// or decrementing it by one in case of an overflow
+        ///
+        /// When expected is different from current by more than one, you might want to
+        /// call this extrinsic more than once.
```
kostekIV, line 81: Actually why wont we fix it in one call?

kostekIV (approved):

pallets/operations/src/lib.rs
Marcin-Radecki: I don't want to do code like this 
```
while expected_counter < current counter {
   inc_consumers();
}
```
just in case there would be some unforeseen changes in pallet balances in the future that might cause happen the loop is not to work as expected, e.g., infinite loop. I know it's almost not possible, yet I want to be on the safe side when placing a loop in the runtime code.

pallets/operations/src/impls.rs
Marcin-Radecki: I tried really to get this match done, but in the end I did not like it like if you look at `x.cmp(&y)`, without checking docs, which order is it, `x > y`, or `< y`? Can reader tell in 1 ms looking at the line of code? They definitely do when they look in the current version. :) 
```
    /// fn f(x: u8, y: u8) {
    ///      match x.cmp(&y) {
    ///          Ordering::Greater => a(),
    ///          Ordering::Less => b(),
    ///          Ordering::Equal => c()
    ///      }
    /// }
```

pallets/operations/src/lib.rs
DamianStraszak: I don't understand the reasoning. We will just remove this call in the future version, right?

DamianStraszak (approved):

PR #1760, opened 2024-06-11
L1-66: Script changes for fixing consumers counter

Marcin-Radecki:
# Description

This PR is a companion for https://github.com/Cardinal-Cryptography/aleph-node/pull/1758 that contains the same logic changes for a pallet balances maintenance script. The script was refactored to a final form where it's composed into small or middle-size modules.

## Type of change

### DIFF
--- a/scripts/accounts-invariants/account_invariants.py
+++ b/scripts/accounts-invariants/account_invariants.py
@@ -1,0 +1,156 @@
+import logging
+from aleph_chain_version import AlephChainVersion
+from utils import format_balance, save_accounts_to_json_file
+from consumers_counter import get_expected_consumers_counter
+
+log = logging.getLogger()
+
+
+def check_account_balances_invariants(account, chain_major_version, ed):
+    """
+    This predicate checks whether an accounts meet pallet balances and general account reference counters predicates.
+
+    :param account: AccountInfo struct (element of System.Accounts StorageMap)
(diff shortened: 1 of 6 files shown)


scripts/accounts-invariants/account_invariants.py
```diff
@@ -0,0 +1,156 @@
...
+    providers = account['providers']
+    consumers = account['consumers']
+    free = account['data']['free']
+    reserved = account['data']['reserved']
+
+    # in both versions, consumers must be 0 if providers are 0; also there is only one provider which is pallet
+    # balance so max possible value of providers is 1
+    account_ref_counter_invariant = (providers <= 1 and consumers == 0) or (consumers > 0 and providers == 1)
+
+    if chain_major_version <= AlephChainVersion.VERSION_11_4:
```
DamianStraszak, line 27: I guess this case will never be true anymore, so maybe just remove?

DamianStraszak (approved):

scripts/accounts-invariants/account_invariants.py
Marcin-Radecki: It's not the head of the chain, yet for now, this script supports querying the state from the past blocks.  There's a functionality to find total issuance imbalance, which uses that past state query feature. It's unlikely somebody will look past 12 version back to 11.4, yet it's not a harm if this code stays as it is now.

PR #1773, opened 2024-06-28
L1-265: Remove the fee_calculation test

obrok:
# Description

The fee_calculation test seems to:

1. Test configuration parameters of `pallet_transaction_payment`
2. Be flaky, possibly by design. What I mean by this (although I don't claim to have proven this) is that we run it without controlling block production, so we can't really guarantee a certain block occupancy, because it might depend on number of CPUs, random timings, etc.

## Type of change


kostekIV (approved):

pmikolajczyk41:
an alternative resolution could be to convert the test into pallet-unit-test-like, where we can control everything, including block production

Marcin-Radecki (approved):
Correct approach to remove something that is flaky in CI

PR #1780, opened 2024-07-10
A0-4318: network rate-limiter for the substrate network

fixxxedpoint:
# Description

Implementation of the rate-limiter algorithm for the substrate-based part of our network code. It limits only the `read` side of connections. ~~Additionally, now both rate-limit parameters for alephbft and substrate networks apply to all open connection (rate-limit for all nodes, not just single connection).~~ An exploit together with e2e tests will be provided in another PR.

## Type of change


finality-aleph/src/network/substrate.rs
```diff
@@ -27,6 +28,18 @@ pub struct ProtocolNetwork {
     last_status_report: time::Instant,
 }
 
+impl Borrow<Box<dyn sc_network::config::NotificationService>> for ProtocolNetwork {
+    fn borrow(&self) -> &Box<dyn sc_network::config::NotificationService> {
+        &self.service
+    }
+}
+
+impl BorrowMut<Box<dyn sc_network::config::NotificationService>> for ProtocolNetwork {
```
ggawryal, line 37: How about creating a type alias for `Box<dyn sc_network::config::NotificationService>` ?

rate-limiter/src/token_bucket.rs
```diff
@@ -44,12 +105,16 @@ impl TokenBucket {
...
     fn update_units(&mut self, now: Instant) -> usize {
         let time_since_last_update = now.duration_since(self.last_update);
-        let new_units = time_since_last_update
+        let mut new_units = time_since_last_update
             .as_micros()
             .saturating_mul(self.rate_per_second as u128)
             .saturating_div(1_000_000)
             .try_into()
             .unwrap_or(usize::MAX);
+        let used_new_units = min(new_units, self.requested);
```
ggawryal, line 114: Out of curiosity: was it computed incorrectly previously?

rate-limiter/src/token_bucket.rs
```diff
@@ -94,68 +163,70 @@ impl TokenBucket {
...
-            None
-        );
+        *time_to_return.borrow_mut() = now + Duration::from_secs(1);
+        assert_eq!(rate_limiter.rate_limit(9), None);
 
-        assert!(rate_limiter
-            .rate_limit(12, now + Duration::from_secs(1))
-            .is_some());
+        *time_to_return.borrow_mut() = now + Duration::from_secs(1);
+        assert!(rate_limiter.rate_limit(12).is_some());
```
ggawryal, line 185: Can we assert there that the value is exactly Some(duration)?

bin/node/src/aleph_cli.rs
```diff
@@ -93,8 +97,12 @@ impl AlephCli {
         self.enable_pruning
     }
 
-    pub fn alephbft_bit_rate_per_connection(&self) -> u64 {
-        self.alephbft_bit_rate_per_connection
+    pub fn alephbft_network_bit_rate(&self) -> u64 {
+        self.alephbft_network_bit_rate
+    }
+
+    pub fn substrate_network_bit_rate(&self) -> u64 {
```
ggawryal, line 104: I'd prefer not to make this configurable (only a constant), as this is too low level for any typical node use case to be modified. Anyway, leaving this up to you, as alephbft is even more low level imho, and it's already there…

ggawryal (commented):
Nice PR, left one question regarding write rate limiting and some nits.

rate-limiter/src/rate_limiter.rs
```diff
@@ -20,25 +23,36 @@ impl Clone for SleepingRateLimiter {
...
         trace!(
             target: LOG_TARGET,
             "Rate-Limiter attempting to read {}.",
             read_size
         );
 
-        let now = Instant::now();
-        let delay = self.rate_limiter.rate_limit(read_size, now);
+        let delay = {
+            let mut rate_limiter_lock = match self.rate_limiter.lock() {
```
deuszx, line 50: Won't this become a bottleneck? You're now taking a lock for every "packet"?

rate-limiter/src/token_bucket.rs
```diff
@@ -44,12 +105,16 @@ impl TokenBucket {
...
-        let new_units = time_since_last_update
+        let mut new_units = time_since_last_update
             .as_micros()
             .saturating_mul(self.rate_per_second as u128)
             .saturating_div(1_000_000)
             .try_into()
             .unwrap_or(usize::MAX);
+        let used_new_units = min(new_units, self.requested);
+        new_units -= used_new_units;
+        self.requested -= used_new_units;
```
deuszx, line 116: Do you need the `requested` field? Wouldn't `available` suffice? You'd call `self.update_units(now)` at the entry to `rate_limit`. It'd calculate how much time has passed since `self.last_update` and add appropriate amount of `new_units` to `self.available`. If you used `AtomicU64` then you could use `fetch_update` – `self.available.fetch_update(..., ..., available.saturating_sub(requested))`. This will return `available` before the call so you can now (lock-free) return to the caller `diff = requested - available` - covering the case where `requested > available` in which case the caller is asked to `sleep(diff)`. 

If you manage to work with a single `AtomicU64` I think it's possible to get rid of the `lock` on the call-side.

rate-limiter/src/token_bucket.rs
```diff
@@ -7,35 +7,96 @@ use log::trace;
...
+impl<F> TimeProvider for F
+where
+    F: FnMut() -> Instant,
+{
+    fn now(&mut self) -> Instant {
+        self()
+    }
+}
+
+pub struct DefaultTimeProvider {}
```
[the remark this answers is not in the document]
fixxxedpoint: changed

rate-limiter/src/token_bucket.rs
```diff
@@ -7,35 +7,96 @@ use log::trace;
...
+    }
+}
+
+impl TimeProvider for DefaultTimeProvider {
+    fn now(&mut self) -> Instant {
+        Instant::now()
+    }
+}
+
+impl Default for DefaultTimeProvider {
```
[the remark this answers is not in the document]
fixxxedpoint: changed

finality-aleph/src/network/substrate.rs
fixxxedpoint: added

rate-limiter/src/token_bucket.rs
fixxxedpoint: check the new version, please

rate-limiter/src/token_bucket.rs
fixxxedpoint: I would leave it this way - it's a basic, kind of black-box sanity check.

bin/node/src/aleph_cli.rs
fixxxedpoint: Let's leave it as it is now.

rate-limiter/src/rate_limiter.rs
fixxxedpoint: I did some benchmarks. It wasn't a bottleneck, but actually I used wrong type of mutex - we prefer version from `parking-lot`.

rate-limiter/src/token_bucket.rs
fixxxedpoint: Note: I think I am using a different definition of `lock-free` here. Did you mean something more like `wait-free` (progress for every thread...per definition on wikipedia)? `fetch_update` can block theoretically...but this not important here.

It is possible to make this algorithm `wait-free`. Main concern here was actually `starvation` of `good` peers by someone sending huge amount of data. With this mutex it behaves like a concurrent queue - `bad` actor can't fully starve `good` nodes, but can slow down everything significantly. That's why I've decided to switch it back to `per-node` rate-limiting version and tried to figure out if there is some other solution of this problem. One problem with per-node approach is that it is harder to choose good enough constants that scale with number of peers (with same hardware) - it is easier to find value like: my hardware can handle 8Mib/s of query bandwidth.

rate-limiter/src/lib.rs
```diff
@@ -1,6 +1,100 @@
...
+
+impl<Write: futures::AsyncWrite + Unpin> futures::AsyncWrite
+    for FuturesRateLimitedAsyncReadWrite<Write>
+{
+    fn poll_write(
+        self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+        buf: &[u8],
+    ) -> std::task::Poll<std::io::Result<usize>> {
+        self.get_inner().poll_write(cx, buf)
```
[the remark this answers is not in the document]
fixxxedpoint: It makes other things more complicated and also makes it difficult for peers to naturally find `consensus` of what is their preferred/best network bandwidth.

rate-limiter/src/token_bucket.rs
deuszx: By _lock-free_ I meant the very naive version – simply don't use `.lock()` calls to `Mutex`. 

Thanks for the comments and changes. I'll study them and learn :)

finality-aleph/src/network/build/mod.rs
```diff
@@ -47,10 +49,17 @@ pub struct NetworkOutput<TP: TransactionPool + 'static> {
     pub system_rpc_tx: TracingUnboundedSender<RpcRequest<TP::Block>>,
 }
 
+pub struct SubstrateNetworkConfig {
+    /// Maximum bit-rate in bytes per second of the substrate network (shared by sync, gossip, etc.).
+    pub substrate_network_bit_rate_per_connection: usize,
```
timorleph, line 54: Nit: this is a very verbose name, considering the name of the struct it should probably be called just `bit_rate_per_connection`.

timorleph (approved):
I only have one miniscule nit, but we should finish the conversation about how much this actually helps before this gets merged.

rate-limiter/src/token_bucket.rs
```diff
@@ -86,55 +69,50 @@ where
...
             .try_into()
             .unwrap_or(usize::MAX);
-        let used_new_units = min(new_units, self.requested);
-        new_units -= used_new_units;
-        self.requested -= used_new_units;
+        self.requested = self.requested.saturating_sub(new_units);
 
-        self.available = self.available.saturating_add(new_units);
-        self.last_update = now;
+        self.available()
```
ggawryal, line 99: Nit: to simplify signature, we can return nothing and instead call `self.available()` one level higher, after returning from this fn.

rate-limiter/src/token_bucket.rs
```diff
@@ -86,55 +69,50 @@ where
...
-    pub fn rate_limit(&mut self, mut requested: usize) -> Option<Duration> {
+    pub fn rate_limit(&mut self, requested: usize) -> Option<Option<Instant>> {
         trace!(
             target: LOG_TARGET,
             "TokenBucket called for {} of requested bytes. Internal state: {:?}.",
             requested,
             self
         );
-        if self.requested > 0 || self.available < requested {
+        if self.available() < requested {
```
ggawryal, line 115: Can't we `update_units` first? This would simplify a bit the whole function.

rate-limiter/src/token_bucket.rs
```diff
@@ -86,55 +69,50 @@ where
...
-        self.available = min(self.available, self.token_limit());
-        self.available
+    fn available(&self) -> usize {
+        self.rate_per_second.saturating_sub(self.requested)
     }
 
     /// Calculates [Duration](time::Duration) by which we should delay next call to some governed resource in order to satisfy
     /// configured rate limit.
-    pub fn rate_limit(&mut self, mut requested: usize) -> Option<Duration> {
+    pub fn rate_limit(&mut self, requested: usize) -> Option<Option<Instant>> {
```
ggawryal, line 108: Nit: I don't quite like nested options. Maybe we can replace `Option<Option<Instant>>` with `Option<Deadline>`, and define `enum Deadline { Never, Instant(Instant) }`? Or just don't care with the case where `rate_limit = 0`, and return max duration, as previously?

rate-limiter/src/token_bucket.rs
ggawryal: Much cleaner indeed

ggawryal (approved):

finality-aleph/src/network/build/mod.rs
fixxxedpoint: changed

rate-limiter/src/token_bucket.rs
fixxxedpoint: changed

rate-limiter/src/token_bucket.rs
fixxxedpoint: Changed. Current version is more clear.

rate-limiter/src/token_bucket.rs
fixxxedpoint: changed

PR #1805, opened 2024-09-03
L1-278: Fix synthetic-network

lesniak43:
# Description

Due to a LLVM bug the `synthetic-network` repository does not compile with the latest `nightly` toolchain. We introduce a local patch to fix the issue.

## Type of change


lesniak43:
@fixxxedpoint will take over

PR #1807, opened 2024-09-05
L1-278: Fix synthetic-network repo compilation

fixxxedpoint:
# Description

Added a custom version of the Dockerfile from the repo of the synthetic-network tool that uses a more specific version of the rust compiler. Also removed an ad-hoc fix for node-js images used by the original Dockerfile.

## Type of change

### DIFF
--- a/docker/Dockerfile.synthetic_network.build
+++ b/docker/Dockerfile.synthetic_network.build
@@ -1,0 +1,39 @@
+FROM node:20.3.0-bookworm
+
+# Rust nightly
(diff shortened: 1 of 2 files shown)


Marcin-Radecki (approved):
Splendid! So I can close this one right? https://github.com/daily-co/synthetic-network/issues/16 It's not needed anymore

lesniak43:
@Marcin-Radecki I'd leave the issue open, they still have a bug in their code.

PR #1809, opened 2024-09-12
L1-282: Exponential inflation

lesniak43:
# Description

Introducing exponential inflation.

## Type of change

### DIFF
--- a/bin/runtime/src/lib.rs
+++ b/bin/runtime/src/lib.rs
@@ -67,7 +67,7 @@ use sp_runtime::{
     transaction_validity::{TransactionSource, TransactionValidity},
     ApplyExtrinsicResult, FixedU128, RuntimeDebug, SaturatedConversion,
 };
-pub use sp_runtime::{FixedPointNumber, Perbill, Permill};
+pub use sp_runtime::{FixedPointNumber, Perbill, Permill, Saturating};
 use sp_staking::{currency_to_vote::U128CurrencyToVote, EraIndex};
 use sp_std::prelude::*;
 #[cfg(feature = "std")]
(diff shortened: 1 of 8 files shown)


pallets/aleph/src/lib.rs
```diff
@@ -68,6 +69,27 @@ pub mod pallet {
...
+        const MILLISECS_PER_YEAR: u64 = 1000 * 3600 * 24 * 36525 / 100;
+        MILLISECS_PER_YEAR
+    }
+
+    #[pallet::storage]
+    pub type AzeroCap<T: Config> = StorageValue<_, Balance, ValueQuery, DefaultAzeroCap>;
+
+    #[pallet::storage]
+    pub type ExponentialInflationHorizon<T: Config> =
+        StorageValue<_, u64, ValueQuery, DefaultExponentialInflationHorizon>;
```
kostekIV, line 91: call to set these storages is missing

DamianStraszak (approved):
Apart from the storage setters this is good to go.

I'm assuming that once we know which/if additional API is necessary to display necessary info in the frontends, then it will be added in a separate PR.

bin/runtime/src/lib.rs
```diff
@@ -441,11 +441,35 @@ parameter_types! {
...
 }
 
-pub struct UniformEraPayout;
+pub struct ExponentialEraPayout;
 
-impl pallet_staking::EraPayout<Balance> for UniformEraPayout {
-    fn era_payout(_: Balance, _: Balance, era_duration_millis: u64) -> (Balance, Balance) {
-        primitives::staking::era_payout(era_duration_millis)
+/// Calculates 1 - exp(-x) for small positive x
+fn exp_helper(x: Perbill) -> Perbill {
```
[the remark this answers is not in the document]
lesniak43: rewrote the function to make the computation more explicit and ensure that `1 - exp(-x) < x`

kostekIV (approved):
looks good.

bin/runtime/src/lib.rs
```diff
@@ -441,11 +441,35 @@ parameter_types! {
...
+    let x2 = x * x;
+    let x3 = x2 * x;
+    let x4 = x2 * x2;
+    let x5 = x4 * x;
+    (x - x2 / 2 + x3 / 6 - x4 / 24 + x5 / 120).min(x)
```
DamianStraszak, line 452: Hmm, what happens in case of overflows in multiplication? Are we sure we never panic?

PR #1822, opened 2024-10-07
L1-294: Safeguards for the inflation parameters setter

lesniak43:
# Description

We introduce safeguards that make it impossible to modify the values of the inflation parameters in a way that the new value differs too much from the old one. The rule is that the new value should lie between half and twice the old value. With regard to the AZERO cap, we look at the remaining issuance instead of the actual cap value, as this parameter influences the payouts directly.

## Type of change

### DIFF
--- a/pallets/aleph/src/lib.rs
+++ b/pallets/aleph/src/lib.rs
@@ -229,6 +229,64 @@ pub mod pallet {
 
             Self::finality_version()
         }
+
+        pub fn check_horizon_upper_bound(
+            new_horizon: u64,
+            current_horizon: u64,
+        ) -> Result<(), &'static str> {
+            match new_horizon > current_horizon.saturating_mul(2).saturating_add(1) {
+                true => {
+                    Err("Horizon too large, should be at most twice the current value plus one!")
+                }
+                false => Ok(()),
+            }
+        }
+
+        pub fn check_horizon_lower_bound(
+            new_horizon: u64,
+            current_horizon: u64,
+        ) -> Result<(), &'static str> {
+            match new_horizon < current_horizon / 2 {
+                true => Err("Horizon too small, should be at least half the current value!"),
+                false => Ok(()),
+            }
+        }
+
+        pub fn check_azero_cap_upper_bound(
+            new_cap: Balance,
+            current_cap: Balance,
+            total_issuance: Balance,
+        ) -> Result<(), &'static str> {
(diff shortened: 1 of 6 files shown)


pallets/aleph/src/traits.rs
```diff
@@ -26,3 +26,17 @@ where
         })
     }
 }
+
+/// Provides the current total issuance.
+pub trait TotalIssuanceProvider<T: Config> {
+    fn get() -> Balance;
+}
+
+impl<T> TotalIssuanceProvider<T> for pallet_balances::Pallet<T>
```
kostekIV, line 35: This could be moved to runtime/lib.rs and implemented for the new struct. The trait definition could be moved to the primitives

pallets/aleph/src/lib.rs
```diff
@@ -284,9 +285,47 @@ pub mod pallet {
...
+            let new_gap = match azero_cap.checked_sub(total_issuance) {
+                Some(new_gap) => new_gap,
+                None => {
+                    return Err(DispatchError::Other(
+                        "AZERO Cap cannot be lower than the current total issuance!",
+                    ))
+                }
+            };
+
+            if horizon_millisecs
```
kostekIV, line 305: Mv  every if to function and name it so reader know immediately what's the check

pallets/aleph/src/lib.rs
```diff
@@ -284,9 +285,47 @@ pub mod pallet {
...
+                > current_horizon_millisecs
+                    .saturating_mul(2)
+                    .saturating_add(1)
+            {
+                return Err(DispatchError::Other(
+                    "Horizon too large, should be at most twice the current value plus one!",
```
kostekIV, line 311: No hard preferences here but you could use new error variant instead of other plus string. Wouldn't it save a little bit of bin size for runtime as well? Not sure

pallets/aleph/src/traits.rs
lesniak43: done

pallets/aleph/src/lib.rs
lesniak43: done

pallets/aleph/src/lib.rs
lesniak43: let's leave it the way it is

Marcin-Radecki (approved):

PR #1843, opened 2024-10-24
L1-315: Updated payout_stakers bench with latest main

Marcin-Radecki:
# Description

This PR updates `payout_stakers` create so it runs with latest `main`, and how to run it against `r-13.4` release. Ultimately, this test should end up in nightly CI. 

## Type of change

### DIFF
--- a/benches/payout-stakers/README.md
+++ b/benches/payout-stakers/README.md
@@ -2,8 +2,17 @@
 
 This is a performance, e2e test for checking rewards for upper bound limit of nominators.
 
-Run by 
+## How to run
+
+1. Start local chain, e.g. via `./scripts/run_nodes.sh -v 1 -n 1 --dont-build`, where `-n`  denotes number of RPC nodes to start
+2. Set `aleph-client` in `Cargo.toml` to point to commit sha which your local `aleph-node` was built from. E.g. below builds `aleph-client` based on `r-13.4` version:
+```
+aleph_client = { git = "https://github.com/Cardinal-Cryptography/aleph-node", rev = "b072940af9b7295a141a89b943a29a74bb21c9f8" }
+```
+3. Run 
 ``` 
-RUST_LOG=aleph_e2e_client=info,aleph-client=info,aleph_client=info,payout_stakers=info cargo run --release -- --validator-count 10
+RUST_LOG=payout_stakers=info cargo run --release -- --validator-count 1
 ``` 
-It needs local nodes to be run (via `scripts/run_nodes.sh` for example.)
+where `1` was number of RPC nodes that will eventually become validators. 
+
+
--- a/benches/payout-stakers/src/main.rs
+++ b/benches/payout-stakers/src/main.rs
@@ -121,7 +121,7 @@ async fn main() -> anyhow::Result<()> {
             Some(validators.iter().map(|keypair| keypair.account_id().clone()).collect()),
             None,
             Some(CommitteeSeats{
-                reserved_seats: validators.iter().count() as u32,
+                reserved_seats: validators.len() as u32,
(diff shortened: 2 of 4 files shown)


benches/payout-stakers/README.md
```diff
@@ -2,8 +2,17 @@
 
 This is a performance, e2e test for checking rewards for upper bound limit of nominators.
 
-Run by 
+## How to run
+
+1. Start local chain, e.g. via `./scripts/run_nodes.sh -v 1 -n 1 --dont-build`, where `-n`  denotes number of RPC nodes to start
+2. Set `aleph-client` in `Cargo.toml` to point to commit sha which your local `aleph-node` was built from. E.g. below builds `aleph-client` based on `r-13.4` version:
+```
```
kostekIV, line 9: :(

kostekIV (approved):

benches/payout-stakers/src/main.rs
```diff
@@ -236,33 +275,35 @@ async fn nominate_validator(
...
-        let validator_address = address.to_string();
-        handles.push(tokio::spawn(async move {
-            let connection = SignedConnection::new(&validator_address, validator).await;
-            connection
-                .bond(MIN_VALIDATOR_BOND, TxStatus::InBlock)
-                .await
-                .unwrap();
-        }));
+/// We assume stash == validator == controller.
+async fn bond_validators_funds_and_choose_controllers(address_ip_without_port: &str, validators: Vec<KeyPair>) {
```
timorleph, line 290: Is `and_choose_controllers` still in any way accurate here?

benches/payout-stakers/src/main.rs
```diff
@@ -21,7 +25,7 @@ use rand::{thread_rng, Rng};
 use sp_keyring::AccountKeyring;
 
 // testcase parameters
-const NOMINATOR_COUNT: u32 = MAX_NOMINATORS_REWARDED_PER_VALIDATOR;
+const NOMINATOR_COUNT: u32 = 2 * MAX_NOMINATORS_REWARDED_PER_VALIDATOR;
```
timorleph, line 28: Perhaps a comment as to why the `2 *` would be nice. Are we testing that some nominators don't get any rewards?

bin/chain-bootstrapper/src/chain_spec/builder.rs
```diff
@@ -116,7 +116,7 @@ fn generate_genesis_config(
         "staking": {
             "forceEra": Forcing::NotForcing,
             "validatorCount":  account_session_keys.len() as u32,
-            "minimumValidatorCount": 4,
+            "minimumValidatorCount": 1,
```
timorleph, line 119: I hate this and I don't trust anything to work with <4 validators, but you can just resolve this comment as I think I have lost this fight already. :P

timorleph (approved):
Some minor comments, other than that looks good.

benches/payout-stakers/src/main.rs
Marcin-Radecki: Good point, not really, I'll fix naming

benches/payout-stakers/src/main.rs
Marcin-Radecki: I added a comment. This particular bench fails when run on r-13.4 version, but it is OK, as those changes prepare this crate to be run in the CI, which is based on `main`, that has paged exposure logic already. In the paged exposure world, all nominators are paid, regardless of how many of them are. In subsequent PR we should add a nighty workflow that runs this bench against `aleph-node` from `main`, and modify this file to call `payout_stalers` in a loop, until `Error::AlreadyClaimed` is received.

PR #1857, opened 2024-11-13
A0-4318: rate-limiter for all

fixxxedpoint:
# Description

Implementation of a shared-bandwidth rate-limiter for both substrate-based and alephbft networks.
In a next PR I will introduce a new e2e-test that should help to verify these changes.

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/bin/node/src/aleph_cli.rs
+++ b/bin/node/src/aleph_cli.rs
@@ -47,13 +47,13 @@ pub struct AlephCli {
     #[clap(long, default_value_t = false)]
     enable_pruning: bool,
 
-    /// Maximum bit-rate in bytes per second of the alephbft validator network.
-    #[clap(long, default_value_t = 4 * 1024 * 1024)]
-    alephbft_network_bit_rate: u64,
+    /// Maximum bit-rate per connection in bytes per second of the alephbft validator network.
+    #[clap(long, default_value_t = 64 * 1024)]
+    alephbft_network_bit_rate_per_connection: u64,
 
-    /// Maximum bit-rate in bytes per second of the substrate network.
-    #[clap(long, default_value_t = 4 * 1024 * 1024)]
-    substrate_network_bit_rate: u64,
+    /// Maximum bit-rate per connection in bytes per second of the substrate network.
+    #[clap(long, default_value_t = 64 * 1024)]
+    substrate_network_bit_rate_per_connection: u64,
 
     /// Don't spend some extra time to collect more debugging data (e.g. validator network details).
     /// By default collecting is enabled, as the impact on performance is negligible, if any.
@@ -97,12 +97,12 @@ impl AlephCli {
         self.enable_pruning
     }
 
-    pub fn alephbft_network_bit_rate(&self) -> u64 {
-        self.alephbft_network_bit_rate
+    pub fn alephbft_network_bit_rate_per_connection(&self) -> u64 {
+        self.alephbft_network_bit_rate_per_connection
     }
 
-    pub fn substrate_network_bit_rate(&self) -> u64 {
-        self.substrate_network_bit_rate
+    pub fn substrate_network_bit_rate_per_connection(&self) -> u64 {
+        self.substrate_network_bit_rate_per_connection
     }
 
     pub fn no_collection_of_extra_debugging_data(&self) -> bool {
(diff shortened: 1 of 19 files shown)


lesniak43 (changes requested):
Nice job! But please, make this PR fit the limit of 300 lines.

finality-aleph/src/network/build/transport.rs
```diff
@@ -0,0 +1,107 @@
...
+    rate_limiter: SharedRateLimiter,
+    config: sc_network::transport::NetworkConfig,
+) -> impl Transport<
+    Output = (
+        PeerId,
+        impl StreamMuxer<Substream = impl Send, Error = impl Send> + Send,
+    ),
+    Dial = impl Send,
+    ListenerUpgrade = impl Send,
+    Error = impl Send,
```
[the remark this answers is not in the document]
fixxxedpoint: `Send` doesn't apply to associated types. :(

I can hide that obscure type within the `network` function (build_network) in `network/build/base.rs`. This file seemed more logical place for it.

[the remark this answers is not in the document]
fixxxedpoint: https://github.com/fixxxedpoint/aleph-node/commit/df03552933c2cbf3792191b07bd53d65ae724116

rate-limiter/src/token_bucket.rs
```diff
@@ -1,211 +1,1209 @@
...
+        Self {
+            max_rate: self.max_rate,
+            peers_count: self.peers_count.clone(),
+            already_requested: None,
+        }
+    }
+
+    fn calculate_bandwidth(&mut self, active_children: Option<u64>) -> NonZeroRatePerSecond {
+        let active_children =
+            active_children.unwrap_or_else(|| self.peers_count.load(Ordering::Acquire));
```
Marcin-Radecki, line 212: While this code looks okay, according to [this](https://en.cppreference.com/w/cpp/atomic/memory_order#Release-Acquire_ordering), I wonder if we really significantly benefit from using non-sequential memory ordering? Ie how many threads will call methods `calculate_bandwidth` / `request_bandwidth` / `notify_idle` and how many such calls we expect? I literally spent 30 minutes now reminding myself what memory ordering was, and I assume each next reader of this code would need to do the same. Normally, such optimizations are once reviewed and tested, wrapped up in some library code, and never touched again, so nobody needs to read complex code. Not saying this does not work, but I think we need some justification to use this optimization here.

bin/node/src/aleph_cli.rs
```diff
@@ -47,9 +47,13 @@ pub struct AlephCli {
...
 
-    /// Maximum bit-rate per node in bytes per second of the alephbft validator network.
-    #[clap(long, default_value_t = 64 * 1024)]
-    alephbft_bit_rate_per_connection: u64,
+    /// Maximum bit-rate in bits per second of the alephbft validator network.
+    #[clap(long, default_value_t = 768 * 1024)]
+    alephbft_network_bit_rate: u64,
+
+    /// Maximum bit-rate in bits per second of the substrate network.
+    #[clap(long, default_value_t = 5*1024*1024)]
```
Marcin-Radecki, line 55: How do we come to reason those defaults are sensible?

Marcin-Radecki (approved):
This was really a splendid chunk of work. Please make sure we add an e2e for this as well.

bin/node/src/aleph_cli.rs
fixxxedpoint: For normal operation it should be plenty. While in major-sync mode, both local tests and on `t3a.xlarge` machines were unable to use more. Using a greater value I was able to kill a node on a `t3a.xlarge` machine by flooding. I can put some additional comment there.

rate-limiter/src/token_bucket.rs
fixxxedpoint: Changed it to `SeqCst`, which according to documentation should be exactly the same (Acquire for loads, Release for stores, Acq+Rel for fetch_modify).

Marcin-Radecki: that's correct, thank you Sir!

bin/node/src/aleph_cli.rs
Marcin-Radecki: Okay

PR #1867, opened 2024-11-29
A0-4341: Bump rust toolchain to 1.79

timorleph:
# Description

Mostly bumps the toolchain, but this also resulted in the detection and removal of some unused code. This PR also removes support for the `try-runtime` subcommand from the node binary – it was deprecated and we are using the external tool already anyway.

## Type of change

Toolchain bump and cleanup.

# Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -23,18 +23,18 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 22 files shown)


Marcin-Radecki (approved):
Thankfully, polkadot 1.6.0 compiles on 1.79.0, which I guess is enough to prove we can merge this PR. Only this [version](https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-stable2407-3) introduce offical Rust compatibility >=1.79.0

mike1729 (approved):

PR #1872, opened 2024-12-03
A0-4573: Add new SignatureSet and AuhtorityVerifier that compile to wasm

mike1729:
As in the title.


finality-aleph/src/abft/mod.rs
```diff
@@ -72,6 +73,18 @@ impl<S: 'static> IntoIterator for SignatureSet<S> {
     }
 }
 
+impl From<SignatureSet<crate::crypto::Signature>> for ScoreSignatureSet {
+    fn from(signature_set: SignatureSet<crate::crypto::Signature>) -> ScoreSignatureSet {
+        let score_sigantures: Vec<(primitives::NodeIndex, primitives::AuthoritySignature)> =
+            signature_set
+                .0
+                .into_iter()
+                .map(|(idx, s)| (idx.0 as u64, s.into()))
```
timorleph, line 82: You could just construct `IndexedSignature` here directly, and then you don't need a lot of the boilerplate (one `Into`, the typing above probably if you just constructed `ScoreSignatureSet` below explicitly).

timorleph (approved):
One improvement suggestion.

finality-aleph/src/abft/mod.rs
mike1729: Good idea!

mike1729: done

finality-aleph/src/abft/mod.rs
```diff
@@ -72,6 +73,24 @@ impl<S: 'static> IntoIterator for SignatureSet<S> {
     }
 }
 
+impl From<SignatureSet<crate::crypto::Signature>>
```
timorleph, line 76: We generally prefer top level imports (perhaps renaming) in the codebase to inline module mentions. This could just be `Signature` and below also `PrimitivesSignatureSet` or something.

finality-aleph/src/crypto.rs
```diff
@@ -100,49 +105,37 @@ impl AuthorityPen {
...
-}
-
 /// Holds the public authority keys for a session allowing for verification of messages from that
 /// session.
 #[derive(PartialEq, Clone, Debug)]
-pub struct AuthorityVerifier {
-    authorities: Vec<AuthorityId>,
-}
+pub struct AuthorityVerifier(
+    primitives::crypto::AuthorityVerifier<AuthorityId, primitives::AuthoritySignature>,
```
timorleph, line 112: Same here re:imports.

pallets/aleph/src/lib.rs
```diff
@@ -127,6 +129,11 @@ pub mod pallet {
     pub(super) type FinalityScheduledVersionChange<T: Config> =
         StorageValue<_, VersionChange, OptionQuery>;
 
+    // Test if runtime compiles
```
timorleph, line 132: Leftover comment?

pallets/aleph/src/lib.rs
```diff
@@ -287,6 +294,12 @@ pub mod pallet {
                 false => Ok(()),
             }
         }
+
+        pub fn verify_multisignature(msg: &Vec<u8>, sgn: &SignatureSet<Signature<T>>) -> bool {
+            let authority_verifier =
+                primitives::crypto::AuthorityVerifier::new(Self::authorities());
+            primitives::crypto::AuthorityVerifier::is_complete(&authority_verifier, msg, sgn)
```
timorleph, line 301: Not sure whether this is also the case in pallet code, some conventions here are different, but in case it is pointing another instance out.

primitives/src/lib.rs
```diff
@@ -413,3 +413,94 @@ pub mod staking {
...
+
+    use parity_scale_codec::{Decode, Encode};
+    use scale_info::TypeInfo;
+    use sp_runtime::RuntimeAppPublic;
+    use sp_std::vec::Vec;
+
+    use super::AuthoritySignature;
+
+    pub type NodeIndex = u64;
+    pub type NodeCount = u64;
```
timorleph, line 428: Hm, perhaps it's better to just use plain `u64`s here (for both), the type names are similar to the actually different types from ABFT and that might be confusing to the reader. We don't get much of of the type names anyway, since they are used in a pretty limited context.

primitives/src/lib.rs
```diff
@@ -413,3 +413,94 @@ pub mod staking {
...
+        /// Verifies whether the message is correctly signed with the signature assumed to be made by a
+        /// node of the given index.
+        pub fn verify(&self, msg: &Vec<u8>, sgn: &S, index: NodeIndex) -> bool {
+            match self.authorities.get(index as usize) {
+                Some(authority) => authority.verify(msg, sgn),
+                None => false,
+            }
+        }
+
+        pub fn node_count(&self) -> usize {
```
timorleph, line 481: Oh, the above comment is even more appropriate in light of this function not even returning `NodeCount`. :P

timorleph (changes requested):
A couple comments. Also, I don't understand what happened to the `aleph_zero.rs` file?

pallets/aleph/src/lib.rs
mike1729: Removed.

finality-aleph/src/crypto.rs
mike1729: done

finality-aleph/src/abft/mod.rs
mike1729: changed

pallets/aleph/src/lib.rs
mike1729: This is a leftover, imported on module level.

primitives/src/lib.rs
mike1729: Removed.

primitives/src/lib.rs
mike1729: Re above.

Marcin-Radecki (approved):

timorleph (approved):

finality-aleph/src/abft/mod.rs
```diff
@@ -72,8 +78,22 @@ impl<S: 'static> IntoIterator for SignatureSet<S> {
     }
 }
 
+impl From<SignatureSet<Signature>> for PrimitivesSignatureSet<AuthoritySignature> {
+    fn from(signature_set: SignatureSet<Signature>) -> PrimitivesSignatureSet<AuthoritySignature> {
+        let score_sigantures: Vec<IndexedSignature<AuthoritySignature>> = signature_set
```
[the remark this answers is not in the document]
mike1729: I'll fix that in a follow up PR today.

PR #1878, opened 2024-12-04
A0-0000: Make aleph-client/check-runtime-metadata.sh check more dev-friendly

Marcin-Radecki:
# Description

Adds `set -e` and more friendly output when calling a script.

### DIFF
--- a/aleph-client/check-runtime-metadata.sh
+++ b/aleph-client/check-runtime-metadata.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
 
 set -e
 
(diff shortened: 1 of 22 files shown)


aleph-client/check-runtime-metadata.sh
```diff
@@ -1,6 +1,12 @@
 #!/usr/bin/env bash
 
+set -e
+
 echo "This script must be run from the aleph-client directory."
+echo "Make sure you have aleph-node running in the background."
```
mike1729, line 6: You can check it by runnig `pidof aleph-node`, but it's not strictly necessary.

mike1729 (approved):

aleph-client/check-runtime-metadata.sh
Marcin-Radecki: Done

PR #1879, opened 2024-12-04
A0-4343: Implement runtime api for storing abft scores on chain.

mike1729:
# Description

Adds runtime api for the client side to store abft performance scores on chain. It's implemented using unsigned transactions since finality committee is considerably smaller than block production committee.

## Type of change

### DIFF
--- a/pallets/aleph/src/lib.rs
+++ b/pallets/aleph/src/lib.rs
@@ -451,7 +451,7 @@ pub mod pallet {
         pub fn unsigned_submit_abft_score(
             origin: OriginFor<T>,
             score: Score,
-            _signature: SignatureSet<Signature<T>>, // We don't check signature as it was already checked
+            _signature: SignatureSet<Signature<T>>, // We don't check signature as it was checked by ValidateUnsigned trait
         ) -> DispatchResultWithPostInfo {
             ensure_none(origin)?;
             Self::check_session_id(score.session_id).map_err(|e| DispatchError::Other(e.into()))?;
--- a/pallets/aleph/src/lib.rs
+++ b/pallets/aleph/src/lib.rs
@@ -39,9 +39,7 @@ pub mod pallet {
         pallet_prelude::{BlockNumberFor, OriginFor},
     };
     use pallet_session::SessionManager;
-    use primitives::{
-        Score, ScoreNonce, SessionInfoProvider, TotalIssuanceProvider,
-    };
+    use primitives::{Score, ScoreNonce, SessionInfoProvider, TotalIssuanceProvider};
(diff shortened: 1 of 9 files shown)


pallets/aleph/src/lib.rs
```diff
@@ -381,6 +451,51 @@ pub mod pallet {
 
             Ok(())
         }
+
+        // fix weight
+        #[pallet::call_index(3)]
+        #[pallet::weight(T::BlockWeights::get().max_block * Perbill::from_percent(10))]
+        /// Stores abft score
+        pub fn unsigned_submit_abft_score(
```
Marcin-Radecki, line 459: When I try to send this ext as unsigned from azero.dev, it gives me `1012: Transaction is temporarily banned` - is it expected?

image

mike1729: This is unsigned xt and you try to send is as signed, right?

Marcin-Radecki: And this is different from `InvalidTransaction::Future` and `InvalidTransaction::Stale` since I can reproduce this errors by sending for future and past session respectively. The UI says in those scenarios  `1010: Invalid Transaction: Transaction will be valid in the future` and `1010: Invalid Transaction: Transaction is outdated`

Marcin-Radecki: I click this button 
image

Marcin-Radecki: Let's leave testing for e2e phase

Marcin-Radecki (approved):

pallets/aleph/src/lib.rs
```diff
@@ -381,6 +443,51 @@ pub mod pallet {
...
+            origin: OriginFor<T>,
+            score: Score,
+            _signature: SignatureSet<Signature<T>>, // We don't check signature as it was already checked
```
timorleph, line 454: This looks very suspicious, even while I think it's correct. Could you a bit more explicitly say that it was checked in `pre_dispatch` and that it's impossible to bypass that at any point in block production or import (as long as that is correct ofc)?

timorleph (approved):
One complaint about a suspicious-looking comment.

Other than that – shouldn't this also bump the "transaction versions" or some such?

Marcin-Radecki:
> Other than that – shouldn't this also bump the "transaction versions" or some such? 

`transaction_version` and `spec_version` is bumped when we cut off the next release branch

pallets/aleph/src/lib.rs
mike1729: Done.

PR #1883, opened 2024-12-06
A0-4577: Implement era abft performance calculation

mike1729:
# Description

Calculate a performance stat for every finalizer at the end of a session and emit an event that a validator is underperforming when a threshold is met. In future, implement banning such validator.

## Type of change

### DIFF
--- a/pallets/committee-management/src/impls.rs
+++ b/pallets/committee-management/src/impls.rs
@@ -106,14 +106,14 @@ fn select_committee_inner<AccountId: Clone + PartialEq>(
     let non_reserved_committee =
         choose_for_session(non_reserved, non_reserved_seats, current_session as usize);
 
-    let mut finality_committee = choose_finality_committee(
+    let mut finalizers = choose_finality_committee(
         &reserved_committee,
         &non_reserved_committee,
         non_reserved_finality_seats,
         current_session as usize,
     );
 
-    let mut block_producers = match (reserved_committee, non_reserved_committee) {
+    let mut producers = match (reserved_committee, non_reserved_committee) {
         (Some(rc), Some(nrc)) => Some(rc.into_iter().chain(nrc.into_iter()).collect()),
         (Some(rc), _) => Some(rc),
         (_, Some(nrc)) => Some(nrc),
@@ -121,15 +121,11 @@ fn select_committee_inner<AccountId: Clone + PartialEq>(
     }?;
 
     // randomize order of the producers and committee
-    shuffle_order_for_session(
-        &mut block_producers,
-        &mut finality_committee,
-        current_session,
-    );
+    shuffle_order_for_session(&mut producers, &mut finalizers, current_session);
 
     Some(SessionCommittee {
-        block_producers,
-        finality_committee,
+        producers,
+        finalizers,
     })
 }
 
@@ -243,8 +239,9 @@ impl<T: Config> Pallet<T> {
         let CurrentAndNextSessionValidators {
             current:
                 SessionValidators {
-                    committee,
+                    producers,
                     non_committee,
+                    ..
                 },
             ..
         } = CurrentAndNextSessionValidatorsStorage::<T>::get();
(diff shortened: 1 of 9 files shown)


Marcin-Radecki (approved):
Changes looks good. We need migration for `committeeManagement.currentAndNextSessionValidatorStorage` which was discussed offline and task was created for it.

pallets/committee-management/src/manager.rs
```diff
@@ -142,6 +143,8 @@ where
             "Result of clearing the `SessionValidatorBlockCount`, {:?}",
             result.deconstruct()
         );
+
+        C::AbftScoresProvider::clear_scores_for_session(end_index);
```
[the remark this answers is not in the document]
mike1729: Storage is precious so we need to clear it periodically. Remember that storage is kept per block so you can read past values until they're pruned.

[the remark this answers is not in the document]
mike1729: We could clear them on era start, do you think that would be ok?

pallets/committee-management/src/impls.rs
```diff
@@ -345,18 +347,49 @@ impl<T: Config> Pallet<T> {
...
+
+        let is_underperforming = |score| score > minimal_expected_performance;
+
+        let finalizers_perf = T::AbftScoresProvider::scores_for_session(session_id)
+            .map(|score| score.points)
+            .unwrap_or(vec![minimal_expected_performance; finalizers.len()])
+            .into_iter()
+            .map(is_underperforming);
+
+        // TODO make sure ordering is correct on validators list
```
timorleph, line 369: This TODO probably shouldn't be here.

If it's about the ABFT order agreeing with the one from the pallets then it should, it should just be the source of that order, unless there is something unnecessarily shuffling it around somewhere in between.

pallets/committee-management/src/lib.rs
```diff
@@ -249,19 +268,65 @@ pub mod pallet {
...
+            underperformed_session_count_threshold: Option<u32>,
+            ban_period: Option<EraIndex>,
+        ) -> DispatchResult {
+            ensure_root(origin)?;
+
+            let mut current_committee_ban_config = FinalityBanConfig::<T>::get();
+
+            if let Some(minimal_expected_performance) = minimal_expected_performance {
+                ensure!(
+                    minimal_expected_performance <= 20,
```
timorleph, line 287: What is this 20 about?

timorleph (approved):
Couple minor comments.

pallets/committee-management/src/impls.rs
```diff
@@ -104,30 +106,26 @@ fn select_committee_inner<AccountId: Clone + PartialEq>(
...
-    let mut finality_committee = choose_finality_committee(
+    let mut finalizers = choose_finality_committee(
         &reserved_committee,
         &non_reserved_committee,
         non_reserved_finality_seats,
         current_session as usize,
     );
 
-    let mut block_producers = match (reserved_committee, non_reserved_committee) {
+    let mut producers = match (reserved_committee, non_reserved_committee) {
```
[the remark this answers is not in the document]
mike1729: I wanted the naming to be consistent.

pallets/committee-management/src/lib.rs
mike1729: I wanted to have some check on the value and I don't expect that lagging more than 20 rounds will be considered acceptable.

mike1729: Changed to value representing not producing units all session

pallets/committee-management/src/manager.rs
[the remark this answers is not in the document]
mike1729: Done.

pallets/committee-management/src/impls.rs
mike1729: I've removed the comment.

PR #1884, opened 2024-12-09
A0-4575: Add ABFT performance scorer

timorleph:
# Description

Adds a scorer that computes how many rounds which ABFT member was behind.

## Type of change

# Checklist:

- I have added tests

### DIFF
--- a/aleph-client/src/aleph_zero.rs
+++ b/aleph-client/src/aleph_zero.rs
@@ -1124,9 +1124,10 @@ pub mod api {
                         "predict_session_committee",
                         types::PredictSessionCommittee { session },
                         [
-                            3u8, 47u8, 188u8, 68u8, 88u8, 141u8, 129u8, 32u8, 29u8, 34u8, 56u8,
-                            60u8, 130u8, 180u8, 83u8, 106u8, 22u8, 51u8, 215u8, 42u8, 3u8, 160u8,
-                            84u8, 125u8, 221u8, 124u8, 236u8, 244u8, 236u8, 138u8, 36u8, 232u8,
+                            24u8, 24u8, 147u8, 150u8, 166u8, 202u8, 198u8, 18u8, 11u8, 149u8, 98u8,
+                            235u8, 33u8, 82u8, 91u8, 218u8, 181u8, 184u8, 209u8, 103u8, 160u8,
+                            165u8, 166u8, 118u8, 40u8, 218u8, 100u8, 175u8, 62u8, 41u8, 129u8,
+                            26u8,
                         ],
                     )
                 }
@@ -1678,9 +1679,10 @@ pub mod api {
                             input_data,
                         },
                         [
-                            202u8, 220u8, 173u8, 198u8, 41u8, 219u8, 66u8, 87u8, 181u8, 141u8,
-                            69u8, 194u8, 9u8, 90u8, 142u8, 250u8, 117u8, 135u8, 19u8, 123u8, 252u8,
-                            180u8, 251u8, 88u8, 86u8, 65u8, 1u8, 98u8, 78u8, 187u8, 159u8, 248u8,
+                            254u8, 77u8, 172u8, 241u8, 229u8, 242u8, 214u8, 150u8, 191u8, 23u8,
+                            65u8, 131u8, 250u8, 201u8, 161u8, 26u8, 7u8, 239u8, 218u8, 46u8, 93u8,
+                            38u8, 55u8, 154u8, 198u8, 77u8, 153u8, 188u8, 203u8, 140u8, 73u8,
+                            201u8,
                         ],
                     )
                 }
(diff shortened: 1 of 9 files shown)


finality-aleph/src/abft/current/performance/scorer.rs
```diff
@@ -0,0 +1,155 @@
+use current_aleph_bft::{NodeCount, NodeMap, Round};
+
+use crate::{abft::current::performance::Batch, UnverifiedHeader};
+
+pub type Score = Vec<Round>;
```
mike1729, line 5: This should be pulled from primitives so that runtime also knows this type.

finality-aleph/src/abft/current/performance/scorer.rs
```diff
@@ -0,0 +1,155 @@
...
+        self.newest_unit_by
+            .size()
+            .into_iterator()
+            .map(|node_id| {
+                self.newest_unit_by
+                    .get(node_id)
+                    // All other units have lower round than head, so the saturating_sub is just
+                    // subtraction.
+                    .map(|unit_round| max_round.saturating_sub(*unit_round))
+                    // If we don't have a unit it's the same as having a unit of round -1.
```
mike1729, line 39: This comment seems off. If we do have a unit of round - 1 then the score won't be max round + 1 unless last unit was a genesis unit.

mike1729 (changes requested):

finality-aleph/src/abft/current/performance/scorer.rs
timorleph: This is supposed to refer to an absolute round -1 (which doesn't really exist, but conceptually), not `max_round-1`, although I don't know how to clarify this. Maybe something like "...as having a virtual unit of round -1." or "We can conceptualize not having a unit as having an impossible unit of round -1."?

mike1729: Nah...I get it now 🤦 to deconfuse the sentence it's enough to change the end to "of round equal to '-1'".

finality-aleph/src/abft/current/performance/scorer.rs
timorleph: Ah, good point, this was written before the PR with `Score` in the primitives was merged and I forgot to edit in the necessary changes.

mike1729 (approved):

finality-aleph/src/abft/current/performance/scorer.rs
```diff
@@ -0,0 +1,153 @@
...
+        let max_round = batch.last().expect("batches always contain a head").round;
+        for unit in batch {
+            // Units are always added in order, so any unit created by an honest node
+            // here has a round greater than any that was included earlier.
+            // This is not necessarily true for forkers, but punishing them is fine.
+            self.newest_unit_by.insert(unit.creator, unit.round)
+        }
+        self.newest_unit_by
+            .size()
+            .into_iterator()
```
Marcin-Radecki, line 30: When I first read this code, I read it wrong so that I though "why it iterates over the same data structure `self.newest_unit_by` using consume iterator and in parallel it reads from it (`self.newest_unit_by.get(node_id)`), but then I realized I read it wrong - the consume iterator is made from `NodeCount` struct, not `NodeMap<Round>`.  Maybe we can improve readability here, by e.g. creating a named variable for ` self.newest_unit_by.size().into_iterator()`? Or simple `for` loop over `0..self.newest_unit_by.size()` would suffice here. My point is, for a reader which does not have this code opened in the IDE, so they can't jump into `.into_iterator()`, this would be more readable.

finality-aleph/src/abft/current/performance/scorer.rs
```diff
@@ -0,0 +1,153 @@
...
+            .map(|node_id| {
+                self.newest_unit_by
+                    .get(node_id)
+                    // All other units have lower round than head, so the saturating_sub is just
+                    // subtraction.
+                    .map(|unit_round| max_round.saturating_sub(*unit_round))
+                    // If we don't have a unit it's the same as having a unit of round equal to -1.
+                    .unwrap_or(max_round + 1)
```
Marcin-Radecki, line 38: Technically, this can overflow, as `max_round` is `u16`  , and `RawScore` is `Vec<u16>`.

Marcin-Radecki (approved):

finality-aleph/src/abft/current/performance/scorer.rs
timorleph: Went with the former, otherwise it's annoying when you cannot `collect` in a loop. Unfortunately the interface for `NodeMap` interates only over present items in general, otherwise this could've been simpler. :/

finality-aleph/src/abft/current/performance/scorer.rs
timorleph: Changed to `saturating_add`, although if we ever hit rounds so close to `u16::MAX` we are in much more trouble anyway. :P

finality-aleph/src/abft/current/performance/scorer.rs
Marcin-Radecki: yeah, I've just faced the same when trying to divide parents into direct and non-direct ones

PR #1886, opened 2024-12-17
Fix for sync config

Marcin-Radecki:
# Description

In aleph-node 13 [polkadot-sdk code](https://github.com/Cardinal-Cryptography/polkadot-sdk/blob/aleph-v1.2.0/substrate/client/network/sync/src/lib.rs#L516) there's a check against 0x0 block hash. This causes to ban nodes on 14 version by nodes on 13 version, causing unwanted behaviour.

## Type of change

### DIFF
--- a/finality-aleph/src/network/base_protocol/config.rs
+++ b/finality-aleph/src/network/base_protocol/config.rs
@@ -6,7 +6,6 @@ use sc_network::{
     NotificationService,
 };
 use sc_network_common::sync::message::BlockAnnouncesHandshake;
-use sp_core::H256;
 use sp_runtime::traits::{Block, Header};
 
 use crate::{BlockHash, BlockNumber};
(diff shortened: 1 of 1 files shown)


finality-aleph/src/network/base_protocol/config.rs
```diff
@@ -47,7 +46,7 @@ where
                 // The best block number, always send a dummy value of 0.
                 0,
                 // The best block hash, always an obviously dummy value.
```
timorleph: Maybe remove this comment and replace the previous one with something like "We always pretend the genesis block is our best block."

timorleph (approved):

finality-aleph/src/network/base_protocol/config.rs
Marcin-Radecki: done

mike1729 (approved):

PR #1887, opened 2024-12-17
A0-4580: Tests for pallet committee management

mike1729:
Missing runtime only tests for pallet committee management.


pallets/elections/src/lib.rs
```diff
@@ -17,7 +17,6 @@ use sp_std::{
     collections::{btree_map::BTreeMap, btree_set::BTreeSet},
     prelude::*,
 };
-
```
timorleph, line 20: Accidental line deletion or is there a reason?

mike1729: Accidental. Restored.

pallets/committee-management/src/mock.rs
```diff
@@ -0,0 +1,406 @@
...
+}
+
+parameter_types! {
+    pub const MinimumPeriod: u64 = 3;
+}
+
+impl pallet_timestamp::Config for TestRuntime {
+    type Moment = u64;
+    type OnTimestampSet = ();
+    type MinimumPeriod = ConstU64<5>;
```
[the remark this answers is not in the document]
mike1729: One was unecessary. Collapsed.

pallets/committee-management/src/impls.rs
```diff
@@ -382,7 +382,7 @@ impl<T: Config> Pallet<T> {
     }
 
     pub(crate) fn calculate_underperforming_validators() {
-        let thresholds = BanConfig::<T>::get();
```
Marcin-Radecki, line 385: Naming: usually we name getter in snake case exactly like its storage is, which would be in that case just `ban_config`. Or do we want to rename storage to `ProducersBanConfig`?

pallets/committee-management/src/lib.rs
```diff
@@ -47,7 +51,7 @@ impl Get<Perquintill> for DefaultLenientThreshold {
     }
 }
 
-const STORAGE_VERSION: StorageVersion = StorageVersion::new(0);
```
Marcin-Radecki, line 50: Why do we bump? We made no breaking changes, so we don't require a migration?

pallets/committee-management/src/tests.rs
```diff
@@ -0,0 +1,204 @@
...
+        let mut session_index = 2;
+        loop {
+            start_session(session_index);
+            if underperf_count == underperformed_session_count_threshold {
+                break;
+            }
+            assert_eq!(CommitteeManagement::banned(), Vec::<AccountId>::new());
+
+            // Make sure underperformer is a producer in every session.
+            let producers = CurrentAndNextSessionValidatorsStorage::<TestRuntime>::mutate(|sv| {
```
Marcin-Radecki, line 99: This test is a little long in terms of code lines, maybe we can extract some code e.g. this seems like a good method candidate like `add_underperformer_to_committee`, which we can re-use in `ban_underperforming_finalizers` test.

Marcin-Radecki (approved):

pallets/committee-management/src/impls.rs
mike1729: I didn't want to rename storage so that we don't have to write a migration. On the other hand, I wanted to differentiate this config from the finality config, hence the name. So I'd rather remove the getter than rename the storage if you don't like violating the convention. WDYT?

pallets/committee-management/src/tests.rs
mike1729: done

pallets/committee-management/src/lib.rs
Marcin-Radecki: Right.

pallets/committee-management/src/impls.rs
Marcin-Radecki: We need to write migration anyway, so we can rename storage.

mike1729: done

PR #1892, opened 2024-12-19
A0-4581: Storage migration for CommitteeManagement pallet

mike1729:


bin/runtime/src/lib.rs
```diff
@@ -1033,6 +1033,8 @@ pub type SignedBlock = generic::SignedBlock<Block>;
 /// BlockId type as expected by this runtime.
 pub type BlockId = generic::BlockId<Block>;
 
+pub type Migration = pallet_committee_management::migration::v1::Migration<Runtime>;
```
Marcin-Radecki, line 1036: Nit: since we now migrate to `2`, this should be `v2`

Marcin-Radecki (approved):

bin/runtime/src/lib.rs
mike1729: fixed.

PR #1897, opened 2024-12-20
A0-4342: Support for multiple kinds of hashes in aggregator

timorleph:
# Description

Makes the aggregator more generic and prepares for gathering multisignatures under hashes of ABFT performance reports. Not yet connected to anything, so technically doesn't change any behaviour.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/aggregation/mod.rs
+++ b/finality-aleph/src/aggregation/mod.rs
@@ -21,16 +21,16 @@ use crate::{
 /// BlockHash and Hash are the same, it has always been this way, but if it ever changes this place
 /// will cause trouble.
 #[derive(PartialEq, Eq, StdHash, Copy, Clone, Debug, Encode, Decode)]
-pub enum EitherHash {
+pub enum SignableTypedHash {
     /// The hash corresponds to a block.
     Block(Hash),
     /// The hash corresponds to an ABFT performance report.
     Performance(Hash),
 }
 
-impl AsRef<[u8]> for EitherHash {
+impl AsRef<[u8]> for SignableTypedHash {
     fn as_ref(&self) -> &[u8] {
-        use EitherHash::*;
+        use SignableTypedHash::*;
         match self {
             Block(hash) => hash.as_ref(),
             Performance(hash) => hash.as_ref(),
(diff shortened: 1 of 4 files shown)


mike1729 (approved):
Nice 🦾

finality-aleph/src/aggregation/mod.rs
```diff
@@ -16,16 +17,37 @@ use crate::{
     Keychain,
 };
 
+/// Either a block hash or a performance report hash. They should never overlap. We assume that
+/// BlockHash and Hash are the same, it has always been this way, but if it ever changes this place
+/// will cause trouble.
+#[derive(PartialEq, Eq, StdHash, Copy, Clone, Debug, Encode, Decode)]
+pub enum EitherHash {
```
Marcin-Radecki, line 24: What about `AggregatedHash`? Or `BroadcastedHash`? More specific name, `EitherHash` sounds like `GenericHash`

finality-aleph/src/aggregation/mod.rs
```diff
@@ -81,17 +103,24 @@ where
...
 
-    pub async fn start_aggregation(&mut self, h: BlockHash) {
+    pub async fn start_aggregation(&mut self, h: EitherHash) {
+        use EitherHash::*;
         match &mut self.agg {
             EitherAggregator::Current(agg) => agg.start_aggregation(h).await,
-            EitherAggregator::Legacy(agg) => agg.start_aggregation(h).await,
+            EitherAggregator::Legacy(agg) => match h {
+                Block(h) => agg.start_aggregation(h).await,
+                Performance(_) => { /* should never happen, but ignoring is fine */ }
```
Marcin-Radecki, line 112: Maybe panic here? Like Jon says here https://youtu.be/qfknfCsICUM?si=HUOv2aXBnkS1M6JE&t=342 . If we are truly right this never happens, this is an unrecoverable error, which means we should panic here

Marcin-Radecki (approved):

finality-aleph/src/aggregation/mod.rs
timorleph: Changed to `SignableTypedHash`, not sure whether that's better, but accurate.

finality-aleph/src/aggregation/mod.rs
timorleph: It should never happen, but it's not really an unrecoverable error, since ignoring it is always fine. That's why there is an `unimplemented!` in the other similar spot currently, that one is an actual problem _if_ it happens, even though it cannot.

PR #1907, opened 2025-01-08
A0-4346: Gather signatures and actually submit scores to chain

timorleph:
# Description

Send performance scores to the aggregator for gathering multisignatures under them and then submits them on chain. This should finish the ABFT performance project on the node side, modulo testing.

## Type of change

# Checklist:

### DIFF
--- a/finality-aleph/src/abft/current/performance/service.rs
+++ b/finality-aleph/src/abft/current/performance/service.rs
@@ -128,7 +128,7 @@ where
                 signatures_from_aggregator,
                 runtime_api,
                 pending_scores: HashMap::new(),
-                nonce: 0,
+                nonce: 1,
                 scorer: Scorer::new(NodeCount(n_members)),
                 metrics,
             },
--- a/finality-aleph/src/abft/current/performance/service.rs
+++ b/finality-aleph/src/abft/current/performance/service.rs
@@ -1,20 +1,30 @@
+use std::collections::HashMap;
+
 use current_aleph_bft::NodeCount;
 use futures::{
     channel::{mpsc, oneshot},
     StreamExt,
 };
 use log::{debug, error, warn};
+use parity_scale_codec::Encode;
+use sp_runtime::traits::Hash as _;
 
 use crate::{
     abft::{
         current::performance::{scorer::Scorer, Batch},
         LOG_TARGET,
     },
+    aleph_primitives::{
+        crypto::SignatureSet, AuthoritySignature, Hash, Hashing, RawScore, Score, ScoreNonce,
+    },
     data_io::AlephData,
     metrics::ScoreMetrics,
     party::manager::Runnable,
-    Hasher, UnverifiedHeader,
+    runtime_api::RuntimeApi,
+    Hasher, SessionId, UnverifiedHeader,
 };
+
+const SCORE_SUBMISSION_PERIOD: usize = 300;
 
 struct FinalizationWrapper<UH, FH>
 where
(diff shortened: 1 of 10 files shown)


finality-aleph/src/party/manager/aggregator.rs
```diff
@@ -154,8 +168,19 @@ where
                     ).await;
                 },
                 None => {
-                    debug!(target: "aleph-party", "Blocks ended in aggregator.");
-                    no_more_blocks = true;
+                    debug!(target: LOG_TARGET, "Blocks ended in aggregator.");
+                    session_over = true;
+                },
+            },
+            maybe_performance_hash = performance_from_scorer.next(), if !no_more_performance && !session_over => match maybe_performance_hash {
```
Marcin-Radecki, line 175: Do we need here `&& !session_over`? What happens if we remove this?

finality-aleph/src/abft/current/performance/service.rs
```diff
@@ -1,21 +1,31 @@
...
+    },
     data_io::AlephData,
     metrics::ScoreMetrics,
     party::manager::Runnable,
-    Hasher, UnverifiedHeader,
+    runtime_api::RuntimeApi,
+    Hasher, SessionId, UnverifiedHeader,
 };
 
+const SCORE_SUBMISSION_PERIOD: usize = 300;
```
Marcin-Radecki, line 27: Will this `aleph-node` version publish some scores actually? I run it locally with below diff included, waited 300s, and nothing happened in the logs, nor in metrics

```
[14:55] marol-Latitude-5521:aleph-node ((e694d21b) *%) | for port in 9615 9616 9617 9618;   curl localhost:9616/metrics 2>/dev/null | grep
 "my_abft_score{"; end
my_abft_score{chain="a0dnet1"} 0
my_abft_score{chain="a0dnet1"} 0
my_abft_score{chain="a0dnet1"} 0
my_abft_score{chain="a0dnet1"} 0
```

[tmp.diff.txt](https://github.com/user-attachments/files/18362442/tmp.diff.txt)

Marcin-Radecki (approved):

finality-aleph/src/abft/current/performance/service.rs
mike1729: No because the first nonce in every session is 0 and should be 1.

mike1729 (approved):
Just one small error, besides that good to go. Congrats for winning the uneven fight with compiler for pushing runtime API💪

finality-aleph/src/abft/current/performance/service.rs
mike1729: Also, remember that you have to specify finality version as current.

finality-aleph/src/abft/current/performance/service.rs
```diff
@@ -87,38 +114,89 @@ where
...
                         Some(batch) => self.scorer.process_batch(batch),
                         None => {
                             error!(target: LOG_TARGET, "Batches' channel closed, ABFT performance scoring terminating.");
                             break;
                         },
                     };
-                    debug!(target: LOG_TARGET, "Received ABFT score: {:?}.", score);
-                    self.metrics.report_score(score[self.my_index]);
-                    // TODO(A0-4339): sometimes submit these scores to the chain.
+                    self.metrics.report_score(points[self.my_index]);
```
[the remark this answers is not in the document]
timorleph: Yeah, it should always return points of the same length, equal to the total number of nodes. We could try handling failures there here, but I don't think it's worth it.

finality-aleph/src/party/manager/aggregator.rs
timorleph: Not strictly. If it wasn't here, it would try submitting performance that is from the previous session. This will also work, by just failing at a later point, but imo it's clearer to already stop here – if you think otherwise I can remove this check.

finality-aleph/src/abft/current/performance/service.rs
```diff
@@ -87,38 +114,89 @@ where
...
         (
             Service {
                 my_index,
+                session_id,
                 batches_from_abft,
+                hashes_for_aggregator,
+                signatures_from_aggregator,
+                runtime_api,
+                pending_scores: HashMap::new(),
+                nonce: 0,
```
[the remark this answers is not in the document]
timorleph: Darn, I thought I had the nonce bump before the creation of the score so this should've worked, but apparently not... and also putting it here is clearer anyway, so fixed. <_<

finality-aleph/src/abft/current/performance/service.rs
timorleph: I checked properly and it seems to be working as intended now. Still requires bumping finality version to 5, but this is by design.

mike1729: You mean bumping in primitives? There is a flag in run_nodes for that now.

PR #1919, opened 2025-01-24
A0-4591: Added major sync metric back (although by different name)

Marcin-Radecki:
# Description

This PR brings back major syncing metric that was unwillingly removed in 14 release. It copies some code from `polkadot-sdk` (metric definition), copy is required since we want different name and `MajorSyncingGauge` constructor is private in `polkadot-sdk`. Other than that implementation approach is taken from the `polkadot-sdk` - the `AtomicBool` is passed from its creation down to when it's used in the metric gauge.

## Type of change

# Testing

I run locally `./scripts/run_nodes.sh -v 4`, then immediately run `watch -n -1 ' curl localhost:9616/metrics | grep major_sync' is one console and in the other 
```
[12:22] marol-Latitude-5521:aleph-node (A0-4591 *%) | tail -f run-nodes-local/node-1.log | grep "major sync"
2025-01-24 12:22:36.317  INFO tokio-runtime-worker aleph-block-sync: Switched to major sync state.
2025-01-24 12:22:46.320  INFO tokio-runtime-worker aleph-block-sync: No longer in major sync state.
```
and I saw correlation when the log appeared, the metric switched its state accordingly.


bin/node/src/service.rs
```diff
@@ -398,6 +398,7 @@ pub fn new_authority(
         validator_port: aleph_config.validator_port(),
         rate_limiter_config,
         sync_oracle,
+        is_major_syncing: major_sync,
```
timorleph, line 401: We are already passing the `SyncOracle` a line above, and it consists exactly of the same `Arc<AtomicBool>`. It's probably better to just add a method to it like `underlying_atomic(&self) -> Arc<AtomicBool>` or something. The you can also change the constructor to just return this and get the atomic by hand.

timorleph (approved):
This can probably be made quite a bit simpler, see the single comment. But other than that it's alright, so if you think my suggestion is bad feel free to merge.

bin/node/src/service.rs
Marcin-Radecki: Done. The reason I did not choose that path is that it breaks `SyncOracle` class invariant in which it is owner of underlying atomic - now every user of the class can clone atomic and set it to `true` for example, despite no major sync is happening. But since this class is not exported in some library, I think it is fine, and code is much sipler now - returning tuple from a constructor when one of its field was underlying atomic also was not perfect SOLID-wise.

timorleph (approved):

PR #1929, opened 2025-02-04
A0-0000: Choose a producer for every block in session

mike1729:


Marcin-Radecki (approved):

pallets/committee-management/src/manager.rs
```diff
@@ -127,7 +130,12 @@ where
         // Notify about elected next session finality committee
         C::FinalityCommitteeManager::on_next_session_finality_committee(finalizers);
 
-        Some(producers)
+        let full_size = C::SessionPeriod::get() as usize;
+        let mut full_aura: Vec<_> = producers.into_iter().cycle().take(full_size).collect();
+        let mut rng = Pcg32::seed_from_u64(new_index as u64);
```
timorleph, line 135: Maybe add a short comment about what is happening here and why, otherwise this code is a bit mysterious.

timorleph (approved):

PR #1934, opened 2025-02-05
A0-0000: Don't assume we have a score submission period

timorleph:
# Description

This failed on the real chain, incl snapshots. Now it should never submit in practice.

## Type of change

# Checklist:

### DIFF
--- a/bin/node/src/service.rs
+++ b/bin/node/src/service.rs
@@ -208,7 +208,7 @@ fn get_aleph_runtime_vars(client: &Arc<FullClient>) -> AlephRuntimeVars {
     let score_submission_period = client
         .runtime_api()
         .score_submission_period(finalized)
-        .expect("should always be available");
+        .unwrap_or(u32::MAX);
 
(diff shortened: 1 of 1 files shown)


bin/node/src/service.rs
```diff
@@ -208,7 +208,7 @@ fn get_aleph_runtime_vars(client: &Arc<FullClient>) -> AlephRuntimeVars {
     let score_submission_period = client
         .runtime_api()
         .score_submission_period(finalized)
-        .expect("should always be available");
+        .unwrap_or(u32::MAX);
```
Marcin-Radecki, line 211: Not `300`? So if someone runs v15 binary on v14 snapshot, they got abft scoring disabled?

timorleph: That is the case anyway, since scoring requires v15 runtime.

Marcin-Radecki: Right.

Marcin-Radecki (approved):

PR #1944, opened 2025-02-12
A0-0000: Double workaround for finality version

timorleph:
# Description

This should make testnet work with a newer version, while retaining the fix on mainnet.

## Type of change

# Checklist:


timorleph:
Not doing this after all.

PR #1947, opened 2025-02-14
A0-0000: Make Deploy to Testnet and Mainnet workflows run from inputs tag

Marcin-Radecki:
It turns out it is more convient to run Deploy to Testnet and Mainnet workflows from input tag, as those workflows can fail while tagged already, like this one https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/13284043237/job/37088626598, and there's no way to re-tag and fix the issue in the workflow. Instead, those should be always called from `main` - if there's an issue in them, we can fix them and run them again against the same tag.

This PR requires some GitHub Actions shenanigans knowledge:
* There's a mess in GHA with regards to checkout out dirs of the same repo, see https://github.com/actions/runner/issues/2816. Theefore, we need to call so called composite acionts via full path, ie `Cardinal-Cryptography/aleph-node/.github/actions/get-docker-image-names@main`
* our action `get-ref-properties` it's a little bit mess too, as first of all it assumes there's some aleph-node cloned beforehand, and secondly parts of its logic requires this cloned repo, and part based on what ref was checked out by GitHub. Therefore, I removed it's usage in one place. This action should be reworked eventually, but it's out of scope of this PR.
* there's one `fetch-depth: '0'` usage, and it's required because by default checkout actions downloads only current ref

Testing: run Deploy to Testnet and Mainnet run with modified version which does not any write/push. 
* https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/13325962070/job/37219219310
* https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/13325942309

Unfortunately, I cannot test on push tag workflow that this PR modifies. Please read carefully dear reviewer, and we'll battle-test this single workflow in the future.

### DIFF
--- a/.github/actions/get-docker-image-names/action.yml
+++ b/.github/actions/get-docker-image-names/action.yml
@@ -13,6 +13,9 @@ inputs:
   dockerhub-repo:
     required: true
     description: URL to public aleph-node repo in DockerHub
+  tag:
+    required: true
+    description: aleph-node git tag to run this action on
 outputs:
   ecr-rc-image:
     description: ECR release candidate image name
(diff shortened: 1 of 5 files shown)


Marcin-Radecki:
@bartoszjedrzejewski  please ignore yaml-validator errors, it's an issue in that tool we'll fix later.

bartoszjedrzejewski (approved):

PR #1963, opened 2025-03-27
A0-0000 Fixed performance e2e flakiness

Marcin-Radecki:
# Description

The way `one_validator_is_dead` is written now, it's disabling both AURA and Aleph keys. Since default production ban config threshold is 3, sometimes we kicked disabled node from the committee which then resulted in perfect score, hence test failed. This PR fixes that by extending production ban config threshold so that kick does not happen.

There's tiny bug in the test which does not influence test run - it printed out wrong account id to std out. Also, I decreased waiting time here and there so both performance tests should run faster now (around 3 minutes).

## Type of change

## Testing

* https://github.com/Cardinal-Cryptography/aleph-node/actions/runs/14104180901

### DIFF
--- a/e2e-tests/src/test/performance.rs
+++ b/e2e-tests/src/test/performance.rs
@@ -12,6 +12,7 @@ use aleph_client::{
     waiting::{AlephWaiting, BlockStatus, WaitingExt},
     AccountId, RootConnection, SignedConnection, TxStatus,
 };
+use aleph_client::pallets::committee_management::CommitteeManagementSudoApi;
 use log::info;
 use primitives::{
     SessionCount, DEFAULT_FINALITY_BAN_MINIMAL_EXPECTED_PERFORMANCE,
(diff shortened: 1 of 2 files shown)


e2e-tests/src/test/performance.rs
```diff
@@ -152,7 +164,7 @@ async fn setup_test(
 
     let validator_keys = get_test_validators(config);
     let reserved_validators = account_ids_from_keys(&validator_keys.reserved);
-    let non_reserved_validators = account_ids_from_keys(&validator_keys.non_reserved);
+    let non_reserved_validators = account_ids_from_keys(&validator_keys.non_reserved).into_iter().take(2).collect::<Vec<_>>();
```
mike1729, line 167: nit: `account_ids_from_keys(&validator_keys.non_reserved)[..2].clone()` seems more idiomatic.

mike1729 (approved):
One nit, feel free to ignore.

e2e-tests/src/test/performance.rs
Marcin-Radecki: Thanks for suggestion, but since it's a cosmetic change I'll leave it as is

